PR #11, opened 2015-04-12
Added the wrapper for the PAGAI tool.

cheshire:


PhilippWendler:
Don't you need to override `cmdline()` to add the `-i` parameter for PAGAI?

cheshire:
Not in the latest version from the repo.
I also need to patch it to accept the ".i" file extension, but that's
another story.

PhilippWendler:
When I merged this, I did not notice that there is no license header in your file.
Once could assume that you meant to distribute this under the Apache 2.0 License because the whole project has this license, but an explicit statement would be better.
Thus, could you please add an appropriate header? 
If possible, we would prefer the Apache 2.0 License like in the other files.

cheshire:
Hi Philipp,

Do you think you could just add it? Whatever license you like.
I keep maintaining that the root "LICENSE" file in the repository is sufficient, and whatever files are added are covered by it already.

PhilippWendler:
I would prefer that the original contributor of the file also adds the license to document this better.

PR #14, opened 2015-05-07
Better statistics output

cheshire:


PhilippWendler:
Adding more information to the statistics is good.

However, the naming should match that of the HTML tables, and not be specific to software verification, because BenchExec should also be able to be used with solvers etc.

Could you please adjust this? Also why is there a merge commit with some changes that are already applied in this pull request?

cheshire:
Because github does this, I think it appears as one commit though if you
merge it.

cheshire:
The previous name is bad, and the fact that it needs a whole sentence in brackets to be understandable speaks for itself.

The problem is that the terms "false positives" and "false negatives" are specific to a classification problem where we usually answer one question, e.g. "is this e-mail SPAM".
In verification, we have two questions instead of one: "can we find a counter-example trace" and "can we find a proof", and the absence of one doesn't imply the other.

Can we compromise on expanding it to "correct true verdicts", "correct false verdicts", "incorrect true verdicts" and "incorrect false verdicts"?

PhilippWendler:
This was now implemented in combination with the renamings.

PR #17, opened 2015-05-15
Enable variable expansion in input xml <column> nodes, and allow html injection in table body

MontyCarter:
These changes were made for the SMACK team to add a link to some intermediate output files that SMACK generates, which contain details about each of our benchmark runs.

After looking at how minor and general the changes were, I thought it might be useful to the community in general, without any crazy changes to the way things are working.

(If you're interested, variables get substituted into <column> pattern, which get passed as 'identifier' to the get_value_from_output() function of objects inheriting the Tool class.  This overridden function then uses the expanded variables to create links to download the intermediate files)

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -607,7 +607,8 @@ class Run():
             self.status = self.runSet.benchmark.tool.determine_result(returncode, returnsignal, output, isTimeout)
         self.category = result.get_result_category(self.identifier, self.status, self.properties)
         for column in self.columns:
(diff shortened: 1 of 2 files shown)


PhilippWendler:
Thank you for this pull request, this is a very nice idea.

I am wondering whether it might be possible to specifically mark certain values as HTML code instead of simply assuming all values to be HTML in order to prevent accidentally introduced invalid HTML. However, I do not know currently how to solve this nicely, and most values appearing in practice will probably never accidentally contain HTML.

BTW: If you like to, you could submit the file for the SMACK integration and have it added to BenchExec itself.

PR #23, opened 2015-06-02
Change base path for relative links in HTML tables

### LINKED ISSUE #21: outputPath & base_dir incorrect for no output file (html to stdout via option "-o -")
When using the "-o -" option, file paths for benchmark source files and run result logfiles are incorrect.

At least for my purposes, all file paths referenced by the html sent to stdout should ALWAYS be relative to the current working directory.  I use this "-o -" to build a table on the fly that is served as if the html physically existed in the directory where table-generator was called.

When using the "-o -" option, if paths referenced by the html sent to stdout aren't always relative paths rooted in the current working directory from where table-generator was called, then we need a way to control the base directory for relative paths.

Here's a little more detail regarding the problem.  I don't have a patch to propose, since I'm not sure how everyone is using this "-o -" option...   I put LOTS of detail here, so only read as much as you need to understand the problem.  ;)

All functions mentioned are from ./benchexec/tablegenerator/\__init__.py.

In the function insert_logfile_names(), the path to each log file is added to its corresponding \<run\>  element in the resultFile xml tree.  This is done using a relative path rooted in the current working directory where table-generator was executed.  (This is good, for my purposes).

However, before table-generator injects things into the template, this base directory gets recalculated.  In main(), on line 1242 of ./benchexec/tablegenerator/\__init__.py, the 'outputPath' is changed to be the same directory as all input xml files if all input xml files are in the same directory.  If all input xml files are not in the same directory, then 'outputPath' is set to the 'DEFAULT_OUTPUT_PATH', which is 'results/'.  

As a result, for the case when input xml files come from different folders, when the relative path is calculated for logfiles (and for original input sourcefiles, for that matter), a "../" is prepended to each file, to step back a level from this "results/" directory.  For the case when input xml files come from the same folder, the relative path to sourcefiles and logfiles is adjusted to be relative to this common folder.  In either case, the path is not correct to the actual location of the folders, unless the input xml files happened to be in the same relative folder structure expected by the if branch on line 1242.

To make this more concrete, here is my setup.

We have a main 'data' folder.  In this folder, we have several folders matching the pattern "exec_\<timestamp\>".  Each of these contains an input.xml file, and a results folder.  In the results folder are the output xml files containing results, as well as the logfiles directory (this results folder is generated by BenchExec, so the contents of this folder are as BenchExec creates them).  From the main 'data' folder, a script called results.py is executed, which searches through all of the "exec_\<timestamp\>" folders to find output xml files.  results.py calls table-generator from this main 'data' folder, and passes in the output xml files selected by the end user to be included in the table (all output xml files are relative to this 'data' folder).  Finally, all input benchmarks are located in data/sv-benchmarks.

Our apache virtual directory (called "dev" at the moment) is rooted at this data folder.  The user opens ./results.py, passing in the filter results which determine which output xml files will be included in the table.  The resulting html from table-generator needs to reference all files relative to this virtual directory root (in this case, ./data/), or else the correct directory isn't selected.  With the current problem, benchmarks end up getting referenced as something like http://\<hostname\>/dev/../../sv-benchmarks/ssh/s3_srvr.blast.03_false-unreach-call.i.cil.c.  Similarly, the logfile link in the 'status' column gets something like http://\<hostname\>/dev/smack.2015-05-28_1417.logfiles/corral_u8.s3_srvr.blast.03_false-unreach-call.i.cil.c.log.

Respectively, these two paths need to be http://\<hostname\>/dev/sv-benchmarks/ssh/s3_srvr.blast.03_false-unreach-call.i.cil.c and http://\<hostname\>/dev/exec_\<timestamp\>/results/smack.2015-05-28_1417.logfiles/corral_u8.s3_srvr.blast.03_false-unreach-call.i.cil.c.log

PhilippWendler:
Fixes #21

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -1204,7 +1204,7 @@ def main(args=None):
     if outputPath == '-':
         # write to stdout
         outputFilePattern = '-'
-        outputPath = None
+        outputPath = '.'
     else:
         outputFilePattern = "{name}.{type}.{ext}"
 


wereh1304-lgtm:
[f.txt](https://github.com/user-attachments/files/21456636/f.txt)

Issue #22, opened 2015-06-01
Table Enhancement: Frozen column and row headers


MontyCarter:
It would be really nice to have tables generated with frozen or fixed column and row headers.  With this, the benchmark file row headers on the left would always be visible, in addition to the tool/execution info column headers always being visible on top.

This is by no means a feature I'm waiting on, so for my purposes it is very low priority, and simply an enhancement.  I just thought it would be nice, as we've started to generate real results, and I've noticed it's difficult to figure out which benchmark input file is causing an erroneous result when there are several runs displayed in the same table.

I haven't really looked at what's involved, and whether this is an extensive overhaul of table-generator.  If it's a big job, don't worry about it...

PhilippWendler:
Yes, this makes a lot of sense, and it would be possible without changing the table-generator at all, just the HTML template (`template.html`). In fact, I have tried to implement this already.

The problem is that something like this is not possible with plain HTML and CSS:
Either you have a table (i.e., something in which the cells in the header align to the cells in the body),
or you have a non-scrolling header. We cannot simply set the table header to non-scrolling,
we need to use a separate tag outside the table for the header and then adjust the width of each cell in the header to the width of the corresponding column in the body using JavaScript. This can get a little bit tricky as it needs to be updated on every size change. So if somebody with experience in JavaScript would be willing to work on this, it would be great, but I cannot make any promises whether I will be able to implement this myself at some point.

cheshire:
http://www.imaputz.com/cssStuff/bigFourVersion.html This is a non-scrolling table header, no?
The disadvantage is having to set a height in advance, but we could e.g. get user screen size with javascript and set the table height to this number.

I think we would need to separate the results into two tables though: one top "header" one (tool/limits/host/options) and the actual "results" one.
Making results one with a non-scrolling header should not be hard, I think.

cheshire:
Actually, there is a plugin which does it out of the box: http://mkoryak.github.io/floatThead/

Opening a benchexec results page in Chrome, then typing in the javascript console: 
`$.getScript('https://cdnjs.cloudflare.com/ajax/libs/floatthead/1.2.13/jquery.floatThead.js');`

followed by

`$("#dataTable").floatThead()`

Does the trick! (everything seems to be working).
We might want to use that, but it might make sense to separate the "header" (with analysis options etc.) into the separate table, so it wouldn't scroll as well. It is also trivial to add the button to turn non-scrolling thead on/off.
Should I implement that?

[...]

PR #30, opened 2015-08-26
Non-scrollable table header.

cheshire:
See  #22 for comments.

### DIFF
--- a/benchexec/tablegenerator/template.html
+++ b/benchexec/tablegenerator/template.html
@@ -163,7 +163,6 @@
     <a href="javascript:createRowFilterButtons()">Filter Rows</a>
     <a href="javascript:createQuantilePlotView()">Quantile Plot</a>
     <a href="javascript:createScatterPlotView()">Scatter Plot</a>
-    <a href="javascript:toggleFloatThead()">Toggle Floating Header</a>
   </span>
   <span class="noscript">
     Please enable JavaScript for advanced features.
@@ -188,7 +187,7 @@ def format_options(str):
 {{for lineName in ['tool', 'limit', 'host', 'os', 'system', 'date', 'runset', 'branch', 'options', 'property', 'title']}}
   {{if lineName in head and head[lineName]}}
   {{py:line = head[lineName]}}
-  <tr id="{{line.id}}" class="{{line.id}} {{if not lineName in ['runset', 'title']}}collapsable-header{{endif}}">
+  <tr id="{{line.id}}" class="{{line.id}}">
 <td colspan="{{count_id_columns}}">{{line.name}}</td>
     {{for cell, width in line.content}}
 <td colspan="{{width}}">{{format_options(cell) if line.id == 'options' else cell}}</td>
@@ -354,7 +353,13 @@ function getNameOfTestDirect(runSet) {
 
 $(document).ready(function() {
   $('body').keydown(hideContentPane);
-  $("#dataTable").floatThead({scrollingTop:'1.6em', useAbsolutePositioning:false, zIndex:3});
+  $("#dataTable").floatThead({
+
+    // Offset after the floating header.
+    scrollingTop:$("#header").outerHeight(true),
+    useAbsolutePositioning:false,
+    zIndex:3
+  });
 });
 
 </script>
@@ -1429,11 +1434,6 @@ function createScatterPlotView() {
   graphData.isRunSet = false;
   showScatterPlot();
 };
-
-function toggleFloatThead() {
-    $(".collapsable-header").toggle();
-    $("#dataTable").floatThead('reflow');
-}
 
 // this function adds the listeners to the table
 $(document).ready(function() {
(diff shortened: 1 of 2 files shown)


cheshire:
Updated again, can we merge now? :P

benchexec/tablegenerator/template.html
```diff
@@ -1423,6 +1430,11 @@
   showScatterPlot();
 };
 
+function toggleFloatThead() {
+    $(".collapsable-header").toggle();
+    $("#dataTable").floatThead('reflow');
```
PhilippWendler: Is this call to reflow really necessary? Doesn't it just slow down the whole toggling (it is quite slow)?

benchexec/tablegenerator/template.html
```diff
@@ -348,6 +354,7 @@
 
 $(document).ready(function() {
   $('body').keydown(hideContentPane);
+  $("#dataTable").floatThead({scrollingTop:'1.6em', useAbsolutePositioning:false, zIndex:3});
```
PhilippWendler: Where this does value `1.6em` come from? The header is still not correctly vertically aligned, and I guess this is the problem.

PhilippWendler:
The feature `Select Columns` is broken with the floating header (header cells are no longer hidden accordingly when the column is hidden).

`#contentPaneBackground` needs to get an appropriate z-index (I guess it's 4), otherwise it won't cover the header as it should (you notice when you open a plot).

When the window is too narrow to have the top bar in one row, the table header hides the top bar and thus makes the features there unusable.

benchexec/tablegenerator/template.html
PhilippWendler: It is actually problematic if there is a change in the width of table cells, e.g., in the following flow:
- load table
- filter rows with status `true`
- toggle header and trigger reflow
- reset row filter such that cells with status `error (something)` reappear

Because the reflow has shrunk the table cells to the width according to their current content, the last step needs to increase the width of the table cells again, and thus the alignment with the header is lost.

cheshire: I think an optimal course of action would be to commit the change without the toggle button, and leave toggling for a separate task.

benchexec/tablegenerator/template.html
cheshire: `1.6em` is indeed a magic value, yet specifying the dimensions in `em` is not that evil, since it scales with the text size (ctrl+ or ctrl-).
There are two options for floatthead: either absolute positioning, or positioning inside the container (with overflow:auto). With the absolute positioning we need to know the offset in advance; I guess one way is to dynamically query for the size of the top-bar with javascript.

cheshire:
Hm interesting, plots did work for me, I did set z-index for them.
Agreed on fixing `select columns`.
What if we bring the top bar inside the `thead` element? That would also solve the problem of magic number and non-perfect offset.

benchexec/tablegenerator/template.html
PhilippWendler: Sure, why not?

PhilippWendler:
Printing seems also broken (the print preview shows the table header over the first data row), at least in Firefox.

> What if we bring the top bar inside the thead element? That would also solve the problem of magic 
> number and non-perfect offset.

I don't understand that solution really. How would it look like?
I think the current header is placed well and makes it clear that this is not part of the table but more like a navigation bar, because this style is common on websites.

cheshire:
I believe the offset problem is fixed. Thus printing and "select columns" are left.
The latter is a bit tricky, since it re-constructs the table...

PhilippWendler:
Offset looks nice now, yes. z-Index also still needs to be set.

Btw, could you please replace

    tbody tr:first-child td { border-top:3px solid black}

with

    thead tr:last-child td { border-bottom:3px solid black}

in the template to keep the border between head and body correctly?

Would you think it makes sense to squash all the commits into one before merging because the history is now quite confusing with features added and removed and the actual change is pretty small?

cheshire:
OK this change fixes the "select column" functionality. Let me look at printing now. Still not following with the z-index, let me check under firefox.

cheshire:
> Would you think it makes sense to squash all the commits into one before merging because the history is now quite confusing with features added and removed and the actual change is pretty small?

I'd prefer not to, given the current volume of commits on benchexec :P the history for the toggle header size might come useful one day.

cheshire:
> Btw, could you please replace

What about `tfoot tr:first-child`?

cheshire:
Printing is tricky, since AFAIK there is no way to know in JS that the "print" button was pressed (or maybe I'm wrong and we can capture this event)? There is a closed ticket about this: https://github.com/mkoryak/floatThead/issues/68

PhilippWendler:
> > Btw, could you please replace

> What about tfoot tr:first-child?

Why would you want to change this?
I would like to keep the thick black borders between head and body, and between body and foot, as they are now. However, with floatThead and the current CSS the upper border scrolls out of sight.

cheshire:
OK printing works both under FF and Chrome, CSS border changed, I still don't follow you with z-index. For me plots work both under FF and Chrome, and I do recall changing the z-index.

cheshire:
z-index fix is done.

PR #36, opened 2015-09-29
Enable using custom hierarchy for cgroups instead of parent one.

IljaZakharov:
I am going to use BenchExec project in the new architecture of LDV project which relies on Docker lightweight virtualization. Docker restricts access to its cgroup hierarchy in running container which precludes BenchExec to count resources. As a solution I added option to make runexec using custom cgroup hierarchy. For example with provided «--cgparent aaa» option runexec will use «/sys/fs/cgroup/cpuacct/aaa» hierarchy and without the option all work as usual.

### DIFF
--- a/benchexec/cgroups.py
+++ b/benchexec/cgroups.py
@@ -47,7 +47,7 @@ MEMORY = 'memory'
 ALL_KNOWN_SUBSYSTEMS = set([CPUACCT, CPUSET, FREEZER, MEMORY])
 
 
-def find_my_cgroups(cgroup_paths=None):
+def find_my_cgroups(cgroup_paths=None, cgparent=None):
     """
     Return a Cgroup object with the cgroups of the current process.
     Note that it is not guaranteed that all subsystems are available
@@ -56,12 +56,13 @@ def find_my_cgroups(cgroup_paths=None):
     A subsystem may also be present but we do not have the rights to create
     child cgroups, this can be checked with require_subsystem().
     @param cgroup_paths: If given, use this instead of reading /proc/self/cgroup.
+    @param cgparent: If given, use this instead of parent mount path.
     """
     logging.debug('Analyzing /proc/mounts and /proc/self/cgroup for determining cgroups.')
     if cgroup_paths is None:
-        my_cgroups = dict(_find_own_cgroups())
+        my_cgroups = dict(_find_own_cgroups(cgparent))
     else:
-        my_cgroups = dict(_parse_proc_pid_cgroup(cgroup_paths))
+        my_cgroups = dict(_parse_proc_pid_cgroup(cgroup_paths, cgparent))
 
     cgroupsParents = {}
     for subsystem, mount in _find_cgroup_mounts():
@@ -89,30 +90,35 @@ def _find_cgroup_mounts():
         logging.exception('Cannot read /proc/mounts')
 
 
-def _find_own_cgroups():
+def _find_own_cgroups(cgparent):
     """
     For all subsystems, return the information in which (sub-)cgroup this process is in.
     (Each process is in exactly cgroup in each hierarchy.)
+    @param cgparent: If given, use this instead of parent mount path.
     @return a generator of tuples (subsystem, cgroup)
     """
     try:
         with open('/proc/self/cgroup', 'rt') as ownCgroupsFile:
-            for cgroup in _parse_proc_pid_cgroup(ownCgroupsFile):
+            for cgroup in _parse_proc_pid_cgroup(ownCgroupsFile, cgparent):
                 yield cgroup
     except IOError:
         logging.exception('Cannot read /proc/self/cgroup')
 
 
-def _parse_proc_pid_cgroup(content):
+def _parse_proc_pid_cgroup(content, cgparent):
     """
     Parse a /proc/*/cgroup file into tuples of (subsystem,cgroup).
     @param content: An iterable over the lines of the file.
+    @param cgparent: If given, use this instead of parent mount path.
     @return: a generator of tuples
     """
     for ownCgroup in content:
         #each line is "id:subsystem,subsystem:path"
         ownCgroup = ownCgroup.strip().split(':')
-        path = ownCgroup[2][1:] # remove leading /
+        if cgparent is not None:
+            path = cgparent
+        else:
+            path = ownCgroup[2][1:] # remove leading /
         for subsystem in ownCgroup[1].split(','):
             yield (subsystem, path)
 
--- a/benchexec/runexecutor.py
+++ b/benchexec/runexecutor.py
@@ -89,6 +89,8 @@ def main(argv=None):
     parser.add_argument("--dir", metavar="DIR",
                         help="working directory for executing the command (default is current directory)")
     parser.add_argument("--version", action="version", version="%(prog)s " + __version__)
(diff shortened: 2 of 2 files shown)


PhilippWendler:
The problem with such a parameter is that the cgroup within which a process is may well be different in multiple hierarchies. For example, a process may be in the following cgroups at the same time:

     /sys/fs/cgroup/cpuset/
     /sys/fs/cgroup/memory/user/23221.user/c4.session

A different solution would be to just execute `runexec` in the correct cgroup, for example by using `cgexec`, just writing the pid to the `tasks` file, or running your whole framework in the appropriate cgroup (this is what we do). Would this be possible for you? If yes, I would prefer to not add the parameter which would only be an incomplete solution.

IljaZakharov:
Running a Docker container I have not any cgroups from the host and even with new created cgroups inside with `runexec` and `cgexec` I have path like:
`/sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11/root` which is not accessible (`root` is a new cgroup and the long number is a container identifier).

The only working use case which I know is to use /sys/fs/cgroup/freezer/, /sys/fs/cgroup/cpuacct/, etc. Is there any other way to make `runexec` use it except my fix?

PhilippWendler:
Sorry, I do not fully understand your comment.

You can use `runexec` within any cgroup you like, you just need to put it in the cgroup you would like it to use. `runexec` will always create sub-cgroups of its current cgroup.

Thus, if you would like to call `runexec --cgparent foo/bar`, you could put `runexec` into the cgroup `foo/bar` yourself, and have the same effect (with the additional advantage that this works with arbitrary cgroup hierarchy layouts, which the parameter cannot do).

IljaZakharov:
Maybe you are right, but I cannot reproduce what you mean. Let me give you an example:
First, I have tried to use `runexec` as it is in a container and got the following error:
```
# runexec ls
2015-09-30 08:01:10,333 - WARNING - Cannot use cgroup hierarchy mounted at /sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11 for subsystem cpuacct, reason: No such file or directory. If permissions are wrong, please run "sudo chmod o+wt '/sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11'".
``` 
In fact I cannot change permissions in the container for this cgroup.
When I use `cgexec` I still got infeasible path:
```
# cgexec -g memory,cpuacct:/ runexec ls
2015-09-30 07:57:10,612 - WARNING - Cannot use cgroup hierarchy mounted at /sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11 for subsystem cpuacct, reason: No such file or directory. If permissions are wrong, please run "sudo chmod o+wt '/sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11'"
```
Or creating new one:
```
# cgcreate -a root:root -t root:root -g cpuacct,memory:root
# cgexec -g memory,cpuacct:/root runexec ls
2015-09-30 08:06:05,749 - WARNING - Cannot use cgroup hierarchy mounted at /sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11/root for subsystem cpuacct, reason: No such file or directory. If permissions are wrong, please run "sudo chmod o+wt '/sys/fs/cgroup/cpuacct/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11/root'".
```

As you can see the problem is that I cannot access cgroups `/864f895...` which belong to the container. My question is how to make `runexec` use cgroup like `/sys/fs/cgroup/cpuacct/`excluding  `/864f895...`.

PhilippWendler:
This is weird. Can you confirm that inside a Docker container `/proc/self/cgroup` lists a cgroup like `cpuacct:/benchexec/864f89580bbafe4fc0a90de72515ff56086a2ca29d2d46784280949648611c11` that does not exist under `/sys/fs/cgroup/cpuacct/`? This would mean that Docker exposes an inconsistent view of the cgroup hierarchy to the processes running in a container, which would mean that probably most cgroups-based software would not work in a container, which I would find surprising.

Which cgroups do actually exist under `/sys/fs/cgroup/cpuacct/` inside the container?
Is the appropriate cgroup hierarchy actually mounted at that directory inside the container?
Is it mounted in read-writable mode?
From this [StackOverflow post](https://stackoverflow.com/questions/32534203/mounting-cgroups-inside-a-docker-container) I assume that you need to assign special privileges to the Docker container to be allowed to use cgroups, did you do that?

IljaZakharov:
Yes, you are right. Below I added commands output. And I always use privileged mode running docker, since without it docker mounts file system in ro mode.

Inside a container I have the following cgroups:
```
# cat /proc/self/cgroup
12:name=systemd:/
...
3:cpuacct:/benchexec/20b6d1b909f3098d128b75aeccc191bdc1ace5ea0a45878046db6dd282370d92
...
# ls /sys/fs/cgroup/cpuacct/
cgroup.clone_children  cgroup.procs  cpuacct.stat  cpuacct.usage  cpuacct.usage_percpu	notify_on_release  tasks
```
At the same time in the host system there is no anything about docker in `/proc/self/cgroup`:
```
$ cat /proc/self/cgroup
12:name=systemd:/user/1000.user/c2.session
11:hugetlb:/user/1000.user/c2.session
...
```
But there are cgroups mounted at `/sys/fs/cgroup/cpuacct/benchexec/`:
```
ls /sys/fs/cgroup/cpuacct/benchexec/
20b6d1b909f3098d128b75aeccc191bdc1ace5ea0a45878046db6dd282370d92  cgroup.clone_children  cgroup.procs  cpuacct.stat  cpuacct.usage  cpuacct.usage_percpu  notify_on_release  tasks
```

PhilippWendler:
Did you do something maybe to have only `/sys/fs/cgroup/cpuacct/benchexec/20b6d1b909f3098d128b75aeccc191bdc1ace5ea0a45878046db6dd282370d92` mountet at `/sys/fs/cgroup/cpuacct` in the container?

I found https://github.com/docker/docker/issues/12394 (the original post, not the further discussion) which shows two options that should both work and let you see and manipulate `/sys/fs/cgroup/cpuacct/benchexec/20b6d1b909f3098d128b75aeccc191bdc1ace5ea0a45878046db6dd282370d92` inside the container like is done in the example given in that post. Can you try out these two solutions?

IljaZakharov:
Seems that approach with explicit mounting of `/sys/fs/cgroup` with read/write access does work, so I am going to stop on this solution. Thank you for fruitful discussion!

PhilippWendler:
Glad to here that it works. So you used the `-v /sys/fs/cgroup:/sys/fs/cgroup` switch and this was enough to make it work? Or did you use the other solution?

I am asking because I would like to add the information how to run in a Docker container to the documentation.

IljaZakharov:
Yes, I used option `-v "/sys/fs/cgroup:/sys/fs/cgroup:rw"` for `docker run` command. Moreover, it works without `--privileged=true` which is much more better.

PR #38, opened 2015-10-07
first attempt at seahorn wrapper

agurfinkel:
Based on the [ufo.py](https://github.com/dbeyer/benchexec/blob/master/benchexec/tools/ufo.py) sample provided in Dirk's email.
Unfortunately, I have no way of testing this. 

I could not find a way to pass specific flags for property, counterexamples, and platform.

### DIFF
--- a/benchexec/tools/seahorn.py
+++ b/benchexec/tools/seahorn.py
@@ -55,6 +55,7 @@ DM-0002198
 """
 import benchexec.util as util
 import benchexec.tools.template
+import benchexec.result as result
 
 class Tool(benchexec.tools.template.BaseTool):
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:
The typical way to handle configuration options of tools within BenchExec is to allow the user to specify them in the benchmark configuration file, not have them hard-coded in the tool wrapper
This makes the wrapper script added to BenchExec useful for your tool in general, not only specifically for SV-Comp. For example, benchmarking users could then choose the target path for the counterexample file themselves.
Thus, could you please remove `--cex` and `-p inline:no_inline`?

The property (file) gets passed to the `cmdline` function in the `propertyfile` parameter, so if your tools wants to get the property passed, you can use this parameter. Any further options such as platform are to be defined in the benchmark configuration file.

You should be able to test this using any Linux machine (even a VM). Are there any specific problems you are experiencing?

PhilippWendler:

agurfinkel:
Hey, we are working on updating the wrapper. This, however, also requires changes in the tool itself, so it has been taken a while.

agurfinkel:
@PhilippWendler please try the pull request again.

PhilippWendler:
Does your tool need to get passed the property file? I assume yes because you asked for it. In this case you should use the `propertyfile` parameter in the `cmdline` function.

Apart from this it looks good now, thanks.

lememta:
@PhilippWendler we have added the support for propertyfile. Can you pull request?

PR #39, opened 2015-10-07
DepthK Class to work on Benchexec

WilliameRocha:
ESBMC+DepthK Tool.

### DIFF
--- a/benchexec/tools/depthk.py
+++ b/benchexec/tools/depthk.py
@@ -1,4 +1,18 @@
-
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+Copyright (C) 2007-2015  Dirk Beyer
+All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
 import subprocess
 import os
 import re
@@ -13,7 +27,7 @@ class Tool(benchexec.tools.template.BaseTool):
 
     def executable(self):
         #Relative path to depthk wrapper
-        return Util.find_executable('../../depthkp/depthk-wrapper.sh')
+        return Util.find_executable('depthk-wrapper.sh')
 
 
     def program_files(self, executable):
@@ -50,7 +64,7 @@ class Tool(benchexec.tools.template.BaseTool):
         if len(output) <= 0:
           return
 
-        output = output[-1].strip()
+        output = output[-1].strip()
         status = ""
         
         if "TRUE" in output:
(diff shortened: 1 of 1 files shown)


PhilippWendler:

Does your tool use the SV-Comp property files? If yes, you should use the `propertyfile` parameter in the `cmdline` function to pass it to your tool.

The path in `executable` seems to be specific to your system, or does DepthK require this directory layout (i.e., executing benchmarks only in a directory from which the tool is reachable in `../../depthk`)? If not, could you please remove the `../../depthk/` part so users can run in easly without requiring a specific directory structure?

WilliameRocha:
**Could you please add a license to the file? We would prefer the Apache 2.0 License, if possible.**
  Apache 2.0 License was added.

**Does your tool use the SV-Comp property files? If yes, you should use the propertyfile parameter in the cmdline function to pass it to your tool.**
 Yes, the form to send propertyfile as parameter is -c propertyfile.prp, for example: depthk-wrapper.sh -c ALL.prp cFile

 During my tests, i used the ``` <propertyfile>${inputfile_path}/ALL.prp</propertyfile> 
parameter on xml file passed to wrapper through Benchexec.

**The path in executable seems to be specific to your system, or does DepthK require this directory layout (i.e., executing benchmarks only in a directory from which the tool is reachable in ../../depthk)? If not, could you please remove the ../../depthk/ part so users can run in easly without requiring a specific directory structure?**

The tool does not require a specific directory,

PhilippWendler:
Thanks for the changes. `<propertyfile>` is indeed the supposed way to pass properties, so you should properly add something like '['-c', propertyfile]' to the command line if `propertyfile` is not None (otherwise your tool won't see the property file).

WilliameRocha:
Dear Philipp,

Sorry. I couldn't understand your claim.

DepthK wrapper script is already prepared to pass properties in the command-line as follows:

$depthk-wrapper.sh -c ALL.prp verification_task_file

Is there something wrong in our call?

Many thanks for your support.

-- Williame

kfriedberger:
Maybe Philipp's intent was something like this:

Please replace line 60:
`return [os.path.relpath(executable, start=workingDir)] + options + [os.path.relpath(sourcefile, start=workingDir)]
`
by the following one:
`return [os.path.relpath(executable, start=workingDir)] + options + ['-c', propertyfile, os.path.relpath(sourcefile, start=workingDir)]
`
and remove any self-written option for the property-file from your XML-file.
Then BenchExec will automatically set the property-file and everything works.

-- Karlheinz

WilliameRocha:
Dear Philip,

the line 60 was updated.

WilliameRocha:
Dear Philipp, 

There is still something to do?

Thanks.

PhilippWendler:
Sorry for the delay, yes, this is what I meant.

Does your tool always require a property file? If not, why not make it optional and add it to the command line only if specified (i.e., if it is not None)? Currently, it will crash if the user specifies no property file.

WilliameRocha:
Hi Philipp,

Yeah. Our tool does require a property file to be specified in the command-line.

Were you able to run our tool? Is there something else that we have to do?

Many thanks for your support.

-- Williame

PhilippWendler:
I did not test it, but the wrapper looks good. Thanks for answering and addressing my questions, I will merge this now.

WilliameRocha:
Thaks for your support, Philipp.

:-)

PR #40, opened 2015-10-07
initial CIVL wrapper script

mitchellgerrard:


PhilippWendler:

mitchellgerrard:
Hi Philipp,

I'm sorry for the very delayed response.  Our tool does not use the SV-Comp property files, so we will not overwrite the cmdline function.  CIVL does have a version number that can be extracted, but when I called _version_from_tool on self, I was getting errors, so we will also not overwrite the version function.

Thanks,

Mitch

PhilippWendler:
Thanks for answering, I merged your pull request.

`_version_from_tool` is just a helper function you can use. If it does not work for you, feel free to write your own version function if you would like to have this feature.

PR #41, opened 2015-10-08
Created an initial version of SMACK's benchexec script

zvonimir:


PhilippWendler:
Does your tool use the SV-Comp property files? If yes, you should use the `propertyfile` parameter in the `cmdline` function to pass it to your tool.

The `getCmdline` function is not necessary anymore.

zvonimir:
Ok, I added support for property file, and I also remove `getCmdline`. I updated our binary at soarlab.org/smack/smack-1.5.1-64.tgz accordingly as well.
Finally, I think I somehow managed to screw up git history on this one, but I don't think that should be a big problem.

PhilippWendler:
Does your tool always require a property file? If not, why not make it optional and add it to the command line only if specified (i.e., if it `is not None`)?

zvonimir:
Based on my understanding of SVCOMP rules, I am not sure that passing in a benchmark without a property file ever makes sense. So if during competition our tool gets a benchmark without a property file, that almost certainly signals that something is wrong with your benchmarking infrastructure. Right?
Please correct me if I am wrong.
Anyhow, my assertion is there to catch potential issues with your benchmarking infrastructure. If you have in mind a situation where a benchmark should be passed in without a property file, please let me know.

PhilippWendler:
Yes, in SV-Comp there will always be a property file. I was thinking about potential users outside of SV-Comp, but if you think your tool will always be run with a property file this is ok.

Thanks for answering, I will merge this now.

PR #42, opened 2015-10-08
Added 2LS parser as provided by Peter Schrammel

tautschnig:
This is just the starting point for the CProver parsers, more work to go into them the next hours...


PhilippWendler:
Unfortunately, it seems that names of Python packages cannot start with numbers, at least it did not work in my quick test. Could you maybe change the name of the module?

peterschrammel:
Michael, you can close this request. It is superseded by #46.

PR #43, opened 2015-10-08
wrapper script for AProVE

cryingshadow:
Adapted from the ufo script - not sure whether this runs since we were not able to install BenchExec (no up-to-date Linux machine available).

### DIFF
--- a/benchexec/tools/aprove.py
+++ b/benchexec/tools/aprove.py
@@ -19,6 +19,7 @@ limitations under the License.
 """
 import benchexec.util as util
 import benchexec.tools.template
+import benchexec.result as result
 
 class Tool(benchexec.tools.template.BaseTool):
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:

For the true/false return values from the `determine_result` function, please use the constants from the `result` module instead of strings.

cryingshadow:
Thanks for the hints!

Do you have a correct example file that I could adapt?

kfriedberger:
> For the true/false return values from the determine_result function, 
> please use the constants from the result module instead of strings.

--> https://github.com/dbeyer/benchexec/blob/master/benchexec/tools/rand.py#L40

> Does your tool have a version number you can extract automatically?

--> https://github.com/dbeyer/benchexec/blob/master/benchexec/tools/cbmc.py#L37

cryingshadow:
Adapted the file to use the result constants. Currently, our tool does not provide a version number and we only participate in one category (both might still change, though).

PhilippWendler:
Thanks.

PR #44, opened 2015-10-08
Benchexec support for the Impara tool

bjowac:
Impara is based on the CProver framework, and therefore
the Python script is almost identical to the one for CBMC.
Compared to CBMC, an additional command-line parameter is provided.

### DIFF
--- a/benchexec/tools/impara.py
+++ b/benchexec/tools/impara.py
@@ -26,7 +26,7 @@ import benchexec.result as result
 
 class Tool(benchexec.tools.template.BaseTool):
     """
-    Tool wrapper for Impara (https://github.com/bjowac/impara).
+    Tool wrapper for impara (https://github.com/bjowac/impara).
     It always adds --xml-ui to the command-line arguments for easier parsing of the output.
     """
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Is the `--eager` option always necessary or is it just the specific configuration you want to run in SV-Comp? The typical way to handle configuration options of tools within BenchExec is to allow the user to specify them in the benchmark configuration file, not have them hard-coded in the tool wrapper
This makes the wrapper script added to BenchExec useful for your tool in general, not only specifically for SV-Comp. Thus I would suggest to remove the `--eager` option and specify it in the system description together with your additional parameters (for machine model etc.).

Does your tool use the SV-Comp property files? If yes, you should use the `propertyfile` parameter in the `cmdline` function to pass it to your tool, if the parameter `is not None`.

bjowac:
Thanks, Philipp. I gladly take up your suggestion of moving the --eager option
into the config files. 
My tool does not parse SVCOMP property files. 
It just looks for VERIFIER_error calls inside the C code.

PR #45, opened 2015-10-09
add symbiotic2 tool

mchalupa:
Add wrapper script for symbiotic2

Signed-off-by: Marek Chalupa <mchqwerty@gmail.com>

### DIFF
--- a/benchexec/tools/symbiotic2.py
+++ b/benchexec/tools/symbiotic2.py
@@ -1,0 +1,112 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2015  Dirk Beyer
+All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Is the `--sv-comp` option always necessary or is it just the specific configuration you want to run in SV-Comp? The typical way to handle configuration options of tools within BenchExec is to allow the user to specify them in the benchmark configuration file, not have them hard-coded in the tool wrapper
This makes the wrapper script added to BenchExec useful for your tool in general, not only specifically for SV-Comp. Thus I would suggest to remove the `--sv-comp` option and specify it in the system description together with your additional parameters (for machine model etc.).

mchalupa:
It specifies just the output format and it also tells symbiotic to read the property file, so I'd need to fix that and send a new link to the tool for pre-run. Is the system description required for the pre-run?

PhilippWendler:
If `--sv-comp` is a general option that always needs to specified anyway because for example the `determine_result` function needs this output then it should be left in and there is no need to change (e.g., for CBMC we have `--xml-ui` in the wrapper because the wrapper expects the output to be in XML format).

`--prp` is used the way it should be anyway.

For questions regarding SV-Comp such as what is needed for the pre-run phase, please ask on the SV-Comp mailing list (I cannot answer this).

mchalupa:
we changed the script behavior and also we bumped the tool's version from 2 to 3, so I'm closing this one and will create new one

PR #47, opened 2015-10-13
adding new property into results.py for handling overflows.

kfriedberger:
The property is used for "signed integers overflow",
which is undefined behavior in ISO-C-standard.

Currently no verifier/tool supports this property, but maybe soon.
Therefore BenchExec should support it first.

### DIFF
--- a/benchexec/result.py
+++ b/benchexec/result.py
@@ -43,7 +43,7 @@ _PROP_MEMTRACK =     'valid-memtrack'
 _PROP_ASSERT =       'assert'
 _PROP_AUTOMATON =    'observer-automaton'
 _PROP_SAT =          'sat'
-_PROP_OVERFLOW =     'signed-integer-overflow'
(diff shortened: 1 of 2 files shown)


PhilippWendler:
Shouldn't `_PROP_OVERFLOW` be `no-overflow` instead of `signed-integer-overflow`?
Otherwise you have `false(signed-integer-overflow)`, which seems to indicate that there is no overflow.

Is it important to distinguish between multiple types of overflows, i.e., should we have `signed-integer` in the property and the file name?

kfriedberger:
renaming finished, for further detail see:
https://github.com/dbeyer/sv-benchmarks/commit/bf7342f89f6feb105315b4ee9e67ec480f75ebbe

PhilippWendler:
I forgot to mention, could you please also update the tests in `test_result.py`?

PR #48, opened 2015-10-13
corrected determine_result

mitchellgerrard:
I was not looping through the list of strings to search for substrings previously, so all the results were returned as 'UNKNOWN'.  Let me know if there is a better Python way to search for the keywords in the 'output' list of strings.

Thanks,

Mitch

### DIFF
--- a/benchexec/tools/civl.py
+++ b/benchexec/tools/civl.py
@@ -31,9 +31,9 @@ class Tool(benchexec.tools.template.BaseTool):
         return 'CIVL'
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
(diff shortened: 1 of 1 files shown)


PhilippWendler:
There are basically 3 possibilities:
 - match complete line patterns with `pattern in output`
 - `any` as you are using it
 - iterate through `output` and check for `pattern in line` for each pattern (that's what most other wrappers do)

Please let me know if you want to change the PR or whether I should merge as it is.

mitchellgerrard:
Hi Philipp,

We will stay with using `any` since it is working on our runs for now.

Thanks,

Mitch

PR #56, opened 2015-10-28
Added driver for Predator-HP

michkot:
Added wrapper script for Predator - Hunting Party, a verification tool.


PhilippWendler:
When I merged this, I did not notice that there is no license header in your file.
Once could assume that you meant to distribute this under the Apache 2.0 License because the whole project has this license, but an explicit statement would be better.
Thus, could you please add an appropriate header? 
If possible, we would prefer the Apache 2.0 License like in the other files.

michkot:
Not a problem, I will add the license header.

PR #57, opened 2015-10-29
wrapper for cascade

Betsy-Wang:
Wrapper for cascade

### DIFF
--- a/benchexec/tools/cascade.py
+++ b/benchexec/tools/cascade.py
@@ -20,7 +20,7 @@ class Tool(benchexec.tools.template.BaseTool):
         return 'Cascade'
 
     def version(self, executable):
-        return '2.0'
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Is there a possibility to extract the version from the tool? With the hardcoded version number, this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "2.0". Thus I think it would be better to automatically get the correct version number or remove it.

Betsy-Wang:
Thanks for the suggestion! Updated!

PhilippWendler:
When I merged this, I did not notice the license header of your file.

As it is, I think we cannot use this file, because you did not even give us the right to redistribute it. Please update the license with something more permissive. If possible, we would prefer the Apache 2.0 License.

PR #59, opened 2015-10-30
Add wrapper script for map2check

hbgit:
Add wrapper script for map2check


PhilippWendler:

For the `version` function, it seems like you could use our helper function `self._version_from_tool` to make the code easier.

hbgit:
Updates done.

PR #60, opened 2015-10-30
skink svcomp16 wrapper

altmattr:
This is our BenchExec wrapper for skink.

### DIFF
--- a/benchexec/tools/skink.py
+++ b/benchexec/tools/skink.py
@@ -1,0 +1,26 @@
+import logging
+import subprocess
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+class Tool(benchexec.tools.template.BaseTool):
+    def executable(self):
+        return util.find_executable('skink.sh')
+
+    def version(self, executable):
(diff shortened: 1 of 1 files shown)


PhilippWendler:

Does your tool need to be given the propery file? If so, you need to overwrite the `cmdline` function.

Is there a possibility to extract the version from the tool? With the hardcoded version number, this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "0.314". Thus I think it would be better to automatically get the correct version number or remove it.

altmattr:
I have added a licence and I am double checking we don't need a property file.  At the moment skink does not report its version but that is something we aim to fix soon.

PhilippWendler:
Version is not required but of course nice to have. Just tell me when you know whether you need a property file and I'll merge this PR.

altmattr:
I can confirm we don't need the property file.

PhilippWendler:
Could you please still remove the hardcoded version number? As I said, I think the hardcoded number is worse than having no number as it could lead to a wrong number being shown.

PR #63, opened 2015-10-30
Add wrapper script for the LCTD verification tool

OlliSaarikivi:
This is a wrapper script for the LCTD tool available at http://users.cse.aalto.fi/osaariki/lctd-svcomp/ . This script is part of our competition contribution "LCTD: Tests-Guided Proofs for C Programs on LLVM".

### DIFF
--- a/benchexec/tools/lctd.py
+++ b/benchexec/tools/lctd.py
@@ -31,6 +31,7 @@ class Tool(benchexec.tools.template.BaseTool):
 
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
         assert len(tasks) == 1
+        assert len(options) == 0
         return [executable] + tasks
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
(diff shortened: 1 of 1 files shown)


PhilippWendler:

OlliSaarikivi:
The tool does not currently support the property files (i.e. it just checks for reachability of calling the LLVM builtin "__assert_fail" function).

The "lctdsvcomp" shell script that this wrapper script calls doesn't have support for passing options to the components of the tool. Would you like for me to still append the options in the wrapper script? Currently trying to pass any options would cause the tool to crash.

The tool doesn't currently have a version number output.

PhilippWendler:
Passing options or at least asserting that `options` is empty would be good. Otherwise someone might try to specify additional options, which would get silently swallowed.

Passing a property file is not necessary if not supported by the tool, same goes for using the version number.

OlliSaarikivi:
I added an assertion that `options` is empty to make it clear none are currently supported.

PR #65, opened 2015-10-30
Update depthk.py

WilliameRocha:
Improve to work with new category "overflow".

### DIFF
--- a/benchexec/tools/depthk.py
+++ b/benchexec/tools/depthk.py
@@ -1,3 +1,6 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
 """
 BenchExec is a framework for reliable benchmarking.
 This file is part of BenchExec.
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Please do not hardcode the version number. Otherwise this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "2.1". Thus it would be better to automatically get the correct version number or remove it.

What is the purpose of the `#!/usr/bin/python` in the first line? To be able to execute this wrapper directly (e.g., for testing), additional code would need to be added.

WilliameRocha:
Is there any problem with script??
Many thanks;

PhilippWendler:
Sorry for not responding, GitHub does not notify me when you add a new commit to the pull request, only when you add a comment. So please always comment after pushing commits.

Please do not remove the license.

Also, the function `version` is now back to the original, but I thought you wanted to fix it because it crashes? It needs to return a string, right now it returns a byte array.

You should be able to test this by running `python3 -m benchexec.test_tool_wrapper depthk` (a new feature I added in the last days).

WilliameRocha:
Follows new version.

PR #66, opened 2015-10-30
wrapper for HipTNT+

letonchanh:


PhilippWendler:

PhilippWendler:
I will merge the pull request now, if you have improvements to your wrapper, please add a new pull request.

PR #69, opened 2015-10-31
DW: added ceagle wrapper script

wanderseeme:
SV-COMP 2016 submission

More info: http://sts.thss.tsinghua.edu.cn/ceagle/

### DIFF
--- a/benchexec/tools/ceagle.py
+++ b/benchexec/tools/ceagle.py
@@ -24,36 +24,21 @@ import benchexec.util as util
 import benchexec.tools.template
 
 class Tool(benchexec.tools.template.BaseTool):
-    """
-    Tool wrapper for Ceagle.
-    It has additional features such as building CPAchecker before running it
-    if executed within a source checkout.
-    It also supports extracting data from the statistics output of CPAchecker
-    for adding it to the result tables.
-    """
 
     def executable(self):
         return util.find_executable('ceagle.sh')
 
-
     def version(self, executable):
-        return '1.0'
+        return self._version_from_tool(executable)
 
     def name(self):
         return 'Ceagle'
 
-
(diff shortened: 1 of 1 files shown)


PhilippWendler:

Is there a possibility to extract the version from the tool? With the hardcoded version number, this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "1.0". Thus I think it would be better to automatically get the correct version number or remove it. To get the version number automatically, you can probably use our helper function `self._version_from_tool`.

Maybe you could also remove the misleading comments mentioning CPAchecker.

wanderseeme:
Many thanks for your response! We tested this wrapper script in our local machines and thought the functionality of it was fine, but your comment is very helpful to us.

1. our tool only participates in Floats category and the property, so we focus our tool on `__VERIFIER_error` reachability by default, which means we do not need a property file to re-specify it; but if it is a must, I'm OK to add it.

2. we designed our tool script "ceagle.sh" as the entry without any options, it just takes verification tasks as inputs and output results (in format `FILE_NAME\tRESULT\tTIME`), so I think it's OK to skip the `options` for now.

3. I will try to fix this version function and thanks for the advice :-)

4. I will fix this asap :-D

Thank you again for your help!

PhilippWendler:
Thanks for responding.

I your tool doesn't use property files, this is not required. Just wanted to make sure it is not accidentally omitted.

Regarding the options, if your tool has no options, then you don't need to pass them, of course. However, to prevent users from accidentally adding options and getting them silently ignored, I would suggest asserting that `options` is empty (or add it to the command line just in case for the future if your script might learn options at one point).

wanderseeme:
Hi, the script is updated, please check :-P

PR #70, opened 2015-10-31
Update esbmc script

mikhailramalho:
Update to the script used by esbmc


PhilippWendler:

mikhailramalho:
Hello Phillipp, thank you for the fast response.

I was adding the following to the xmls:

  <rundefinition name="esbmc_k_induction">
    <option name="-c">${inputfile_path}/ALL.prp</option>
  </rundefinition>

but your suggestion makes this much simple.

PR #71, opened 2015-10-31
Submit wrapper script for Lazy-CSeq

lamthientruc:
Lazy-CSeq wrapper for BenchExec
http://users.ecs.soton.ac.uk/gp4/cseq/cseq.html


PhilippWendler:
Please change the file name to `lazycseq.py` or similar. This is the name that users of BenchExec will have to use to choose the tool.

Is there a possibility to extract the version from the tool? With the hardcoded version number, this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "1.0-svcomp16". Thus I think it would be better to automatically get the correct version number, or remove it completely if there is no such way. To get the version number automatically, you can probably use our helper function `self._version_from_tool`.

lamthientruc:
Thanks for the response. 

I have updated the script, could you please tell me is that look ok now?

Thanks,
Truc

PhilippWendler:
Looks fine, yes.

PR #72, opened 2015-10-31
Tool wrapper for Ceagle Absref

cgcgbcbc:


PhilippWendler:

Does your tool need to be given the propery file? If so, you need to overwrite the cmdline function.

cgcgbcbc:
@PhilippWendler Hi, License added. My tool can only target reachability property, so it doesn't need property file, does it?

PhilippWendler:
That's correct.

PR #74, opened 2015-10-31
update lazycseq.py output interpretation

lamthientruc:
Update lazycseq.py script for correct interpretation of the output.


lamthientruc:
Hi, 

I have just submitted the script for another tool of our (UL-CSeq, in the script ulcseq.py). 
I am really sorry for the inconvenience if this makes things confused.

Many thanks,
Truc

PhilippWendler:
No problem. Btw, if you have multiple similar tools, you could also inherit from a common super class to avoid the duplicate code.

PR #75, opened 2015-11-01
Add VVT tool class

hguenther:


PhilippWendler:

Note that catching exceptions in `determine_result` is dangerous. If you have a bug in this function, the exception will get silently swallowed and the result marked as "unknown".

Is there a possibility to extract the version from the tool? With the hardcoded version number, this could lead to confusing and misleading benchmark results if a newer version of your tool is used and BenchExec labels it with "prerelease". Thus I think it would be better to automatically get the correct version number, or remove it completely if there is no such way. To get the version number automatically, you can probably use our helper function `self._version_from_tool`.

hguenther:
Thanks for the feedback!
Since we're only competing in categories with error calls and have no way of parsing the propertyfile right now anyway, we don't need the property file.
We need quite a lot of preprocessing before we can parse SVComp benchmarks, which is why we need to call a script instead of our real tool, which doesn't do any option processing so we neither have any options that can be processed, nor a version that can be queried. I hope this can be alleviated in later releases.

PhilippWendler:
In this case I would recommend that you assert that the list of options is empty, otherwise any specified options will get silently swallowed, which can be confusing for users (or just append the options anyway to make the tool wrapper future-proof).

PhilippWendler:
I will merge your pull requests now, if you have further improvements to your tool class, please add a new pull requests.

PR #78, opened 2015-11-02
Add wrapper script for HIPrec

muoitranduc:
This is a wrapper script for the HIPrec tool available at http://loris-7.ddns.comp.nus.edu.sg/~project/hiprec/ . This script is part of our competition submission "HIPrec: Verifying Recursive Programs with an Unfolding-based Satisfiability Solver".

### DIFF
--- a/benchexec/tools/hiprec.py
+++ b/benchexec/tools/hiprec.py
@@ -28,20 +28,15 @@ import sys
 import os
 import re
 
+import benchexec.result as result
 import benchexec.util as util
 import benchexec.tools.template
-import benchexec.result as result
-
+from benchexec.model import SOFTTIMELIMIT
 
 sys.dont_write_bytecode = True # prevent creation of .pyc files
 
 if __name__ == "__main__":
     sys.path.append(os.path.join(os.path.dirname(__file__), os.path.pardir, os.path.pardir))
-
-import benchexec.result as result
-import benchexec.util as util
-import benchexec.tools.template
-from benchexec.model import SOFTTIMELIMIT
 
 REQUIRED_PATHS = [
                   "hiprec",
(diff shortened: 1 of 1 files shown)


PhilippWendler:

muoitranduc:
Dear Philipp Wendler,
Please note that the commit on: https://github.com/muoitranduc/benchexec/commit/f30999d571e73e6b3f284de9132836c8c1b447e3
is not the final version. We just notice there are some issues on the wrapper script. We are fixing it asap. Sorry for any inconvenience.

PhilippWendler:
No problem. GitHub does not even notify me for new commits anyway, so just add a comment when you are ready for merging.

muoitranduc:
Dear Philipp Wendler,
We added some restrictions in the latest version. It's failed on the codacy checking, but without any issue. Please take a look at it. If this commit is truly bug-free, please help in merging it. Otherwise, all feedback are welcomed since I'm quite inexperienced in this stuff.
Thanks.

PhilippWendler:
Looks good so far, thanks for updating the request. Please ignore the Codacy check failure, I don't really know what they understand under "commit quality", but they have some useful static-analysis checks (like the one that found the undeclared variable above), that's why it is enabled.

From my point it could be merged, I have just two questions which maybe you would like to address (both optional though):

muoitranduc:
Does your tool need to be given the propery file? If so, you need to use the propertyfile parameter of the cmdline function.
--> Our tool does not require a property file to run.

Does your tool have a version number you can extract automatically? Then it would be great to overwrite version and do so (makes it nicer for users, but is not required). You can probably simply call self._version_from_tool.
--> For now, I guess the tool does not have a version number so we can ignore it.

Many thanks for your feedback.

TM.

PhilippWendler:
Thanks for answering, I will merge the pull request.

PR #84, opened 2015-11-04
Added Wrapper script for Forest

pablo-aledo:
This is our BenchExec wrapper script

### DIFF
--- a/benchexec/tools/forest.py
+++ b/benchexec/tools/forest.py
@@ -1,0 +1,46 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+Copyright (C) 2007-2015  Dirk Beyer
+All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool wrapper for Forest
+    """
+    def executable(self):
+        return util.find_executable('forest')
+
+    def name(self):
+        return 'Forest'
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        assert len(tasks) == 1, "only one inputfile supported"
+        return [executable] + ["-propertyfile", propertyfile] + ["-svcomp", "-svcomp_only_output"] + [ tasks[0] ] + ["2>/dev/null"]
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        status = result.RESULT_UNKNOWN
+        if "TRUE" in output:
+            status = result.RESULT_TRUE_PROP
+        if "FALSE_REACH" in output:
+            status = result.RESULT_FALSE_REACH
+        if "FALSE_DEREF" in output:
+            status = result.RESULT_FALSE_DEREF
+        if "FALSE_FREE" in output:
+            status = result.RESULT_FALSE_FREEE
+
+        return status
--- a/benchexec/tools/forest.py
+++ b/benchexec/tools/forest.py
@@ -33,7 +33,7 @@ class Tool(benchexec.tools.template.BaseTool):
 
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
         assert len(tasks) == 1, "only one inputfile supported"
-        return [executable] + ["-propertyfile", propertyfile] + options + ["-svcomp_only_output"] + [ tasks[0] ] + ["2>/dev/null"]
+        return [executable] + ["-propertyfile", propertyfile] + options + ["-svcomp_only_output"] + [ tasks[0] ]
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
         status = result.RESULT_UNKNOWN
--- a/benchexec/tools/forest.py
+++ b/benchexec/tools/forest.py
@@ -28,9 +28,12 @@ class Tool(benchexec.tools.template.BaseTool):
     def name(self):
         return 'Forest'
 
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
         assert len(tasks) == 1, "only one inputfile supported"
-        return [executable] + ["-propertyfile", propertyfile] + ["-svcomp", "-svcomp_only_output"] + [ tasks[0] ] + ["2>/dev/null"]
+        return [executable] + ["-propertyfile", propertyfile] + options + ["-svcomp_only_output"] + [ tasks[0] ] + ["2>/dev/null"]
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
         status = result.RESULT_UNKNOWN


PhilippWendler:
Are the `-svcomp` and `-svcomp_only_output` options always necessary or are they just the specific configuration you want to run in SV-Comp? The typical way to handle configuration options of tools within BenchExec is to allow the user to specify them in the benchmark configuration file, not have them hard-coded in the tool wrapper
This makes the wrapper script added to BenchExec useful for your tool in general, not only specifically for SV-Comp. Thus I would suggest to remove both optione and specify it in the system description together with your additional parameters (for machine model etc.).

pablo-aledo:
El 05/11/2015 10:04, "Philipp Wendler" <notifications@github.com> escribió:
>
> Are the -svcomp and -svcomp_only_output options always necessary or are
they just the specific configuration you want to run in SV-Comp?

They are specific for sv-comp.

> The typical way to handle configuration options of tools within BenchExec
is to allow the user to specify them in the benchmark configuration file,
not have them hard-coded in the tool wrapper
> (except if your tool absolutely requires these options).

Who is "the user" in this case?. Do I have to modify the "benchmark
configuration file" ?. Sorry, but I don't know where this file is.

> This makes the wrapper script added to BenchExec useful for your tool in
general, not only specifically for SV-Comp. Thus I would suggest to remove
both optione and specify it in the system description together with your
additional parameters (for machine model etc.).
>
> Does your tool have a version number you can extract automatically?

No, but I can add it. For the moment there's only one person working in the
project, and we release only a version a year, for sv-comp.

Then it would be great to overwrite version and do so (makes it nicer for
users, but is not required). You can probably simply call
self._version_from_tool.

Do you prefer that I implement a way to show the version from the tool, or
call _version_from_tool ?

PhilippWendler:
> > Are the -svcomp and -svcomp_only_output options always necessary or are
> > they just the specific configuration you want to run in SV-Comp?
> 
> They are specific for sv-comp.
>  The typical way to handle configuration options of tools within BenchExec
> > is to allow the user to specify them in the benchmark configuration file,
> > not have them hard-coded in the tool wrapper
> > (except if your tool absolutely requires these options).
> Who is "the user" in this case?. Do I have to modify the "benchmark
> configuration file" ?. Sorry, but I don't know where this file is.

Sorry if I was unclear, let me explain a little bit more.
When you use BenchExec for benchmarking, you specify everything about the benchmark in a certain file (the benchmark definition file): which input files are used, which tool is used, which resource limits, which configuration options for the tool etc. BenchExec then takes this information and determines how often it has to execute the benchmarked tool and with which command lines. For the latter question, it uses the tool wrapper like the one from this pull request: given an input file and some user-specified options, what is the exact command line that needs to be executed.

Now, if the tool wrapper always adds certain configuration options to the command line, nobody will be able to use BenchExec to benchmark your tool in a different configuration. Thus, the typical way is that the tool wrapper does not enforce such options (except if they are absolutely required for this tool). Then everyone who wants to benchmark your tool can decide and specify freely which configuration options to use: SV-Comp will be able to use `-svcomp` if you tell them so in your submitted system description, and other users will be able to use other options.

You do not need to write this benchmark definition file, but the SV-Comp organizers will need to do this (and everybody else who wants to use BenchExec for benchmarking). If you are interested in how this looks, you can look at our [example file](https://github.com/dbeyer/benchexec/blob/master/doc/benchmark-example-cbmc.xml).

> > Does your tool have a version number you can extract automatically?
> No, but I can add it. For the moment there's only one person working in the
> project, and we release only a version a year, for sv-comp.
> 
> > Then it would be great to overwrite version and do so (makes it nicer for
> > users, but is not required). You can probably simply call
> > self._version_from_tool.
> 
> Do you prefer that I implement a way to show the version from the tool, or
> call _version_from_tool ?

Actually, it would be both. BenchExec just would like to have a version that it can show in the benchmark results. If your tool does not have a version that is ok for BenchExec, but it would be nice to have. How the version is determined depends on the function `version` of your tool wrapper. The typical way is that this function starts the tool with some special parameter and extracts the version from the tool output. Because this is so common, we provide a helper function `_version_from_tool` that does exactly this: start your tool with `--version` (by default) and return the output.

So if you want to provide a version number, you would need to do both: let your tool have a version number, and add some way to the tool wrapper to extract the version from the tool (possibly using our helper function).

pablo-aledo:
2015-11-05 12:34 GMT+01:00 Philipp Wendler <notifications@github.com>:
>
> Now, if the tool wrapper always adds certain configuration options to the
> command line, nobody will be able to use BenchExec to benchmark your tool
> in a different configuration. Thus, the typical way is that the tool
> wrapper does not enforce such options (except if they are absolutely
> required for this tool).
>
Then everyone who wants to benchmark your tool can decide and specify
> freely which configuration options to use: SV-Comp will be able to use
> -svcomp if you tell them so in your submitted system description, and
> other users will be able to use other options.
>
>
OK, Now I think I understand better. So, the parameter -svcomp is specific
for the competition and maybe should be integrated in the benchmark
definition file. However, the -svcomp_output_only formats the output in
such a way that is readable by the "determine_result" function, and
therefore I think it should be left in the wrapper script.


> Actually, it would be both. BenchExec just would like to have a version
> that it can show in the benchmark results. If your tool does not have a
> version that is ok for BenchExec, but it would be nice to have. How the
> version is determined depends on the function version of your tool
> wrapper. The typical way is that this function starts the tool with some
> special parameter and extracts the version from the tool output. Because
> this is so common, we provide a helper function _version_from_tool that
> does exactly this: start your tool with --version (by default) and return
> the output.
>
So if you want to provide a version number, you would need to do both: let
> your tool have a version number, and add some way to the tool wrapper to
> extract the version from the tool (possibly using our helper function).
>
OK, I'll do this and update the wrapper script. I've not find any example
of tool that uses the --version in the wrapper script, though. Can you
point me to an example?, or this doesn't need to be implemented in the
wrapper script?.

For all the changes we have discussed, do I need to create another pull
request, or can I update this one?.

Regards.


> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dbeyer/benchexec/pull/84#issuecomment-154036837>.
>



-- 
_                                      =   (
                                        255,
                                      lambda
                               V       ,B,c
                             :c   and Y(V*V+B,B,  c
                               -1)if(abs(V)<6)else
               (              2+c-4*abs(V)**-0.4)/i
                 )  ;v,      x=1500,1000;C=range(v*x
                  );import  struct;P=struct.pack;M,\
            j  ='<QIIHHHH',open('M.bmp','wb').write
for X in j('BM'+P(M,v*x*3+26,26,12,v,x,1,24))or C:
            i  ,Y=_;j(P('BBB',*(lambda T:(T*80+T**9
                  *i-950*T  **99,T*70-880*T**18+701*
                 T  **9     ,T*i**(1-T**45*2)))(sum(
               [              Y(0,(A%3/3.+X%v+(X/v+
                               A/3/3.-x/2)/1j)*2.5
                             /x   -2.7,i)**2 for  \
                               A       in C
                                      [:9]])
                                        /9)
                                       )   )
-. .----------------------------------------------------
  Y
  ,,  ,---,
 (_,\/_\_/_\
   \.\_/_\_/>
   '-'   '-'

PhilippWendler:
> OK, Now I think I understand better. So, the parameter -svcomp is specific for the competition and maybe should be integrated in the benchmark definition file. However, the -svcomp_output_only formats the output in such a way that is readable by the "determine_result" function, and therefore I think it should be left in the wrapper script.

Yes, this is exactly how it should be.

> I've not find any example of tool that uses the --version in the wrapper script, though. Can you point me to an example?, or this doesn't need to be implemented in the wrapper script?.

Actually, most wrappers do this, for example the ones for CBMC and CPAchecker. Almost all wrappers use our helper function, I think the wrapper for SMTInterpol is currently the only one that does not.

> For all the changes we have discussed, do I need to create another pull request, or can I update this one?.

Just update this request, that's fine.

pablo-aledo:
Changes made. I hope this can be merged :smiley:

PhilippWendler:
The changes look good, thanks.

However, I just noticed the command-line paramer `2>/dev/null`. Is this really on purpose? BenchExec will not use a shell to run your tool to avoid all the nasty escaping issues, thus your tool will end up with `2>/dev/null` as literal input option instead of redirecting stderr to `/dev/null`.

There is no way to tell BenchExec to ignore stderr of a tool because I think the user should get all output. If you really want to do this, you need to adjust your tool. If your tool really should get `2>/dev/null` as input parameter this is ok of course. Sorry for not noticing this earlier, the long line width hid this.

pablo-aledo:
What I want to do with "2>/dev/null" is to ignore the output of stderr and not receiving it in the "determine_result" function. I don't want this to be passed as a parameter :stuck_out_tongue_closed_eyes:. I presume from your last message there's no way do differentiate in "determine_result" what comes from stdout and what comes from stderr?. Are both of them combined in the same string and this is what we get?. How are lines separated?, by a '\n' char?. Thanks.

PhilippWendler:
Yes, both stdout and stderr of the tool are combined together and there is no way to differentiate between them. This is because (a) I think the full output of the tool should be given to the user, BenchExec should not hide anything, and (b) I do not know any way of keeping stdout and stderr separate in a meaningful way. Yes, we could capture stdout and stderr separately, but then we would loose the time relationship between the stdout lines and the stderr lines. This makes it very hard and sometimes impossible to interpret the output of tools that write both on stdout and stderr.

The parameter `output` of `determine_result` is a list, not a single string. Each entry of the list is one line of output of the tool.

pablo-aledo:
OK, I've removed the redirection. It should still work because I check for the answer in the list via "in" operator in Python. Please tell me if there's any other concern.
Pablo

PR #91, opened 2015-11-08
CBMC wrapper with SV-COMP capabilities

tautschnig:


PhilippWendler:
Does it actually make sense to keep using `--xml-ui` in any case? Or can the output be parsed similarly well without it? Then we could remove it completely.

tautschnig:
I'd leave that entirely up to you as all the previous commits to cbmc.py are yours :-)
Using XML seems like a more solid solution, it's just that I was lazy and didn't want to mess up the parsing for SV-COMP in the last hours before the ultimate deadline.

PhilippWendler:
Well, you are the expert on CBMC, so please do not hesitate to overwrite any of my stuff. If you agree that XML seems more solid in principle we can leave it as it is for now, and maybe at some point in the future it would be great if we could unify this.

PR #93, opened 2015-11-12
Add the option to filter rows with certain file names

stieglma:
With this change one can not only filter rows with specific values for the given table columns, but also after the name of the file. Javascript regular expressions can be used, e.g. ```.*loop.*``` will hide all files which do not have loop in the file/path name.

### DIFF
--- a/benchexec/tablegenerator/template.html
+++ b/benchexec/tablegenerator/template.html
@@ -751,6 +751,20 @@ function createRowFilterButtons() {
       $('<li>').append(button).append(label).appendTo(sublist);
     }
   }
+  var fileFilter = $('<ul>').text("Filter benchmark files by Name (wildcards * and . allowed)").appendTo(buttonList);
+  var inputField = $('<input>', {type:'text', name:'filter', id:target});
+  var button = $('<button>').text("apply file filter")
+                            .click(function(){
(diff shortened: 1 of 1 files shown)


PhilippWendler:
When I press `c` in the input box, the dialog closes.

Why is there an alert box with the pattern?

`*` and `.` are not wildcards. Wildcards are what the shell uses. Here a regexp is used.
Instead of `benchmark files` we use the term `tasks`.
So the label should probably read `Task filter regexp`

stieglma:
The alert box is a left over debugging output, I forgot to remove it. The label will also be renamed, about closing the box with ```c``` (```f``` is also not recognized): This is done from somewhere else, I guess here:
````javascript
function cKeyPressed(event) {
  // only "c" is pressed, not in combination with Ctrl-key
  return (event.keyCode == 67 && event.ctrlKey == false);
}

function fKeyPressed(event) {
  // only "f" is pressed, not in combination with Ctrl-key
  return (event.keyCode == 70 && event.ctrlKey == false);
}
````
And I don't know what I can do about it, but both should be disabled when typing in the filter field.

PhilippWendler:
The last change broke the HTML tables (#97), could you please work on a fix for that?

PR #94, opened 2015-11-17
walkaround for darwin when loading glibc

cgcgbcbc:
In mac os x, should use libc.dylib instead of libc.so.6


PhilippWendler:
Does BenchExec actually work on OS X? I thought that OS X lacks cgroups and thus BenchExec will not work at all.

cgcgbcbc:
> Does BenchExec actually work on OS X? I thought that OS X lacks cgroups and thus BenchExec will not work at all.

You are right... after this fixing I find that cgroups are missing in OS X...

PR #95, opened 2015-11-17
Support for OS X

cgcgbcbc:
skip any function that requires cgoups, thus making benchexec work as a simple test runner in osx.

### DIFF
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -36,6 +36,7 @@ import resource
 import subprocess
 import threading
 import time
+import platform
 
 from .model import CORELIMIT, MEMLIMIT, TIMELIMIT, SOFTTIMELIMIT
 from . import cgroups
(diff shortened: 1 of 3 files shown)


PhilippWendler:
There are some conceptual issues with this approach:

The CPU time measurement that will be reported can be wrong (cf. Section 3.1 of our [paper](http://www.sosy-lab.org/%7Edbeyer/Publications/2015-SPIN.Benchmarking_and_Resource_Measurement.pdf)). Maybe the CPU time should get completely removed instead of reporting a potentially wrong value.

Ignoring resource limits (even with a warning) could be problematic in case that users do not notice the warning and think they are being enforced. Maybe the resource limits should never get ignored if specified, and OS X users need to avoid specifying resource limits (even if the benchmark definition file contains limits, they can be disabled on the command line with `-T -1 -M -1 -c -1`.

The biggest problem I see is that BenchExec does not work as a meaningful test runner (what seems to be your goal) without cgroups, even with these changes. From a test runner, I would expect for example that it properly cleans up, e.g., terminates all started processes and child processes. BenchExec relies on cgroups for this, so with these changes there would be no process cleanup at all. Even if some code is added for process cleanup in this case, this cannot be done reliably without cgroups (cf. Section 3.3 of our [paper](http://www.sosy-lab.org/%7Edbeyer/Publications/2015-SPIN.Benchmarking_and_Resource_Measurement.pdf)).

Thus BenchExec would be left with a minimal functionality only: starting some processes and collecting their output, without measurements, without limits, and without proper cleanup. I wonder whether this is actually helpful and worth the additional complexity in the code. What is your usecase for this? Can't you use a Linux system?

Thus I am skeptical whether such a change is something that we would like to have in BenchExec. 

Are there alternative ways on OS X to ensure the necessary requirements for reliable benchmarking as listed in our paper? If such methods exist and could get added to BenchExec, this would certainly be interesting.

In case we decide we do add something like this, there are some technical problems with this pull request:

Special-casing OS X everywhere is error prone and will create problems during future development. This would need to be done in a different way that does not make the main code more complex (e.g., by a "fake" Cgroups object that does nothing). [Prefer polymorphism over conditionals.](https://sourcemaking.com/refactoring/replace-conditional-with-polymorphism)

Loading a different library in `oomhandler.py` for OS X is useless, because the OOM handler only works with cgroups. It should not get loaded at all if there are no cgroups.

We need tests that ensure that it works on OS X and keeps working (otherwise developers will fail to remember to handle OS X where necessary). Probably this can be done by adding an appropriate Mac build to the TravisCI config.

cgcgbcbc:
Thanks for your reply!


> What is your usecase for this?

My use case is that I use osx for developing (although I use linux machine as CI server), and I often need to run test against local changes, although I've written some makefile target for this, but I find benchexec's output are much more beautiful. 



> Are there alternative ways on OS X to ensure the necessary requirements for reliable benchmarking as listed in our paper?

I'll try to look into OS X's sandbox and investigate if it meets those requirements.



> by a "fake" Cgroups object that does nothing

Thanks for your advice! If OS X is able to ensure the requirements, I'll try to "mimic" the Cgroups object and implement in OS X ways, otherwise I could only write a fake Cgoups object

PhilippWendler:
>> Are there alternative ways on OS X to ensure the necessary requirements for reliable benchmarking as listed in our paper?
> 
> I'll try to look into OS X's sandbox and investigate if it meets those requirements.

Thanks. That would be great.

>> by a "fake" Cgroups object that does nothing
> 
> Thanks for your advice! If OS X is able to ensure the requirements, I'll try to "mimic" the Cgroups object and implement in OS X ways, otherwise I could only write a fake Cgoups object

Note that we have not yet decided whether we would want to include such a "fake" implementation (see my concerns at the beginning of my post), so I cannot guarantee that this would get added. But we would discuss this internally.

cgcgbcbc:
Seems that Apple only has launchd, which only has limited choices of resources limit like: cpu core, soft memory, hard memory, numbers of process etc. And it does not have any cpu time limit.

So I think this pr should be closed. And as for OSX/Windows, I think using VM/docker is enough

PR #103, opened 2015-12-18
added possibility to wrote stdout and stderr to different files

jwacalex:
added two parameters for writing stdout and/or stderr to different paths and not merge them to a generic output.

### DIFF
--- a/benchexec/runexecutor.py
+++ b/benchexec/runexecutor.py
@@ -649,8 +649,8 @@ class RunExecutor(object):
         and writes the output to a file.
         @param args: the command line to run
         @param output_filename: the file where the output should be written to
-        @param stdout_filename: dedicated file where the stdout output should be written to
-        @param stderr_filename: dedicated file where the stderr output should be written to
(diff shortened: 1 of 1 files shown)


PhilippWendler:
This pull request mixes output file writing with sudo args, could you please clean this up?

Also, why 3 files instead of just 2?

jwacalex:
#104 was splitted because I do not want to mix up both requests as one in upstream. if you want both changes we will create a new request which combines them.

it writes three files, one for stdout one for stderr and the still existing output log which might contain more information besides the pure stdout/err

PhilippWendler:
Each of the pull requests should add exactly one feature, so please clean up this pull request.

I don't really see the need for the the three files, but if you prefer this I think this is ok.
Please move their whole setup to the utility function, though, to avoid making the main function any longer than necessary.

Tests for this functionality would be good, too.

jwacalex:
regarding the pull requests the intention was to have them splitted but I broke something while creating them and merging too much commits.

jwacalex:
closing this request, because repository needs a clean-up due to some messy advent calendar fixes. there will be a new request from the current master

PR #104, opened 2015-12-18
added possibility to add custom sudo arguments

jwacalex:


PhilippWendler:
What kind of arguments would you supply here? Maybe they are generally useful?

Or is there an equivalent option that could be defined in the `sudoers` file instead?

jwacalex:
in my case it is necessary to specifiy ``-u`` and ``-r`` to set the correct selinux context since i'm using sudo from root.  

regarding pullrequest #103: both are splited because I don't know if upstream will add one or both of them. and yes, there there was an issue while generating the patchfiles

PhilippWendler:
`-u` should be set by runexec for you correctly, isn't it?

For `-r`, I think this could be useful for others as well? In this case a special parameter for the role could be added instead of `sudo_args`. This would allow to keep the interface independent from the command-line arguments of sudo and let us switch to a different "backend" if we wish to do so one day.

PhilippWendler:
Also please check the TravisCI build failure.

jwacalex:
``u`` was the wrong here, sorry for that.

in our case we need to provide two parameters in an dynamic way, because of that we have changed the methods signature

PhilippWendler:
Setting `user` per run is not supported, and I don't think I would change that. Your change also does not allow that. I think for the SELinux role this would be handled in the same way?

You can however create a fresh instance of `RunExecutor` if you need a different user per run of course.

jwacalex:
i agree with you, moving the definition of ``role`` and ``type`` to the same point where the ``user`` is defined would make it more consistent. But on the other hand it would make very specific for the advent calendar usecase, so I thought it might be better to configure additional sudo parameters per run

PhilippWendler:
What is `type`?

I think SELinux is not that specific, I would be fine with adding support for SELinux features to RunExecutor. Maybe we could also have some helpful documentation on how to set up SELinux for cases like this. Do you have some text on this?

jwacalex:
i've attached the relevant parts of the man page from an selinux aware sudo
```
     -r role, --role=role
                 Run the command with an SELinux security context that includes the specified role.
     -t type, --type=type
                 Run the command with an SELinux security context that includes the specified type.  If no type is specified, the
                 default type is derived from the role.

```

a description of *roles* and *types* can be found in the [selinux-wiki](http://selinuxproject.org/page/BasicConcepts).

If it is useful for your project I will provide a new pull request with proper documentation and the changes we have discussed here.

PhilippWendler:
Ah, you did not mention `type` before, so I did not know it was also SELinux specific. A pull request for both would be great. Thanks!

jwacalex:
i'll close this request and create a new one as soon as is finished

PR #106, opened 2015-12-24
Add new dependencies for SymDIVINE

yaqwsx:
There were added new dependecies for SymDIVINE in order to run it without packages from PPA Ubuntu repositories.

### DIFF
--- a/benchexec/tools/symdivine.py
+++ b/benchexec/tools/symdivine.py
@@ -28,7 +28,8 @@ class Tool(benchexec.tools.template.BaseTool):
     SymDIVINE info object
     """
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Please do not hardcode version numbers, this will inevitably lead to outdated and wrong information. If you cannot determine the version number, just do not override the method. No version number is better than a wrong one.

yaqwsx:
It was just a fast fix for SV-COMP and I thought version number is mandatory in the moment. There was a problem with libraries linking in the wrapper that couldn't be simply solved without changing the SymDIVINE package. After SV-COMP I'll update SymDIVINE package and fix this wrapper.

PhilippWendler:
@yaqwsx Some time has passed now, do you still want to update this tool-info module?

PR #117, opened 2016-01-15
Update systemd install docs section

mk-fg:
This is a sqashed list changes for relatively-small systemd-related INSTALL.md section:

* Don't create a separate script to go with systemd unit, keep it all contained within.
* Use `set -e` in the script, so it'd fail properly if any/all commands in it don't work.
* Do not add script itself to the cgroup, to avoid accounting for it in any way.
* Add mention for `groupadd benchexec` command, required for this script to make sense.
* Avoid flooding logs (esp. on boot, where it's bad as it is) with pointless verbose output from chgrp/chmod, drowning-out any potential errors in hundred lines of noise.
* Drop `Type=simple` unit line, as it is the default for systemd services, and has always been.
* Drop `*Accounting=true` unit lines, as these are redundant with `Delegate=true`.
* Remove `systemctl daemon-reload` command, as it is redundant with `systemctl enable benchexec-cgroup` that follows it.

Squashed together due to each being very minor tweak by itself, to avoid spamming git-log with this stuff.

Please feel free to pick only the changes you think make sense (no need for attribution), or I can tweak the commit based on the comments before merging of course, to leave some of this stuff as-is or add more info on all/any of these to commit msg/log.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -136,43 +136,37 @@ The following steps are necessary:
    This setting needs a reboot to take effect,
    and [potentially a regeneration of your initramdisk](http://www.freedesktop.org/software/systemd/man/systemd-system.conf.html#Options).
 
- * Put the following into a file `/usr/local/sbin/benchexec-cgroup.sh`.
-   By default, this gives permissions to use the BenchExec cgroup
-   to users of the group `benchexec`, please adjust this as necessary.
-   Do not forget to make the file executable.
-```
-#!/bin/bash
-# Fill in set of allowed CPUs and memory regions (default is empty).
-cp /sys/fs/cgroup/cpuset/cpuset.cpus /sys/fs/cgroup/cpuset/system.slice/
-cp /sys/fs/cgroup/cpuset/cpuset.mems /sys/fs/cgroup/cpuset/system.slice/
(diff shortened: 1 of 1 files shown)


mk-fg:
One more minor tweak I forgot to mention:

* Use `exec ...` on the last line to avoid extra "bash" process (~2-3 MiB) hanging around for no reason.

PhilippWendler:
Great, thank you very much! I do not have much experience with systemd, so any testing and improvements for systemd systems is highly appreciated. Also thanks for explaining this so nicely.

Adding the script itself to the cgroup was done because we had problems with systemd removing the created cgroup at some time, if the systemd version is too old for `Delegate`. It does not hurt because BenchExec will always create sub-cgroups and use only them for accounting.

A similar reason (systemd too old for `Delegate`) is why we added `*Accounting`.

As far as I know keeping these two points as they were does not hurt and might help people with older systemd, am I correct? If yes, would you like to update your PR accordingly?

mk-fg:
Yes, definitely shouldn't hurt, will re-add them in a moment. Thanks for review.

mk-fg:
Restored `CPUAccounting=true`, `MemoryAccounting=true` and `echo $$$$ > system.slice/benchexec-cgroup.service/tasks`, with `$$$$` being a bit weird because of systemd treating `$$` as escaping for its own templating.

Tested on RPi2 / ArchLinuxARM, incl. on boot, seem to work fine either way, "sleep" pid gets left in the cgroup in the latter case, as expected.

PR #135, opened 2016-06-17
Add property file to run specific required files.

ottseb:


PhilippWendler:
`Benchmark.add_required_file` should be unnecessary now, you could remove it.

PR #138, opened 2016-07-06
Table-generator: Add possibility to diff over arbitrary columns

stephanlukasczyk:
In its current state, the table generator is able to create a table with the differences between run sets.  This currently only considers the `status` column.

This commit extends the functionality in order to use one (or more) arbitrary columns that are defined in the table definition XML file for the comparison for the diff table.

To activate the feature, you need version 1.7 of the table DTD and add the attribute `relevantForDiff` to the column, for example:

    <column title="col1" relevantForDiff="true">Matching String</column>

If the attribute is not specified at any column, the status column will be used to calculate the differences.  Otherwise only the columns with the attribute present will be used.

The PR also provides an updated DTD that needs to be published if the PR is accepted. Furthermore, it updates the documentation such that the feature is also mentioned there.


stephanlukasczyk:
I've updated the PR and also added a test case for the new functionality.

PhilippWendler:
Updating [doc/table-generator.xml](https://github.com/sosy-lab/benchexec/blob/master/doc/table-generator.xml) as documentation would be great. Otherwise this looks fine, I'll merge it.

stephanlukasczyk:
Pushed on more commit to add this documentation as well. Sorry for missing this file.

PR #141, opened 2016-09-10
Fix location of logfiles.zip

stieglma:
Creating a table with additional columns parsed from the logfiles fails due to a wrong path:

WARNING: Could not find logfile
'test/results/predicate_bitprecise_CNF_KIND.2016-09-04_1153.logfiles/CNF_KIND-path.linux-4.2-rc1.tar.xz-43_2a-drivers--usb--host--xhci-hcd.ko-entry_point_true-unreach-call.cil.out.c.log'
nor log archive
'file:///test/results/predicate_bitprecise_CNF_KIND.2016-09-04_1153.logfiles.zip'.

It seems that the ```file:///``` is too much. I changed this behaviour to match the one for unzipped logfiles. Now I can create the table with the additional columns without problems.

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -754,8 +754,7 @@ class RunResult(object):
             log_file_url = Util.make_url(log_file)
             url_parts = urllib.parse.urlparse(log_file_url, allow_fragments=False)
             log_zip_path = os.path.dirname(url_parts.path) + ".zip"
-            log_zip_url = urllib.parse.urlunparse((url_parts.scheme, url_parts.netloc,
-                log_zip_path, url_parts.params, url_parts.query, url_parts.fragment))
+            log_zip_url = Util.make_url(log_zip_path)
             path_in_zip = urllib.parse.unquote(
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Can you please add a test case that shows the broken behavior and the fix?

Some simiilar to `test_simple_table_xml_with_columns` in `integration_test.py` should probably work.

stieglma:
The attached zip-file contains a ```test/results/...``` folder with a table-definition, the necessary logfiles and the .bz2 file. When I use the tablegenerator shipped with CPAchecker I get the errors I outlined in my first post in this pull request. With the changes I have proposed creating the table works correctly.

If there is something else breaking this is of course not good, but then we still need to find another way to fix this issue. Unfortunately my freetime is not really much at the moment, so I can't help with that. My tries to put this into a unit test with the already existing files was not successful, so perhaps we have to add the files I attached here with one or two new tests.

[example.zip](https://github.com/sosy-lab/benchexec/files/542718/example.zip)

PhilippWendler:
Thanks for the example that allowed me to reproduce it. I committed a fix in d0a07f666929285d3bbcf41340489bb71ece465e.

PR #144, opened 2016-11-07
update ceagle and remove ceagle_absref

cgcgbcbc:


benchexec/tools/ceagle.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
PhilippWendler: I think this is unnecessary because the file is not an executable python script anyway,

benchexec/tools/ceagle.py
```diff
@@ -24,19 +25,8 @@
 
 class Tool(benchexec.tools.template.BaseTool):
 
-    REQUIRED_PATHS = [
```
PhilippWendler: Could please keep and update the `REQUIRED_PATHS` list? It contains all paths that are part of your tool and are necessary to execute.

benchexec/tools/ceagle_absref.py
```diff
@@ -1,55 +0,0 @@
-# coding=utf-8
```
PhilippWendler: Why remove this file? Again, it will make it difficult for other users to use this tool.

benchexec/tools/ceagle.py
```diff
@@ -45,17 +35,23 @@ def name(self):
...
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        return [executable] + options + tasks
+        spec = ["--property-file=" + propertyfile] if propertyfile is not None else []
+        return [executable] + options + spec + tasks
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
 
         status = result.RESULT_UNKNOWN
         stroutput = str(output)
+        # strresult = stroutput.split("\n")[-1]
```
PhilippWendler: This line is unnecessary.

benchexec/tools/ceagle.py
```diff
@@ -24,19 +25,8 @@
...
-                  "parsece.sh",
-                  "svcore",
-                  "svie",
-                  "verifier.py",
-                  "z3"
-                  ]
-
     def executable(self):
-        return util.find_executable('ceagle.sh')
+        return util.find_executable('sv-ceagle')
```
PhilippWendler: Changing the executable name will make it impossible to run previous versions of Ceagle with a newer BenchExec, which is unfortunate if people want to reproduce old results.

Is this change really necessary? Do you have a consistent entry point to your tool that is stable over the years?

cgcgbcbc: This change is necessary, since we use a wrapper shell script for making the output following svcomp convention and preprocessing files last year, but this year we implement all these inside our binary.

I understand your concern, so how about we re-submit a new module called ceagle_sv this year without changing previous ones?

benchexec/tools/ceagle_absref.py
cgcgbcbc: This variant of ceagle has been integrated into our main tool, and the old one are not maintained any more.

PR #145, opened 2016-11-07
Update to call new esbmc wrapper

mikhailramalho:


PhilippWendler:
Changing the executable name in the method `executable` will make it impossible to run previous versions of ESBMC with a newer BenchExec, which is unfortunate if people want to reproduce old results. Is this change really necessary? Do you have a consistent entry point to your tool that is stable over the years?

mikhailramalho:
The script is not the tool, it's a wrapper to call our binary. 

The script was only rewritten in python, the behavior is identical to last year's script.

PhilippWendler:
So can there be a solution that avoids loosing compatibility with previous versions of ESBMC?

mikhailramalho:
Since we always provided the package with the wrapper script, there won't be any compatibility issues.

PhilippWendler:
Sorry, I don't understand. In ESBMC from SV-Comp'16 there is no `esbmc-wrapper.py`, so running this version for comparison purposes would not be possible any more. It would be nice to have a solution that avoids this.

mikhailramalho:
As I said, the only thing that changed is the wrapper.

If someone wants to run esbmc with the options we used in sv-comp'16, the only have to get the binary from our website and run benchexec with the xml config from benchmark-defs. The script is irrelevant in this case, because the contain the exactly same options.

PhilippWendler:
Sorry, I seem to be missing something. As far as I understand, executing ESBMC with BenchExec without the wrapper script is not possible at all. So to execute the old ESBMC version, somebody would to download the new ESBMC version and take the wrapper script, download the old version and take the binary, combine these too and run it, and before that they would need to know that this is what they should do.

Of course it is doable, but it also sounds somewhat complicated and risks the reproducibility of old results, something which is an explicit goal of SV-Comp. I especially would want to avoid that next year perhaps there is yet another wrapper script that needs to be used, so we end up with a tight coupling of versions between BenchExec and ESBMC. That's why I am asking whether there is a single stable entry point to ESBMC that could be used to avoid this problems. If not, then we have to bite the bullet, but I wonder whether this wouldn't be good for your tool anyway? Why do you even need a wrapper script?

mikhailramalho:
Phillips, we use the script because that's how we chose to run our tool on the competition. Defining all our options on the xml is terrible in our opinion, so we chose to set our options from the script based on the property file. The last script was a awful, so we changed it.

I'll say again, if someone wants to reproduce the behaviour from last year's competition, they should only download the binary (v2.1). We'll provide the script in our website, they only need to download both the script and the binary, and place on the same folder. We'll update [the xml](https://github.com/sosy-lab/sv-comp/blob/master/benchmark-defs/esbmc.xml) to reflect the same behaviour from sv-comp16.

mikhailramalho:
@PhilippWendler, are we done with this?

I submitted our tool to the competition yesterday and it won't work unless this is merged.

PR #149, opened 2016-11-07
Create depthk.py

WilliameRocha:


benchexec/tools/depthk.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/python
```
PhilippWendler: This script is not executable, so this line is superfluous.

benchexec/tools/depthk.py
```diff
@@ -0,0 +1,159 @@
...
+            status = result.RESULT_FALSE_MEMTRACK
+        elif 'FALSE' in output:
+            status = result.RESULT_FALSE_REACH
+        else:
+            status = result.RESULT_UNKNOWN
+
+        # else:
+        #    status = result.RESULT_ERROR
+
+        return status
```
PhilippWendler: This method should not return an empty string or `None`. Please change that `result.RESULT_ERROR` is returned if no result can be determined (as it was before).

benchexec/tools/depthk.py
```diff
@@ -0,0 +1,159 @@
...
+            if identifier == 'k' and line.startswith('Bound k:'):
+                matchbound = re.search(r'Bound k:(.*)', line)
+                return matchbound.group(1).strip()
+            if identifier == 'Step' and line.startswith('Solution by:'):
+                matchstep = re.search(r'Solution by:(.*)', line)
+                return matchstep.group(1).strip()
+
+        return '-'
+
+    def addColumnValues(self, output, columns):
```
PhilippWendler: The method `addColumnValues` was only used a long time ago. It was replaced by `get_value_from_output`, so please remove it.

dbeyer:
@WilliameRocha What is the ETA for the fixes?

WilliameRocha:
Ok, I will create a new file and submit it.

Thanks

PR #150, opened 2016-11-07
Update smack.py

Guoanshisb:
We updated the wrapper to SMACK so memory safety bugs can be reported.

### DIFF
--- a/benchexec/tools/smack.py
+++ b/benchexec/tools/smack.py
@@ -75,6 +75,12 @@ class Tool(benchexec.tools.template.BaseTool):
             return result.RESULT_TRUE_PROP
         elif re.search(r'SMACK found an error.*', splitout):
             return result.RESULT_FALSE_REACH
+        elif re.search(r'SMACK found an invalid pointer dereference', splitout):
+            return result.RESULT_FALSE_DEREF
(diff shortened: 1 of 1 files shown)


benchexec/tools/smack.py
```diff
@@ -75,6 +75,12 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
             return result.RESULT_TRUE_PROP
         elif re.search(r'SMACK found an error.*', splitout):
             return result.RESULT_FALSE_REACH
+        elif re.search(r'SMACK found an invalid pointer dereference', splitout):
+            return result.RESULT_FALSE_DEREF
+        elif re.search(r'SMACK found an invalid memory deallocation', splitout):
+            return result.RESULT_FALSE_FREE
+        elif re.search(r'SMACK found a memory leak', splitout):
+            return result.RESULT_FALSE_MEMTRACK
```
PhilippWendler: Btw., you can use `'text' in splitout` to search for a substring in Python without involving a regexp engine (should be more efficient).

dbeyer:
@Guoanshisb Can you please implement the suggestion by @PhilippWendler and file a new pull request for this?  Thanks!

PR #153, opened 2016-11-08
Add wrapper scripts for two tools: Lazy-CSeq-Abs and Lazy-CSeq-Swarm

trucnguyenlam:


PhilippWendler:
By now there are quite a lot of tool-info modules for CSeq. Wouldn't it make sense to have a single entry point for all these tools and distinguish them by passing a parameter? Then less scripts and less such Python modules would be needed. For example, CPAchecker also exists in a lot of different configurations, but all are accessible via the same entry point.

Of course if your tools are actually completely different and just share the name, this would not apply.

trucnguyenlam:
Hi @PhilippWendler,

I am afraid that I may not be able to merge these tools as one since their source structures are different and it may take time to solve the dependency problem.

Regards,
Truc

PR #154, opened 2016-11-08
Specification for Yogar-CBMC

yinliangze:


benchexec/tools/yogar-cbmc.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
```
PhilippWendler: This line serves no purpose, please remove.

benchexec/tools/yogar-cbmc.py
```diff
@@ -0,0 +1,61 @@
...
+import benchexec.util as util
+import benchexec.tools.template
+
+class Tool(benchexec.tools.template.BaseTool):
+
+    def executable(self):
+        return util.find_executable('yogar-cbmc')
+
+    def version(self, executable):
+        return '0.1'
```
PhilippWendler: Hard-coding a version number in this module is not a good idea. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

If this is not possible, please remove the method completely.

benchexec/tools/yogar-cbmc.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def name(self):
+        return 'yogar-cbmc'
```
PhilippWendler: Is this the human-readable name that you use as name for your tool? It does not need to be the name of the binary here, it could also be a different string that you like more.

benchexec/tools/yogar-cbmc.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        options += ['--no-unwinding-assertions']
+        options += ['--error-label', 'ERROR']
+        options += ['--mm', 'sc']
+        options += ['--32']
+        return [executable] + options + tasks
```
PhilippWendler: Please do not hard-code parameters here, if they are not strictly always necessary. By adding them here, you would force these parameters onto all people that want to use BenchExec for benchmarking your tool.

In fact, adding `--m32` would probably make your tool behave wrong on all 64-bit benchmarks.

Furthermore, if you want to participate in SV-Comp with this tool, note that this tool-info module would also pass wrong parameters there (because in SV-Comp, you are not supposed to check for label `ERROR`).

benchexec/tools/yogar-cbmc.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+
+        status = result.RESULT_UNKNOWN
+        stroutput = str(output)
+        # strresult = stroutput.split("\n")[-1]
```
PhilippWendler: This last line is irrelevant, please remove.

benchexec/tools/yogar-cbmc.py
```diff
@@ -28,24 +27,17 @@ class Tool(benchexec.tools.template.BaseTool):
...
 
-    def version(self, executable):
-        return '0.1'
-
     def name(self):
-        return 'yogar-cbmc'
+        return 'Yogar-CBMC'
 
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
         options += ['--no-unwinding-assertions']
```
PhilippWendler: Does your tool really always need this option? Or would there be cases where people might want to execute your tool without this option? Then please also remove it.

PR #155, opened 2016-11-08
add Ultimate Taipan, update required paths, print whole line for "unknown error" case

danieldietsch:


benchexec/tools/ultimate.py
```diff
@@ -72,7 +72,7 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
                 status = result.RESULT_UNKNOWN
                 break
             elif line.startswith('ERROR'):
-                status = 'ERROR'
+                status = line
```
PhilippWendler: The result of this method should be a reasonably short status code (1-3 words) that makes sense to print in a table cell. Would returning the whole line here exceed this, or is it guaranteed to be short?

benchexec/tools/ultimateautomizer.py
```diff
@@ -25,18 +25,22 @@ class Tool(ultimate.UltimateTool):
     REQUIRED_PATHS = [
                   "artifacts.xml",
                   "AutomizerTermination.xml",
-                  "Automizer.xml",
                   "AutomizerWitnessValidation.xml",
+				  "Automizer.xml",
```
PhilippWendler: Here and in many other lines of the `REQUIRED_PATHS` list in this file and the others there is inconsistent indentation. Python is quite sensitive about this, thus please make it consistent. Maybe you used tabs? (Spaces only, please)

danieldietsch:
@PhilippWendler : You are right, I fixed those issues.

PhilippWendler:
If you want to, you can have different status values depending on the error, it should just not be too much for a table cell. I don't know your tool, so I can't say whether this will be the case. If you look at `cpachecker.py`, you can see that we have for example something like `ERROR (parsing failed)` there, which is both helpful and short enough.

danieldietsch:
Yeah, I know, we have some grouping there. But that line will contain unexpected errors and their message can be quite long, so it should probably only be used internally.

PR #156, opened 2016-11-08
Update SymDIVINE wrapper.

yaqwsx:


yaqwsx:
It could be extracted. I'll implement it.

PR #163, opened 2016-11-27
Update ceagle.py

cgcgbcbc:
add additionalEnv: PATH

### DIFF
--- a/benchexec/tools/ceagle.py
+++ b/benchexec/tools/ceagle.py
@@ -64,6 +64,6 @@ class Tool(benchexec.tools.template.BaseTool):
             status = result.RESULT_UNKNOWN
 
         return status
-    
+
     def environment(self, executable):
-        return {"additionalEnv": {"PATH": ":" + executable}}
+        return {"additionalEnv": {"PATH": ":."}}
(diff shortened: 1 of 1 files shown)


dbeyer:
@cgcgbcbc Don't you want to write something like:
return {"additionalEnv": {"PATH": ":."}}

@PhilippWendler Can you please advice?

The extracted directories needs to be added to $PATH for Ceagle.

PhilippWendler:
I agree, adding an executable to `PATH` won't work.

PhilippWendler:
Though in general I advice against changing PATH in order to execute bundled tools like in this case.
It is generally trivial to ensure that scripts etc. call bundled tools from the correct directory instead of relying on PATH.

This has two advantages:
1. It makes it easier for users to run your tool, because they don't need to ensure PATH is correct all the time.
2. It ensures that actually the bundled tool gets executed, and not some other binary with the same name that happens to be installed on the system.

cgcgbcbc:
@dbeyer Thanks for pointing out! I've fixed it. It's strange that it works when I write that way.

@PhilippWendler Thank you for your advice! I'll figure out how to find the binary's location even when current working directory is outside the binary's location. Can we make this as a walkaround before we improve our tool?

cgcgbcbc:
@PhilippWendler Finally I found the solution to this problem. Now whether call the bundled tools from PATH or not (i.e. called with a relative/absolute  path pointing to the file) does not matter now. Should I then remove what this pr added?

dbeyer:
@PhilippWendler Your advice should be conserved in an FAQ or so.  What would be a good place for where to have that?

PR #165, opened 2016-12-04
2LS: propertyfile handling and further error codes

peterschrammel:


PhilippWendler:
This code won't work as expected.

`self.options` is not guaranteed to stay unchanged between corresponding calls to `cmdline` and `determine_result`, because there will be other calls to `cmdline` in between.

PR #167, opened 2016-12-04
2LS: Fix use of self.options

peterschrammel:
self.options is not guaranteed to stay unchanged between
corresponding calls to cmdline and determine_result, because
there will be other calls to cmdline in between
(as pointed out by PhilippWendler).


PhilippWendler:
This will fail if no propertyfile is given, right? In this case, it would be good to add an assertion to fail early.

peterschrammel:
Without propertyfile the "2ls" script will complain now and `determine_result` will detect this error.

PR #174, opened 2016-12-10
Update smack.py

Guoanshisb:
This pull request is created to add output status for signed integer overflows and to refactor the code as suggested by @PhilippWendler in #150.

Please let me know if pull request like this is allowed at this moment. We found sanity check run results sent to us show that the current script in the benchexec repo does not parse the output of SMACK correctly for memory safety violations.


PhilippWendler:
Improvements to BenchExec are of course always allowed, why wouldn't they? Thanks!

PR #178, opened 2016-12-15
allow more complex names for column titles, when filtering

kfriedberger:
for correct/wrong/unknown results.

@dbeyer 
please test this PR for your tables, then apply it.


dbeyer:
Done already in commit 47a789aa55d.

PR #180, opened 2016-12-15
add support for merging hostnames in HTML tables.

kfriedberger:
With an increasing number of machines (e.g., apollon001-150) the table
gets crowded with many hostnames. As all machines are identical and a
user does not read all hostnames, we can also merge the hostnames.

Example: ['test', 'pc1', 'pc2', 'pc3'] -> ['test', 'pc*']

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -71,7 +71,6 @@ TEMPLATE_NAMESPACE={
    'json': Util.to_json,
    'create_link': Util.create_link,
    'format_options': Util.format_options,
-   'format_host': Util.format_host,
    }
 
 _BYTE_FACTOR = 1000 # bytes in a kilobyte
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -963,7 +963,7 @@ def get_table_head(runSetResults, commonFileNamePrefix):
                 runSetResult.attributes[key] = Util.prettylist(map(round_to_MHz, values))
 
             elif key == 'host':
-                runSetResult.attributes[key] = Util.prettylist(values, True)
+                runSetResult.attributes[key] = Util.prettylist(Util.merge_entries_with_common_prefixes(values))
 
             else:
                 runSetResult.attributes[key] = Util.prettylist(values)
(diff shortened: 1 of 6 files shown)


benchexec/tablegenerator/__init__.py
```diff
@@ -962,6 +962,9 @@ def round_to_MHz(value):
                         return value
                 runSetResult.attributes[key] = Util.prettylist(map(round_to_MHz, values))
 
+            elif key == 'host':
+                runSetResult.attributes[key] = Util.prettylist(values, True)
```
PhilippWendler: `True` is pretty non-intuitive here, better have something like `merge_common_values=True`.

benchexec/tablegenerator/util.py
```diff
@@ -97,17 +97,18 @@ def open_url_seekable(path_url, mode='rt'):
         return io.TextIOWrapper(response)
 
 
-def split_number_and_unit(s):
+def split_number_and_unit(s, reverse=False):
```
PhilippWendler: Adding a parameter `reverse` to this method might technically make sense, but not semantically. For example, `reverse` could also mean "recombine number and unit". What this method is now used for has nothing to do with "units".

A more general interface should be designed that allows both use cases, or they should be kept separate.

benchexec/tablegenerator/util.py
```diff
@@ -192,14 +193,30 @@ def to_json(obj):
...
 
 
-def prettylist(list_):
+def prettylist(list_, merge=False):
+    """
+    Filter out duplicate values while keeping order.
+
+    In 'merge' mode we try to also match post-fixed names.
+    This might be useful in cases of several similar values like hostnames.
+    Example: ['test', 'pc1', 'pc2', 'pc3'] -> ['test', 'pc*']
```
PhilippWendler: This explanation does not mention that this only collapses number suffixes and nothing else. This makes me wonder whether this is the right strategy.

Maybe we should just use `commonprefix` instead of removing the number? Maybe only if there is a certain number of different values with the same prefix?

kfriedberger:
@PhilippWendler 
Are there further request? Or can we merge this?

benchexec/tablegenerator/util.py
```diff
@@ -104,11 +103,22 @@ def split_number_and_unit(s):
     Splitting is done from the end, so the split is where the last digit
     in the string is (that means the prefix may include non-digit characters,
     if they are followed by at least one digit).
+    In 'reverse' mode we split the element into the string and a postfix number.
```
PhilippWendler: This comment is outdated.

PR #188, opened 2016-12-19
Update depthk.py

WilliameRocha:


PhilippWendler:
Hard-coding version numbers in the tool module is not ok.

Please also do not file multiple conflicting pull requests, this just wastes time.

PR #192, opened 2016-12-19
Changes on Version Method

WilliameRocha:
Changes on version method to fix a bug


PhilippWendler:
As I already said: Hard-coding version numbers in the tool-info module is not ok.

Please do not file additional pull requests with the same change.

PR #193, opened 2016-12-19
Changes in determine_result

avritichauhan:
Changes in determine_result


PhilippWendler:
Why should these additional branches for result values be removed? They look fine to me?

dbeyer:
@avritichauhan Please test your tool-info module with example log output from your tool and revise accordingly via a new pull request. Also see Philipp's comment above.

PR #198, opened 2016-12-28
Score calculation for unconfirmed true

dbeyer:
This is perhaps specific for SV-COMP 2017. May be should not go into master branch.


dbeyer:
Merged manually in the meanwhile.

PR #203, opened 2017-01-02
Measure CPU energy consumption

PhilippWendler:
This is done using the tool [power-gadget_benchexec](https://github.com/sosy-lab/power-gadget_benchexec).


benchexec/intel_cpu_energy.py
```diff
@@ -0,0 +1,86 @@
+import warnings
```
PhilippWendler: Here our typical header with license and `from future import` is missing. There should also be a comment like in runexecutor that this module needs to support Python 2.

benchexec/runexecutor.py
```diff
@@ -656,7 +663,7 @@ def _setup_ulimit_time_limit(self, hardtimelimit, cgroups):
     def execute_run(self, args, output_filename, stdin=None,
                    hardtimelimit=None, softtimelimit=None, walltimelimit=None,
                    cores=None, memlimit=None, memory_nodes=None,
-                   environments={}, workingDir=None, maxLogfileSize=None,
+                   environments={}, energy_measurement=None, workingDir=None, maxLogfileSize=None,
```
PhilippWendler: Why is `energy_measurement` a parameter instead of just an implementation detail of `RunExecutor`?

leostrakosch: There is old code using some script `read-energy.sh`, so I wanted to make energy measurements interchangeable. It was kind of arbitrary, we can also change that.

PhilippWendler: I understand.

The old code is not relevant anymore. In the future, we will probably want more flexibility in order to allow more measurements, but we don't know yet how this should look, so we can just do this later.

In any case, this should probably be a parameter to the constructor, and not to `execute_run`, and appended to the parameter list for backwards compatibility (`RunExecutor` is public API).

So, please, for now just hard-code this.

PR #204, opened 2017-01-03
Add a specific error result for witnesses that are structurally incorrect witnesses

mdangl:
Add a specific error result for witnesses that are structurally incorrect witnesses.
This implements a feature for witness validation requested for SVCOMP'17.


PhilippWendler:
There are two cases that seem to handle similar situations, are both necessary?

For consistency with the other results, it would be good to include a space between `error` and the explanation.

mdangl:
> There are two cases that seem to handle similar situations, are both necessary?

Yes.

> For consistency with the other results, it would be good to include a space between error and the explanation.

Yes, definitely. I fixed that. Thanks for the review.

PR #219, opened 2017-01-16
Refined error output of witness2test

tautschnig:


benchexec/tools/witness2test.py
```diff
@@ -111,5 +111,8 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
                         status = result.RESULT_FALSE_REACH
                 elif 'UNKNOWN' in output:
                     status = result.RESULT_UNKNOWN
+        elif output and \
+             re.match(r'^INVALID WITNESS FILE', output[-1].strip()) is not None:
```
PhilippWendler: Could you please change the indendation to be a multiple of 4 spaces?

And while we are at it,
```
elif (output and
     re.match(...) is not None):
```
is the recommended way to have line breaks in expressions.

tautschnig:
Thanks a lot for the feedback. I believe I have fixed this as requested.

PR #224, opened 2017-02-14
symbiotic4.py: update the tool object

mchalupa:
we have a new directories hierarchy and outcome presentation

Signed-off-by: Marek Chalupa <xchalup4@fi.muni.cz>

### DIFF
--- a/contrib/plots/README.md
+++ b/contrib/plots/README.md
@@ -16,6 +16,16 @@ but the accumulated score achieved with these results.
 Such plots are for example used by the
 [International Competition on Software Verification](http://sv-comp.sosy-lab.org/2015/results/).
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Is there a chance to make this backwards compatible? Would be nice if people could use a current BenchExec to benchmark current and older versions of Symbiotic.

PhilippWendler:
[Here](https://github.com/sosy-lab/benchexec/pull/241#issuecomment-301045423) is a suggestion how to make such changes backwards compatible, this might work in your case, too?

mchalupa:
Pushing new commits closed this issue for some reason. I'll start a new pull-request.

Issue #166, opened 2016-12-04
Provide Run Options to determine_result of Tool-Info Module


PhilippWendler:
The tool-info module for CBMC assigns values to `self.options` and reads them in `determine_result`. This is problematic because there can be several calls to `cmdline` before a call to `determine_result`, so the value of `self.options` might not belong to the run for which `determine_result` is called. `determine_result` should instead look at the output of the tool to detect the mode in which it was called.

The tool-info module for Impara contains the same code and should also be fixed.

tautschnig:
Wouldn't it be possible to provide the options to `determine_result`? The code in `determine_result` is fixable, but `parseXML` relies on that information as well as a result may or may not be inconclusive, depending on the options chosen by the user.

PhilippWendler:
It would be possible, yes, but it would be backwards-incompatible with existing tool-info modules, so I'd like to avoid it.

Could we detect from the CBMC output whether this option was specified?

tautschnig:
No, CBMC does not output how loops are being handled. (And neither would it output other options that could be of interest, such as 32/64 bits.) Overall I think it is dangerous to look at output without information about command line options, unless tools are forced to reproduce all configuration options as part of the output.

I do see that this would imply changes across all tool-info modules, and understand the concerns in doing so. In my opinion, going for kwargs would be the best approach as it adds flexibility for future developments.

MartinSpiessl:
> I do see that this would imply changes across all tool-info modules, and understand the concerns in doing so. In my opinion, going for kwargs would be the best approach as it adds flexibility for future developments.

I agree on that point.
@PhilippWendler  is it really that difficult to add `**kwargs` to all the implementations of `determine_result`. This should be straight forward and I do not see how this could possibly break existing code.

Turns out I would really need this feature (options beeing passed to `determine_result`) for classifying the output of metaval correctly. Since metaval has options to use a different verifier backend, and the result determination depends on which verifier backend is chosen, there is really no way to choose the right action in `determine_result` for the job if I do not know the options. As a workaround, I can add an ugly heuristic which will guess which tool was executed based on the output now.

PhilippWendler:
> @PhilippWendler is it really that difficult to add `**kwargs` to all the implementations of `determine_result`. This should be straight forward and I do not see how this could possibly break existing code.

Of course it breaks existing code, after all you yourself say that `**kwargs` needs to be added to all existing implementations.

MartinSpiessl:
[...]

Issue #217, opened 2017-01-12
Feature: support for multiproperty verification


mutilin:
This year we have added new demo category for SV-COMP'2017 called `c/ldv-multiproperty`. Each task in the category contains several properties to check. The properties are listed in the corresponding property file (`c/ldv-multiproperty/ALL.prp`)

```
    CHECK( init(main()), LTL(G ! call(__VERIFIER_error_<property>())) )
    ...
```
For each task we need to get verification result. The problem that we need verification result not only for the whole task, but for each property separately. Currently BenchExec provides only one verdict (`status`) for the verification task.
It looks like we need some extension of BenchExec to support the following:

1. Get status for each property separately.
2. Set correspondence between witness file and violated property with status=unsafe.
3. Specify ideal verdicts for the properties.

For items 2 and 3 it seems that we have a solution. For item 2 we can extract property from `violated_property` tag in the xml witness file. For 3 we have decided to add a file (`<task>.verdict`) for each verification task in the following format:
`
   <property>:<verdict>
`
See examples in `c/ldv-multiproperty`.

For the first item we have an idea that the verdict should be represented in the benchmark result xml file, ilke
```
    <column title="status" property="<property1>" value="<status1>" />
    …
    <column title="status" property="<propertyN>" value="<statusN>" />
```

In BenchExec it may be implemented as the new interface function, like `determine_multiproperty_result`, which returns a map from property to status.

PhilippWendler:
> 2. Set correspondence between witness file and violated property with
> status=unsafe.
> 
> For items 2 and 3 it seems that we have a solution. For item 2 we can
> extract property from /violated_property/ tag in the xml witness file.

What exactly do you mean with this?
Does BenchExec need to handle witness files for multiproperty verification?

Currently, BenchExec itself does not know about witness files, only Dirk's additional scripts for witness validation do. Of course these scripts need to be updated, but I can't say anything about them.

> 3. Specify ideal verdicts for the properties.
>
> For 3 we have decided to add a file (<task>.verdict) for each
> verification task in the following format:
> 
>     <property>:<verdict>
> 
> See examples in c/ldv-multiproperty.

Wouldn't it be a good idea to use a standard language for the format of
this file, such as YAML?

There is also a more general discussion of how to store metadata about
verification tasks in sosy-lab/sv-benchmarks#213. 
If this is actually implemented in a way that adds a metadata file for
each verification task, this should of course also contain the verdicts
you are proposing.

[...]

PR #226, opened 2017-02-17
Added support of multi-property verification

vmordan:
Implementation of #217.


benchexec/model.py
```diff
@@ -719,7 +725,14 @@ def after_execution(self, exitcode, forceTimeout=False, termination_reason=None)
             output = []
 
         self.status = self._analyze_result(exitcode, output, isTimeout, termination_reason)
-        self.category = result.get_result_category(self.identifier, self.status, self.properties)
+        if self.multiproperty:
+            for property in self.properties:
+                short_property = property.replace(result._PROP_CALL_SPECIFIC, '')
```
mutilin: what for do we need this line?

vmordan: This line removes `unreach-call` from property name, therefore tool searches for property by error function name or by name of any other categories ( 'valid-deref', 'valid-free', 'valid-memtrack', 'no-overflow', 'no-deadlock', 'termination'). This is correct for CPAchecker wrapper (if property is not equal to 'valid-deref', 'valid-free', 'valid-memtrack', 'no-overflow', 'no-deadlock', 'termination', it is considered as 'unreach-call'), but I am not sure, if it will be correct for general tool (see my comment in https://github.com/sosy-lab/benchexec/issues/217).

mutilin: If it is correct only for CPAchecker, then let cpachecker wrapper to do this job. For general case I suggest to pass 'unreach-call,error_function_name' as a property.

mutilin:
Please, have a look at failures https://travis-ci.org/sosy-lab/benchexec/builds/202661335
There I see that `yaml` module is missing. How should we install it?
```
 File "/home/travis/build/sosy-lab/benchexec/benchexec/result.py", line 25, in <module>

import yaml

ImportError: No module named yaml
```

vmordan:
Issue with missing module was resolved.

benchexec/tools/cpachecker.py
```diff
@@ -182,6 +182,30 @@ def isOutOfNativeMemory(line):
...
+        @param requiredProperty: property, for which status is expected
+        @return: status of CPAchecker after executing a run for specified property
+        """
+        for line in output:
+            # 	Property __VERIFIER_error_linux_kernel_rcu_srcu: UNKNOWN
+            match = re.match('.*Property (.*): (TRUE|FALSE|UNKNOWN).*', line)
+            if match:
+                foundProperty = match.group(1)
+                verdict = match.group(2)
+                if foundProperty in requiredProperty:
```
PhilippWendler: It would be better to match the property name exactly. Why is this not possible?

benchexec/tools/template.py
```diff
@@ -151,6 +151,24 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
+        Otherwise an arbitrary string can be returned that will be shown to the user
+        and should give some indication of the failure reason
+        (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
+        @param returncode: the exit code of the program, 0 if the program was killed
+        @param returnsignal: the signal that killed the program, 0 if program exited itself
+        @param output: a list of strings of output lines of the tool (both stdout and stderr)
+        @param isTimeout: whether the result is a timeout
+        (useful to distinguish between program killed because of error and timeout)
+        @param property: property, for which status is expected
+        @return a non-empty string, usually one of the benchexec.result.RESULT_* constants
```
PhilippWendler: The implementation of this method in CPAchecker's tool-info module seems to violate this specification, because it returns `STR_FALSE` instead of one of the `RESULT_*` constants (and `STR_FALSE` is not even something that should be used outside the `result` module).

What is the intended behavior here?

benchexec/result.py
```diff
@@ -55,6 +57,8 @@
 _PROP_LABEL =        'unreach-label'
 # currently used by SV-COMP (http://sv-comp.sosy-lab.org/2016/rules.php):
 _PROP_CALL =         'unreach-call'
+# extends _PROP_CALL for any error function:
+_PROP_CALL_SPECIFIC ='unreach-call,'
```
PhilippWendler: Why a separate constant? And what does the term `specific` mean here? Can you find a better name?

benchexec/outputhandler.py
```diff
@@ -731,9 +734,11 @@ def add_result(self, run):
         self.counter += 1
         self.dic[run.category] += 1
         self.dic[(run.category, result.get_result_classification(run.status))] += 1
-        self.score += result.score_for_task(run.identifier, run.properties, run.category, run.status)
+        self.score += result.score_for_task(run.identifier, run.properties, run.category, run.status,
+                                            run.multiproperty_statuses)
         #if run.properties:
-        self.max_score += result.score_for_task(run.identifier, run.properties, result.CATEGORY_CORRECT, None)
+        self.max_score += result.score_for_task(run.identifier, run.properties, result.CATEGORY_CORRECT, None,
+                                                run.multiproperty_statuses)
```
PhilippWendler: This is probably wrong. `max_score` does not depend on the actual tool output, which I presume is present in `run.multiproperty_statuses`.

benchexec/result.py
```diff
@@ -314,16 +332,93 @@ def get_result_classification(result):
...
+    ideal_statuses = {}
+    specification_file = filename + _YAML_EXTENSION
+    with open(specification_file, 'r') as f:
+        try:
+            content = yaml.load(f)
+            for (property, status) in content.get("correct results").items():
+                ideal_statuses[property] = status
+        except yaml.YAMLError as exc:
+            # Wrong format, ideal statuses were failed to obtained.
+            pass
```
PhilippWendler: Swallowing errors silently is not good.

benchexec/result.py
```diff
@@ -314,16 +332,93 @@ def get_result_classification(result):
...
+            score += _SCORE_UNKNOWN
+    return score
+
+
+def _get_yaml_ideal_statuses(filename):
+    # Multiple properties are being checked.
+    # We need to get results from specific YAML file.
+    ideal_statuses = {}
+    specification_file = filename + _YAML_EXTENSION
+    with open(specification_file, 'r') as f:
```
PhilippWendler: This just crashes if the file is not present or cannot be read. Proper error handling should be added.

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+
+In this case property name is equal to corresponding SV-Comp category:
+`valid-deref`, `valid-free`, `valid-memtrack`, `no-overflow`, `no-deadlock`, `termination`.
+
+Also different error functions may be used in `unreach-call` properties:
+
+    CHECK( init(main()), LTL(G ! call(error_function_1())) )
+    CHECK( init(main()), LTL(G ! call(error_function_2())) )
+
+In this case property name contains error function name: `unreach-call,error_function_1`.
```
PhilippWendler: Is there a specific reason why you use this format for the property name?

I think it is a little bit weird, as I would interpret `unreach-call,error_function_1` as something like `unreach-call` or `error_function_1`, not as `unreach-call` for `error_function_1`.

doc/run-results.md
```diff
@@ -82,6 +82,7 @@ More complex values are represented as a `dict`.
     If the `category` is `CATEGORY_ERROR`, the `status` is a human-readable string with more information
     about which kind of error occurred,
     e.g., whether the tool terminated with an error code, the time limit was hit, etc.
+    In case of multi-property verification status is expected for each property.
```
PhilippWendler: "expected" sounds weird here, because this does not describe what BenchExec expects, but what BenchExec outputs. So please describe how the result looks like.

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+Note that verification tool may not support each combination.
+
+### Provide results in xml file
+For each given property `benchexec` provides separated status as a result
+in the following format:
+
+    <column title="status (<property_name>)" value="true|false|unknown"/>
+
+In case of global errors (parsing failed, out of memory, etc.) each property
+gets unknown status.
```
PhilippWendler: Does this make sense? Having a lot of redundant information seems weird. Would it be better maybe to avoid having a result for each property if there is a global problem?

Also, what is the main `status` in case of multi-property verification? Is it present and what does it contain?

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+Multi-property verification is aimed at checking several properties
+in one verification run and providing status for each property.
+For more information about multi-property verification
+please read our paper [On-The-Fly Decomposition of Specifications in Software Model Checking](https://www.sosy-lab.org/~dbeyer/spec-decomposition/2016-FSE.On-the-Fly_Decomposition_of_Specifications_in_Software_Model_Checking.pdf).
+
+### Set multi-property verification mode
+In order to use multi-property verification in `benchexec` 
+tag `<multiproperty>` should be set to `true`.
+Note that verification tool also should support multi-property verification
+in order to provide any meaningful results.
```
PhilippWendler: It would be better if BenchExec would not provide misleading or wrong results in case the tool does not support multi-property verification. The tool-info module should be able to detect this and report an error instead of non-meaningful results, right?

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+### Specify several properties
+The property file (which is specified by `<propertyfile>` tag) 
+may contain several properties, for example:
+
+    CHECK( init(main()), LTL(G valid-free) )
+    CHECK( init(main()), LTL(G valid-deref) )
+    CHECK( init(main()), LTL(G valid-memtrack) )
+
+In this case property name is equal to corresponding SV-Comp category:
+`valid-deref`, `valid-free`, `valid-memtrack`, `no-overflow`, `no-deadlock`, `termination`.
```
PhilippWendler: I don't understand this sentence. Furthermore, it would be better to explain BenchExec by itself, such that readers do not need to know what SV-Comp does (adding an additional hint for those that know SV-Comp is nevertheless good, of course).

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+### Get correct results
+Correct results for each task are placed in YAML configuration file
+`<task_file_name>.yml` in the following format:
+
+    name:
+      - <...>
+    sources:
+      - <...>
+    correct results:
+      property1: true|false|unknown
```
PhilippWendler: What is `property1` etc.?

What happens if a property is not listed here but in the property file?
What happens if a property is listed here but not in the property file?
Why is there a need to have `unknown` as a result?

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+gets unknown status.
+
+### Get correct results
+Correct results for each task are placed in YAML configuration file
+`<task_file_name>.yml` in the following format:
+
+    name:
+      - <...>
+    sources:
+      - <...>
```
PhilippWendler: Is information from `name` and `sources` even used? What should it contain? Isn't this redundant if the name of the yml filed needs to match the input-file name anyway?

doc/multiproperty.md
```diff
@@ -0,0 +1,60 @@
...
+in the following format:
+
+    <column title="status (<property_name>)" value="true|false|unknown"/>
+
+In case of global errors (parsing failed, out of memory, etc.) each property
+gets unknown status.
+
+### Get correct results
+Correct results for each task are placed in YAML configuration file
+`<task_file_name>.yml` in the following format:
```
PhilippWendler: I don't really like this implicit coupling of input file and yaml file. So far we avoid such implicit name requirements in the input of benchexec, and make all names explicit. Would this be good here as well?

If the yaml file is what defines a task, shouldn't the yaml file be also specified as the task, and benchexec should read the input files from it?

I guess this is what #213 is about and where @delcypher might want to comment.

benchexec/result.py
vmordan: Fixed in 7a66ba0

benchexec/result.py
vmordan: Fixed in 7a66ba0

benchexec/outputhandler.py
vmordan: Fixed in 8c0938f

benchexec/outputhandler.py
```diff
@@ -737,8 +737,7 @@ def add_result(self, run):
         self.score += result.score_for_task(run.identifier, run.properties, run.category, run.status,
                                             run.multiproperty_statuses)
         #if run.properties:
-        self.max_score += result.score_for_task(run.identifier, run.properties, result.CATEGORY_CORRECT, None,
-                                                run.multiproperty_statuses)
+        self.max_score += result.score_for_task(run.identifier, run.properties, result.CATEGORY_CORRECT, None, {None:None})
```
PhilippWendler: Is `{None:None}` really what the method expects? Why is `None` an allowed property?

Wouldn't it make more sense to forbid `None` and use `{}` as empty parameter?

doc/benchmark.dtd
```diff
@@ -15,6 +15,7 @@
 
 <!ELEMENT option (#PCDATA)>
 <!ELEMENT propertyfile (#PCDATA)>
+<!ELEMENT multiproperty (#PCDATA)>
```
vmordan: Fixed in 7604dcc.
In order to use multi-property verification attribute `kind` of tag `propertyfile` should be set to `multiproperty`. Otherwise all properties will be considered as a single composite property.

benchexec/outputhandler.py
vmordan: Fixed in 3824f3e

doc/multiproperty.md
vmordan: Removed in 300b3c9 as excess information for multi-property verification.

doc/benchmark.dtd
PhilippWendler: Do we have another value for `kind` in order to allow explicitly setting the default value for users that want to document this?

Also this attribute needs to be specified in the DTD.

PhilippWendler: Do we have another value for `kind` in order to allow explicitly setting the default value for users that want to document this?

Also this attribute needs to be specified in the DTD.

doc/multiproperty.md
vmordan: Does format `unreach-call(error_function_1)` fit better for such properties?

benchexec/tools/template.py
vmordan: It was suggested to use the following format for correct results and results.xml file: 
`propertyK: true|false|unknown`
For example:
`unreach-call(__VERIFIER_error_linux_usb_urb): false`
If we use `RESULT_FALSE_REACH`, which is used in CPAchecker by default, there will be duplication:
`unreach-call(__VERIFIER_error_linux_usb_urb): false(unreach-call)`
For other properties RESULT_FALSE_* constants contain full property name, for example:
`valid-deref: false(valid-deref)`
What do you think about such duplications (in property names and in RESULT_FALSE_* constants)?

doc/multiproperty.md
vmordan: In case of global error having unknown result for each property will help to get list of checked properties in results.xml file.

Main status can be useful in multi-property verification in case of unexpected termination of a verifier. For example, main status may be unknown (timelimit, out of java memory, etc.), whereas some properties (but not all) get statuses during verification (true or false or even unknown due to internal time limitations). In this case main status will help to distinguish, which properties get status during verification and which still were checking in the moment of verifier termination. Also, as I understand, main status is required for backward compatibility with single property verification in BenchExec.
On the other hand, main status does not make much sense, if all properties get true or false statuses. Currently it is false, if any property was violated, and true otherwise. In this case it is suggested to ignore main status in multi-property verification.

doc/benchmark.dtd
vmordan: Currently only value `multiproperty` is used for obtaining several statuses. By default attribute `kind` is not set.

PhilippWendler: Yes, but it would be good if users could explicitly select the default behavior.

Also, are there checks that reject any invalid values for this property?

doc/multiproperty.md
vmordan: Tool-info module should override function `determine_result_for_property` in order to support multi-property verification. Added in documentation in 8c83026.

doc/multiproperty.md
vmordan: We have the following suggestion - pass yaml file directly to the `BenchExec` (for example, in `<include>` or some new tag), each yaml file should contain the following required attributes:
```YAML
sources:
  - <verification task>
correct results:
  property1: true|false|unknown
  ...
  propertyN: true|false|unknown
```
Thus `BenchExec` can obtain ideal results for each passed verification task (for example, C file) and then pass this information in run.

Do you agree with this suggestion?

doc/benchmark.dtd
vmordan: Default value of `kind` attribute was set to `composite` (check all properties as a single composite property) and corresponding check for that attribute was added in 88bfeae.

benchexec/result.py
vmordan: Currently property name should be equal to _PROP_* constants, thus for property:
`LTL(G ! call(__VERIFIER_error()))`
property name is `_PROP_CALL`.
If we extend this property to any error function:
`LTL(G ! call(<error_function>()))`,
then information about <error_function> is missing in `_PROP_CALL` name.
Can we extend old format to support always property with any error function? 
In this case by default property name will be the following:
`_PROP_CALL(__VERIFIER_error)`
Also only one constant `_PROP_CALL` will be required.

doc/multiproperty.md
vmordan: Unknown result shows, that some property cannot be checked (its status was not obtained). Ideally such unknown results should be as less as possible. In this case the task for verification tool is to not spend all resources on properties with unknown result and thus to find statuses for other properties (true or false). Removing such properties with unknown result will significantly simplify the verification task and will make it less challenging.

vmordan: > What happens if a property is not listed here but in the property file?

This property will be added to results.xml file, but it will not be used for computation of a score and determining if result is correct.

> What happens if a property is listed here but not in the property file?

Caused exception fixed in 5405543. Such property should not be used in score computation, task category is CATEGORY_MISSING.

PhilippWendler: So is there a difference between not listing a property in the yaml file and setting it to unknown? (It seems like this behaves identically.)

doc/multiproperty.md
PhilippWendler: Ok, seems good.

benchexec/tools/template.py
PhilippWendler: I think the most important thing is to have it consistently. So I would not like `unreach-call(__VERIFIER_error_linux_usb_urb): false` in combination with `valid-deref: false(valid-deref)`.

On the other hand, `valid-deref: false(valid-deref)` seems unnecessarily redundant anyway. So I would prefer `valid-deref: false` for the yaml file.

For the result XML (and the HTML tables) I am not sure. You might have experience looking at such tables: Is repeating the property name in cases like `valid-deref: false(valid-deref)` helpful, or does it clutter the table too much?

For the return value of `determine_result_for_property`, we could either simply use `false` as well, or we use `false(valid-deref)`. The latter would make it behave the same as `determine_result`, which I like, because in the future we could merge the two methods maybe?

doc/multiproperty.md
PhilippWendler: I haven't really thought about how this format should look like, so I would like to get other people discussing this in sosy-lab/sv-benchmarks#213.

benchexec/result.py
PhilippWendler: These constants are purely internal, so reorganize them just as you wish, as long as output is not affected.

doc/multiproperty.md
vmordan: Yes, currently there is no difference between not listing a property in the yaml file and setting it to `unknown`. Therefore, it is possible to remove status `unknown` from yaml file.

Such statuses could be helpful only if yaml file is used to create the property file, but as I understand, this functionality is unnecessary.

benchexec/tools/template.py
vmordan: In yaml file correct false results are specified in "shorten" form:
```YAML
valid-deref: false
unreach-call(__VERIFIER_error_linux_usb_urb): false
```
Function `determine_result_for_property` returns "full" false result: 
`false(valid-deref)` for property `valid-deref` and
`false(unreach-call)` for property `unreach-call(__VERIFIER_error_linux_usb_urb)`.
Same is added in result XML file for generating tables.
During comparison of results all false properties (`false(valid-deref)`, `false(unreach-call)`, etc.) are considered as `RESULT_CLASS_FALSE`.

PhilippWendler: Ok. Just to understand it correctly, the last sentence does *not* mean that if a tool returns `false(unreach-call)` for a `false(valid-deref)` property, this will be treated as correct? (It needs to be treated as wrong even if the result class is the same).

PhilippWendler:
`property_kind` should probably not be a value like the others in the result XML. After all, it is not an output value of the run. For example, it never makes sense to have this in a column in the table (but it would end up there with `table-generator --all-columns`).

Btw., how does all this work if only a part of a benchmark is marked as multiproperty? Can benchexec handle it if some `<tasks>` tags are multiproperty and some not? Can table-generator handle the results that come from such a benchmark? It it is not possible or desirable to handle this, how is this detected and what happens?

vmordan:
> Btw., how does all this work if only a part of a benchmark is marked as multiproperty?

It is possible to specify different values for attribute `kind` in `propertyfile` tag, so it will be different in different runs. For example:

```XML
  <rundefinition name="multiproperty">
    <propertyfile kind="multiproperty">ALL.prp</propertyfile>
  </rundefinition>
  
  <rundefinition name="composite">
    <propertyfile kind="composite">ALL.prp</propertyfile>
  </rundefinition>

  <tasks>
    <include>...</include>
  </tasks>
```
will specify 2 rundefinitions (with multi-property verification and with 1 composite property) for the same tasks.

vmordan:
> `property_kind` should probably not be a value like the others in the result XML.

Yes, this tag was excessive and was removed in e8ac6dc.

vmordan:
I think all requested changes were made. Do you have any more comments about current implementation of multi-property verification?

PhilippWendler:
We have discussed this internally now, and have a proposal that is slightly different. Sorry for the late changes, but we noticed that doing this differently would lead to some advantages, for example for #234.

We noticed that we will always need a short identifier for a property (for humans and tables) and a long LTL string (e.g., for #234). So far there was a mapping between short identifier and long LTL string implemented in BenchExec, but we prefer to remove this "magic" knowledge from BenchExec. Furthermore, such an implicit mapping becomes awkward with multiple properties (now we need pattern matching and start defining a specific format for the short identifier), and impossible for "real LTL" verification.
So we want to make this mapping between short identifier and long LTL string explicit. Our suggestion is to use the file name of the property file (without directory and extension) as short identifier. This is simple, easy to use, and with the least amount of effort for the users. We can reasonably expect that file names of property files are unique within a benchmark suite (the same requirement already exists for input files). This change implies that each property file defines only a single property (except for memsafety, which is still special but probably will become a regular "multi-property verification" in the future). You will then have multiple `<propertyfile>` tags in the benchmark XML and the tools will get multiple property files on the command line. This is nicely backwards compatible for the existing use cases. The metadata file for a benchmark will look like in your proposal.

Example:
sources/dummy.yml:
```
input_files:
  - dummy.c
properties:
  - unreach-call_error1:
    - result: true
  - unreach-call_error2:
    - result: true
  - termination:
    - result: false
```

Benchmark XML:
```
...
<tasks><include>sources/dummy.yml</include></tasks>
<propertyfile>props/unreach-call_error1.prp<propertyfile>
<propertyfile>props/unreach-call_error2.prp<propertyfile>
...
```

props/unreach-call_error1.prp (similarly for props/unreach-call_error2.prp):
```
CHECK( init(main()), LTL(G ! call(error_function_1())) )
```

Example tool command line:
```
verifier --property-file props/unreach-call_error1.prp --property-file props/unreach-call_error2.prp sources.c
```

Example tool output:
```
Result for CHECK( init(main()), LTL(G ! call(error_function_1())) ): false
Result for CHECK( init(main()), LTL(G ! call(error_function_2())) ): true
```

The tool and the tool-info modules for BenchExec will only use the long LTL string, they do not need to know about the short identifier at all. Currently this is awkward, because tools are given the property files with the long LTL string as input, but are expected to produce as output a result that uses the short identifier, so not only BenchExec but also all tools need to implement this implicit mapping. In the future, this is no longer necessary, so tools are made simpler and less SV-Comp specific. For backwards compatibility of existing tools that output results with the short identifier, the tool-info module can easily support this and translate it to the long LTL string.

Please note that we suggest different names for a few of the yaml identifiers (cf. example above).
The handling of multi-property verification in result XML files and tables should stay like they are in this PR.

These changes have the following advantages compared to the current state of the PR:
- Supports multiple arbitrary LTL properties (a feature that will certainly soon be requested if we add arbitrary LTL properties as well as multi-property verification).
- Less redundant property files (one for each property, not one for each combination).
- No flag `kind` with awkward names for `propertyfile` attribute is needed.
- Implementation of implicit mapping between short identifiers and long LTL string in BenchExec can be removed.
- BenchExec does not need to know about and enforce a specific format of (LTL) properties, it can just treat them as opaque strings. This also makes it easier to adopt BenchExec in other communities.
- Users can choose property names flexibly (SV-Comp can simply continue to use their existing names). We do not need to define, implement, and enforce a specific format.
- Easier enabling/disabling of properties for a verifier on the command line (no file needs to be edited).

Would this fit well for your use case?

danieldietsch:
Can you clarify some things for me?
 * What is a short identifier? 
 * How does benchexec map between the yaml file and the .prp file? 
 * Why should we keep the .prp files if we could switch to yaml completely?

PhilippWendler:
> What is a short identifier?

A short identifier for a property that is used where the full LTL formula is too unwieldy, e.g., in cells in the result tables. This is not new to the proposal, current examples of these identifiers are `unreach-call`, `valid-free`, etc.

> How does benchexec map between the yaml file and the .prp file?

Only indirectly. BenchExec always starts with the benchmark XML. This defines both the property file(s) as well as the yaml file(s) with full file names including paths. Then BenchExec knows all property entries in the yaml files and the names of all property files, and can match them.

> Why should we keep the .prp files if we could switch to yaml completely?

We need a way to pass the set of to-be-verified properties to the verifier. We can't use the proposed yaml files for multiple reasons, the most important being that the yaml file contains the expected result of the verification. Also we don't want to force tools to parse yaml.
Because the LTL formulas can be large, this should be a file, and not a command-line parameter.
So for passing the properties to the verifier we need some format, and using the existing property-file format has the advantage of being backwards-compatible.

danieldietsch:
Ok, regarding the short names, it looks like you say you just use the filename of the .prp file as short name, right?

PhilippWendler:
Yes, this is what we think is the most simple and easy way to use.

danieldietsch:
Regarding the short names: Yes, I agree. I also see no better way.

But there is another issue (at least for Ultimate): I take it for SV-COMP this would mean that we would remove all the true/false labellings from filenames and migrate them to appropriate .yaml files?

As we use the SV-COMP files not only for benchmarking but also for testing, our test result decider framework has to be updated s.t. it can understand .yaml and .prp files and match them to the appropriate Ultimate results.
From our view, tests get suddenly a lot more complicated because they are no longer some property, a file, and a configuration, but a .yaml file with a matching property subset. 

It would be easier if the .yaml file itself would contain all information and .prp files would perhaps be generated from it for tool execution (if backwards-compatibility is important). 
Tools could then chose if they want to use the new .yaml interface (e.g., get one .yaml file) or prefer to be called with multiple .prp files in their tool implementation. 
One could also transform the .yaml file easily for tool execution:
 * strip the `- result` part 
 * strip all the properties that are not enabled by the rundefinition.

PhilippWendler:
> But there is another issue (at least for Ultimate): I take it for SV-COMP this would mean that we would remove all the true/false labellings from filenames and migrate them to appropriate .yaml files?

This is not strictly required, but I assume that SV-Comp would do this, yes.
(I am also in favor of it: having these labels makes the file names very long, and its leads to many renamings, which makes it hard to keep track of the history).

> As we use the SV-COMP files not only for benchmarking but also for testing, our test result decider framework has to be updated s.t. it can understand .yaml and .prp files and match them to the appropriate Ultimate results.
> From our view, tests get suddenly a lot more complicated because they are no longer some property, a file, and a configuration, but a .yaml file with a matching property subset.
> 
> It would be easier if the .yaml file itself would contain all information and .prp files would perhaps be generated from it for tool execution (if backwards-compatibility is important).
> Tools could then chose if they want to use the new .yaml interface (e.g., get one .yaml file) or prefer to be called with multiple .prp files in their tool implementation.
> One could also transform the .yaml file easily for tool execution:
> 
>     strip the - result part
>     strip all the properties that are not enabled by the rundefinition.

This would be possible, yes.
However, it has two disadvantages:
- If tools need to parse yaml files, this is more complicated than the existing format. Consider for example tools written in C, or where the logic for handling SV-Comp tasks is implemented in bash scripts (this is surprisingly common). The authors of these tools would probably not be happy with a new format.
- This solution requires temporary files to be produced by BenchExec for the verification run (either the stripped yaml, or the property file). Currently, you can open the view with the tool output from a BenchExec table, copy the first line, paste it on the command line, and rerun the tool with the verification task in the same configuration as BenchExec did. This is very useful, for trying out tools, testing, debugging etc. (we use this extensively for investigating failures in our continuous integration, for example). With temporary files, you would need to manually create a file (with the correct content, which can be non-trivial) for every verification run you want to reproduce manually. Because of this effort we prefer a solution without temporary files.

For your tests, yes, you would have one additional yaml file for every C file, just like in SV-Comp. Note that if you have files that conform to the existing naming scheme, these yaml files could be auto-generated (I assume that a script for this will be developed anyway to migrate the sv-benchmarks repository). Maybe this is a solution for you?

Btw, you could maybe also use BenchExec as a test runner? We use BenchExec for all integration tests of CPAchecker. If you are interested in this, or would need additional features for this, just contact me (please not in this issue to keep the discussion on topic).

danieldietsch:
> If tools need to parse yaml files, this is more complicated than the existing format. Consider for example tools written in C, or where the logic for handling SV-Comp tasks is implemented in bash scripts (this is surprisingly common). The authors of these tools would probably not be happy with a new format.

I think for multi-property verification, all tools have to adapt. We currently use a python script as wrapper that converts .prp in the proper Ultimate settings. This will just get more complicated. 

> This solution requires temporary files to be produced by BenchExec for the verification run (either the stripped yaml, or the property file). Currently, you can open the view with the tool output from a BenchExec table, copy the first line, paste it on the command line, and rerun the tool with the verification task in the same configuration as BenchExec did. This is very useful, for trying out tools, testing, debugging etc. (we use this extensively for investigating failures in our continuous integration, for example). With temporary files, you would need to manually create a file (with the correct content, which can be non-trivial) for every verification run you want to reproduce manually. Because of this effort we prefer a solution without temporary files.

Agreed, this may be harder on benchexec tests. My issue is mainly: we should avoid having a benchmark set where only a combination of .yaml, .prp, actual C input files, and benchexec's .xml files defines the set of verification tasks. 
In my proposal, one does not necessarily need to create temporary files. The arguments could just be given to the tool module and tool developers could then decide what to do with them. As convenience, one could provide .prp and .yaml file generators. I would prefer to get rid of both and just encode the translation from benchexec's verification task to Ultimate's commandline. 

In the SV-COMP repository it would be nice to have one .yaml file per C file (or at some time C project) that describes all properties that could be checked on this file.

PhilippWendler:
> I think for multi-property verification, all tools have to adapt.

Yes, but only for multi-property verification, which probably not all tools will support. For existing use cases the current proposal is such that only the benchmark structure and BenchExec need to change, but no tool (this is a requirement for using SV-COMP'17 verifiers in SV-COMP'18).

> We currently use a python script as wrapper that converts .prp in the proper Ultimate settings. This will just get more complicated.
> [...]
> The arguments could just be given to the tool module and tool developers could then decide what to do with them.

This is of course always possible, regardless of how the metadata format changes. In fact, this is already possible with the existing system. After all, the tool-info module can read the property file and act accordingly.
However, I would advise against doing so. By putting this logic into the tool-info module instead of into the tool, it becomes complicated to run the tool on SV-Comp benchmarks without BenchExec. While we would certainly appreciate it, unfortunately not all paper authors use BenchExec when they execute experiments like comparing against a state-of-the-art tool, and if it becomes complicate to do this comparison, paper authors will make errors in their attempts. Thus I think it is in the best interest of all tool authors to make their tool as easy as possible to use with the SV-Comp benchmarks, both with and without BenchExec. For example, you wouldn't want to use paper authors to use a wrong configuration of your tool because they incorrectly interpreted the logic in the tool-info module.

(For CPAchecker, we found a simple and convenient solution: we have meta-configuration files (e.g., one for SV-Comp'16, one for SV-Comp'17, etc.) that define a mapping which actual configuration file should be used for which property. CPAchecker then first reads the meta-configuration and the property and chooses the appropriate configuration. This is easy to implement, flexible and extensible. If you are interested in learning more about this, just contact us.)

> we should avoid having a benchmark set where only a combination of .yaml, .prp, actual C input files, and benchexec's .xml files defines the set of verification tasks.

It would be great if you could write up a full description of how your proposal would look like like I did yesterday. This would make discussion much easier, I think.

As far as I see the situation, eliminating the property files requires either temporary files or putting important logic into the tool-info module. Both make it hard(er) to use the benchmark suite outside of SV-Comp/without BenchExec. Or did I overlook a solution that does not have this disadvantage?
For all other files, I think it is obvious that we need them, right?

Specifically for LTL verification, we assume that formulas are always passed to the tool as a file anyway, because putting potentially large formulas into a command-line argument can be impossible (some OS have quite small limits for command-line arguments).

I am also not quite sure why this proposal is a disadvantage for your use case (as far as I have understood it). If you have your own benchmark programs for tests, and you are not using BenchExec anyway, nobody forces you to use property and yaml files anyway. If you want to use the SV-Comp benchmark suite for tests, you can do so, but then the property and yaml files would exist anyway.

danieldietsch:
> If you want to use the SV-Comp benchmark suite for tests, you can do so, but then the property and yaml files would exist anyway.

This is the main issue. As soon as the changes discussed here are integrated into benchexec, the SV-COMP structure will change and I will be forced to use .yml files and .prp files to get the expected test result. Today, SV-COMP does not require .yml files and does not require parsing of .prp files to get expected results. It is enough to look at the file names (for testing / benchmarking with things other than benchexec). 
Essentially, I have to add .prp file parsing and .yml file parsing and resolution of files inside those things to Ultimate. 

> By putting this logic into the tool-info module instead of into the tool, it becomes complicated to run the tool on SV-Comp benchmarks without BenchExec. While we would certainly appreciate it, unfortunately not all paper authors use BenchExec when they execute experiments like comparing against a state-of-the-art tool, and if it becomes complicate to do this comparison, paper authors will make errors in their attempts
[...]
For example, you wouldn't want to use paper authors to use a wrong configuration of your tool because they incorrectly interpreted the logic in the tool-info module.

Paper authors should either try to replicate our setting and use benchexec or have to use the interface of the tool. They should never look at the things contained in the tool-info module. Anyways, this is a risk I am willing to take. 

Today, Ultimate's SV-COMP logic is already quite complicated because we do not (and cannot) support .prp files natively. We use a wrapper script for that. Which is wrapped by benchexecs tool-info wrapper. It is not an ideal situation, but it works. 

My proposal would be the same as above except: 
 * Do not use .prp files. Encode everything in one .yml file per input 
 * Change the `spec` parameter in the tool module. It is no longer a path to a .prp file but rather an array of .prp file contents (one entry for each property that should be checked, extracted from .yml and rundefinitions combination) 
 * (Optional) Provide a simple method that takes a .prp entry of this array, and returns a file (or just the absolute path) to a .prp file (created in the temporary folder of the system). 
 * Convention for SVCOMP: the .yml file has the same name as the input file but with the additional suffix .yml.

delcypher:
@PhilippWendler 
```yaml
input_files:
  - dummy.c
properties:
  - unreach-call_error1:
    - result: true
  - unreach-call_error2:
    - result: true
  - termination:
    - result: false
```

That should be this, no?

```yaml
input_files:
  - dummy.c
properties:
  unreach-call_error1:
    result: true
  unreach-call_error2:
    result: true
  termination:
    result: false
```

Verification task names are unique so `properties` should map to a dictionary, not a list. Similarly the properties of each verification task should be a dictionary, not a list.

delcypher:
@PhilippWendler 
> If tools need to parse yaml files, this is more complicated than the existing format. Consider for example tools written in C, or where the logic for handling SV-Comp tasks is implemented in bash scripts (this is surprisingly common). The authors of these tools would probably not be happy with a new format.

User's of bash can use utilities (e.g. [yaml-cli](https://github.com/pandastrike/yaml-cli)) to extract what they need from the YAML file.

For C. There are libraries for parsing YAML files. It's certainly not fun. C is a horrible language to work with 😛

PhilippWendler:
> That should be this, no?
> 
>     input_files:
>       - dummy.c
>     properties:
>       unreach-call_error1:
>         result: true
>       unreach-call_error2:
>         result: true
>       termination:
>         result: false
> 
> Verification task names are unique so properties should map to a dictionary, not a list. Similarly the properties of each verification task should be a dictionary, not a list.

Yes, of course. Sorry, I find the yaml syntax confusing and I never know when to put `-` and when not.

PhilippWendler:
> As soon as the changes discussed here are integrated into benchexec, the SV-COMP structure will change and I will be forced to use .yml files and .prp files to get the expected test result.

I think you won't need the property files just for this, if you care only about the SV-Comp tasks and their restricted set of standard properties, the yaml files should be enough. And you will only need this in the test runner, not in your tool itself.

> Today, SV-COMP does not require .yml files and does not require parsing of .prp files to get expected results. It is enough to look at the file names (for testing / benchmarking with things other than benchexec).

I have not yet seen any proposal how we could support features like verification of several properties, verification of (maybe several) arbitrary LTL specifications, and verification of programs with several input files without something like these yaml files for metadata. Even your own proposal does not get rid of the yaml files and the need for parsing them.

So as long as there is no alternative proposal that eliminates the need for metadata files completely, we have to accept the fact that we will need parsing of some metadata files in benchmarking frameworks and test runners. There is no way to continue putting information only in file names.

> Paper authors should either try to replicate our setting and use benchexec or have to use the interface of the tool. They should never look at the things contained in the tool-info module.

That is exactly my point.
However, the way I understood you, you would like to put logic of the form "if we have prop_A then use config_A.cfg, else ..." into the tool-info module. Then, paper authors need to understand this logic if they want to use your tool on the SV-Comp benchmark suite without BenchExec. This is what I suggest to avoid, but of course you might want to take that risk, and this is fine (neither the current state nor any of the proposals prevents you from doing so).

> My proposal would be the same as above except:
> 
>  -  Do not use .prp files. Encode everything in one .yml file per input
>  -  Change the spec parameter in the tool module. It is no longer a path to a .prp file but rather an array of .prp file contents (one entry for each property that should be checked, extracted from .yml and rundefinitions combination)

This point is a purely technical point that is completely independent from any of the proposals. BenchExec can (and probably will) implement this in any case, even if we would stay with the current state.

>  -  (Optional) Provide a simple method that takes a .prp entry of this array, and returns a file (or just the absolute path) to a .prp file (created in the temporary folder of the system).
>  -  Convention for SVCOMP: the .yml file has the same name as the input file but with the additional suffix .yml.

The last point won't be possible in general. There will be programs with multiple input files, and there will be input files with several yaml files (e.g., an input file that is parameterizable via `#define`s can be used for several verification tasks when non-preprocessed files are accepted).

Your proposal does not describe how the user can define which properties of those listed in the yaml file should be actually verified.

And the most important question that is missing in your proposal: How do you pass the properties to the tool?
I see 4 possibilities for doing this:
- command-line argument: unwieldy, can easily hit OS limits on command-line length
- temporary files: manual re-execution of verification runs becomes very tedious
- not pass property at all, only parse them in tool-info module: tool is difficult to use without BenchExec
- real (non-temporary) property file: requires an additional small number of property files (irrelevant compared to large number of yaml files), benchmarking frameworks and test runners may need to open two files instead of one

Of these 4 possibilities, my preference is clearly the last, because it is the only of these possibilities that does not have a disadvantage for users of the benchmark suite and verification tools, only for the people that write benchmarking frameworks and test runners (which should be only few people).

danieldietsch:
> I have not yet seen any proposal how we could support features like verification of several properties, verification of (maybe several) arbitrary LTL specifications, and verification of programs with several input files without something like these yaml files for metadata. Even your own proposal does not get rid of the yaml files and the need for parsing them.

I like the yml files. I am all for it. I just want to remove the .prp files. 

> The last point won't be possible in general. There will be programs with multiple input files, and there will be input files with several yaml files (e.g., an input file that is parameterizable via #defines can be used for several verification tasks when non-preprocessed files are accepted).

Programs with multiple input files should be put in a folder, the .yml file should specify the folder. 
With #defines, I do not see how any of the proposals would tackle this, except if the .yml files would also specify some preprocessing step. Then, it would become the same as multiproperty. 

> Your proposal does not describe how the user can define which properties of those listed in the yaml file should be actually verified.

You are right. One needs an additional name for a property. Probably something like 
````
properties:
    name-of-property:
    property: CHECK( init(main()), LTL(G ! call(error_function_1())) )
    result: true
...
````
The name of the property is then used in the rundefinition instead of the path to the .prp file. 

> command-line argument: unwieldy, can easily hit OS limits on command-line length

I personally never encountered such a limit, but I guess it could happen. 
For example, Ultimate currently generates the following commandline for some arbitrary task 
````
java -Xmx12G -Xms1G -jar /storage/ultimate/releaseScripts/default/UAutomizer-linux/plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar -data /storage/ultimate/releaseScripts/default/UAutomizer-linux/data -tc /storage/ultimate/releaseScripts/default/UAutomizer-linux/config/AutomizerReach.xml -i ../../../trunk/examples/svcomp/loop-acceleration/array_true-unreach-call3_true-termination.i -s /storage/ultimate/releaseScripts/default/UAutomizer-linux/config/svcomp-Reach-32bit-Automizer_Default.epf --cacsl2boogietranslator.entry.function main --witnessprinter.witness.directory /storage/ultimate/releaseScripts/default/UAutomizer-linux --witnessprinter.witness.filename witness.graphml --witnessprinter.write.witness.besides.input.file false --witnessprinter.graph.data.specification CHECK( init(main()), LTL(G ! call(__VERIFIER_error())) )

 --witnessprinter.graph.data.producer Automizer --witnessprinter.graph.data.architecture 32bit --witnessprinter.graph.data.programhash 4090a90876c90ba415ea4bea59e2b23309410708
````
The line break is intentional. We just pass it through Python and I do not know how Python handles this; it just does. 

Real .prp files 
 * increase maintenance, 
 * introduce an additional possibility for errors: you have to check consistency between multiple files when creating benchmarks,
 * are a burden on tool developers, which are as far as I understand it, the majority of the benchexec users, and 
 * are not easily extendable (which is why this discussion exists) 

As far as I understand it, .prp files are only here for legacy reasons. They are simply the format CPAChecker used since more or less the beginning of SVCOMP and benchexec.  
Now is a chance to replace it by a more convenient format with a lot less hassle.

**Comment by PhilippWendler** [comment]
> With #defines, I do not see how any of the proposals would tackle this, except if the .yml files would also specify some preprocessing step. Then, it would become the same as multiproperty.

Yes, specifying `#define` values in the yaml files is one of the further proposals in the pipeline.

> > Your proposal does not describe how the user can define which properties of those listed in the yaml file should be actually verified.
> 
> You are right. One needs an additional name for a property. Probably something like
> 
>     properties:
>         name-of-property:
>             property: CHECK( init(main()), LTL(G ! call(error_function_1())) )
>             result: true

In this proposal, `name-of-property` can be an arbitrary string, right?
This is a disadvantage, because now it is very easy to loose consistency between all the yaml files in a benchmark suite. There could be some verification tasks with a property name `unreach-call`, some with `unreach`, some with `unreach-call(__VERIFIER_error)`, and this would for example make result tables (where we need to show these property names) inconsistent. Using property files is a way to encourage people to use consistent property names (without restricting the flexibility to deviate if desired), and I think we agree that consistency is good to have.


>> command-line argument: unwieldy, can easily hit OS limits on command-line length
> 
> I personally never encountered such a limit, but I guess it could happen.

https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string

You didn't say anything regarding temporary files or handling in the tool-info module, so I assume that you agree that we don't want to rely on these two possibilities?

> Real .prp files
> 
>  -   increase maintenance,

Maybe a little bit, but I guess that compared to the thousands of files that we already have, and the thousands of yaml files that we are going to introduce, keeping the dozen of property files won't be a big problem.

>  -   introduce an additional possibility for errors: you have to check consistency between multiple files when creating benchmarks,

As explained above, this gets worse without property files: With property files, you need to have consistency between each yaml file and the few central property files. If you get it wrong, BenchExec will throw an error. Without property files, there should still exist consistency between all these yaml files, without a central place that defines the canonical name, and without automatic inconsistency checks.

>  -   are a burden on tool developers, which are as far as I understand it, the majority of the benchexec users, and

If you mean developers of verifiers, I do not understand this argument. Verifiers always need to provide a way to specify the to-be-verified property. How are property files any more difficult for verifier developers than any of these other 3 possibilities I listed?

>  -   are not easily extendable (which is why this discussion exists)

I see no difference in extendability between the proposal of yaml+property files and just yaml files.

> As far as I understand it, .prp files are only here for legacy reasons. They are simply the format CPAChecker used since more or less the beginning of SVCOMP and benchexec.

That's not true at all, where did you get this impression? Property files are not specific to CPAchecker at all, and CPAchecker supports them precisely because SV-Comp introduced them. BenchExec also implements handling of property files because of SV-Comp.

This of course means that property files can be removed if they are no longer needed, and I would absolutely be in favor of removing them, but all alternative solutions proposed so far have worse disadvantages than the property files IMO.

**Comment by danieldietsch** [comment]
> In this proposal, name-of-property can be an arbitrary string, right?
This is a disadvantage, because now it is very easy to loose consistency between all the yaml files in a benchmark suite. There could be some verification tasks with a property name unreach-call, some with unreach, some with unreach-call(__VERIFIER_error), and this would for example make result tables (where we need to show these property names) inconsistent. Using property files is a way to encourage people to use consistent property names (without restricting the flexibility to deviate if desired), and I think we agree that consistency is good to have.

Yes, this is indeed a disadvantage assuming we only care for the 5 or so hardcoded properties used in SV-COMP. 
Perhaps this is our main misunderstanding. I think about more properties because I have our LTL benchmarks in mind and multiple error functions for multi-property verification. 
For example, for our LTL alone each ltl-eca example has 50 different properties. The coolant example is similar: its actually only two or three different files with 8 or so properties each. 
Now, these properties are truly program-dependent because they use expressions over the variables of the program, so there can be no reuse. 

I have no good proposal that definitely solves the naming problem. But property files do not solve this either. Only convention will keep people from renaming .prp files. But you still have to care for the right paths in your rundefinition file so that benchexec can pick up the properties. 
You do not have to do that if you do not use filenames as shortnames and paths that end in filenames (but without .prp) as paths and as matching anchor for the selection of a certain subset of properties. 

> > > command-line argument: unwieldy, can easily hit OS limits on command-line length
> >
> > I personally never encountered such a limit, but I guess it could happen.
> 
> https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string

So, basically you care that for a platform you will likely never support (Windows) were the limit is very high (32768 chars in Windows 7, unclear in Windows 10)? That does not seem like a very good argument against command-line usage. 

> You didn't say anything regarding temporary files or handling in the tool-info module, so I assume that you agree that we don't want to rely on these two possibilities?

I think there are only two possibilities (use .prp files or not) and handling in the tool-info module is required for both. Temporary files is something that one could use to retain backwards compatibility by creating .prp files from the benchexec input in the tool module. 

> Maybe a little bit, but I guess that compared to the thousands of files that we already have, and the thousands of yaml files that we are going to introduce, keeping the dozen of property files won't be a big problem.

See above. Perhaps this is the main misunderstanding. I expect a lot of .prp files because of LTL and differently labelled error functions for multi-property verification. One should expect that, because we will get more and more properties, and those will be more complex than the old ones. 

> Without property files, there should still exist consistency between all these yaml files, without a central place that defines the canonical name, and without automatic inconsistency checks.

I am not sure. The consistency check could enforce the same short name for all hardcoded properties by just collecting all the .yml files, creating a dictionary from property to shortname and enforcing that the shortname is unique for each unique property. It could also enforce this for some regex that captures kinds of properties (reachability with different function names, LTL).
It would actually be easier and more reliable because you do not have to open the .prp file to see whether someone changed the content and kept the name. 

> >are a burden on tool developers, which are as far as I understand it, the majority of the benchexec users, and
> 
> If you mean developers of verifiers, I do not understand this argument. Verifiers always need to provide a way to specify the to-be-verified property. How are property files any more difficult for verifier developers than any of these other 3 possibilities I listed?

I mean they force developers to support something they may not want to support or cannot support. I can only imagine that many other tools do not really care for having arbitrary files that define their behavior. And if they have files, they usually have one config file, not many. Having .prp files just forces them to adopt this format for a very small -- albeit important -- aspect of a verifier tool configuration.

> > are not easily extendable (which is why this discussion exists)
>
> I see no difference in extendability between the proposal of yaml+property files and just yaml files.

You are right. If we have the .yml files we can just extend these and keep the .prp files around because they are already around. 

> > As far as I understand it, .prp files are only here for legacy reasons. They are simply the format CPAChecker used since more or less the beginning of SVCOMP and benchexec. 
>
> That's not true at all, where did you get this impression? Property files are not specific to CPAchecker at all, and CPAchecker supports them precisely because SV-Comp introduced them. BenchExec also implements handling of property files because of SV-Comp.

Oh, ok. I probably just assumed it was like that. Because Dirk started the SV-COMP and already had a tool, and the .prp files also came from there. 

I can understand that there is a lot of infrastructure that currently depends on .prp files. However, for these cases, one can easily generate the .prp file. There will be backwards compatibility and there will be the benefit of easier maintenance. If you only care about the dozen or so hardcoded .prp files, you can just auto-generate them with different names (they will have unique names) to a fixed place, so you can rerun things that depend on .prp files easily. 
Cleanup is also easy: just delete all .prp files. 

I actually cannot see anything that would be worse by removing .prp files, I see only benefits. If you are still convinced everything will be in a much worse state than with them, can you please restate why you think that?

**Comment by PhilippWendler** [comment]
> Perhaps this is the main misunderstanding. I expect a lot of .prp files because of LTL and differently labelled error functions for multi-property verification. One should expect that, because we will get more and more properties, and those will be more complex than the old ones.

Yes, I did not expect a ratio of 1:8 or even 1:50 for program to property files. If this really happens, it is unfortunate that so many files are needed, but I still think the disadvantages of the alternatives are worse.

> So, basically you care that for a platform you will likely never support (Windows) were the limit is very high (32768 chars in Windows 7, unclear in Windows 10)? That does not seem like a very good argument against command-line usage.

Well, BenchExec will not support Windows, but why shouldn't verification tools support Windows? And even on Linux, if I have to pass these LTL strings to tools on the command line when I am manually using a verification tool, I would certainly not find that user friendly. Just imagine you are going to experiment with or debug a tool. Do you really always want to open the yaml file in a text editor and copy'n'paste the LTL string to the command line, instead of just pointing the verifier to a certain file?

> >> are a burden on tool developers, which are as far as I understand it, the majority of the benchexec users, and
> 
>  > If you mean developers of verifiers, I do not understand this argument. Verifiers always need to provide a way to specify the to-be-verified property. How are property files any more difficult for verifier developers than any of these other 3 possibilities I listed?
> 
> I mean they force developers to support something they may not want to support or cannot support. I can only imagine that many other tools do not really care for having arbitrary files that define their behavior. And if they have files, they usually have one config file, not many. Having .prp files just forces them to adopt this format for a very small -- albeit important -- aspect of a verifier tool configuration.

I still do not understand this. Nobody forces tools to use config files, that was just an explanation of our solution, which I find easy and convenient. "files that define their behavior" are already used by all existing verifiers, after all they read the to-be-verified program from a file. How is the property different from the program? Handling program and property similar would actually be a nicer interface than handling them differently, I would argue.

> I can understand that there is a lot of infrastructure that currently depends on .prp files.

I never made that argument, and I am not saying that we need property files for backwards compatibility, I am saying that I see no better solution.

> I actually cannot see anything that would be worse by removing .prp files, I see only benefits. If you are still convinced everything will be in a much worse state than with them, can you please restate why you think that?

I think I have explained this several times already: The removal of property files forces us to either pass long LTL strings on the command line, or to use temporary files, or to put significant logic in the tool-info module. All these possibilities have significant drawbacks for the users of verification tools, which outweigh the small drawbacks of property files for the implementors of benchmarking frameworks. No alternative has been suggested so far that avoids these drawbacks.

Please, make a clear proposal on what in your opinion should be the recommende way to pass property information from BenchExec to tools in the future, i.e., what kind of interface should tool authors implement for next SV-Comp?

**Comment by danieldietsch** [comment]
I, as a tool provider, do not want to use .prp, .yml, or .xml files as interface for my users. 
Ultimate also does not use contents of .prp files to specify which property should be checked. Ultimate provides a wrapper script that is used by the Ultimate benchexec tool to be able to participate in SV-COMP. 
I am in favour of passing long LTL strings on the command line. But I want to allow other tool developers the flexibility of using their very own user interface.

Storage of verification task metadata should -- in my opinion -- happen at just one place. 
Telling different tools what they have to verify was a matter of telling them "use this input file and that property". The properties were stored in files. 
Until now, we had only some low number of different properties and expected results. We also did not need any additional metadata. 

This changes with the introduction of LTL, multi-property verification, and possibly special preprocessor directives. Therefore, we need more and more complex metadata. 
I expect that in the future we want to tell a tool "use this set of input files, this set of header files, these includes, these preprocessor flags, and then verify properties p_1 to p_n where you should treat p_i as a property that consists of three sub-properties and if any is violated, you do not need to tell me the result of the other two".

You asked for a clear proposal. Here it is.  Note that this proposal only wants to deal with the scenario in which there is one input file and multiple properties, but can be easily extended to hold other metadata as well. 

#### Changes to SV-COMP rules 
Lets look at the SV-COMP 17 rules.
SV-COMP 17 (cf. [rules section](https://sv-comp.sosy-lab.org/2017/rules.php)) uses as interface to the tools the following two descriptions 
> Properties
>
> The specification to be verified for a program "path/filename" is given either in a file with the name "path/filename.prp" or in a file "Category.prp".
> 
> [...]
> 
> Parameters
>
> The competition environment feeds the candidates with parameters when started for a verification run. There is one global set of parameters that can be used to tune the verifier to the benchmark programs. This set of (command-line) parameters have to be defined in the competition contribution paper. One parameter defines the specification file. There are categories that need to be verified with the information that a 64-bit architecture is used, thus, if the verifier has a parameter to handle this aspect, it needs to be defined. There are also categories for which an additional parameter is allowed to specify the simple memory model.

I propose to change the first part from 
> The specification to be verified for a program "path/filename" is given either in a file with the name "path/filename.prp" or in a file "Category.prp".

to

> The specification to be verified for a program "path/filename" is given as a parameter of the competition environment.

In the second part I would change 
> One parameter defines the specification file.

to 

> One parameter defines the specification.

#### Metadata storage and verification task generation  

One will have multiple ``<propertyname>`` tags in the benchmark XML and the tools will get multiple properties on the command line. The metadata file for a benchmark will look like this: 

Example:
sources/dummy.yml:
````yaml
input_files:
  - dummy.c
architecture:
    64bit
properties:
    unreach-call_error1:
        property: 
            CHECK( init(main()), LTL(G ! call(error_function_1())) )
        result: true
    unreach-call_error2:
        property: 
            CHECK( init(main()), LTL(G ! call(error_function_2())) )
        result: true
    termination:
        property: 
            CHECK( init(main()), LTL(F end) )    
        result: false
    memsafety:
        property: 
            CHECK( init(main()), LTL(G valid-free) )
            CHECK( init(main()), LTL(G valid-deref) )
            CHECK( init(main()), LTL(G valid-memtrack) ) 
        result: true
    ltl1
        property:
            CHECK( init(main()), LTL((G "x==1")) )     
        result: true
````

Benchmark XML:
````xml
<tasks><include>sources/dummy.yml</include></tasks>
<property>memsafety<property>
<property>termination<property>
````

Example tool command line:
````
verifier --property memsafety="CHECK( init(main()), LTL(G valid-free) ),CHECK( init(main()), LTL(G valid-deref) ),CHECK( init(main()), LTL(G valid-memtrack) )" --property termination="CHECK( init(main()), LTL(F end) )" --architecture 64bit --input dummy.c
````
Note that tool developers can implement their own command line passing method. They can even generate new configuration files on the fly. This allows for greater flexibility in tool user interfaces than before. 

Example tool output:
````
Result for memsafety: true
Result for termination: false
````

#### Benefits 
 * Only one place to store all the metadata for a single file or project
 * Easier maintenance 
 * Other test runners do not need to replicate complicated .yml / .prp relation (no need to resolve paths, match filenames, etc.) 
 * Tool developers can use their preferred interface without being forced to read additional configuration files. 
* Easier consistency checks: a consistency check can enforce the same short name for all hardcoded properties by just collecting all the .yml files, creating a dictionary from property to shortname and enforcing that the shortname is unique for each unique property. It could also enforce this for some regex that captures kinds of properties (reachability with different function names, LTL).
This is easier than a consistency check for .prp files were someone would need to open the .prp file to see whether someone changed the content and kept the name.

#### Arguments brought against it 
> passing long LTL strings may hit a limit on certain plattforms 

So far, only Windows has a rather high limit (32k chars). Windows cannot be supported by benchexec. 
> tool users want to use .prp files 

Our users do not want to use .prp files. They are too restrictive. They prefer our command line switches. 
> we do not want to have complicated logic for the interface between tool and benchexec

We already have that logic. It will not change.
> researchers cannot replicate because it is too hard to understand the necessary logic in the tool module if we pass .prp contents around instead of paths to .prp files

We already have such logic, but in wrapper scripts. Researchers should either use benchexec to replicate findings made with benchexec (or else, whats the point) or read the system description that explains how a tool is called. 
> It is harder to keep short names of properties consistent without the "tax" benchmark providers have to pay if they add new .prp files with the same content as already existing ones. It is hard to implement automatic consistency checks. 

It is actually easier. See above.

**Comment by PhilippWendler** [comment]
Thanks for explaining your proposal. It seems we disagree of the relative weighting of the (dis)advantages of the two proposals. I have asked @dbeyer for comments, and I would like to hear feedback from other people as well.

Just one last note: It does not matter whether we remove property files or not, you will always be able to pass command-line strings with LTL to Ultimate. As I said already several times, you can already do so right now. Just send us a pull request that changes the tool-info module to put the content of the property file into a command-line argument. So this is not an argument against property files.

**Comment by PhilippWendler** [comment]
@kfriedberger has found a solution that I think is good.

Because the two proposals differ only in the handling of property files, we should first implement everything else (i.e., adopt the yaml files), and do not change anything regarding the property files. Then, when we have an implementation, we can gain experience and see whether property files provide an advantage or a disadvantage in practice. If we find it is the latter, it will be a simple change to remove property files: Just delete them and put their content into the `property` field of the yaml files.

**Comment by danieldietsch** [comment]
I just realized another problem with the .prp files regarding SV-COMP. 
Currently, benchmark maintainers provide .prp files for their benchmark set. I presume in the future they should do that as well. But because the matching between .yml files and .prp files relies on the filename (rather than the path) of the property file, one has to make sure that all .prp files have a unique name.
For LTL this means we have to add numbers or input filenames or something to all of the .prp files. 
If someone wants to add new benchmarks he has to make sure that he does not impact the naming of all the other .prp files.

**Comment by vmordan** [comment]
There are several questions about this suggestion.

> Our suggestion is to use the file name of the property file (without directory and extension) as short identifier.

1. Do we need to support situation, in which property file (short identifier) contains several properties (LTL strings)? For example:

```
property_file_1:
  ltl_string_1
  ltl_string_2
```
and
```
property_file_2:
  ltl_string_3
```

2. What kind of result is expected from verification tool in this case - status for each property file (property_file_1 is considered as `composite`), i.e.
```
Result for 'property_file_1': false|true|unknown
Result for 'property_file_2': false|true|unknown
```
or status for each LTL string, i.e.
```
Result for 'ltl_string_1': false|true|unknown
Result for 'ltl_string_2': false|true|unknown
Result for 'ltl_string_3': false|true|unknown
```
3. Do we need full backward compatibility?
Should we consider single property verification as a special case of multi-property verification or not?
For example, do we need to keep `false-unreach-call` in file names instead of yaml files? 
Also do we need to specify short property names in `false` verdicts, for example `false(unreach-call)`:
```
Result for 'ltl_string_1': false(property_file_1)|true|unknown
```

**Comment by danieldietsch** [comment]
I would think:
1. Yes
2. I would say one property file has one result (as it is the case with memory safety), so ``composite``
3. As I understand it, we want to get rid of the results in the filenames. I dont know about the short names.

**Comment by PhilippWendler** [comment]
1. No (except for the existing MemSafety property)
2. One result per property file
3. Backwards compatibility should be kept in the way that it is possible to execute for example SV-Comp'17 benchmark definitions with a new BenchExec.

**Comment by vmordan** [comment]
> No (except for the existing MemSafety property)

Currently there are 3 different false statuses in category `MemSafety`. 
Do we need to return 3 different results for this property:

```
Result for 'valid-free': false|true|unknown
Result for 'valid-deref': false|true|unknown
Result for 'valid-memtrack': false|true|unknown
```

vmordan:
If several property files are considered, then tool-info module interface should be changed (`propertyfile` in [`cmdline`](https://github.com/sosy-lab/benchexec/blob/master/benchexec/tools/template.py#L113)), which will affect almost all tool-info modules. Are you agree with this?

PhilippWendler:
No, we need to keep backwards compatibility with existing tool-info modules. I am thinking about how to handle this in the best way.

vmordan:
I added backward compatibility for single-property verification in the following way. 
If only one property is specified, then `propertyfile` is a string in `cmdline` (i.e., tool-info module interface is the same).
If several properties are specified, tool-info module should be changed to support new interface.
Currently it supports both single-property verification (existing tool-info modules) and multi-property verification (CPAchecker branches [cmav-2](https://svn.sosy-lab.org/software/cpachecker/branches/cmav-2/) and [muauto](https://svn.sosy-lab.org/software/cpachecker/branches/muauto/)).

PhilippWendler:
This solution has the problem that is functionally backwards-compatible, but if a user specifies multiple properties for a tool which does not support it (and whose tool-info module has not been updated accordingly), it will produce something strange instead of a clear error message. Furthermore, all tools that support multiple properties now need this `isinstance` check instead of just being able to handle an arbitrary number of properties.

I have a different idea of how to support this and I am going to implement it soon (the same idea will also be usable to fix #166).

mutilin:
@PhilippWendler
Any news on this?

PhilippWendler:
To sum up yesterday's discussion: Support for tasks defined in Yaml files has been independently added to BenchExec in the meantime, so the support for multi-property verification would need to be implemented on top of that, which should simplify it. Probably a new PR is better, though, so I am closing this.

Another suggestion: For a first iteration I would avoid adding a parameter to the `<propertyfile>` tag. We could simple say that a single file with several properties defines composite property, and for multi-property verification each property has to be in a separate file.

PR #234, opened 2017-03-13
Preparing for LTL demo category

danieldietsch:
Let benchexec support valid-ltl property (similar to valid-deref). 

One issue: LTL properties allow arbitrary formulas inside the LTL part of a check annotation. Without changing the matching code in result.py matching may become brittle. 

Intention would be: If nothing else matches and LTL is present, assume valid-ltl property.

### DIFF
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -18,15 +18,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import functools
 import logging
+import os
+import re
 import subprocess
+import sys
 
 import benchexec.result as result
 import benchexec.tools.template
 import benchexec.util as util
-import functools
-import os
-import re
 from benchexec.model import MEMLIMIT
 
 _OPTION_NO_WRAPPER = '--force-no-wrapper'
@@ -47,6 +48,7 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         "config",
         "configuration",
         "cvc4",
+        "cvc4nyu",
         "cvc4-LICENSE",
         "features",
         "LICENSE",
@@ -89,33 +91,34 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         launcher_jar = self._get_current_launcher_jar(executable)
 
         cmds = [
-            ["java", "-jar", launcher_jar, "-data", "@noDefault", "-ultimatedata", data_dir, "--version"],
-            ["java", "-jar", launcher_jar, "-data", data_dir, "--version"],
+            ["java", "-jar", launcher_jar, "-data", "@noDefault", "-ultimatedata", data_dir, "--version"],  # 2
+            ["java", "-jar", launcher_jar, "-data", data_dir, "--version"],  # 1
         ]
 
-        api = len(cmds)
+        self.api = len(cmds)
         for cmd in cmds:
-            version = self._query_ultimate_version(cmd, api)
+            version = self._query_ultimate_version(cmd, self.api)
             if version != '':
                 return version
-            api = api - 1
-        return ''
+            self.api = self.api - 1
+        raise NameError("Could not deterime Ultimate version")
 
     def _query_ultimate_version(self, cmd, api):
         try:
             process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             (stdout, stderr) = process.communicate()
         except OSError as e:
-            logging.warning('Cannot run Java to determine Ultimate version (API {0}): {1}'.format(api, e.strerror))
+            logging.warning('Cannot run Java to determine Ultimate version (API {0}): {1}'
+                            .format(api, e.strerror))
             return ''
         if stderr:
-            logging.warning('Cannot determine Ultimate version, error output (API {0}): {1}'.format(api,
-                                                                                                    util.decode_to_string(
-                                                                                                        stderr)))
+            logging.warning('Cannot determine Ultimate version (API {0}). Error output: {1}'
+                            .format(api, util.decode_to_string(stderr)))
             return ''
         if process.returncode:
             logging.warning(
-                'Cannot determine Ultimate version, exit code (API {0}): {1}'.format(api, process.returncode))
+                'Cannot determine Ultimate version (API {0}). Exit code : {1}\nCommand was {2}'
+                    .format(api, process.returncode, ' '.join(cmd)))
             return ''
 
         version_ultimate_match = _ULTIMATE_VERSION_REGEX.search(util.decode_to_string(stdout))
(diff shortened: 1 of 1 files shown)


danieldietsch:
Should I change something with this pull request or is there a reason for not merging it?

PhilippWendler:
My main issue is with the statement "If nothing else matches and LTL is present, assume valid-ltl property.".

First, this means that any typo or other change in property files would lead BenchExec to silently treat them as a valid-ltl property. For example, if someone writes `LTL(F end)` this is interpreted as termination property by BenchExec, whereas `LTL( F end )` is interpreted as valid-ltl property. This is not robust. Even if we make matching of existing properties more flexible, this does not help the case where accidental typos like `LTL(F en)` are made, which would better be reported as an error instead of being treated as valid-ltl property.

Second, a similar problem arises if we add some new specific properties in the future. For example, if we introduce `LTL(foo)` for property "foo", this changes the semantics of `LTL(foo)` from valid-ltl property to foo property.

In general, it is better to never have the case where something can either be one of a few specific hard-coded values, or a completely arbitrary string, without the possibility to distinguish between these two cases. So we should use different syntax for these two cases, though I do not have a concrete proposal for the valid-ltl case. Another way would be to remove the hard-coded properties completely, but this would mean that we cannot use any names like `unreach-call` anymore and need to switch completely from having verdicts in file names to verdicts in yaml files.

PhilippWendler:
Oh, and is it really good to use a generic name like `valid-ltl` for this property? This would mean that you can never have two different LTL properties for the same program file, because there would be no possibility to distinguish between their expected verdicts. And if a user accidentally specifies a wrong property file for a task, we could not detect this. Maybe it is better to avoid this and accept only verdicts specified in yaml files as discussed in https://github.com/sosy-lab/sv-benchmarks/issues/213, https://github.com/sosy-lab/benchexec/issues/217, and https://github.com/sosy-lab/benchexec/pull/226 for LTL properties, and have the full property be repeated in the yaml file such that BenchExec can ensure that the correct expected verdict is used? Maybe this also means that we need to discuss again how the yaml format should look like to support arbitrary LTL as property names?

Or should there be a way for the user to assign a name for a specific LTL property, and then BenchExec uses this name like it uses the existing hard-coded names like `unreach-call`?

danieldietsch:
One important aspect (in particular for @dbeyer)  was that we stay backwards-compatible. 
You are right, of course, with regard to LTL being the "default" case. I think the only way would be to change the prefix (``LTL``) to something different and unique. It is easily doable and preserves backwards compatibility. I am open to this. What do @dbeyer, @Heizmann, @Langenfeld say?

Regarding ``valid-ltl``: I don't see any different solution that can also be backwards-compatible. In particular, we can only encode so much properties in the filename. 
I agree that it would be nice to have this, and especially for LTL. If you look at the sanity checks in the other pull request (https://github.com/sosy-lab/sv-benchmarks/pull/458), you see that a lot of input files have the same content, simply because the property specified by the matching .prp file is different. 
I would say we accept this shortcoming for now. 
Keep in mind that we would like to run the demo category before we meet at TACAS. I do not think we can agree on a good format before then. 
My hope is that this demo category will show the need for a more flexible format in the future.

Heizmann:
The need for an LTL syntax that does not hide typos in the classical properties is indeed an annoying problem that I overlooked. 
I still cannot see a simple solution here. I would propose that we change our keyword from ``LTL`` to ``INDIVIDUAL-LTL``.

I also think that we should use the keyword ``valid-ltl`` in filenames until we switch to verdicts in yaml files.

PhilippWendler:
Sorry for the delay, we have now discussed this further, in combination with the request for multi-property verification (#226). We assume that this would also be interesting for LTL properties, so we need to find a solution that supports both features at the same time.

The current proposal (cf. https://github.com/sosy-lab/benchexec/pull/226#issuecomment-300243909) would take care of support for LTL verification automatically, because BenchExec would no longer need to parse property files. It would also allow users to give a name to each specific property, and this name will be used for example in tables (instead of just `valid-ltl`). What do you think of this?

As for the distinction between the specific properties defined by SV-Comp and other, arbitrary properties, it might still be a good idea, at least as a convention. Tools can use this for parsing and better error messages, for example.
Our idea here would be to simply wrap all non-hardcoded properties in an extra set of parentheses, i.e., `LTL((some arbitrary formula))`, but `LTL(F end)`.

danieldietsch:
I just commented on #226 ([comment](https://github.com/sosy-lab/benchexec/pull/226#issuecomment-300259490)). I like the basic idea: make benchexec property-agnostic. 

But I would still prefer to separate LTL from built-in keywords clearly. I would like to introduce the keyword BUILTIN(...) for the old, non-LTL properties and LTL for the LTL properties. 

Double parenthesis seem like a very unnatural way for this separation.

I just noticed that our proposal does not even talk about parenthesis although we use them in our properties. Our syntax is not
> \phi := "P" 
>          | \phi || \phi 
>          | \phi && \phi 
>          | !\phi | G \phi 
>          | F \phi 
>          | \phi U \phi 
>          | \phi W \phi 
>          | \phi R \phi
> where \phi is a temporal formula and "P" is an atomic proposition.

but rather
> \phi := "P" 
>          | \phi || \phi 
>          | \phi && \phi 
>          | !\phi | G \phi 
>          | F \phi 
>          | \phi U \phi 
>          | \phi W \phi 
>          | \phi R \phi
>          | (\phi)
> where \phi is a temporal formula and "P" is an atomic proposition.

PhilippWendler:
> I just commented on #226 (comment). I like the basic idea: make benchexec property-agnostic.

Good :-)

> But I would still prefer to separate LTL from built-in keywords clearly. I would like to introduce the keyword BUILTIN(...) for the old, non-LTL properties and LTL for the LTL properties.
> 
> Double parenthesis seem like a very unnatural way for this separation.

At first glance maybe yes, I should have explained this a little bit more.
We noticed that the existing property format would be the most natural approach for specifying arbitrary LTL properties, but if we do this and introduce a different keyword `BUILTIN(...)`, this whole proposal becomes backwards-incompatible. This would for example mean that an SV-COMP'17 verifier could not be used directly in SV-COMP'18 (something that SV-COMP has recently declared it might want to do). Introducing a new keyword for the new properties (like `INDIVIDUAL-LTL`) would look awkward.

The double parentheses would have the following advantages:
- Fully backwards compatible.
- Forwards compatible: If we decide in the future that we want to drop the distinction between predefined and arbitrary properties, or remove all predefined properties, this would just extend the set of valid properties, but all existing properties keep the same semantics (if parentheses are allowed).
- No new potentially unclear keyword.
- Tools can distinguish between predefined and arbitrary properties, but if they don't want to, they can treat them the same.

However, because it looks like that this decision will be irrelevant for BenchExec and more important for the SV-Comp community, I would suggest to move the discussion of this point to the sv-benchmarks repository, maybe to https://github.com/sosy-lab/sv-benchmarks/pull/458?

danieldietsch:
Ok, I understand. 

But regarding LTL: Our benchmarks would then change to only one file, but as many (.yaml/.prp) tuples as single ltl verification tasks. Or can we somehow specify one .yaml file with all the properties for the file, and in the rundefinition we make file / .prp tuples?

PhilippWendler:
If you have one C file with several LTL properties, you would have the following structure:
- the C file
- one .prp file for each LTL formula
- one yaml file with the expected result for each property (note that you could also use several yaml files instead)

This is similar to your original proposal, right?

In the benchmark XML file, you would then specify the yaml file and the set of property files you want to use. You can also specify several yaml files, for which the same properties will be used, or several yaml files, each with their own property file(s), though the latter case is probably more effort to write. However, I don't see a way to make this simpler without loosing some features.

danieldietsch:
Sounds good. I think from the LTL perspective, this is something we can get behind. 

Should we keep this pull request open and wait until the benchexec changes are completed and then update it?

PhilippWendler:
> Sounds good. I think from the LTL perspective, this is something we can get behind.

Nice. Thanks for your feedback!

> Should we keep this pull request open and wait until the benchexec changes are completed and then update it?

Yes, this makes sense.

Issue #220, opened 2017-01-22
Fix paths to input files in result XML


PhilippWendler:
The paths to the input files for a run in the result XML file of `benchexec` are wrong, they are relative to the current directory where `benchexec` got executed, but they should be (as all relative paths!) relative to the directory of the file in which they are present.

This is easy to reproduce with `benchexec doc/benchmark-example-true.xml`. In `results/benchmark-example-rand.*.results.xml.bz2` the paths in `<run files="..." name="...">` will start with `doc/`, but they should start with `../doc/`.

The responsible line is probably 285 in `benchexec/outputhandler.py`.
We should also investigate what happens for tasks specified with `<withoutfile>`.
And we should check whether and how a test for this can be written.

PhilippWendler:
Fixed by #240.

PR #240, opened 2017-05-09
Fix path handling in benchexec run result xml files

leostrakosch:
According to #220 , path references in result xmls are fixed.
This influences the <run> tags in created xml files.

1. Files of run tasks are correctly referenced in the result xml.
2. Run tasks that are specified `<withoutfile>` do not point to any file anymore.
Previously, the text inside the `<withoutfile>` tag, e.g. for `<withoutfile>dummy</withoutfile>` the text `dummy`, was used as a source file in the task list and result xml. Now no sourcefile is referenced.
3. A small integration test to check this is added, which checks the column names of the run result, the run information inside the `<run>` tags and the values of columns `status`, `category`, `exitcode` and `returnvalue` against the expected results.
\
I'm open for proposals to perform this test in a more elegant way.

### DIFF
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -309,8 +309,7 @@ class OutputHandler(object):
         for run in runSet.runs:
             run.resultline = self.format_sourcefile_name(run.identifier, runSet)
 
-            adjusted_sourcefiles = [util.relative_path(s, xml_file_name) for s in run.sourcefiles]
-            if adjusted_sourcefiles:
+            if run.sourcefiles:
                 adjusted_identifier = util.relative_path(run.identifier, xml_file_name)
             else:
                 # If no source files exist the task doesn't point to any file that could be downloaded.
@@ -318,8 +317,11 @@ class OutputHandler(object):
                 adjusted_identifier = run.identifier
 
         # prepare XML structure for each run and runSet
-            run.xml = ET.Element("run",
-                                 {"name": adjusted_identifier, "files": "[" + ", ".join(adjusted_sourcefiles) + "]"})
+            run_attributes = {'name': adjusted_identifier}
+            if run.sourcefiles:
+                adjusted_sourcefiles = [util.relative_path(s, xml_file_name) for s in run.sourcefiles]
+                run_attributes['files'] = '[' + ', '.join(adjusted_sourcefiles) + ']'
+            run.xml = ET.Element("run", run_attributes)
             if run.specific_options:
                 run.xml.set("options", " ".join(run.specific_options))
             if run.properties:
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -310,7 +310,13 @@ class OutputHandler(object):
             run.resultline = self.format_sourcefile_name(run.identifier, runSet)
 
             adjusted_sourcefiles = [util.relative_path(s, xml_file_name) for s in run.sourcefiles]
-            adjusted_identifier = util.relative_path(run.identifier, xml_file_name)
+            if adjusted_sourcefiles:
+                adjusted_identifier = util.relative_path(run.identifier, xml_file_name)
+            else:
+                # If no source files exist the task doesn't point to any file that could be downloaded.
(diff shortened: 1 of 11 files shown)


benchexec/test_benchexec_integration.py
```diff
@@ -27,11 +27,15 @@
...
 import zipfile
+
+from xml.etree import ElementTree
+
 sys.dont_write_bytecode = True # prevent creation of .pyc files
 
 here = os.path.dirname(__file__)
 base_dir = os.path.join(here, '..')
 bin_dir = os.path.join(base_dir, 'bin')
+tmp_dir = os.path.join(here, 'test_integration')
```
PhilippWendler: Can a real temp dir be used instead of a subdirectory of the source directory? It is better not to write files into the source directory.

benchexec/outputhandler.py
```diff
@@ -308,9 +309,17 @@ def output_before_run_set(self, runSet):
...
+                adjusted_identifier = util.relative_path(run.identifier, xml_file_name)
+            else:
+                # If no source files exist the task doesn't point to any file that could be downloaded.
+                # In this case, the name doesn't have to be adjusted because it's no path.
+                adjusted_identifier = run.identifier
+
         # prepare XML structure for each run and runSet
             run.xml = ET.Element("run",
-                                 {"name": run.identifier, "files": "[" + ", ".join(run.sourcefiles) + "]"})
+                                 {"name": adjusted_identifier, "files": "[" + ", ".join(adjusted_sourcefiles) + "]"})
```
PhilippWendler: Could you please remove the attribute for files completely for `<withoutfile>` runs instead of using `[]` as content? The format of this attribute is pretty ugly anyway (we will probably change this), so better avoid using it completely where possible.

benchexec/test_benchexec_integration.py
leostrakosch: This is one of the things I don't like ;-)
The alternative to this is having an 'expected file' that has wrong files-references for the directory it is in, but correct references if it were in the temporary directory.

Since this makes the 'expected file' basically a bad example, I preferred this way, but I understand your point.
So should I change this?

leostrakosch:
Actually, sorry for the too short description in the pull request above.

> What are the things you do not like about the current integration test?

In addition to the temporary directory being located beneath the source directory, which is mentioned above:
1. I don't like that I'm only checking a specified list of columns, since I didn't find a good way
to check the values of cputime/walltime and the likes.
Maybe a 'blacklist' of columns not to check makes more sense than a whitelist (even though I don't think so)? Or some other approach to checking the run results?

2. There's a directory `benchexec/test_integration/expected` now, containing the expected results xml, but the integration tests are still in `benchexec/` . In my opinion it would be more structured to put all integration tests into `benchexec/test_integration`, but I didn't want to change the existing project structure without asking.

3. There's only one very specific integration test that's only looking at TRUE results and with all files in the same directory - it would be better to add more diverse tests, but since the benchmark examples in `doc/` are currently used for the integration tests, I didn't want to just add another, more complicated example and diverge from the current way the tests are constructed.
My proposal would be to copy the existing files to `benchexec/test_integration` and add some more integration tests, but if you say the current test is sufficient from your point of view, I'd rather look into other things.

benchexec/test_benchexec_integration.py
PhilippWendler: All paths are relative, right? So there could be a copy of the directory structure re-created in the temp directory, such that all paths stay valid relative to each other?

PhilippWendler:
> There's a directory benchexec/test_integration/expected now, containing the expected results xml, but the integration tests are still in benchexec/ . In my opinion it would be more structured to put all integration tests into benchexec/test_integration, but I didn't want to change the existing project structure without asking.

Feel free to change whatever you think fits better. We can still discuss it in this or a separate PR. You can also have a look at the directory structure of the integration tests of table-generator, it is probably good if these directories are organized similarly.

PhilippWendler:
For the remaining issues, these can be addressed in the future and don't need to be solved immediately.

PR #241, opened 2017-05-11
Updates to Ultimate tools with new parameter format 

danieldietsch:
Just a small update to account for some changes in Ultimate's CLI format. Also, finally using config and data folder for Ultimate (makes REQUIRED_PATHS easier)

### DIFF
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -57,7 +57,7 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         if executable == None: 
             raise Exception('No executable specified')
         
-        cmdline = [executable] + ['--full-output']
+        cmdline = [executable]
         
         if spec != None: 
             cmdline = cmdline + ['--spec'] + [spec]
@@ -67,10 +67,7 @@ class UltimateTool(benchexec.tools.template.BaseTool):
             
         if options:
             for option in options:
-                if option in ['32bit','64bit']:
-                    cmdline = cmdline + ['--architecture'] + [option]
-                else:
-                    cmdline = cmdline + [option]
+                cmdline = cmdline + [option]
 
         return cmdline
 
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -34,7 +34,25 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, tasks, spec, rlimits):
-        return [executable] + [spec] + options + ['--full-output'] + tasks
+        if executable == None: 
+            raise Exception('No executable specified')
+        
+        cmdline = [executable] + ['--full-output']
+        
+        if spec != None: 
+            cmdline = cmdline + ['--spec'] + [spec]
+            
+        if tasks:
+            cmdline = cmdline + ['--file'] + tasks            
+            
+        if options:
+            for option in options:
+                if option in ['32bit','64bit']:
+                    cmdline = cmdline + ['--architecture'] + [option]
+                else:
+                    cmdline = cmdline + [option]
+
+        return cmdline
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
         for line in output:
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -27,6 +27,26 @@ class UltimateTool(benchexec.tools.template.BaseTool):
     Abstract tool info for Ultimate-based tools.
     """
 
+    REQUIRED_PATHS = [
+              "artifacts.xml",
+              "config",
+              "configuration",
+              "data",
+              "features",
+              "p2",
+              "plugins",
+              "LICENSE",
+              "LICENSE.GPL",
+              "LICENSE.GPL.LESSER",
+              "README",
+              "Ultimate",
+              "Ultimate.ini",
+              "Ultimate.py",
+              "z3",
+              "mathsat",
+              "cvc4",
+              ]
+
     def executable(self):
         return util.find_executable('Ultimate.py')
 
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -21,6 +21,10 @@ limitations under the License.
 import benchexec.util as util
 import benchexec.tools.template
 import benchexec.result as result
+import os.path
+import subprocess
+import logging
+import re
 
 class UltimateTool(benchexec.tools.template.BaseTool):
     """
@@ -46,6 +50,11 @@ class UltimateTool(benchexec.tools.template.BaseTool):
               "mathsat",
               "cvc4",
               ]
+    
+    SVCOMP17_WRAPPER_VERSIONS = {'f7c3ed31'}
+    SVCOMP17_ULTIMATE_VERSIONS = {'0.0.1'}
+    SVCOMP17_FORBIDDEN_FLAGS = {'--full-output','--architecture'}
+    ULTIMATE_VERSION_REGEX = re.compile('^Version is (.*)$', re.MULTILINE)
 
     def executable(self):
         return util.find_executable('Ultimate.py')
@@ -54,6 +63,16 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, tasks, spec, rlimits):
+        return self.determine_commandline_version(executable)(executable, options, tasks, spec, rlimits)
+    
+    def cmdline_svcomp17(self, executable, options, tasks, spec, rlimits):
+        for flag in self.SVCOMP17_FORBIDDEN_FLAGS:
+            if flag in options:
+                options.remove(flag)
+
+        return [executable] + [spec] + options + ['--full-output'] + tasks
+    
+    def cmdline_current(self, executable, options, tasks, spec, rlimits):
         if executable == None:
             raise Exception('No executable specified')
 
@@ -69,7 +88,41 @@ class UltimateTool(benchexec.tools.template.BaseTool):
             cmdline = cmdline + options
 
         return cmdline
+    
+    def determine_commandline_version(self, executable):
+        bin_python_wrapper = [util.find_executable('Ultimate.py')]
+        bin_ultimate = ['java', '-jar' , os.path.join(os.path.dirname(os.path.realpath(bin_python_wrapper[0])), 'plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar')]
+        version_wrapper = self.get_version(bin_python_wrapper + ['--version'])
+        version_ultimate = self.get_version(bin_ultimate + ['--version'])
+        version_ultimate_match = self.ULTIMATE_VERSION_REGEX.search(version_ultimate)
+        if not version_ultimate_match:
+            raise RuntimeError('Could not obtain Ultimate version')
 
+        version_ultimate = version_ultimate_match.group(1)
+        
+        msg = 'Detected Ultimate version {0} / {1}'.format(version_ultimate, version_wrapper)
+        if version_wrapper in self.SVCOMP17_WRAPPER_VERSIONS and version_ultimate in self.SVCOMP17_ULTIMATE_VERSIONS:
+            logging.info('{0}. Using SVCOMP17 compatibility mode'.format(msg))
+            return self.cmdline_svcomp17
+        else:
+            logging.info(msg)
+            return self.cmdline_current
+    
+    def get_version(self, command):
+        try:
+            process = subprocess.Popen(command,
+                                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            (stdout, stderr) = process.communicate()
+        except OSError as e:
+            logging.warning('Cannot run {0} to determine version: {1}'.
+                            format(command, e.strerror))
+            return ''
+        if process.returncode:
+            logging.warning('Cannot determine {0} version, exit code {1}'.
+                            format(command, process.returncode))
+            return ''
+        return util.decode_to_string(stdout).strip()
+    
     def determine_result(self, returncode, returnsignal, output, isTimeout):
         for line in output:
             if line.startswith('FALSE(valid-free)'):
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -54,20 +54,19 @@ class UltimateTool(benchexec.tools.template.BaseTool):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, tasks, spec, rlimits):
-        if executable == None: 
+        if executable == None:
             raise Exception('No executable specified')
-        
+
         cmdline = [executable]
-        
-        if spec != None: 
+
+        if spec != None:
             cmdline = cmdline + ['--spec'] + [spec]
-            
+
         if tasks:
-            cmdline = cmdline + ['--file'] + tasks            
-            
+            cmdline = cmdline + ['--file'] + tasks
+
         if options:
-            for option in options:
-                cmdline = cmdline + [option]
+            cmdline = cmdline + options
 
         return cmdline
 
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -25,6 +25,7 @@ import os.path
 import subprocess
 import logging
 import re
+from functools import lru_cache
 
 class UltimateTool(benchexec.tools.template.BaseTool):
     """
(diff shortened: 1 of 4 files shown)


benchexec/tools/ultimate.py
```diff
@@ -27,14 +27,52 @@ class UltimateTool(benchexec.tools.template.BaseTool):
...
+            
+        if tasks:
+            cmdline = cmdline + ['--file'] + tasks            
+            
+        if options:
+            for option in options:
+                if option in ['32bit','64bit']:
+                    cmdline = cmdline + ['--architecture'] + [option]
+                else:
+                    cmdline = cmdline + [option]
```
PhilippWendler: I would suggest here to leave the special handling of `--architecture` out. Users would be expected to use `<option name="--architecture">32bit</option>` in the benchmark XML file.

danieldietsch: This translates from the old 32bit / 64bit parameter to a named one for our wrapper script. I do not think this should change anything for users.

danieldietsch:
Probably not, no. Someone would have to use an older version of benchexec. 

But for the future: 
When does benchexec start measuring? I could try to check the version of Ultimate before creating the actual commandline and depending on the version select different code. But this will take time and thus skim the actual measurement.

PhilippWendler:
Using an old version of BenchExec is not recommended. Users should always use the latest version of BenchExec for the best results.

BenchExec measures exactly the execution of the tool. Of course, all preparations such as determining version and command-line arguments are not included.

benchexec/tools/ultimate.py
PhilippWendler: Well, it breaks for users that specify `--architecture 32bit` in the benchmark definition. I suggest that tool-info modules should not interfere with the command-line arguments given by the user, except where necessary (e.g., input files, any options that are necessary such that the tool-info module can do its job of extracting results, etc.).

benchexec/tools/ultimate.py
```diff
@@ -27,14 +27,52 @@ class UltimateTool(benchexec.tools.template.BaseTool):
...
 
     def version(self, executable):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, tasks, spec, rlimits):
-        return [executable] + [spec] + options + ['--full-output'] + tasks
+        if executable == None: 
+            raise Exception('No executable specified')
+        
+        cmdline = [executable] + ['--full-output']
```
PhilippWendler: Does this break if users specify `--full-output` in their benchmark definition?

danieldietsch:
Ah, so you want to move the changes with CLI options to the [benchmark-defs](https://github.com/sosy-lab/sv-comp/tree/master/benchmark-defs), right? I can do that.

benchexec/tools/ultimate.py
```diff
@@ -27,14 +27,49 @@ class UltimateTool(benchexec.tools.template.BaseTool):
...
+        
+        if spec != None: 
+            cmdline = cmdline + ['--spec'] + [spec]
+            
+        if tasks:
+            cmdline = cmdline + ['--file'] + tasks            
+            
+        if options:
+            for option in options:
+                cmdline = cmdline + [option]
```
PhilippWendler: This can be simplified now.

PhilippWendler: cmdline += options

benchexec/tools/ultimate.py
```diff
@@ -27,14 +27,49 @@ class UltimateTool(benchexec.tools.template.BaseTool):
...
+        if executable == None: 
+            raise Exception('No executable specified')
+        
+        cmdline = [executable]
+        
+        if spec != None: 
+            cmdline = cmdline + ['--spec'] + [spec]
+            
+        if tasks:
+            cmdline = cmdline + ['--file'] + tasks
```
PhilippWendler: Is this really the intended way if the are multiple files in `tasks`?
If Ultimate does not support multiple files, it would be good to `assert` this.

danieldietsch: It works as intended for now.

benchexec/tools/ultimate.py
danieldietsch: Wouldn't that produce just a list of elements instead of a list of lists?

PhilippWendler: The desired outcome is a list of strings, not a list of lists. Both versions of the code have the same outcome, but replacing the if and the loop with the one line is of course easier to maintain.

danieldietsch:
Now its good, right?

PhilippWendler:
I thought you would add a version check for backwards compatibility?

danieldietsch:
No, only for the future. We did not have a reliable versioning until now.

PhilippWendler:
Ok. I am going to think a little bit about how we want to handle such cases in BenchExec in the future in general, and maybe ask some colleagues for ideas (maybe somehow add explicit versioning to the tool-info module?). If we don't find anything better, I will merge this soon.

danieldietsch:
Sounds good. In general, this is a hard problem, especially for research tools. Most of them do not have reliable release management or really any release management -- we are just trying to introduce something in that direction in Ultimate.

PhilippWendler:
We have decided that we indeed prefer to not break benchmarking older tool versions, especially because of SV-Comp.

Even though previous versions of Ultimate have no real version numbers, is it at least possible to distinguish between old and new versions? I could image for example that new version numbers match a certain pattern, and if it does not match, it is an old version? Or do old versions always look like a plain git hash and can be detected from this?

If there is really no way to distinguish between old and new, we would prefer to at least mark the specific version number of the Ultimate submission(s) to SV-Comp as using the old format, such that users can benchmark at least these versions, but we hope that this is not necessary.

Sorry for the inconvenience, in the future we will establish this policy upfront for new tool-info modules such that hopefully it will be easier.

danieldietsch:
How far back do you want to go?

danieldietsch:
BTW, it is rather pressing that we resolve this -- at the moment, the current Ultimate tools cannot be run with benchexec.

PhilippWendler:
For users that want to replicate old experiments from papers it is of course nice if as many versions as possible are supported, but if you think this is infeasible or impractical, then leave it out (I have no idea how difficult it would be).

For running current Ultimate with BenchExec, do you have some concrete case where you would like to have an official BenchExec release with this change? (E.g., a paper where you would like to say we used Ultimate x.y and BenchExec z.foo?) I could make a new release in this case.

danieldietsch:
No, not necessarily a release -- but the LDV guys use benchexec with CPAChecker and Ultimate for their work and want to use our newest release. They cannot do that before this pull request gets accepted (or they would have to use my fork). 

I will try to integrate backwards compatibility for SVCOMP 17, but it might take some time to get it right.

danieldietsch:
Ok, now I have added the desired backwards compatibility. Note that I am not sure how to approach the REQUIRED_PATHS variable; it also changed and newer versions have different required paths. As this seems static, I do not know when to set it.

PhilippWendler:
`REQUIRED_PATHS` is just an implementation detail of the method `program_files`, which is inherited from the super class. So this method can be overwritten with one that returns the correct set of paths.

Why does `determine_commandline_version` does something that is much more difficult than the `version` method? I would have expected `cmdline` method to simply call `version` and check the result.

The file name `plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar` looks like something that might change again at any time, and we would have the same problem soon again. Is this really the required way to get the version from Ultimate?

For `cmdline_svcomp17`, I am not sure whether this should change the user-specified command-line arguments. Ignoring `--full-output` is a good fix for the old method which could lead to `--full-output` being used twice, but why is `--architecture` removed? If users want to benchmark an old version of Ultimate, they need to specify correct command-line arguments for the old version anyway, I guess?

Now we should probably cache the result of the `version` method, because `cmdline` will be called thousands of times, and we don't want to start Ultimate each time. Luckily, Python makes this trivial, you can just add [https://docs.python.org/dev/library/functools.html#functools.lru_cache](`@functools.lru_cache`)

The log message `Using SVCOMP17 compatibility mode` sounds like this is something negative. Is there any reason why users would need to be aware of this? Otherwise I would suggest to turn it into a debug message. If users need to be warned, it should probably even be a warning, but in this case the `cmdline` method must log this only once, not for each call (remember, there will be thousands of calls to this method).

danieldietsch:
> ``REQUIRED_PATHS`` is just an implementation detail of the method ``program_files``, which is inherited from the super class. So this method can be overwritten with one that returns the correct set of paths.

Ok, if ``REQUIRED_PATHS`` is only used there, sure. 

> Why does determine_commandline_version does something that is much more difficult than the version method? I would have expected cmdline method to simply call version and check the result.

The version method fixes the arguments. Ultimate itself has to be called via Java with more arguments. 

> The file name plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar looks like something that might change again at any time, and we would have the same problem soon again. Is this really the required way to get the version from Ultimate?

Yes. Or as long as we use RCP 3.xx comatibility and do not upgrade to E4 etc. 
But I expect that you require backward compatibility to all versions from now on, so it wont matter. 

> For cmdline_svcomp17, I am not sure whether this should change the user-specified command-line arguments. Ignoring --full-output is a good fix for the old method which could lead to --full-output being used twice, but why is --architecture removed? If users want to benchmark an old version of Ultimate, they need to specify correct command-line arguments for the old version anyway, I guess?

I dont now what the scenario will be. But the old version does not support ``--architecture xxx`` but only ``xxx`` as was required at some point. So, if one specifies ``--architecture`` Ultimate will crash anyways. 

> Now we should probably cache the result of the version method, because cmdline will be called thousands of times, and we don't want to start Ultimate each time. Luckily, Python makes this trivial, you can just add @functools.lru_cache

I will implement this. 

> The log message Using SVCOMP17 compatibility mode sounds like this is something negative. Is there any reason why users would need to be aware of this? Otherwise I would suggest to turn it into a debug message. If users need to be warned, it should probably even be a warning, but in this case the cmdline method must log this only once, not for each call (remember, there will be thousands of calls to this method).

You are right, I will change it to a debug message.

PhilippWendler:
>> Why does determine_commandline_version does something that is much more difficult than the version method? I would have expected cmdline method to simply call version and check the result.
> 
> The version method fixes the arguments. Ultimate itself has to be called via Java with more arguments.

Sorry, I can't follow here. What do you mean by `fixes the arguments`?
Why is the user-visible version number not the same one you use for making the distinction?

>>     For cmdline_svcomp17, I am not sure whether this should change the user-specified command-line arguments. Ignoring --full-output is a good fix for the old method which could lead to --full-output being used twice, but why is --architecture removed? If users want to benchmark an old version of Ultimate, they need to specify correct command-line arguments for the old version anyway, I guess?
> 
> I dont now what the scenario will be. But the old version does not support --architecture xxx but only xxx as was required at some point. So, if one specifies --architecture Ultimate will crash anyways.

Of course, users who wish to benchmark an old version of Ultimate will need to use command-line arguments that fit the old version anyway. So why have special handling of `--architecture`? Or does this mean that with this change old and new Ultimate versions become fully compatible and users do not even need to distinguish between them? That would be nice, of course (but was never my intention when requesting backwards compatibility).

danieldietsch:
``version`` calls ``version_from_tool`` which calls ``subprocess.Popen([executable, arg]`` -- you can give only one argument, we need many for the version number from Ultimate itself. 

In the future, we want to drop our python wrapper script and integrate result-detection as well as toolchain / settings selection into the tool-info module. Then, the wrapper script version number will be gone. Therefore, I already integrated two version providers: the one of the past and the one we will be using in the future. 

> Of course, users who wish to benchmark an old version of Ultimate will need to use command-line arguments that fit the old version anyway. So why have special handling of --architecture? Or does this mean that with this change old and new Ultimate versions become fully compatible and users do not even need to distinguish between them? That would be nice, of course (but was never my intention when requesting backwards compatibility).

At least regarding the arguments used in SVCOMP, they become fully compatible. Other arguments are not supported by the old version. 
Only future versions will support full control over Ultimate (when we get rid of the wrapper script).

PhilippWendler:
> version calls version_from_tool which calls subprocess.Popen([executable, arg] -- you can give only one argument, we need many for the version number from Ultimate itself.
> 
> In the future, we want to drop our python wrapper script and integrate result-detection as well as toolchain / settings selection into the tool-info module. Then, the wrapper script version number will be gone. Therefore, I already integrated two version providers: the one of the past and the one we will be using in the future.

If you eliminate the wrapper script, the tool-info module will need to be changed again anyway, right? After all, it currently always calls `Ultimate.py`.
And I still don't understand why `version` should use a different logic for determining the version than `cmdline`. If you want to make the latter future-proof, then the same should be applied to `version`.

Sorry that this change needs so much discussion, I would have never assumed this to be so complex.
I had expected the code to just look something like this without further changes:
```
def cmdline(...):
    if self.version(executable) == "...":
        return cmdline_svcomp17(...)
    elif re.matches(self.version(executable), ...):
        return cmdline_current(...)
    else:
        throw Exception()
```

For `program_files`, it would be the same logic, just taking `REQUIRED_PATHS` in one branch and `REQUIRED_PATHS_SVCOMP17` in the other branch (no need to check tool names here, simply define `REQUIRED_PATHS_SVOMP17` in the tool-specific subclasses, exactly like it is currently).

Btw., do you really want to tell users to call `java -jar plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar --version` if they want to get a version number from Ultimate in the future?

danieldietsch:
In the current version, the ``--version`` command of Ultimate yields one part of the version string and the wrapper script's ``--version`` yields the other. In the future I will combine this in one string. I do not see the need to adapt the version command to this also, but I can do it.  

In the current version, the tools only define their name. The initial change tried to eliminate this duplication by moving tool-specific configuration to their own folders. I do not want to reverse that, but if you insist I can do it. I thought that thanks to Pythons magic, it wont make any difference. 

We do not want to tell users this complex command line, but we currently do. You could also use ./Ultimate, but the launcher has a bug that tries to open X although we are in the commandline. This will delay startup by about 7s. We will fix that in the future by migrating to E4. But in the old versions, it has to be like that. And in the intermediate too. 

I would very much appreciate it if you could just merge this: It is an intermediate state that will change. I just wanted to give the LDV guys a way to use the current benchexec version with the current Ultimate version. 

It will take some time to get to the final CLI interface (because of the E4 migration), and I intent to do that before SVCOMP18.

PhilippWendler:
> In the current version, the --version command of Ultimate yields one part of the version string and the wrapper script's --version yields the other. In the future I will combine this in one string. I do not see the need to adapt the version command to this also, but I can do it.

If you want to show this nice new version number to benchmarking users, then you need to adapt the version command (and I would certainly suggest to do so).

> In the current version, the tools only define their name. The initial change tried to eliminate this duplication by moving tool-specific configuration to their own folders.

Sure, that's fine. But why move the old sets of required paths to ultimate.py from the tool-specific modules? Moving them just complicates the code.

> I would very much appreciate it if you could just merge this: It is an intermediate state that will change. I just wanted to give the LDV guys a way to use the current benchexec version with the current Ultimate version.

I can't merge something if it breaks use cases like benchmarking a tool from last year's submission. After all, there are other people as well who want to use BenchExec. And I am sure that you also want the experience that people have when benchmarking Ultimate to be as nice as possible.

Furthermore, one can always have a tool-info module locally on one's own machine and [tell BenchExec to use this one](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#specifying-a-tool-for-benchexec), it is not required that a tool-info module is bundled with BenchExec. So there is no need to rush with the merge this before it is properly implemented.

However, of course it is more convenient if the current BenchExec version also supports the current Ultimate version, and I, too, want this to happen as soon as possible. Thus, I have made a new proposal on how the tool-info module could support both versions and committed this to the branch [ultimate-support](https://github.com/sosy-lab/benchexec/compare/sosy-lab:05660cd...sosy-lab:fbf91ee). Can you please have a look at this and check if this is what Ultimate needs and what you have in mind? As far as I see this should implement exactly what was discussed in this thread, just in a much simpler way. I tested it with Automizer of both the SV-COMP'17 submission and your latest release on GitHub.

The only things I did not know when implementing this were whether `--full-output` is required for a current Ultimate as well, and how you would like the combined version number to look. Both can be easily changed, of course.

danieldietsch:
> I can't merge something if it breaks use cases like benchmarking a tool from last year's submission. After all, there are other people as well who want to use BenchExec. And I am sure that you also want the experience that people have when benchmarking Ultimate to be as nice as possible.

It does not do that anymore. 
But anyways, your branch does the same. Just merge this and we are done (until I remove the wrapper and integrate it into the tool-info module). 

>The only things I did not know when implementing this were whether --full-output is required for a current Ultimate as well, and how you would like the combined version number to look. Both can be easily changed, of course.

I moved ``--full-output`` to [sv-comp defs](github.com/sosy-lab/sv-comp/pull/42) (if they accept my pull-request, that is). So the new version should not have that anymore. It does only help with debugging anyways. Now users have to specify it if they run into trouble.

PR #243, opened 2017-05-16
Fix: Tablegenerator HTML shows as much zeros as were measured during execution

leostrakosch:
If a measured value is zero, table generator shows the full number of measured digits.
Example:
0.000 was measured. Display '.000' in HTML.

This has the following advantage over showing a single '0':
The user knows, by looking at the table, the precision of the measured 0-value.


PhilippWendler:
I am not sure if we prefer this or the current output of just `0`. @dbeyer?

PhilippWendler:
We decided that we want to have `.00000` which as much decimal places as the maximum of the other cells in this column. (If none of the other cells has any decimal places, this implies that the result is `0`.)

leostrakosch:
I added the intended behavior - do we also want this for tooltips? Currently, it's the same behavior.

PhilippWendler:
Ah, this was easy because we need the total width anyway for alignment, right? I didn't think of this and thought it would be much more complex.

Did you have a look at the HTML table in the integration test to verify that it indeed looks correct?
Should we add unit tests?

I don't fully understand your question about tooltips. I think that we should format them in the same logic as the cells, just without omitting the prefixed `0` for values less than `1`. What do you mean with "same behavior"?

leostrakosch:
>Ah, this was easy because we need the total width anyway for alignment, right? I didn't think of this and thought it would be much more complex.

Yupp.

>Did you have a look at the HTML table in the integration test to verify that it indeed looks correct?
Should we add unit tests?

I looked at the tables and they look good, but I can also add some unit tests.

>I don't fully understand your question about tooltips. I think that we should format them in the same logic as the cells, just without omitting the prefixed 0 for values less than 1. What do you mean with "same behavior"?

Exactly that, and that's the way it's done - I just wanted to make sure :-)

PhilippWendler:
Thanks, I will merge this, now. You can commit unit tests directly afterwards if you like.

PR #244, opened 2017-05-17
Add wall-time stats to tex output

mdangl:
This adds stats about the wall time to the tex output.
Thanks in advance for all feedback.

### DIFF
--- a/contrib/statistics-tex.py
+++ b/contrib/statistics-tex.py
@@ -21,6 +21,7 @@ limitations under the License.
 
 import argparse
 import collections
+import itertools
 import re
 import string
 import sys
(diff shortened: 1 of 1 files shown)


contrib/statistics-tex.py
```diff
@@ -58,25 +65,33 @@ class StatAccumulator(object):
...
+        return "\n".join(itertools.chain.from_iterable(
+            ["\StoreBenchExecResult{%s}{Count}{}{%s}" % (name, self.count)]
+          + [
+                r"\StoreBenchExecResult{%s}{%s}{}{%s}" % (name, time_name, time_stats.sum),
+                r"\StoreBenchExecResult{%s}{%s}{Avg}{%s}" % (name, time_name, time_stats.avg),
+                r"\StoreBenchExecResult{%s}{%s}{Median}{%s}" % (name, time_name, time_stats.median),
+                r"\StoreBenchExecResult{%s}{%s}{Min}{%s}" % (name, time_name, time_stats.min),
+                r"\StoreBenchExecResult{%s}{%s}{Max}{%s}" % (name, time_name, time_stats.max),
+                r"\StoreBenchExecResult{%s}{%s}{Stdev}{%s}" % (name, time_name, time_stats.stdev)
+              ] for (time_name, time_stats) in [
```
PhilippWendler: Are you sure this works correctly? I would suspect that the `+` has a higher binding priority than the `for`, and thus the `Count` value appears once for each time.

mdangl: Right; fixed it. Thanks!

PR #245, opened 2017-05-18
Update tool-info for Ultimate

PhilippWendler:
The current version of Ultimate needs a different command-line syntax.

We continue supporting the SV-COMP'17 version of Ultimate for backwards compatibility, but other old versions are not supported anymore (but could be by whitelisting their version numbers).

This replaces #241.


PhilippWendler:
Approved by @danieldietsch in https://github.com/sosy-lab/benchexec/pull/241#issuecomment-302504421.

PR #248, opened 2017-06-01
Implement a CLI option to set particular CPU cores for BenchExec

IljaZakharov:
This adds an option to provide the set of particular CPU cores identifiers to BenchExec. The option implements separating available CPU cores for independent BenchExec instances to avoid usage of the same CPUs.

### DIFF
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -152,6 +152,13 @@ class BenchExec(object):
                           metavar="N",
                           help="Limit each run of the tool to N CPU cores (-1 to disable).")
 
+        parser.add_argument("-s", "--allowedCores", dest="coreset",
+                            default=None,
+                            metavar="n",
+                            nargs='+',
+                            type=int,
+                            help="Limit each run of the tool to a subset of particular CPU cores.")
+
         parser.add_argument("--user",
                             dest="users",
                             action="append",
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -152,12 +152,11 @@ class BenchExec(object):
                           metavar="N",
                           help="Limit each run of the tool to N CPU cores (-1 to disable).")
 
-        parser.add_argument("-s", "--allowedCores", dest="coreset",
-                            default=None,
-                            metavar="n",
-                            nargs='+',
-                            type=int,
-                            help="Limit each run of the tool to a subset of particular CPU cores.")
+        parser.add_argument("--allowedCores",
+                          dest="coreset", default=None, type=util.parse_int_list,
+                          help="Limit the set of cores BenchExec will use for all runs "
+                               "(Applied only if the number of CPU cores is limited).",
+                          metavar="N,M-K",)
 
         parser.add_argument("--user",
                             dest="users",
--- a/benchexec/resources.py
+++ b/benchexec/resources.py
@@ -68,7 +68,7 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups, coreSet=None):
 
     @param coreLimit: the number of cores for each run
     @param num_of_threads: the number of parallel benchmark executions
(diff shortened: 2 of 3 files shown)


benchexec/benchexec.py
```diff
@@ -152,6 +152,13 @@ def create_argument_parser(self):
                           metavar="N",
                           help="Limit each run of the tool to N CPU cores (-1 to disable).")
 
+        parser.add_argument("-s", "--allowedCores", dest="coreset",
```
PhilippWendler: I would prefer going without a short option because this option is not one that will be used by many people (I guess).

benchexec/benchexec.py
```diff
@@ -152,6 +152,13 @@ def create_argument_parser(self):
...
+                            default=None,
+                            metavar="n",
+                            nargs='+',
+                            type=int,
+                            help="Limit each run of the tool to a subset of particular CPU cores.")
```
PhilippWendler: Can you write "Limit the set of cores BenchExec will use for all runs" instead, please? The current description sounds like one run will be able to use all the given cores.

benchexec/benchexec.py
```diff
@@ -152,6 +152,13 @@ def create_argument_parser(self):
                           metavar="N",
                           help="Limit each run of the tool to N CPU cores (-1 to disable).")
 
+        parser.add_argument("-s", "--allowedCores", dest="coreset",
+                            default=None,
+                            metavar="n",
+                            nargs='+',
```
PhilippWendler: Instead of using `nargs`, I would suggest making it possible to give a `1-3,5` style list. We have an existing utility method `parse_int_list` for this.

benchexec/resources.py
```diff
@@ -68,11 +68,20 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups):
...
+    @param coreSet: the list of CPU cores identifiers provided by a user
     @return a list of lists, where each inner list contains the cores for one run
     """
     try:
         # read list of available CPU cores
         allCpus = util.parse_int_list(my_cgroups.get_value(cgroups.CPUSET, 'cpus'))
+
+        # Filter CPU cores according to the list of identifiers provided by a user
+        if coreSet:
+            invalid_cores = list(map(str, sorted(set(coreSet).difference(set(allCpus)))))
```
PhilippWendler: Why mapping to `str` here?

benchexec/resources.py
```diff
@@ -68,11 +68,20 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups):
 
     @param coreLimit: the number of cores for each run
     @param num_of_threads: the number of parallel benchmark executions
+    @param coreSet: the list of CPU cores identifiers provided by a user
```
PhilippWendler: Please mention that `None` makes benchexec use all cores.

IljaZakharov:
Changed option a bit as you requested and removed the redundant cast. Found out that the option works only if the CPU core limit is also provided (added the notion of this to the help message).

benchexec/benchexec.py
```diff
@@ -152,6 +152,12 @@ def create_argument_parser(self):
                           metavar="N",
                           help="Limit each run of the tool to N CPU cores (-1 to disable).")
 
+        parser.add_argument("--allowedCores",
+                          dest="coreset", default=None, type=util.parse_int_list,
+                          help="Limit the set of cores BenchExec will use for all runs "
+                               "(Applied only if the number of CPU cores is limited).",
```
PhilippWendler: I actually like this because it means that most users will keep using the normal option for the core limit instead of using this "manual" option. But could you please change it such that benchexec aborts with an invalid-configuration error if this option is given without a core limit? Silently ignoring this option seems very dangerous.

IljaZakharov:
As BenchExec uses the option if the limitation of cores is provided either within XML or CLI option, I added termination only if both are missing.

PR #253, opened 2017-08-28
Fix #252. Full content of column cell is considered for classification.

### LINKED ISSUE #252: table-generator fails for cells with comma in them
A comma in a cell causes the following stack trace when table-generator attempts to determine the column type:
```
  File "/usr/lib/python3.5/concurrent/futures/process.py", line 175, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/usr/lib/python3.5/concurrent/futures/process.py", line 153, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/usr/lib/python3.5/concurrent/futures/process.py", line 153, in <listcomp>
    return [fn(*args) for args in chunk]
  File "scripts/../lib/python-benchmark/BenchExec-1.10-py3-none-any.whl/benchexec/tablegenerator/__init__.py", line 641, in load_result
    result.collect_data(options.correct_only)
  File "scripts/../lib/python-benchmark/BenchExec-1.10-py3-none-any.whl/benchexec/tablegenerator/__init__.py", line 533, in collect_data
    column.type, column.unit, column.source_unit, column.scale_factor = get_column_type(column, self)
  File "scripts/../lib/python-benchmark/BenchExec-1.10-py3-none-any.whl/benchexec/tablegenerator/__init__.py", line 380, in get_column_type
    return _get_column_type_heur(column, column_values)
  File "scripts/../lib/python-benchmark/BenchExec-1.10-py3-none-any.whl/benchexec/tablegenerator/__init__.py", line 300, in _get_column_type_heur
    scaled_value = float(Util.remove_unit(str(value))) * column_scale_factor
ValueError: could not convert string to float: '0,4,1,5,2,6,3,7'
```

This can be easily triggered by running a job on multiple cores in the VerifierCloud and using `table-generator --all-columns`.

leostrakosch:
Fixes a conversion error that appeared later on if the cell value started
with a sequence that fits the measurement regex. (#252)

To avoid future errors we should discuss whether the regular expression to match
measurements is good enough:
https://github.com/sosy-lab/benchexec/blob/61877f109d4cdcf252201731e3d84cbd38e2ddc1/benchexec/tablegenerator/__init__.py#L79

1. the expression does not allow white space in front of a measurement - I guess this is fine, but just to be on the safe side.

2. if a float is given in scientific notation, it expects the exponent to always have a sign
(`([eE][-\+]...)`) .
Should we change this to `([eE][-\+]?...)` ?

3. similar to first, it allows exactly one (arbitrary) whitespace between number and unit.
Should we be more strict, only allowing a space, or no whitespace at all?

4. We currently don't allow numbers or the slash (`/`) to be included in the unit name. Maybe we should allow this?

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -315,7 +315,7 @@ def _get_column_type_heur(column, column_values):
         if value is None or value == '':
             continue
 
-        value_match = REGEX_MEASURE.match(str(value))
+        value_match = REGEX_MEASURE.fullmatch(str(value))
 
         # As soon as one row's value is no number, the column type is 'text'
         if value_match is None:
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -76,7 +76,7 @@ TEMPLATE_NAMESPACE={
 _BYTE_FACTOR = 1000 # bytes in a kilobyte
 
 # Compile regular expression for detecting measurements only once.
-REGEX_MEASURE = re.compile('([-\+])?(\d+)(\.(0*)(\d+))?([eE]([-\+])(\d+))?\s?([a-zA-Z]*)')
+REGEX_MEASURE = re.compile('([-\+])?(\d+)(\.(0*)(\d+))?([eE]([-\+])(\d+))?\s?([a-zA-Z/]*)\s*')
 GROUP_SIGN = 1
 GROUP_INT_PART = 2
 GROUP_DEC_PART = 3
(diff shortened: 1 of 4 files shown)


PhilippWendler:
I think the expression should allow what could be parsed afterwards, i.e., what the Python number parser accept. If we trim whitespace, the regexp should also allow this. Maybe we should not even use a regexp, but simply try parsing and see if it fails? Would this be significantly slower?

Slashes in units would be good, but why would we need numbers in units? Exponents are not nicely representable anyway with plain text.

leostrakosch:
I thought about the try-except approach too.
If we did that, we would still have to trim the potential unit before parsing, and check whether the potential unit fits our 'unit pattern'. So in the end, we would still need some pattern matching.

> but why would we need numbers in units? Exponents are not nicely representable anyway with plain text.

People often represent exponents  in units by just omitting the '^'.
But since this regular expression only matters for native benchexec values, it is very unlikely that
we will ever need this.

PhilippWendler:
Right. I am merging this in order to have the original fix. Further improvements can be made afterwards.

PR #255, opened 2017-09-06
statistics-tex supports multiple input files

leostrakosch:
This change allows the user to hand `contrib/statistics-tex.py` multiple result files at once for processing, which allows easier handling of multiple runs at once and which is common practice on the command line.

### DIFF
--- a/contrib/statistics-tex.py
+++ b/contrib/statistics-tex.py
@@ -93,6 +93,39 @@ class StatAccumulator(object):
                 ]]
             ))
 
+
+class StatsCollection(object):
+    def __init__(self, prefix_list, total_stats, category_stats, status_stats):
+        self.prefix_list = prefix_list
+        self.total_stats = total_stats
+        self.category_stats = category_stats
+        self.status_stats = status_stats
+
+
+def load_results(result_file, status_print):
+    run_set_result = tablegenerator.RunSetResult.create_from_xml(
+            result_file, tablegenerator.parse_results_file(result_file))
+    run_set_result.collect_data(False)
+
+    total_stats = StatAccumulator()
+    category_stats = collections.defaultdict(StatAccumulator)
+    status_stats = collections.defaultdict(lambda: collections.defaultdict(StatAccumulator))
+    for run_result in run_set_result.results:
+        total_stats.add(run_result)
+        category_stats[run_result.category].add(run_result)
+        if status_print == "full":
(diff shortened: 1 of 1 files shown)


leostrakosch:
I just found a bug in this while using it, but didn't find a fix yet,
so please don't accept the request for now.

PhilippWendler:
What is the intended advantage of this code over just running `statistics-tex.py` multiple times and concatenating the output? From the description I assume that the advantage is just convenience and the output would be the same, but the size of the code changes indicates something else.

leostrakosch:
Most of the code changes are just due to indentation/code movement.

In addition to convenience, the header is only created once, so it's cleaner than simple concatenation.

leostrakosch:
The script should work as expected, now,
but I don't understand the failed CI-build.

PhilippWendler:
The build failures are unrelated (we don't even check the `contrib` directory).

Why are all results loaded first and then printed together? Wouldn't it make more sense to just load and dump each file in a loop?

leostrakosch:
I wanted to reduce the time span in which output is generated.
Reading a result file can take a few seconds on my machine,
and I prefer the behavior that all content is written in one operation, compared to
'part of the output is written' - wait 5 seconds - 'part of the output is written' - wait 5 seconds - ...

PhilippWendler:
Why do you prefer this? I would actually prefer the other way: Why wait longer for output than necessary? This way you can interrupt the process and still have at least some output. Also it gives better feedback that the process is actually doing something meaningful and how far it is.

leostrakosch:
It's more likely that I'd wonder why the process didn't finish after writing output.
The output per result file is still written all at once if you write it per file, and if you only require the output for part of the result files, why not just specify those result files you need?

>Also it gives better feedback that the process is actually doing something meaningful

I agree with that, but
> and how far it is.

this can be misleading unless we guarantee that output for results is written in the order of the given result files.
Otherwise, one could see output for one result file and assume that all (in the parameter list) prior
files were already handled.
And I don't think that we want to guarantee a specific order, since handling result files in parallel would be the next step (which would be quite easy with the current code).

So I'd still prefer the output to be written at once, but since I'll know the behavior of the script either way, I wouldn't mind changing it, if you ask me to.

PhilippWendler:
Ok, so I suggest to change the script to work in parallel, but still keep the output in the same order as the input files, because deterministic output is always a good thing. This should be easy using the `map` method of the `multiprocessing` module.

PR #264, opened 2017-10-17
symbiotic: update the tool-info object

mchalupa:
Directories hierarchy changed. Follow-up PR for #224 

This PR makes the tool-info object check version and parse the results accordingly (as suggested in #224). The only thing that does not fit me here is that the file is named symbiotc4.py, but now the version of symbiotic is 5. Wouldn't it be better to leave symbiotic4.py as it is and add a new file symbiotic.py for generic version of Symbiotic?

### DIFF
--- a/benchexec/tools/symbiotic.py
+++ b/benchexec/tools/symbiotic.py
@@ -1,0 +1,131 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2015  Dirk Beyer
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Yes, I think that would make sense. I would prefer to have a generic `symbiotic.py` that works with all future versions of Symbiotic, but if you expect that Symbiotic 6 will change input and output again, then maybe we could also go with adding `symbiotic5.py`.

If it is not too much effort, could you make `symbiotic.py` support Symbiotic 4, too (by delegating to the methods from `symbiotic4.py`)?

PR #265, opened 2017-10-19
Update map2check.py tool-info

hbgit:


PhilippWendler:
Hi, is it possible to make these changes in a backwards-compatible way?
Otherwise it will be very hard for users for example to compare older versions against a current one.

You could detect which script is present and act accordingly. For `REQUIRED_PATHS` you can just use the union of old an new required files.

benchexec/tools/map2check.py
```diff
@@ -27,20 +28,48 @@ class Tool(benchexec.tools.template.BaseTool):
...
     def program_files(self, executable):
         executableDir = os.path.dirname(executable)
-        return [executableDir]
+
+        if self._get_version() == 6:
+            paths = REQUIRED_PATHS_6
+        elif self._get_version() > 6:
+            paths = REQUIRED_PATHS_7_1
+
+        return [executableDir] + paths
```
PhilippWendler: I just happened to see this now: Aren`t the files from `REQUIRED_PATHS` inside your `executableDir`? If yes, this is redundant. In the old code it seems that `REQUIRED_PATHS` was completely unused. I would prefer to remove `executableDir` from the result of `program_files` to improve this.

benchexec/tools/map2check.py
```diff
@@ -51,14 +51,14 @@ def executable(self):
 
 
     def program_files(self, executable):
-        executableDir = os.path.dirname(executable)
+        #executableDir = os.path.dirname(executable)
```
PhilippWendler: Please delete unnecessary code instead of commenting it out.

PR #266, opened 2017-10-20
Added tool description for CPA-BAM-Slicing

schrodibear:
The only difference from basic CPAchecker is executing the wrapper rather than cpa.sh

### DIFF
--- a/benchexec/tools/cpa-bam-slicing.py
+++ b/benchexec/tools/cpa-bam-slicing.py
@@ -1,0 +1,203 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2015  Dirk Beyer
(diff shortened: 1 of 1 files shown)


PhilippWendler:
If the only difference to `cpachecker.py` is the use of `wrapper.sh`, then this should be refactored such that all duplicate code is avoided.

Btw., are you sure that you want to call the main entry point of your tool `wrapper.sh` and not something more descriptive?
If your tool is based on CPAchecker, you could also integrate your extensions into for example `cpa.sh` itself such that no new tool-info module is needed.

schrodibear:
I moved slicer invocation to cpa.sh in the submission and changed tool info in the benchmark definition to just "cpachecker" ([here](https://github.com/sosy-lab/sv-comp/pull/51)). So this PR can be closed.

PR #268, opened 2017-10-21
Added tool description for VIAP

VerifierIntegerAssignment:
Please commit the tool description for VIAP.

### DIFF
--- a/benchexec/tools/viap.py
+++ b/benchexec/tools/viap.py
@@ -1,0 +1,64 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2015  Dirk Beyer
+All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import benchexec.tools.template
+import benchexec.result as result
+import os
+class Tool(benchexec.tools.template.BaseTool):
+
+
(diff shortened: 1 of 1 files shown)


benchexec/tools/viap.py
```diff
@@ -0,0 +1,49 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import benchexec.tools.template
+import benchexec.result as result
+import os
+class Tool(benchexec.tools.template.BaseTool):
+    REQUIRED_PATHS = ["viap_tool.py","viap_svcomp.py","config.properties","SyntaxFilter.py","graphclass.py","commandclass.py","packages"]
+    def executable(self):
+        return 'viap_tool.py'
```
PhilippWendler: This won't work if a user has installed your tool somewhere else than the current directory. Why not use `Util.findExecutable` like the other modules?

benchexec/tools/viap.py
```diff
@@ -0,0 +1,49 @@
...
+        spec = ['--spec=' + propertyfile]
+        return [executable] + options + spec + tasks
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        status = result.RESULT_UNKNOWN
+        stroutput = str(output)
+        if "True" in stroutput:
+            status = result.RESULT_TRUE_PROP
+        elif "False" in stroutput:
+            status = result.RESULT_FALSE_REACH
+        return status
```
PhilippWendler: This looks quite fragile, it would return `true` or `false` if `True` or `False` appear anywhere in the output. For example, suppose that there is some error message of your tool that includes the source file name or part of the C code, and there these strings appear. Is this really what you intended?

VerifierIntegerAssignment:
I have made changes as suggested

benchexec/tools/viap.py
```diff
@@ -0,0 +1,55 @@
...
+import benchexec.tools.template
+import benchexec.result as result
+import os
+class Tool(benchexec.tools.template.BaseTool):
+
+    REQUIRED_PATHS = ["viap_tool.py","viap_svcomp.py","config.properties","SyntaxFilter.py","graphclass.py","commandclass.py","packages"]
+
+    def executable(self):
+        return 'viap_tool.py'    
+
```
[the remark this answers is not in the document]
PhilippWendler: I am not sure this code works correctly.
`viap_tool.py` as well as all other required files are in the same directory, right?
Then the `program_files` implementation from the super class will work correctly. Your code would be necessary if the other files are in a parent directory from `viap_tool.py`, but then it wouldn't make sense to look for `viap_tool.py` in that parent directory.

PR #274, opened 2017-10-31
Optional separate file for stderr of run and optional header information in output files

fheck:
As previously discussed in #103, we would like to have stderr written to a different file than stdout. Further, we want the possibility to remove the header runexec is writing in the output files. The proposed parameters default to the current behavior but allow finer control.

### DIFF
--- a/benchexec/runexecutor.py
+++ b/benchexec/runexecutor.py
@@ -608,7 +608,7 @@ class RunExecutor(containerexecutor.ContainerExecutor):
         return run_environment
 
 
-    def _setup_output_file(self, output_filename, args):
+    def _setup_output_file(self, output_filename, args, write_header = True):
         """Open and prepare output file."""
         # write command line into outputFile
         # (without environment variables, they are documented by benchexec)
@@ -616,9 +616,12 @@ class RunExecutor(containerexecutor.ContainerExecutor):
             output_file = open(output_filename, 'w') # override existing file
         except IOError as e:
             sys.exit(e)
-        output_file.write(' '.join(map(util.escape_string_shell, self._build_cmdline(args)))
-                          + '\n\n\n' + '-' * 80 + '\n\n\n')
-        output_file.flush()
+
+        if write_header:
+            output_file.write(' '.join(map(util.escape_string_shell, self._build_cmdline(args)))
+                              + '\n\n\n' + '-' * 80 + '\n\n\n')
+            output_file.flush()
+
         return output_file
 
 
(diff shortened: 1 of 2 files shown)


benchexec/runexecutor.py
```diff
@@ -692,8 +695,8 @@ def _setup_file_hierarchy_limit(
 
     # --- run execution ---
 
-    def execute_run(self, args, output_filename, stdin=None,
-                   hardtimelimit=None, softtimelimit=None, walltimelimit=None,
+    def execute_run(self, args, output_filename, error_filename=None, stdin=None, write_header=True,
```
PhilippWendler: Inserting parameters in the middle breaks backwards compatibility.

benchexec/runexecutor.py
```diff
@@ -704,7 +707,9 @@ def execute_run(self, args, output_filename, stdin=None,
         and writes the output to a file.
         @param args: the command line to run
         @param output_filename: the file where the output should be written to
+        @param error_filename: the file where the error output should be written to (default: same as output_filename)
         @param stdin: What to uses as stdin for the process (None: /dev/null, a file descriptor, or a file object)
+        @param stdin: Write informational headers to the output and the error file
```
PhilippWendler: wrong parameter name

benchexec/runexecutor.py
fheck: Ok, had not thought about this, should I change it for the internal call as well?

fheck:
No, we don't need it for the command line interface because we call runexec directly from Python code, but I will add it there too and make the other requested changes.

benchexec/runexecutor.py
PhilippWendler: For consistency, maybe yes.

PhilippWendler:
> No, we don't need it for the command line interface because we call runexec directly from Python code, but I will add it there too and make the other requested changes.

No, please don't add an external interface as long as it is not necessary.

fheck:
Another question, regarding `_reduce_file_size_if_necessary`: my understanding is, that the limit the file size is 'packed' to, should semantically include every output from the process, therefore I should call it in a manner that looks at both files and limits them to the same degree, not a fixed limit each.

```
        max_outfile_size = max_output_size
        if error_filename is not None:
            # limit both files to the same degree
            error_filezize = os.path.getsize(error_filename)
            output_filezize = os.path.getsize(output_filename)
            filesize_relation = error_filezize / output_filezize
            max_outfile_size = int((1 - filesize_relation) * max_output_size)
            max_errorfile_size = int(filesize_relation * max_output_size)

            _reduce_file_size_if_necessary(error_filename, max_errorfile_size)

        _reduce_file_size_if_necessary(output_filename, max_outfile_size)
```
So the sum of both file sizes is in the limit, is that the correct understanding of the semantics?

PhilippWendler:
> Another question, regarding `_reduce_file_size_if_necessary`: my understanding is, that the limit the file size is 'packed' to, should semantically include every output from the process, therefore I should call it in a manner that looks at both files and limits them to the same degree, not a fixed limit each.

That would be possible, but if it is complex to do, simply applying the limit individually would also be acceptable. The limit is not 100% precise anyway, it is just a safe guard against a tool filling the whole disk.

Issue #272, opened 2017-10-25
Do not add trailing zeroes for numeric values in table-generator


PhilippWendler:
If a user specifies for example three significant digits for a table column, but the value in the input file has only two digits (e.g., `1.1`), table-generator should not output `1.10` but `1.1` to not give a wrong indication of the precision of the value.

PhilippWendler:
I wrote some unit tests for this, and I am surprised to see that this seems to be almost implemented already. @leostrakosch Is this correct and can you confirm that this is implemented? Did I overlook anything in the tests, or are there other interesting tests for this functionality that should be added? If yes, could you please add them?

It seems that the only incorrectly formatted number is the value zero, for which trailing zeros get added, probably as a result of #243. Given that the rest of this issue is already implemented, I guess doing it for zero should be easy. Could you please do this and adjust the unit tests that are marked with `FIXME` accordingly?

leostrakosch:
I intended to implement it without adding additional zeros, so if it isn't working, it's an "unintended" bug.

> It seems that the only incorrectly formatted number is the value zero, for which trailing zeros get added, probably as a result of #243

I did it on purpose for zeros with the changes in #243 because I thought that that's what we wanted.
So just to reiterate and to be on the safe side:
`no. of zeros after decimal = min(no. of zeros after decimal in input value, max no. of decimal places in column)`

Right?

PhilippWendler:
Yes, this is correct. And if the result of your formula is 0, we use `0` as output.

PR #275, opened 2017-11-04
Fix issue #272

leostrakosch:
Fixes issue #272 .
If a value is zero, as many decimal digits available as possible are displayed, up to the existing max number of decimal digits in the column


PhilippWendler:
I don't understand why it changes this line to `0` from `.00`:

https://github.com/sosy-lab/benchexec/blob/9f9ded330c2394ba79a97bb04b0538a6a0baa037/benchexec/tablegenerator/test_integration/expected/test.2015-03-03_1613.results.predicateAnalysis.correct-only.html#L121

However, the value in this cell needs to be `-` anyway.

PR #278, opened 2017-11-08
extended REQUIRED_PATHs s.t. it gets the content of directories

danieldietsch:
silent change of treatment of REQUIRED_PATHS addressed


PhilippWendler:
This kind of change is not necessary because when a directory is specified, all files that are in the directory structure below this directory are used.

Issue #113, opened 2016-01-14
Make back button of browser close overlay pane in HTML panels


PhilippWendler:
Some users expect to be able to click on a link (such as the log files in tables) and use the back button to get back to the original page, but because we open the target of the link in an overlay pane with JavaScript, this is not possible. As it is possible to manipulate the back button of the browser through JavaScript, we should do so and make the back button close the overlay pane (or, even better, let the overlay pane show the previous content if there is one).

kfriedberger:
Should this also work for plots?
For example: first show the plot for cputime, then for walltime, then for memory.
Should the back button then also show the old plots?

There would be (at least) 2 possibilities how to implement this: 
1) We store the complete history as a list in JavaScript. The elements of the list could be callbacks to show the plot (or logfile) or something.
2) Use the URL and encode the current view within the URL of the table. The showing a plot or sourcefile would modify the current URL. 
Example: 
`.../table.html?show=plot&title=cputime`, 
`.../table.html?show=source&id=main.c`, 
`.../table.html?show=log&id=main.c.log`. 
Maybe changing the URL can be done without reloading the table. In this case the browser should already handle the back-button feature, as the history of URLs is part of the browsers history. This feature would also allow us to directly use URLs to show the plot (as view above a table).

PhilippWendler:
I think it is not necessary for switching the shown columns inside a plot (users do not expect drop-down fields to affect history), but for switching between different plot types (by clicking on links) it could be done.

The second solution was what I had in mind, although I do not know if we maybe can do this without modifying the url (and keep more state like plot configuration than what is encoded in URL).

PhilippWendler:
Seems like this is easily possible: https://developer.mozilla.org/de/docs/Web/Guide/DOM/Manipulating_the_browser_history#Adding_and_modifying_history_entries

We can store arbitrary state, and we do not need to fiddle with the URL.

kalpit4088:
@PhilippWendler  can you tell me how to test pushState method on my local machine. Because whenever I try to test it on localhost it gives me this error - "SecurityError: The operation is insecure."

PhilippWendler:
Sorry, I never used this function myself. I suggest to look at its documentation and maybe StackOverflow. This [question](https://stackoverflow.com/q/13348766/396730) seems relevant.

PhilippWendler:
Might be solved using the technique from #483?

PR #288, opened 2018-02-19
back button added

kalpit4088:
fixes    https://github.com/sosy-lab/benchexec/issues/113

### DIFF
--- a/example-table/svcomp-simple-cbmc-cpachecker.table.html
+++ b/example-table/svcomp-simple-cbmc-cpachecker.table.html
@@ -116,6 +116,13 @@
     z-index:5;
   }
 
+  #gobackbtn {
+    position: fixed;
+    top:6%; left:90.2%;
+    z-index: 6;
+    display: none;
+  }
+
   .scriptonly {
     display: none; // only enabled via JavaScript
   }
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Thanks for your contribution!

It seems that this doesn't address #113 though, right? #113 is about making the "back" button of the browser have an effect, this adds a new back button in the table. Is this intentional? Do you think it is worth having this additional button? Users can already close the modal for example by clicking somewhere on the background outside of it.

kalpit4088:
I don't think that we can do that (making the "back" button of the browser have an effect). This is the article which I read [https://forums.asp.net/t/1882709.aspx?how+to+enable+back+button+of+browser](url) . It says that back button of browser remains enable by default. I think the reason why we can't use back button of browser using "window.history.back()" because url is not changing when we open/close html overlay. Hence I thought to create the back button.

kalpit4088:
@PhilippWendler I am open to any other suggestions on anything that can be done.

PhilippWendler:
This forum discussion seems quite old. I have read [here](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Adding_and_modifying_history_entries) (which is linked from the issue) that it should be possible what is described in #113. Or do you have any information about why this technique is not usable for what we would need?

PhilippWendler:
Please do not mix unrelated features in the same PR.

Are you planning on updating this PR with a new fix for #113? Otherwise I would close this PR.

kalpit4088:
I am working on enhancement part and will make an update soon. So I am closing this Pull Request.

Issue #123, opened 2016-01-22
Lazy-load JavaScript libraries where possible


PhilippWendler:
The HTML tables produced by `table-generator` include several libraries like JQPlot and zip.js that are only sometimes necessary. These could be loaded on demand to reduce the initial loading time of the table.

PhilippWendler:
Some notes on how to do his are [here](https://www.igvita.com/2014/05/20/script-injected-async-scripts-considered-harmful/). Probably just declaring these scripts `async` is good, but this doesn't solve the problem that they could load later than required for slow connections and fast users.

PR #291, opened 2018-02-22
fixes lazy load of js library like JQPlot.

kalpit4088:
fixes https://github.com/sosy-lab/benchexec/issues/123

### DIFF
--- a/example-table/svcomp-simple-cbmc-cpachecker.table.html
+++ b/example-table/svcomp-simple-cbmc-cpachecker.table.html
@@ -201,9 +201,6 @@
     <a href="javascript:createScatterPlotView()">Scatter Plot</a>
     <a href="javascript:toggleHeader()" id="toggleHeaderButton" data-shown-text="Shrink Header" data-hidden-text="Expand Header">Shrink Header</a>
   </span>
-  <span class="noscript">
-    Please enable JavaScript for advanced features.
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Thanks for your submission!

Unfortunately, this PR contains code related to some other feature, please remove this.

Also, it breaks the [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) that we have been using, this should be avoided.

Could you please explain a little bit more how your chosen approach of lazy loading works? Does it work in all browsers? Would it also be possible by a more modern approach such as declaring the scripts as `async`, or would that have other disadvantages?

kalpit4088:
I am working on enhancement part and will make an update soon. So I am closing this Pull Request.

kalpit4088:
This approach works similarly as it works with "async" attribute, The only difference is that the async attribute works fine on modern browsers – supported by Firefox 3.6+, IE 10+, Chrome 2+, Safari 5+, iOS 5+, Android 3+, however  for a deeper browser support the script I wrote works perfectly. So yeah the way I have used "async" method works for all browsers.

kalpit4088:
I think I've put subresource integrity part inside every script tag. But if it doesn't work then pls let me know I'll fix this too.

PR #292, opened 2018-02-23
lazyload js problem resolved

### LINKED ISSUE #123: Lazy-load JavaScript libraries where possible

kalpit4088:
fixes #123

### DIFF
--- a/example-table/svcomp-simple-cbmc-cpachecker.table.html
+++ b/example-table/svcomp-simple-cbmc-cpachecker.table.html
@@ -1573,15 +1573,17 @@ $(document).ready(function() {
 
 <!-- ################## Everything related to showing plots of the data ################## -->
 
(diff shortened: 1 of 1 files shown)


kalpit4088:
This approach works similarly as it works with "async" attribute, The only difference is that the async attribute works fine on modern browsers – supported by Firefox 3.6+, IE 10+, Chrome 2+, Safari 5+, iOS 5+, Android 3+, however for a deeper browser support the script I wrote works perfectly. So yeah the way I have used "async" method works for all browsers.

kalpit4088:
@PhilippWendler  I think I've put sub-resource integrity part inside every script tag. But if it doesn't work then please let me know I'll fix this too.

PhilippWendler:
We can ignore all older browsers than the ones you have listed, so I would prefer the simpler solution that works only in these browsers.

I cannot understand how the subresource-integrity should work in your code. How would the browser now that it should check the hash of the loaded scripts? Did you test it by changing the hash and verifying that the script indeed fails to load?

kalpit4088:
@PhilippWendler  okay so i'll do it with that way. So that subresource-integrity can also work.

PhilippWendler:
I just noticed now that this PR attempts to update the example table in the gh-pages branch, which is auto-generated. PRs need to be made for the template in the master branch.

PR #295, opened 2018-04-09
CBMC: report "OUT OF MEMORY" instead of generic ERROR(6)

tautschnig:
Out-of-memory is reported and can be detected; this had always been the
case for the XML output, now it's done for plain-text output as well.

Also fixes a missing newline character in invalid-arguments cases.

### DIFF
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -143,8 +143,11 @@ class Tool(benchexec.tools.template.BaseTool):
                 elif 'UNKNOWN' in output:
                     status = result.RESULT_UNKNOWN
 
-        elif returncode == 64 and 'Usage error!' in output:
+        elif returncode == 64 and 'Usage error!\n' in output:
             status = 'INVALID ARGUMENTS'
+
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Does CBMC really exit with code 0 in case of error?

And when would such an OOM actually happen? If BenchExec kills CBMC due to OOM, it will set the result correctly anyway.

tautschnig:
Why would you expect an exit code 0?

It’s ok if benchexec kills, but CBMC’s wrapper script also does its own memory limiting via ulimit.

PhilippWendler:
Sorry, I misread the 6 to be a 0 (font was too small).

Btw.: limiting the virtual address space with ulimit will create a stronger memory limit than BenchExec's one (if both are set to the same value). So I am not sure why one would set this limit?

tautschnig:
The internal ulimit might indeed be redundant nowadays. It made sense when we returned the best/last guess, but nowadays we just say UNKNOWN in those cases so it’s no different. Thanks for the heads-up that we might be limiting ourselves more than necessary.

PR #298, opened 2018-05-04
added new command line argument -ultimatedata to ultimate

danieldietsch:


PhilippWendler:
Thanks!

Do older versions of Ultimate support this as well, or will they at least ignore it?

danieldietsch:
They should ignore it. But I can try and make another version check and only supply it for newer versions.

PhilippWendler:
If they ignore it then its fine for me.

danieldietsch:
Now it should also work for all the older versions.

PR #299, opened 2018-05-14
Update Verifuzz.py

animeshbchowdhury:
Added falsification keywords for various subcat.

### DIFF
--- a/benchexec/tools/verifuzz.py
+++ b/benchexec/tools/verifuzz.py
@@ -49,8 +49,18 @@ class Tool(benchexec.tools.template.BaseTool):
         lines = " ".join(output)
         if "VERIFUZZ_VERIFICATION_SUCCESSFUL" in lines:
             return result.RESULT_TRUE_PROP
-        elif "VERIFUZZ_VERIFICATION_FAILED" in lines:
+        elif "FALSE(unreach-call)" in lines:
             return result.RESULT_FALSE_REACH
+        elif "FALSE(no-overflow)" in lines:
+            return result.RESULT_FALSE_OVERFLOW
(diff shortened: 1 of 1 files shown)


PhilippWendler:
This would break backwards compatibility with old Verifuzz versions, right?
In this case I would suggest to keep checking for the old string, too.

animeshbchowdhury:
We are currently in development zone. Surely it would break backward compatibility of old verifuzz versions. But, we are okay with this change.

PhilippWendler:
But it would be sad if other users cannot run older versions of Verifuzz anymore with a newer BenchExec, for example if they want to reproduce results from previous SV-COMP.

animeshbchowdhury:
As such, we haven't yet made our tool available for other users. We are using it for our own experimentation on SVCOMP benchmarks. We are planning to participate from this time. Once the tool is available for other users for evaluation, I would make sure whatever changes we do, it will preserve backward compatibility of tool.

PhilippWendler:
Ok, in this case I suggest you submit a PR to BenchExec for the updated tool-info module once it is finalized. You can use BenchExec with any tool-info module in your local `PYTHONPATH`, these modules do not need to be integrated into BenchExec itself. Just put your Python module somewhere your Python interpreter finds it, and use the fully-qualified name of the module in your benchmark definitions (cf. last paragraph of the [documentation](https://github.com/sosy-lab/benchexec/blob/6ecad757bdc4db170724fadac8ff61ea8a67bd13/doc/tool-integration.md)).

animeshbchowdhury:
Okay. I am currently using at my end in the same way, you mentioned. In that case, if anything more comes up later, its better to submit the PR one time, after finalizing everything.

Issue #100, opened 2015-12-15
Users need the way to specify their own wall time limits in benchmarks


eunovm:
At the moment all benchmarks implicitly have the default wall time limit that is 30 seconds greater than a specified CPU time limit. Likely users need/want develop new benchmarks with explicit wall time limits. For old-style benchmarks new versions of BenchExec can issue warnings if wall time limits aren't specified.
BTW, runexec already has an option to specify wall time limits explicitly, so its direct users can use it already.

PhilippWendler:
This makes sense and could certainly be added. So far nobody has actually requested it, so it seems to be more of a "nice to have", though. I will probably wait with this until there is somebody who would like to use it. Please comment here if you do want to use this.

There would be no need for warnings, though. The wall time limit would stay on by default and could be disabled manually if desired.

cheshire:
I would really like to have a walltime-limit feature (as opposed to CPU time).
From informal conversations with SV-COMP participants I gather that a _lot_ of people would prefer to measure timeouts by walltime rather than CPU time. Adding this feature to Benchexec would be a first step in evaluating how much difference that would make.

This is anecdotal and not really supported, but I have talked to about eight people who would prefer wall time limit instead of CPU time limit, and I haven't seen a single person arguing in favor of CPU time limit.

PhilippWendler:
Yes, as I said, if someone approaches me with a concrete usecase I would be willing to add this, but so far nobody has done that. Note that for evaluating the difference between the limits you can just run the benchmarks once and then analyze the results to see how they would look like if the other limit would have been applied.

Of course, if somebody sends a pull request, I would be willing to merge it.

PR #301, opened 2018-06-25
Add an CLI option and an attribute for XML benchmark files to set a w…

IljaZakharov:
Added a CLI option and XML attribute to set wall time limitations to implement feature #100.

### DIFF
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -140,7 +140,7 @@ class BenchExec(object):
                             dest="walltimelimit", default=None,
                             help='Wall time limit for each run, e.g. "90s" '
                                  '(overwrites wall time limit from XML file, '
-                                 'use "-1" to disable time limits completely)',
+                                 'use "-1" or the default value to use CPU time limit plus a few seconds))',
                             metavar="SECONDS")
 
         parser.add_argument("-M", "--memorylimit",
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -136,6 +136,13 @@ class BenchExec(object):
                                'use "-1" to disable time limits completely)',
                           metavar="SECONDS")
 
+        parser.add_argument("-W", "--walltimelimit",
+                            dest="walltimelimit", default=None,
+                            help='Wall time limit for each run, e.g. "90s" '
+                                 '(overwrites wall time limit from XML file, '
+                                 'use "-1" to disable time limits completely)',
+                            metavar="SECONDS")
+
         parser.add_argument("-M", "--memorylimit",
                           dest="memorylimit", default=None,
                           help="Memory limit, if no unit is given MB are assumed (-1 to disable)",
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -140,7 +140,7 @@ class BenchExec(object):
                             dest="walltimelimit", default=None,
                             help='Wall time limit for each run, e.g. "90s" '
                                  '(overwrites wall time limit from XML file, '
-                                 'use "-1" or the default value to use CPU time limit plus a few seconds))',
+                                 'use "-1" to use CPU time limit plus a few seconds, such value is also used by default)',
                             metavar="SECONDS")
 
         parser.add_argument("-M", "--memorylimit",
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -29,7 +29,7 @@ import threading
 import time
 from queue import Queue
 
-from benchexec.model import CORELIMIT, MEMLIMIT, TIMELIMIT, SOFTTIMELIMIT
+from benchexec.model import CORELIMIT, MEMLIMIT, TIMELIMIT, SOFTTIMELIMIT, WALLTIMELIMIT
 from benchexec import cgroups
 from benchexec import containerexecutor
 from benchexec.resources import *
@@ -274,6 +274,7 @@ class _Worker(threading.Thread):
                 result_files_patterns=benchmark.result_files_patterns,
                 hardtimelimit=benchmark.rlimits.get(TIMELIMIT),
                 softtimelimit=benchmark.rlimits.get(SOFTTIMELIMIT),
+                walltimelimit=benchmark.rlimits.get(WALLTIMELIMIT),
                 cores=self.my_cpus,
                 memory_nodes=self.my_memory_nodes,
                 memlimit=memlimit,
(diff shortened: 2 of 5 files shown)


PhilippWendler:
Is there a concrete use case for this?
I have asked for one in #100, but nobody has replied in the last two years.

IljaZakharov:
The use case is running verification in OpenStack cloud virtual machines. On practice it is not possible there to reserve CPU cores fairly and sometimes, when other users of OpenStack heavily use it, we get wall timeouts. Also we do not want to wait for verification results forever in this case disabling wall time limitation at all.

PhilippWendler:
Doesn't that mean that the measurement results are completely unreliable, too? That is not really a scenario for which BenchExec is intended, its main focus is on reliable benchmarking, i.e., correct measurements.

I fear that adding features for unreliable scenarios would lead people to misuse BenchExec in such scenarios (and claim that their results are valid because they are produced with BenchExec) instead of executing their benchmarks in scenarios where proper benchmarking is possible.

Would it make sense to provide instead a mode where BenchExec just executes the tasks with some (approximate) resource limits but without measurements? I.e., just as a task runner, not as a benchmarking tool?

IljaZakharov:
On the one hand, I agree that in this case our main goal is finding bugs or development & testing but not precise measurements. So for this case it is nice to have a mode which would allow just running tools, record resource consumption and kill tasks according to limits without checking specific requirements relevant to fair measurements. On the other hand, I doubt that these different modes should have different feature sets and that this particular feature that allows setting wall time limits is useless at benchmarking. Anyway Benchexec already implies wall time limit implicitly and we just want to have an ability to modify it. Moreover, I am not sure that such new mode will appear in the nearest future.

eunovm:
Any measurement is subjected to errors. Regarding measuring computational resources, you constantly try to minimize errors from inside BenchExec and by implying some rules and requirements to the environment (some examples are here: https://github.com/sosy-lab/benchexec/blob/master/doc/benchmarking.md). I am sure that for benchmarking you reached very accurate results.

Sometimes errors may be bigger, but that may be not a problem for a user. For instance, instead of, say, error 0.02% (assume this the best accuracy you can achieve during benchmarking, perhaps, it is even unknown, since nobody knows the ideal case), one gets error 0.2%. Of course, it 10 times more and it is not welcome in some cases, e.g. for fair comparison of verification tools. But it is still too small for users to complain about it somehow. On average this error may be equivalent to just ~0.05% of deviations in results that is very little. (I took the number of timeouts for the winner of last SV-COMP, category _Overall_, divided it to the total number of verification tasks in that category and multiplied the result by 0.2%)

The bad thing I saw is that we do not know error values. If it is not 0.2%, but 2% or maybe even 10%, it does can be an issue for more users, since verification results or something else can vary quite considerably for different launches with the same tools and settings. Moreover, these error values can change dynamically, e.g. if one overloads CPUs from time to time (not necessary by solving additional verification tasks, but also due to other activities, e.g. by watching some videos). @PhilippWendler, what do you think, is it possible to develop and to run some tests to measure these errors somehow? I keep in mind something like measuring an Internet connection speed. Ideally, it would be great if BenchExec will count errors in addition to measuring computational resources and provide this additional measurement to a user for individual judgment. Perhaps I am wrong, but it seems to be enough to count additional context switches due to extra process preemptions that happen due to CPUs overload. Better is to accumulate a CPU time spent on this. This knowledge is useful in any case, for benchmarking as well. 

Anyway, this discussion seems to be almost independent from the given pull request, since it just adds one more mean for killing tools, that existed implicitly already as @IljaZakharov mentioned. Why not to move the discussion into a dedicated issue rather than to prevent completing the pull request.

PhilippWendler:
We have decided to add this feature and I will look at the changes soon in more detail.

benchexec/benchexec.py
```diff
@@ -136,6 +136,13 @@ def create_argument_parser(self):
                                'use "-1" to disable time limits completely)',
                           metavar="SECONDS")
 
+        parser.add_argument("-W", "--walltimelimit",
+                            dest="walltimelimit", default=None,
+                            help='Wall time limit for each run, e.g. "90s" '
+                                 '(overwrites wall time limit from XML file, '
+                                 'use "-1" to disable time limits completely)',
```
PhilippWendler: This is not correct, right? If -1 is passed, the wall-time limit is `None`, which means that the standard wall-time limit will be used. Or did I overlook something?

(In any case, the text that says 'use "-1" to disable time limits completely' is no longer correct, of course.)

IljaZakharov: Yes, sorry - this is a mistake.

What would you like to see as a default behavior? Disabling the limit completely or using the default value which is used currently? And seems that currently all timelimit options do not accept -1. Is this a bug or an intended scenario and help is incorrect?

PhilippWendler: Hm, -1 should work actually, and it does for me:
```
benchexec doc/benchmark-example-rand.xml --timelimit -1
```
What exactly does not work?

For wall-time limit, I would say the behavior of -1 should be the current default value.

IljaZakharov: I see, seems that I have introduced a bug to this somehow too. Ok, I will fix -1 and wall option then.

benchexec/benchexec.py
```diff
@@ -140,7 +140,7 @@ def create_argument_parser(self):
                             dest="walltimelimit", default=None,
                             help='Wall time limit for each run, e.g. "90s" '
                                  '(overwrites wall time limit from XML file, '
-                                 'use "-1" to disable time limits completely)',
+                                 'use "-1" or the default value to use CPU time limit plus a few seconds))',
```
PhilippWendler: What is meant by "or the default value" here?

And actually, the "overwrites wall time limit from XML file" is also wrong because there is no wall-time limit in the XML file.

So I guess it would be best to write this explanation from scratch in a meaningful way for users.

IljaZakharov: I have added an option to XML format too, so this should be correct.

Is it Ok to make the following description: "Wall time limit for each run, e.g. "90s" (overwrites wall time limit from XML file, 'use "-1" to use CPU time limit plus a few seconds, such value is also used by default)" ?

PhilippWendler: I don't see the limit in the XML? But I also think that it is not that important, fixing the documentation would be enough for me.

IljaZakharov:
It is possible to specify in XML a new attribute as follows (implemented in the first commit):
``<benchmark walltimelimit="10" tool="cpachecker">``

And I changed the doc as proposed above.

PhilippWendler:
Ah, I hadn't noticed, but of course `handle_limit_value` does this automatically.

In this case, we need to add the attribute to the `doc/benchmark.dtd` file, otherwise it will not be legal XML, and we need to add documentation to `doc/benchmark.xml`, e.g.
```
walltimelimit="*optional wall-time limit, use unit 's', 'min', etc. (default: CPU time plus a few seconds)*"
```

PR #303, opened 2018-07-10
WIP: Let table generator recognize NaN values

### LINKED ISSUE #302: "nan" as a column value crashes table-generator
My understanding is that
1. any string is supposed to be a valid value of the ```value``` attribute of the ```column``` tag within a ```run``` tag in a result file and
2. there is some special magic that computes interesting statistics (sum, min, max, average, median, ...) for columns where the values are numbers.

However, if I put the string "nan" as a column value (but not for any other non-numeric string!), the table generator crashes:
```ValueError: cannot convert float NaN to integer at line 252 column 82```
I attached an [example XML file](https://github.com/sosy-lab/benchexec/files/2180602/nan-error.xml.zip) that triggers the error.

I presume the reason is that one part of the code tries to graciously interpret "nan" (in any capitalization) as the floating point (non-)value ```NaN```, while another part is not able to handle it.
Of course, this raises the question *how* to handle ```NaN```. Two options would be:
1. treat it like any other non-numeric string
2. use ```NaN``` logic for the summary rows, i.e. any column with at least one ```NaN``` would produce a ```NaN``` summary (sum, min, max, average, median, ...).

For what it's worth, in *my* case the ```NaN``` was a bug in my benchmarks, so I had to fix it on my end anyway, but in the general case, I would not expect ```NaN```s to crash the table generator.

leostrakosch:
If a column contains a NaN value, it is still
considered a column containing measurements,
but all statistics will be NaN.

Outstanding issues:
- The NaN value is not formatted for CSV output, i.e., if the result xml contains values `NaN`, `nan` and `Nan`, they will have the same different capitalization in the CSV.
- I wasn't able to take a proper look at the integration test yet.

Fixes #302

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -79,16 +79,17 @@ TEMPLATE_NAMESPACE={
 _BYTE_FACTOR = 1000 # bytes in a kilobyte
 
 # Compile regular expression for detecting measurements only once.
-REGEX_MEASURE = re.compile('\s*([-\+])?(\d+)(\.(0*)(\d+))?([eE]([-\+])(\d+))?\s?([a-zA-Z/]*)\s*$')
+REGEX_MEASURE = re.compile('\s*([-\+])?(?:([Nn][aA][Nn])|(\d+)(\.(0*)(\d+))?([eE]([-\+])(\d+))?\s?([a-zA-Z/]*))\s*$')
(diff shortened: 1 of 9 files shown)


PhilippWendler:
This should also cover `+-Inf`.

leostrakosch:
I'll handle +- Inf in #304 , which also contains this PRs content

PR #304, opened 2018-07-11
Consistent handling of column value formatting for special float values

### LINKED ISSUE #302: "nan" as a column value crashes table-generator

leostrakosch:
A simple check at the beginning of the method to check whether the column is a count or measure.

If a column contains a NaN value, it is still
considered a column containing measurements,
but all statistics will be NaN.

Outstanding issues:
* The NaN value is not formatted for CSV output, i.e., if the result xml contains values NaN, nan and Nan, they will have the same different capitalization in the CSV.

Fixes #302 and bugs that were uncovered through it.


PhilippWendler:
Thanks.

Do you know why this did not cause more problems before?

Tests will be added together with those in #303?

leostrakosch:
See https://github.com/sosy-lab/benchexec/issues/302#issuecomment-404116010 and the comment following it for an explanation.

For this specific change, we can't add any tests other than NaN and +-INF (are there others?) because
all other cases should be covered by the existing code. SoI don't think that this if-check is actually necessary, but it just makes things more explicit.

PhilippWendler:
Good idea to remove the try-catch. Now it seems we didn't understand it fully yet ;-)

Maybe this column with `1,2,3` is mistakenly detected as numeric by the regexp?

leostrakosch:
Worse, I don't even understand what my goal of this test case was:

https://github.com/sosy-lab/benchexec/blob/f162c2d14ef486c7f68b78302897f15f0d444c1b/benchexec/tablegenerator/test_numberformat.py#L44-L47

I'm testing the following: If value "1,2,3" occurs in a column of measurements,
the formatted value also is "1,2,3".
But if the value "1,2,3" occurs in a column, we should never assume that the column contains measurements!
In my opinion, this test case should be changed to actually expect the `ValueError`.

PhilippWendler:
Hm, the commit message was "Add tests for tablgenerator: Cell value starts with measurement-like string, but is none".
Such a test makes sense, and I would suggest to just replace the test with a new test with the same intention but that tests that the column type is text.

benchexec/tablegenerator/columns.py
```diff
@@ -162,7 +162,7 @@ def format_value(self, value, isToAlign=False, format_target="html"):
         if format_target not in POSSIBLE_FORMAT_TARGETS:
             raise ValueError('Unknown format target')
 
-        if value is None:
+        if value is None or value == '':
```
PhilippWendler: This should be `if value:`

leostrakosch:
I've merged the branches because of larger changes I did for comprehension and testability.

benchexec/tablegenerator/columns.py
leostrakosch: But we have to do proper formatting if `value == 0`, and `not value` would also be true for `value == 0`.

benchexec/tablegenerator/columns.py
```diff
@@ -259,13 +280,24 @@ def _format_number(number, initial_value_sig_digits, number_of_significant_digit
...
-    if number != 0:
+    if number == 0:
+        formatted_value = '0'
+        if max_digits_after_decimal > 0 and initial_value_sig_digits > 0:
+            formatted_value += '.' + '0' * min(max_digits_after_decimal, initial_value_sig_digits)
+
+    elif isnan(number):
+        formatted_value = 'NaN'
+    elif isinf(number):
+        formatted_value = 'Inf'
```
PhilippWendler: This looks like it converts all infinities to 'Inf', i.e., positive infinity, doesn't it?

benchexec/tablegenerator/util.py
```diff
@@ -180,10 +180,13 @@ def format_options(options):
     return '<span style="display:block">' + '</span><span style="display:block">'.join(line for line in lines if line.strip()) + '</span>'
 
 def to_decimal(s):
-    # remove whitespaces and trailing units (e.g., in '1.23s')
     if s:
-        s, _ = split_number_and_unit(s.strip())
-        return Decimal(s) if s else None
+        if s.lower() == 'nan':
+            return Decimal(s)
+        else:
```
PhilippWendler: Missing case for infinity here?

benchexec/tablegenerator/columns.py
leostrakosch: Yupp, thanks for the hint!

benchexec/tablegenerator/util.py
leostrakosch: Yes, this lead to statistics ignoring (+-) infinity.

PhilippWendler: But that is not correct. If there is `+Inf` in the values, of course the max value and the average in the statistics must be `+Inf`. Median and min should still have proper values, though.

leostrakosch: It should already be fixed in 691d0bd61 . I forgot to add '(erroneously)' :wink:

benchexec/tablegenerator/__init__.py
```diff
@@ -1121,30 +1122,43 @@ def __str__(self):
...
+            stdev = nan
+        elif inf in values:
+            values_sum = inf
+            mean = inf
+            stdev = inf
+        elif -inf in values:
+            values_sum = -inf
+            mean = -inf
+            stdev = inf
+        else:
```
PhilippWendler: Don't the calculations in the else branch automatically calculate the correct results?

leostrakosch: Unfortunately not, there's an invalid operation for the type Decimal on `-inf - -inf`, as well as any operation on `nan`.
As an alternative, we could first cast all Decimal values to floats and only use the else-branch,
but I expected that we use Decimals to get a higher precision than floats.

PR #309, opened 2018-08-26
Enable Java assert property

peterschrammel:


PhilippWendler:
This breaks existing users, right? While the Java mode is not really widely used, there are users (CPAchecker) and they probably have to change how they use BenchExec soon anyway (when the new metadata files come), so I don't want to make them change twice. If you prefer to have this PR merged now, please make it backwards compatible, but you could also wait until the new property format exists (which should happen in time for next SV-COMP).

PR #310, opened 2018-08-26
JBMC tool

peterschrammel:


PhilippWendler:
I wonder whether it makes sense to let the JBMC class inherit from the CBMC class, like we do with other frameworks? This would have the advantage that we don't need to call "private" (in Python convention) functions from other modules, and we don't need to manually pass around the tool instance.

I guess the question is, should the JBMC module automatically get improvements and changes made to the CBMC module, because JBMC is a different flavor of CBMC, or is it more like a fork and will probably diverge in the future?

peterschrammel:
> let the JBMC class inherit from the CBMC class

Let's try that. Done.

PR #312, opened 2018-08-27
Tool-info module for TBF

lembergerth:
tbf currently doesn't understand property files and I don't plan to implement this in the immediate future,
but I thought it would still be nice to add the tool-info module to benchexec.
Currently, when a propertyfile is provided, a warning is printed.

### DIFF
--- a/benchexec/tools/tbf.py
+++ b/benchexec/tools/tbf.py
@@ -46,8 +46,7 @@ class Tool(benchexec.tools.template.BaseTool):
         return util.find_executable('tbf', 'bin/tbf')
 
     def version(self, executable):
-        stdout = self._version_from_tool(executable)
-        return stdout.strip()
+        return self._version_from_tool(executable)
 
     def name(self):
         return 'TBF'
--- a/benchexec/tools/tbf.py
+++ b/benchexec/tools/tbf.py
@@ -1,0 +1,101 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2018  Dirk Beyer
+All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import benchexec.result as result
+import benchexec.util as util
+import benchexec.tools.template
+import logging
+import os
+
+from benchexec.model import SOFTTIMELIMIT
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for tbf (https://github.com/sosy-lab/tbf).
+    """
+
+    REQUIRED_PATHS = ["tbf", "lib"]
+
+    def program_files(self, executable):
+        files = super().program_files(executable)
+
+        if 'bin/' in executable:
+            # Get parent of directory of `executable`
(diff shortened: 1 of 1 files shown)


benchexec/tools/tbf.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Tool info for tbf (https://github.com/sosy-lab/tbf).
+    """
+
+    REQUIRED_PATHS = ["tbf", "lib"]
+
+    def program_files(self, executable):
+        files = super().program_files(executable)
+
+        if 'bin/' in executable:
```
PhilippWendler: This seems very fragile. Please use a proper check of what you actually want to do (maybe the last path segment is `bin`?).

But why do you need this check at all? What should `base_dir` be if the check returns `False`? Currently the tool-info module will just crash, right?

benchexec/tools/tbf.py
```diff
@@ -0,0 +1,101 @@
...
+            base_dir = os.path.dirname(os.path.dirname(executable))
+
+        return files + [os.path.join(base_dir, p) for p in self.REQUIRED_PATHS]
+
+    def executable(self):
+        return util.find_executable('tbf', 'bin/tbf')
+
+    def version(self, executable):
+        stdout = self._version_from_tool(executable)
+        return stdout.strip()
```
PhilippWendler: `_version_from_tool` also calls `strip`, so this can be simplified to `return self._version_from_tool(executable)`.

benchexec/tools/tbf.py
leostrakosch: My assumption was that `executable` is only the name of the executable (without path)
and I wanted to check whether the fallback had been used. But since `find_executable` includes the PATH-path for executables that are in `$PATH`, this won't work.

I wanted to distinguish between an installed version of tbf and a local directory (i.e., use of the fallback). In the first case, the python module must not be transferred because we may assume that it is installed on the system, and in the second case, it must be transferred.
(but the else-case is actually missing).

What will benchexec do if a program file cannot be found?
Would it be enough to completely remove the condition, resolve the executable correctly with `abspath`, and just always try to transfer `../tbf` and `../lib`?
Or should I check whether `../tbf/__init__.py` exists and include `REQUIRED_FILES` only then?

benchexec/tools/tbf.py
leostrakosch: Thanks!

benchexec/tools/tbf.py
PhilippWendler: Benchexec does nothing special if a program file cannot be found. It will just attempt to start the run and report something like "program cannot be found".

The question is: what do we assume the user wants if TBF is installed on the system where BenchExec is executed (instead of having it in a local directory): Should we assume that the tool is probably installed in the same place on the worker, too (and nothing should be transferred), or should we not assume this and still transfer everything?
A "middle way" of transferring parts does not seem meaningful.

I would suggest to transfer everything (if we can do this), because only then users will get the same version of TBF as installed locally.

leostrakosch: > A "middle way" of transferring parts does not seem meaningful.

Hm, if the python modue "tbf" is installed on the system, I could get its location and transfer that.
But there are two issues I see, and because of which I wouldn't want to do it:

  1. the directory will not just be transferred to the root directory of the container, but will stay in the same (relative?) path, is this right? If this is true, then we would also have to adjust the PYTHONPATH so that the worker finds that module, and that seems quiet error-prone and complicated. And
  2. if we wanted to also transfer the libraries in `lib`, we would have to hardcode the dependencies inside the tool-info module, because these may also be installed in arbitrary locations.

Thus, I strongly tend to require that tbf is executed from a directory with the repository's structure. This would greatly simplify things.

PhilippWendler: Can this be solved such that the PR can be merged?

lembergerth: Yes, sorry. I had the commit lying around, but waited (at first) whether you have anything to add - and then I forgot about it.

PR #315, opened 2018-09-14
Output results for each block unconditionally

mchalupa:
When only one block of benchmarks is executed,
the only output is the overall results
(with no "block" attribute in it).
However, when the results are processed automatically,
it is convenient to have an output for each block
no matter how many blocks we have.
This patch removes the special condition for single blocks
so that the output files have always the same structure.

### DIFF
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -489,15 +489,15 @@ class OutputHandler(object):
 
         self.add_values_to_run_set_xml(runSet, cputime, walltime, energy)
 
-        # write results to files
+        # write results to the one file with overall results
         self._write_pretty_result_xml_to_file(runSet.xml, runSet.xml_file_name)
 
-        if len(runSet.blocks) > 1:
-            for block in runSet.blocks:
-                blockFileName = self.get_filename(runSet.name, block.name + ".xml")
-                self._write_pretty_result_xml_to_file(
-                    self.runs_to_xml(runSet, block.runs, block.name),
-                    blockFileName)
+        # write results divided by blocks
+        for block in runSet.blocks:
+            blockFileName = self.get_filename(runSet.name, block.name + ".xml")
+            self._write_pretty_result_xml_to_file(
+                self.runs_to_xml(runSet, block.runs, block.name),
+                blockFileName)
 
         self.txt_file.append(self.run_set_to_text(runSet, True, cputime, walltime, energy))
 


PhilippWendler:
I think it would be confusing (and useless) to write the exact same output file twice.

PhilippWendler:
The current logic is:
- BenchExec always writes the full results to `...results.xml`.
- If you have invested the time to split your tasks into subsets, you might be interested in evaluating them individually, so BenchExec additionally produces the results files for individual blocks.

The first point is what most users need, use, and rely on, so this should not be changed.

Actually I think that these additional output files for blocks should go away completely and be replaced by additional filtering capabilities of the HTML table (and maybe table-generator, too).

For the current issue, do you have a concrete example where the current scheme is problematic?

PhilippWendler:
Ok, but this seems like it is better solved by consistently using the `block` attribute in the result XML file, right?

mchalupa:
Well, yes, that was the aim of the PR, but I wanted to do this more generally -- for the case that user wants to distinguish between "overall" results and "block" results (based on the presence of `block` attribute).
Anyway, I'm completely ok with getting rid of the overall results when there is just one block (which would amount to "consistently using the `block` attribute", right?).

mchalupa:
> I am not sure if there is a misunderstanding. I do not want to get rid of the "overall" results, just add the block attribute to the XML file that is written in case only a single block of tasks is executed.

That is what exactly happens when you do not emit the "overall" results but emit only the block file when the block is just one (since the files differ only in the `block` attribute)

> [...] This results in differently named result XML files (as expected), but when generating tables from this everything works as it should.

I do not use table generator, I parse the XML files with our scripts.

> And furthermore, while trying this out I noticed that I was completely wrong when stating the current behavior of BenchExec. BenchExec already does create a file named ...results.<block>.xml if restricted to a single block (and it creates only this file).

Yes, but this file does not contain the `block` attribute.

PhilippWendler:
> > I am not sure if there is a misunderstanding. I do not want to get rid of the "overall" results, just add the block attribute to the XML file that is written in case only a single block of tasks is executed.
> 
> That is what exactly happens when you do not emit the "overall" results but emit only the block file when the block is just one (since the files differ only in the `block` attribute)

But writing to the same file twice still doesn't make sense, so the implementation should be different from this PR.
Furthermore, benchexec writes partial results files during run execution, and with this PR the block attribute would only be set in the final file, not in the intermediate files.

So I added a different solution in 542e7d1fe90ddefcc2afd6dc48898fa490cda041.

> > [...] This results in differently named result XML files (as expected), but when generating tables from this everything works as it should.
> 
> I do not use table generator, I parse the XML files with our scripts.

Ok, this is an important information that would have prevented some earlier misunderstandings.

Though I wonder why your scripts rely on the block attribute if table-generator doesn't?

PR #316, opened 2018-10-12
Updated Verifuzz.py

animeshbchowdhury:
Required for qualification run in SVCOMP 2019


benchexec/tools/verifuzz.py
```diff
@@ -49,8 +49,18 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
         lines = " ".join(output)
         if "VERIFUZZ_VERIFICATION_SUCCESSFUL" in lines:
             return result.RESULT_TRUE_PROP
-        elif "VERIFUZZ_VERIFICATION_FAILED" in lines:
+        elif "FALSE(unreach-call)" in lines:
```
PhilippWendler: This change makes the tool-info module non-backwards compatible. How about checking for both strings?

animeshbchowdhury: Hi Phillip,

I will check this once with our testing. I think it is doable. Once done, I will re-raise the PR.

PR #318, opened 2018-10-17
add new tool tabol

xuechaofelix:
Add a new tool, called "tabol", for termination analysis

### DIFF
--- a/benchexec/tools/tabol.py
+++ b/benchexec/tools/tabol.py
@@ -19,7 +19,7 @@ import benchexec.util as util
 import benchexec.tools.template
 
 class Tool(benchexec.tools.template.BaseTool):
-    REQUIRED_PATHS = []
+    REQUIRED_PATHS = ['tabol.sh','tabol.jar','output/','tools']
 
     def executable(self):
         return util.find_executable('tabol.sh')
@@ -28,7 +28,9 @@ class Tool(benchexec.tools.template.BaseTool):
         return 'TABOL'
     def cmdline(self, executable, options, tasks, propertyfile, rlimits):
         return [executable] + options + tasks
-
+    
+    def version(self, executable):
+        return self._version_from_tool(executable)
     def determine_result(self, returncode, returnsignal, output, isTimeout):
 
         status = result.RESULT_UNKNOWN
--- a/benchexec/tools/tabol.py
+++ b/benchexec/tools/tabol.py
@@ -1,7 +1,7 @@
 """
 BenchExec is a framework for reliable benchmarking.
 This file is part of BenchExec.
-Copyright (C) 2007-2015  Dirk Beyer
+Copyright (C) 2007-2018  Dirk Beyer
 All rights reserved.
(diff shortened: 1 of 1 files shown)


benchexec/tools/tabol.py
```diff
@@ -0,0 +1,47 @@
...
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+
+        status = result.RESULT_UNKNOWN
+        stroutput = str(output)
+
+        if isTimeout:
+            status = 'TIMEOUT'
+        elif 'TABOL_TRUE' in stroutput:
+            status = result.RESULT_TRUE_PROP
+		elif 'TABOL_FALSE' in stroutput:
```
PhilippWendler: The indentation is wrong here, doesn't this create a syntax error?

benchexec/tools/tabol.py
```diff
@@ -0,0 +1,47 @@
...
+        return 'TABOL'
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+		
+        return "java -jar " + [executable] + tasks
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+
+        status = result.RESULT_UNKNOWN
+        stroutput = str(output)
```
PhilippWendler: `output` is a list of lines, so `str(output)` will give something like `['line1', 'line2']`. This will in principle work, but might be unexpected during future changes. Something like `"\n".join(output)` might be more intuitive (or just loop over the lines).

benchexec/tools/tabol.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+import benchexec.result as result
+import benchexec.util as util
+import benchexec.tools.template
+
+class Tool(benchexec.tools.template.BaseTool):
+
+    def executable(self):
+        return util.find_executable('tabol.jar')
```
PhilippWendler: Did you test that this works? This should require `tabol.jar` to be executable, which JARs usually aren't.

What is the output of the [test](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#testing-the-tool-integration)?

xuechaofelix: I changed the jar file to a shell file now! then it works.

benchexec/tools/tabol.py
```diff
@@ -19,29 +19,27 @@
 import benchexec.tools.template
 
 class Tool(benchexec.tools.template.BaseTool):
+    REQUIRED_PATHS = []
```
PhilippWendler: Could you please add a proper set of required paths? E.g., `tabol.jar`, `tabol.sh` and whatever else is necessary.

This is not necessary for BenchExec itself, but for being able to automatically distribute the tool across various machines.

PhilippWendler:
Thanks for the fixes. What about the version number?

benchexec/tools/tabol.py
xuechaofelix: Thanks for your suggestion! Then, I add some requried path and the version number.

PR #321, opened 2018-10-18
Add tool-info module for CBMC Path

karkhaz:
This module is identical to CBMC's tool-info module, except that it
returns "CBMC Path" as the tool name. The `--paths` switch is specified
in the wrapper script `cbmc`, which invokes `cbmc-binary`.

### DIFF
--- a/benchexec/tools/cbmc-path.py
+++ b/benchexec/tools/cbmc-path.py
@@ -1,0 +1,30 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+Copyright (C) 2007-2018  Dirk Beyer
+All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


karkhaz:
Ping @tautschnig

PhilippWendler:
I suggest to use the regular tool-info module for CBMC, if there is no necessary change?

Otherwise it would be better to inherit from the CBMC module, like it is done for JBMC, for example.

PhilippWendler:
The name is used only for human-readable purposes. If you look at an [example table](https://sv-comp.sosy-lab.org/2018/results/results-verified/META_MemSafety.table.html), the result of `name()` is used for the top-most row (together with the version number). The other occurrences of tool names already come from different places.

karkhaz:
Okay, then I would recommend to keep a separate tool-info module that inherits everything from CBMC apart from `name()`, which I've just pushed.

PR #322, opened 2018-10-19
Tool-info pinaka.py added for SVCOMP 2019

EtiChaudhary:


PhilippWendler:
Please add a license to the file. If you are ok with it, you can copy the license from e.g., `template.py` (this transfers copy right to Dirk Beyer and assigns the Apache 2.0 License).

Does your tool have a version number? If yes, please add a `version` function that reads the version number from the tool and returns it (you can use the utility function from `template.py`).

EtiChaudhary:
I have added the license.
No we don't have a version number. The version function has been intentionally left out.

Thanks.

PR #323, opened 2018-10-19
Add PeSCo tool info for 2019

cedricrupb:


PhilippWendler:
Please do not copy code, instead use inheritance to reuse the code. You can check `cpa-witness2test.py` to see how to inherit from other tool-info modules, and then overwrite only those functions you need.

PR #324, opened 2018-10-19
Version that uses only jpf-core; URL update for existing file.

cyrille-artho:
new file:   jpf-core.py: Copy of jpf.py that uses only jpf-core.
modified:   jpf.py: Updated repo URL.

### DIFF
--- a/benchexec/tools/jpf-core.py
+++ b/benchexec/tools/jpf-core.py
@@ -27,8 +27,7 @@ class Tool(benchexec.tools.template.BaseTool):
 
     REQUIRED_PATHS = [
                   "jpf-core/bin",
-                  "jpf-core/build",
-                  "jpf-sv-comp"
+                  "jpf-core/build"
                   ]
     def executable(self):
         return util.find_executable('jpf-core-sv-comp')
--- a/benchexec/tools/jpf-core.py
+++ b/benchexec/tools/jpf-core.py
@@ -21,19 +21,17 @@ import benchexec.result as result
 
 class Tool(benchexec.tools.template.BaseTool):
     """
-    Tool info for JPF with symbolic extension
+    Tool info for JPF (plain jpf-core)
     (https://github.com/javapathfinder/jpf-core/).
     """
 
     REQUIRED_PATHS = [
-                  "jpf-core/bin/jpf",
+                  "jpf-core/bin",
                   "jpf-core/build",
-                  "jpf-symbc/lib",
-                  "jpf-symbc/build",
                   "jpf-sv-comp"
                   ]
     def executable(self):
-        return util.find_executable('jpf-sv-comp')
+        return util.find_executable('jpf-core-sv-comp')
 
 
     def version(self, executable):


PhilippWendler:
Why are two tool-info modules needed for JPF? Both return "JPF" as name, this would be confusing for users.

It there is only a single tool, it should ideally be handled by a single tool-info module. If there are two distinct tools, there names should be different.

Furthermore, please do not copy code, instead use inheritance to reuse the code. You can check `jbmc.py` to see how to inherit from other tool-info modules, and then overwrite only those functions you need (if two tool-info modules are actually needed).

benchexec/tools/jpf.py
```diff
@@ -22,7 +22,7 @@
 class Tool(benchexec.tools.template.BaseTool):
     """
     Tool info for JPF with symbolic extension
-    (https://babelfish.arc.nasa.gov/hg/jpf).
+    (https://github.com/javapathfinder/jpf-core/).
```
peterschrammel: If we have both spf and jpf now then I'd prefer renaming this file to `spf.py`

peterschrammel: The name returned should then be `SPF` in line 44 below.

benchexec/tools/jpf-core.py
```diff
@@ -0,0 +1,60 @@
...
+    def executable(self):
+        return util.find_executable('jpf-core-sv-comp')
+
+
+    def version(self, executable):
+        return 'rev '+self._version_from_tool(executable).split()[2]
+
+
+    def name(self):
+        return 'JPF'
```
peterschrammel: Is there a JPF (other than SPF) that is not JPF-core? If so, then I'd return `JPF-core` here.

benchexec/tools/jpf-core.py
```diff
@@ -0,0 +1,60 @@
...
+import benchexec.result as result
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for JPF (plain jpf-core)
+    (https://github.com/javapathfinder/jpf-core/).
+    """
+
+    REQUIRED_PATHS = [
+                  "jpf-core/bin",
```
peterschrammel: I'm not sure what benchexec is using this information for. But afaik, the top-level directory in the verifier archive is stripped off. Hence the path should be `bin` and `build` respectively. Or, if everything in the current zip file is put into an additional directory, then these paths would be correct.

PhilippWendler:
This has now conflicts in `jpf.py`.

Is JPF Core based on SPF? Then please let `jpf-core.py` inherit from `spf.py`. Similarly if SPF is based on JPF Core.

Btw., from @peterschrammel's explanation I understood that JPF Core is basically JPF, right? So shouldn't this tool be named "JPF" and the tool-info module `jpf.py`?

cyrille-artho:
Philipp Wendler wrote:
> This has now conflicts in |jpf.py|.
> 
> Is JPF Core based on SPF? Then please let |jpf-core.py| inherit from 
> |spf.py|. Similarly if SPF is based on JPF Core.
> 
> Btw., from @peterschrammel <https://github.com/peterschrammel>'s 
> explanation I understood that JPF Core is basically JPF, right? So 
> shouldn't this tool be named "JPF" and the tool-info module |jpf.py|?
> 
In that case, SPF should inherit from JPF. Please also rename jpf-core.py 
to jpf.py; I'll update that in the submission form later.

-- 
Regards,
Cyrille Artho - http://artho.com/
If someone had told me I would be Pope one day, I would have studied harder.
		-- Pope John Paul I

peterschrammel:
@cyrille-artho, @PhilippWendler, please have a look at https://github.com/sosy-lab/benchexec/pull/333 to check whether this is what you intended.

PR #326, opened 2018-10-20
Improve error reporting in DIVINE 4 tool module

vlstill:


benchexec/tools/divine4.py
```diff
@@ -114,7 +114,7 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
         (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
         """
 
-        if not output:
+        if not output or output[-1] == None:
```
PhilippWendler: This doesn't make sense: `output` is a list of strings, it will never contain `None`.
Furthermore, one cannot compare with `None` using `==` in Python ([explanation](https://stackoverflow.com/q/14247373/396730)).

vlstill:
@PhilippWendler I have removed the test for `None` (it was moved, and wrongly, and I am no longer sure why it was there).

PR #328, opened 2018-10-21
Rename JPF to SPF

peterschrammel:
JPF is now JPF-core. Therefore it makes more sense to call the symbolic version SPF.


peterschrammel:
@dbeyer, this one would be required to run SPF.

PhilippWendler:
How does this relate to #324?

(Note that for SV-COMP, if you submit a tool twice in two different configurations, you should not add two tool-info modules to BenchExec, just one. Tool-info modules should only be added for separate tools.)

peterschrammel:
To resolve the confusion:
- Initially, I created a jpf.py for the SPF tool (symbolic execution version of JPF)
- Then Cyrille also wanted to submit the 'plain' version of JPF that he called JPF-core.
- Then we came to the conclusion that this is confusing and jpf.py should be renamed spf.py

So, these are _different_ tool variants (_more_ than just different configurations).

jpf-core.py could inherit from spf.py or vice versa. This is doable.

peterschrammel:
To resolve the confusion:
- Initially, I created a jpf.py for the SPF tool (symbolic execution version of JPF)
- Then Cyrille also wanted to submit the 'plain' version of JPF that he called JPF-core.
- Then we came to the conclusion that this is confusing and jpf.py should be renamed spf.py

So, these are _different_ tool variants (_more_ than just different configurations).

jpf-core.py could inherit from spf.py or vice versa. This is doable.

peterschrammel:
To resolve the confusion:
- Initially, I created a jpf.py for the SPF tool (symbolic execution version of JPF)
- Then Cyrille also wanted to submit the 'plain' version of JPF that he called JPF-core.
- Then we came to the conclusion that this is confusing and jpf.py should be renamed spf.py

So, these are _different_ tool variants (_more_ than just different configurations).

jpf-core.py could inherit from spf.py or vice versa. This is doable.

peterschrammel:
To resolve the confusion:
- Initially, I created a jpf.py for the SPF tool (symbolic execution version of JPF)
- Then Cyrille also wanted to submit the 'plain' version of JPF that he called JPF-core.
- Then we came to the conclusion that this is confusing and jpf.py should be renamed spf.py

So, these are _different_ tool variants (_more_ than just different configurations).

jpf-core.py could inherit from spf.py or vice versa. This is doable.

PR #332, opened 2018-10-23
make memcleanup property orthogonal to other memsafety properties.

kfriedberger:


dbeyer:
This is a follow-up PR for #330.

PR #341, opened 2018-11-10
AProVE: extract version number

jhensel:


PhilippWendler:
It is not necessary to cache the results of these functions, BenchExec will call them only once anyway (and if caching was useful, we would do so centrally).

There is also no reason to have `_aprove_version` as a separate method, so I suggest to inline it for simplicity.

PhilippWendler:
While you are at it, you might want to consider adding documentation as [suggested here](https://github.com/sosy-lab/benchexec/commit/bd82323567b2c8ebade40b8778460de1dda53308#diff-564acd0f58dd345e8d161cb791d6af53). Would be great. An example is [here](https://github.com/sosy-lab/benchexec/blob/bd82323567b2c8ebade40b8778460de1dda53308/benchexec/tools/cpachecker.py).

PhilippWendler:
I implemented the simplifications in 59be692104. If you would like to add the suggested documentation this would still be great.

jhensel:
Thanks! I just opened PR #350 for the documentation.

PR #361, opened 2018-11-15
First try to find DIVINE in relative locations.

vlstill:
@dbeyer, I am not sure what was the problem in your testing, I have fixed search for DIVINE in current directory (if it is the verifier archive submitted), it the problem is something else please let me know what is the problem.

### DIFF
--- a/benchexec/tools/divine4.py
+++ b/benchexec/tools/divine4.py
@@ -53,6 +53,11 @@ class Tool(benchexec.tools.template.BaseTool):
         and most implementations will look similar to this one.
         The path returned should be relative to the current directory.
         """
+        def is_exec(fpath):
+            return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Thinking about it, the tool-info module seems to pretty confusing. For example, `cmdline` should typically start the `executable`, but it starts something else.

Could you please explain the expected layout of your tool installation and what is considered the executable, such that this can be cleaned up?

vlstill:
There are two parts, the DIVINE itself and an SV-COMP wrapper, `divine-svc`, version has to be read from DIVINE directly, but programs should be analyzed using the wrapper. In the submitted archive, both live in the `divine-svc` directory, which is the only directory in the archive.

PhilippWendler:
@vlstill Thanks for the answer. I have updated the tool-info module using this information and improved and simplified the code. Please review #367 to make sure that it matches your intention.

PR #363, opened 2018-11-15
updated spf.py script

yannicnoller:


PhilippWendler:
This now duplicates two methods from `jpf.py`, what is the reason for that?

For `version` there should be error handling. But doesn't the `version` method from `jpf.py` make more sense (it calls `jpf -version`) than reading from `.version`?

The version `1` should not be hard-coded, not even as fallback (if in a future version the file `.version` is not present anymore, a wrong version would be shown).

yannicnoller:
Hi Phillip,

JPF and SPF are different tools. Using the version from JPF makes no sense.

SPF is no versioning system so far. We talked to Dirk, and we agreed that for now the solution with the .version file is okay.

We adapted the jpf.py script for SPF, so there might be some overlap for the functions, but as I a said, these are different tools, so we want to / need to have our own script.

Thanks
Yannic

dbeyer:
Just for the record:  I wrote the following on Nov 15, 20:39 CET:
```
Hi Yannic,

it is short, because it inherits from jpf.py.

Please do not use the .version file.  Cyrille changed his mind and
now uses the proper call of the tool.

Just run a test of the tool-info module and if everything is good that we are happy.

Best,
Dirk

> Hi Dirk,
>
> we can add a .version file, this is no problem. Are we also supposed to change the benchexec/tools/spf.py file?
> 
> The current code for spf is very short, compared to the other tools:


```

dbeyer:
The desired solution is definitely to call `jpf -version`.
The previous `.version` alternative was created when the assumption was that jpf does not have version support. Now we know that it apparently has a parameter to give a version string, which is really much better.

corinus:
Hi:
SPF is different than JPF so we can not use jpf -version.

corinus:
I think .version file is fine for now until we find a better solution. 
We can not use jpf -version because it would not be true.

yannicnoller:
@PhilippWendler I agree on the error handling. We will submit an improved version later. Versioning is something we need to implement in SPF, then we can improve the script.

yannicnoller:
@PhilippWendler @dbeyer I updated our spf.py script. It does not call our tool via the jpf-sv-comp bash script in the archives repository. The current solution their gets the timestamp of jpf-symbc.jar file; this timestamp is set at compilation time and won't change during the experiment runs.

Here is the output.log for the test script. There are some warning left, but I think this are okay for now? Please check.

[output.log](https://github.com/sosy-lab/benchexec/files/2590298/output.log)

PR #365, opened 2018-11-16
Update version number.

franck44:


PhilippWendler:
Hard-coding a version number in this module is not possible. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

franck44:
Ok I can add a version option to our tool but I would like a pre-run to be triggered asap.
Could you accept this skink.py for the time being with the hard coded version and I'll update the skink.py and our archive over the WE and push them on Monday. I am away and won't be able to download/upload 7GB in any reasonable time otherwise.

kfriedberger:
To avoid the large overhead of transferring several GB of data, you can use an existing old repository of archives, create a new branch from an old master, commit the new version of the tool into that new branch and push this new branch onto the server. This works without pulling anything. 
Then create a merge-request from your branch via the website. The CI might fail due to an outdated check-script, but it is worth a try. And the organizers could even merge a failing CI if the result after the merge is expected to be valid.

PhilippWendler:
@franck44 Note that BenchExec is a general tool, and not specific for SV-COMP and its pre-runs.

However, this PR it its current form (with hard-coded version number) would not even help for the SV-COMP pre-runs: Version numbers are required now for SV-COMP in order to be able to identify the tool version used in a specific pre-run (among other reasons), but if the version number is hard-coded in the tool-info module this goal is not achieved at all, because now all pre-runs will just be labeled with "2.0" instead of "" as version number (effectively no difference).

And worse: if later on somebody uses this version of BenchExec to benchmark a different version of your tool, they will get benchmark results with a wrong version number, which can lead to all sorts of misunderstandings and wrong conclusions.

So a hard-coded version number does not even have no advantage at all over having no version number, but is even strictly worse.

PR #373, opened 2018-11-19
Pinaka.py updated to handle added categories

EtiChaudhary:


benchexec/tools/pinaka.py
```diff
@@ -37,14 +37,23 @@ def version(self, executable):
     def name(self):
         return "Pinaka"
 
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        if propertyfile:
+            options = options + ['--propertyfile', propertyfile]
+
+        self.options = options
```
PhilippWendler: This line seems useless, please remove.

EtiChaudhary: Done!

PR #377, opened 2018-11-25
Various updates for Ultimate tools

danieldietsch:


PhilippWendler:
While you are at it, could you please replace the remaining cases of `NameError`, too?
`BenchExecException` would be ok for those, I guess.

PR #378, opened 2018-11-26
Fix Ultimate tool-info

danieldietsch:
I made some stupid mistakes -- these commits fix the Ultimate tools.

### DIFF
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -25,13 +25,12 @@ import re
 import subprocess
 import sys
 
-from benchexec import BenchExecException
-from benchexec.tools.template import UnsupportedFeatureException
-
 import benchexec.result as result
 import benchexec.tools.template
 import benchexec.util as util
+from benchexec import BenchExecException
 from benchexec.model import MEMLIMIT
+from benchexec.tools.template import UnsupportedFeatureException
 
 _OPTION_NO_WRAPPER = '--force-no-wrapper'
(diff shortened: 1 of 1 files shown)


PhilippWendler:
I see that `'-Xss4m'` is also added in the code further down, is it really intended to have it in both places?

danieldietsch:
For versions it is not really necessary, but I wanted to have it for consistency. Further down it is necessary, because there we have to rebuild the string if someone specifies a memory limit.

danieldietsch:
By the way, is it expected that benchexec converts ``memlimit="8GB"`` to ``8000000000`` bytes in ``rlimits`` ?
Thats whats happening in this line:
 https://github.com/sosy-lab/benchexec/blob/ec1a3193d16ff7279ebe5b42bf2e48dc82e6413f/benchexec/tools/ultimate.py#L215

PhilippWendler:
Yes, benchexec handles the units for such parameters and converts them to bytes. This way tool-info modules do not need to care about supporting units in the user input.

danieldietsch:
But 8GB (Gigabyte) are 8589934592 Byte, not 8000000000 Byte.

PhilippWendler:
No, 8 GB (SI units) are 8000000000 Byte, 8 GiB are 8589934592 Byte. BenchExec always uses SI units as [documented](https://github.com/sosy-lab/benchexec/blob/master/doc/INDEX.md#units).

danieldietsch:
Ok, got it. Then I use GiB.

PR #380, opened 2018-11-27
symbiotic: update tracking phases

mchalupa:


PhilippWendler:
I am waiting for response from the PR author.

mchalupa:
> Maybe it makes sense to make the output of the tool structured in a way that the tool-info module can recognize it without a hard-coded list of phases?

What do you mean by this?

> But if you prefer, I can merge this now, too.

Yes, please, merge it, I can send a follow-up PR.

PhilippWendler:
I do not know, because it depends on your tool, but certainly there is a way to have the output in a format such that the tool-info module can extract the latest phase without needing to know the names of all existing phases?

PR #385, opened 2018-12-04
Add result 'cover' for test-case generators.

lembergerth:
The verdict signalizes that the used tool created a test suite
that tries to achieve a given coverage goal.
(as opposed to looking for a property violation and saying TRUE/FALSE.)

### DIFF
--- a/benchexec/result.py
+++ b/benchexec/result.py
@@ -107,6 +107,8 @@ RESULT_SAT =                'sat'
 """task is satisfiable"""
 RESULT_UNSAT =              'unsat'
 """task is unsatisfiable"""
+RESULT_COVER =              'cover'
+"""test suite for coverage criterion created"""
 
 # List of all possible results.
 # If a result is not in this list, it is handled as RESULT_CLASS_ERROR.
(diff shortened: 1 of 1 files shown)


lembergerth:
If I use benchexec with a test-suite generator, the results TRUE_PROP and FALSE_PROP
do not fit, as we are not checking a program property,
but trying to cover a coverage criterion.

Currently, the only possibility in such a case is to use result "UNKNOWN",
as any result not known to benchexec will be interpreted as an error (according to the doc).

PhilippWendler:
Would returning `cover` and having it treated as error a problem for your use case?
Or having it return `true` or `unknown`?

lembergerth:
> Would returning `cover` and having it treated as error a problem for your use case?

No, but wouldn't table-generator color all of these results in pink/violet then?
I think it would be nice to be able to visually differentiate between error and success in the created HTML tables, if possible.

> Or having it return true or unknown?

I'd resort to unknown if the COVER as a result really didn't fit benchexec.
I'm afraid `true` may give a wrong impression for people from communities where this has a specific meaning, e.g., people from the SV-COMP community.

Benchexec has result-values for property checking (TRUE, FALSE) and for checking satisfiability (SAT, UNSAT), so wouldn't it be consistent to also add a result-value for test-suite generation, to natively support this additional application?

PR #388, opened 2018-12-11
Tool-Info for test-generation with CPAchecker

Eiram:
Do not report TRUE, UNKNOWN, TIMEOUOT, etc., but DONE, which better describes what test-generation is doing and better fits to the rules of the testing competition.

### DIFF
--- a/coveritest.py
+++ b/coveritest.py
@@ -1,0 +1,37 @@
+# BenchExec is a framework for reliable benchmarking.
+# This file is part of BenchExec.
+#
+# Copyright (C) 2007-2018  Dirk Beyer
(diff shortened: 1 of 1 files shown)


dbeyer:
Re: tool name
Using the change in PR #398, the tool name `CoVeriTest` can be shown using the tool-info module `cpachecker.py`, and this tool-info module is most likely not needed anyway.

dbeyer:
Seems unnecessary after the change in PR #398.

PhilippWendler:
I had discussed with @Eiram that this PR is unnecessary now anyway.

PR #393, opened 2018-12-28
Tool-info module for tbf testsuite validator

lembergerth:
This PR provides a tool-info module for the tbf test-suite validator, as used in Test-Comp.


lembergerth:
Thanks for the quick merge!

PR #396, opened 2018-12-30
Add cpa-tiger Tool Info

SRuland:
Add cpa-tiger Tool Info

### DIFF
--- a/benchexec/tools/cpa-tiger.py
+++ b/benchexec/tools/cpa-tiger.py
@@ -25,7 +25,3 @@ class Tool(cpachecker.Tool):
 
     def name(self):
         return 'CPA-tiger'
-
-    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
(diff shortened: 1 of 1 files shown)


benchexec/tools/cpa-tiger.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Tool(cpachecker.Tool):
+    """
+    Tool info for CPA-tiger.
+    """
+
+    def name(self):
+        return 'CPA-tiger'
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
```
PhilippWendler: Why the override of this method? It seems redundant?

dbeyer:
This was indeed an override with identical code.

dbeyer:
Using the change in PR #398, this tool-info module is most likely not needed anyway.

dbeyer:
Seems unnecessary after the change in PR #398.

Issue #384, opened 2018-12-03
Use container mode in tool-info module for version information


PhilippWendler:
Many tool-info modules execute the tool to get the version number. In order to prevent the tool from (accidentally) modifying files, this execution should happen in a container, just like the actual run execution.

PhilippWendler:
A tricky part of this is that of course container mode should be used only if it is also active for the actual runs, but the configuration that contains this information is not available to the tool-info module.

dbeyer:
Related to issue #173.

PhilippWendler:
Actually not related. #173 is about a cgroup, which is completely independent from the container mode, and would not be implemented in a similar manner.

dbeyer:
> A tricky part of this is that of course container mode should be used only if it is also active for the actual runs, but the configuration that contains this information is not available to the tool-info module.

Why?  I would say that independently from how `runexec` starts the tool, `benchexec` itself should be executed as safely and securely as possible.

PhilippWendler:
Container mode is not possible on all systems. So if the user passes `--no-container`, we must not use the container mode, neither for the actual runs nor for getting the version information.

dbeyer:
> Actually not related. #173 is about a cgroup, which is completely independent from the container mode, and would not be implemented in a similar manner.

The abstract idea is very related ;-) and my comment in that issue #173 was actually referring to this aspect already, as I interpreted the title of the issue a bit more widely. I will drop a comment there.

dbeyer:
> Container mode is not possible on all systems.

I know. But `benchexec` can find out itself.

> So if the user passes `--no-container`, we must not use the container mode, neither for the actual runs nor for getting the version information.

Not necessarily.  I would interpret the option `--no-container` as relating to how the process invocation happens for the experiment execution, by `runexec`.  For the `benchexec` execution, I would simply assume that it is executed as securely as possible.

In other words, just from the user perspective, if name spaces are available, I would expect `benchexec` to execute itself in a new name space, and if name spaces are not available, I would expect it to print a warning that (a) files are not guaranteed to stay valid and (b) the tool is executed with a full view of the host system.

... just my immediate thoughts from SV-COMP execution experience.

kfriedberger:
What about the following solution:
If container-mode is wanted and available, then use it for the version-info.
Otherwise check whether any file or directory that would be sent into the VerifierCloud was modified and warn the user about it. 
[...]

PR #398, opened 2018-12-30
Get tool name from executable

### LINKED ISSUE #397: Tool-info module with flexible tool name
https://github.com/sosy-lab/benchexec/blob/ac991eabb53cfd1e4e7540b0b0d0abd2b1afa5ee/benchexec/tools/cpachecker.py#L92

Is it possible to ask the binary for the tool name, instead of hardcoding the tool name in the tool-info module?

Then we would not need additional tool-info modules for CPAtiger, CoVeriTest, etc.

dbeyer:
Get tool name from executable, instead of having it hard-coded in the tool-info module; fixes #397.

### DIFF
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -83,15 +83,15 @@ class Tool(benchexec.tools.template.BaseTool):
 
     def version(self, executable):
         stdout = self._version_from_tool(executable, '-help')
-        line = next(l for l in stdout.splitlines() if l.startswith('CPAchecker'))
-        line = line.replace('CPAchecker' , '')
+        line = next(l for l in stdout.splitlines() if l.startswith('CPA'))
+        line = line.replace(self.name() , '')
         line = line.split('(')[0]
         return line.strip()
 
     def name(self):
         executable = self.executable()
         stdout = self._version_from_tool(executable, '-help')
-        line = next(l for l in stdout.splitlines() if l.startswith('CPAchecker'))
+        line = next(l for l in stdout.splitlines() if l.startswith('CPA'))
         line = line.split(' ')[0]
         return line.strip()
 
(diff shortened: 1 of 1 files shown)


dbeyer:
The code in functions `name` and `version` of the tool-info module is quite brittle and should be refactored soon.

Best would be to ask CPAchecker for an option `-version` to give name and version number,
in order to make the parsing in those two functions of the tool-info module unnecessary.

PhilippWendler:
> Best would be to ask CPAchecker for an option `-version` to give name and version number,
> in order to make the parsing in those two functions of the tool-info module unnecessary.

What difference would that make? The output of `-version` would look the same as for `-help`, just without the rest of the help text, which is ignored anyway.

Furthermore, even if CPAchecker would get such an option, the tool-info module would need to keep using `-help` for backwards compatibility.

However, the tool-info module for CPAchecker relies on the fact that `self.executable()` is called only once (because this method builds CPAchecker), but now it is called three times. This needs to be changed.

dbeyer:
The following comment should be treated independently from whether the tool-info module
returns a constant string as tool name or asks the executable for its name.
(I currently tent to revert this change.)

> However, the tool-info module for CPAchecker relies on the fact that `self.executable()` is called only once (because this method builds CPAchecker), but now it is called three times. This needs to be changed.

I would say a tool-info module should not rely on assumptions such as how many times a tool is called.
In fact, a tool should not have a state. Ideally, once issue #384 is implemented,
there will be no difference no matter how often a tool is called.

Also, building is not a concern of BenchExec and is not expected to be done in the tool-info module.
I am actually not strictly against this (nice convenience feature), but if this feature gets in the way and causes problems, then it should be removed.

dbeyer:
Depends on issue #412.

PR #403, opened 2019-01-11
KLEE: Updates for cloud-based run environment; Parsing of output

MartinNowack:


benchexec/tools/klee.py
```diff
@@ -52,6 +59,15 @@ def version(self, executable):
         line = line.split('(')[0]
         return line.strip()
 
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        if "memlimit" in rlimits:
+            options += ["--max-memory="+str(rlimits["memlimit"])]
+        if "walltimelimit" in rlimits:
+            options += ["--max-walltime="+str(rlimits["walltimelimit"])]
+        if "timelimit" in rlimits:
```
PhilippWendler: Probably you want to check for `softtimelimit` here. The soft time limit defines how long a run should take at most (results produced later are not counted), whereas `timelimit` is the limit after which the tool is forcefully killed.

benchexec/tools/klee.py
```diff
@@ -52,6 +59,15 @@ def version(self, executable):
         line = line.split('(')[0]
         return line.strip()
 
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        if "memlimit" in rlimits:
```
PhilippWendler: You can import constants for these strings from `benchexec.model`.

MartinNowack:
@PhilippWendler Thanks a lot for the feedback. I changed the code according to your suggestions.

PR #405, opened 2019-01-13
Fixes #404 (bug in method 'score_for_task')

### LINKED ISSUE #404: Method 'score_for_task' fails for properties that are not a SAT or SV-COMP property
If the given property is not a SAT and not an SV-COMP property, then method score_for_task calls _svcomp_score, which in turn fails.

dbeyer:


PhilippWendler:
Replaced by ec2b865.

PR #406, opened 2019-01-13
Let table generator support custom score-columns in result XMLs.

lembergerth:
Currently, table-generator computes a score internally and does not use columns titled 'score', if they exist in the given result XML.
This makes it impossible to create own score-schema without using names that are different from the established 'score'.
This PR lets table-generator favor such columns over an internally computed score.

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -697,10 +697,10 @@ class RunResult(object):
 
             if column.title.lower() == 'score':
                 if value is None:
+                    # If no score column exists in the xml, take the internally computed score,
+                    # if available
                     value = str(score) if score is not None else None
                 else:
-                    # If a score column exists in the xml, we use that column's value for the score, instead of
-                    # the computed score.
                     score = float(value)
             values.append(value)
 
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -698,7 +698,7 @@ class RunResult(object):
             if column.title.lower() == 'score' and value is None and score is not None:
                 # If no score column exists in the xml, take the internally computed score,
                 # if available
-                value = str(score) if score is not None else None
+                value = str(score)
             values.append(value)
 
         sourcefiles = sourcefileTag.get('files')
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -700,12 +700,6 @@ class RunResult(object):
                     # If no score column exists in the xml, take the internally computed score,
                     # if available
                     value = str(score) if score is not None else None
-                else:
-                    try:
-                        score = float(value)
-                    except ValueError:
-                        logging.error("Column 'score' is selected, but contains non-number value: %s", value)
-                        exit(1)
             values.append(value)
 
         sourcefiles = sourcefileTag.get('files')
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -695,8 +695,7 @@ class RunResult(object):
 
                     value = get_value_from_logfile(logfileLines, column.pattern)
 
-            if column.title.lower() == 'score':
-                if value is None:
+            if column.title.lower() == 'score' and value is None and score is not None:
                     # If no score column exists in the xml, take the internally computed score,
                     # if available
                     value = str(score) if score is not None else None
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -681,9 +681,7 @@ class RunResult(object):
 
         for column in listOfColumns: # for all columns that should be shown
             value = None # default value
-            if column.title.lower() == 'score':
-                value = str(score) if score is not None else None
-            elif column.title.lower() == 'status':
+            if column.title.lower() == 'status':
                 value = status
 
             elif not correct_only or category == result.CATEGORY_CORRECT:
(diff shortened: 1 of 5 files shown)


dbeyer:
Very nice.
And once issue #404 is fixed, `value = str(score)` will work in all cases.

benchexec/tablegenerator/__init__.py
```diff
@@ -697,6 +695,13 @@ def read_logfile_lines(log_file):
 
                     value = get_value_from_logfile(logfileLines, column.pattern)
 
+            if column.title.lower() == 'score':
+                if value is None:
+                    value = str(score)
+                else:
+                    # If a score column exists in the xml, we use that column's value for the score, instead of
+                    # the computed score.
+                    score = float(value)
```
dbeyer: Nit-picking: I would actually prefer the branches the other way around:
`if value is not None:` and then
as comment "If a score value does not exist in the XML, then we use the computed score."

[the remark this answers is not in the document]
dbeyer: > Why is score a float here?

Scores are floats, not integers. Mind the normalization and coverage scores.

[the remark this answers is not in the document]
dbeyer: > What happens if there is a score column, but it contains text? Won't the current code crash table-generator?

Add exception handling to the float conversion?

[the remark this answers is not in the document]
lembergerth: I missed the exception handling, but also chose "float" because I had the notion in mind that scores must be numbers.
But if a user decides to store some other values in a column called score, should this be forbidden?

When a user actually uses numbers, this decision won't make any difference: Even if we choose the string-Type here, if the score is a number, it should be correctly recognized by table-generator during output formatting.

PhilippWendler:
This has conflicts now because the `score` value can now be `None`, please adjust accordingly.

benchexec/tablegenerator/__init__.py
lembergerth: I prefer the branching that way because "value is None" is a simpler expression than "value is not None".
But I adjusted the comment as you suggested, since it better describes the real logic behind it.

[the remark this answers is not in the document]
lembergerth: I've added exception handling and a small integration test.

lembergerth:
> This has conflicts now because the score value can now be None, please adjust accordingly.

I've adjusted the code - would you mind having another look?

PhilippWendler:
Another question: with this PR, the value of `RunResult.score` in table-generator is changed depending on whether a score column is present in the table-definition. This seems like a confusing side effect: whether a column is shown or not should not influence anything else, right?

So a better solution might be to always fill the score value from the result XML if it is present there.

However, I am currently confused why the internal `RunResult.score` value is changed at all? I think that table-generator won't even show this value anywhere, because scores are shown (in the `status` column) only for SV-COMP properties, right? So could we remove the change of this value completely and just change the value of the score column in this PR?

Or do you want to override the score value that table-generator adds to the `status` column of results with SV-COMP properties?

dbeyer:
I would like to do the 'usual' calculations with the scores (normalization of categories, etc.),
besides showing the scores and their statistics in the tables such as this one:
https://test-comp.sosy-lab.org/2019/results/results-verified/coveritest.2019-01-12_1839.results.test-comp19_prop-coverage-branches.ReachSafety-BitVectors.xml.bz2.merged.xml.bz2.table.html

PhilippWendler:
> I would like to do the 'usual' calculations with the scores (normalization of categories, etc.),
> besides showing the scores and their statistics in the tables such as this one:
> [test-comp.sosy-lab.org/2019/results/results-verified/coveritest.2019-01-12_1839.results.test-comp19_prop-coverage-branches.ReachSafety-BitVectors.xml.bz2.merged.xml.bz2.table.html](https://test-comp.sosy-lab.org/2019/results/results-verified/coveritest.2019-01-12_1839.results.test-comp19_prop-coverage-branches.ReachSafety-BitVectors.xml.bz2.merged.xml.bz2.table.html)

Yes, this should all work by treating the `score` column just like any numeric column, no need to set `RunResult.score`.

dbeyer:
In the end, I need the following code to work:
```
        run_set_result = tablegenerator.RunSetResult.create_from_xml(results_file, tablegenerator.parse_results_file(results_file))
        for run_result in run_set_result.results:
            score += run_result.score
```

PhilippWendler:
Ok, this would mean that `RunResult.score` would actually need to be changed, but I am not sure that we should base internal implementation decisions of BenchExec on other scripts (this is not an officially stable API). The script could just collect the score value from the column in `run_result.columns` instead, anyway.

dbeyer:
See this comment: https://github.com/sosy-lab/benchexec/pull/406#discussion_r247938372

benchexec/tablegenerator/__init__.py
```diff
@@ -697,6 +695,11 @@ def read_logfile_lines(log_file):
 
                     value = get_value_from_logfile(logfileLines, column.pattern)
 
+            if column.title.lower() == 'score':
+                if value is None:
+                    # If no score column exists in the xml, take the internally computed score,
+                    # if available
+                    value = str(score) if score is not None else None
```
PhilippWendler: Can be simplified to:
```suggestion
            if column.title.lower() == 'score' and value is None and score is not None:
                # If no score column exists in the xml, take the internally computed score, if available
                value = str(score)
```

Please also check the failing test.

lembergerth: That's way nicer, thanks for the hint!

lembergerth:
I've adjusted the failing test - it was due to the fact that the score-value is not entered into the status-column, anymore.

benchexec/tablegenerator/__init__.py
PhilippWendler: The `if` in the assignment can be removed.

PhilippWendler: Please fix the indentation, too.

lembergerth: D'oh, sorry.

PR #414, opened 2019-01-20
Simplify method program_files for tbf and tbf_testsuite_validator

lembergerth:
Do not run the template's `program_files` method first,
as this produces an invalid, and thus unnecessary, file list.

Create relative paths instead of absolute paths.
This is required by the interface and was ignored before.

Also use the util-methods of benchexec to support potential wildcards in future changes.

### DIFF
--- a/benchexec/tools/tbf.py
+++ b/benchexec/tools/tbf.py
@@ -34,13 +34,10 @@ class Tool(benchexec.tools.template.BaseTool):
     REQUIRED_PATHS = ["tbf", "lib"]
 
     def program_files(self, executable):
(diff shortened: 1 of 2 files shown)


PhilippWendler:
Since 6b15783b19489bc4515d043b7d9ea74818d1dc16, there is a new utility method. You should be able to just call

        return self._program_files_from_executable(
            executable, self.REQUIRED_PATHS, parent_dir=True)

Could you please check and adjust the code if possible?

lembergerth:
Thanks, works like a charm!

PR #415, opened 2019-01-20
Let test-comp versions of tbf use given property files

lembergerth:
This is currently the case for all versions tagged with 'testcomp',
so we use this to differentiate whether parameter '--spec' is viable.


PhilippWendler:
I can merge this, but is the proposed version check a good idea? In the title it says "newer versions", but I assume that not all newer versions will have `testcomp` in the version string?

lembergerth:
I only want the test-comp versions to support the property for now, and I'm not sure yet whether I will merge this into the main-version of tbf.

The title was quickly written down - I'm only working on test-comp versions, at the moment :wink:

I'll have to adjust this once I decide that these property files make sense for tbf outside of the context of test-comp.

Issue #402, opened 2019-01-09
clickable filenames if yaml-file is shown as overlay in html-table


kfriedberger:
For the previous sv-benchmarks with the plain C-source-files the link to the sourcefile was showing an overlay window containing the plain source of the analyzed C-source-file. With the new yaml-scheme only the yaml is shown, which is nice, but not very helpful.

**Wanted feature**:
Parse the loaded yaml-file with some Javascript, replace filenames with links, and open those links in the overlay window, such that a user can show the C-source-file in the overlay.
The same could be done for property-files.
It would be sufficient to only replace those parts of the shown yaml, where filenames are expected.

After the above got implemented, having a "back to yaml"-button would be a nice bonus feature.

kfriedberger:
I myself and several students want to look at the table and have a possibility to directly access the sourcecode. This is not possible since a several months now. I would either like to have #419 merged or at least see that someone is really working on this. 

Philipp only said that #419 is not good enough, but currently I have not yet seen any case where it would not work. Could Philipp provide a better solution or at least give a hint on his idea of a better solution. What are the required features for this (beyond the given description above)?

PhilippWendler:
> I myself and several students want to look at the table and have a possibility to directly access the sourcecode. This is not possible since a several months now. I would either like to have #419 merged or at least see that someone is really working on this.

Currently all people who are active on BenchExec unfortunately have more important tasks, and thus I cannot make any ETA promises.

> Philipp only said that #419 is not good enough, but currently I have not yet seen any case where it would not work.

This is wrong, it was already in my very first reponse: https://github.com/sosy-lab/benchexec/pull/419#issuecomment-458616730

> Could Philipp provide a better solution or at least give a hint on his idea of a better solution. What are the required features for this (beyond the given description above)?

A better solution is suggested already in your very own issue description (https://github.com/sosy-lab/benchexec/issues/402#issue-397575406), which I agreed on (https://github.com/sosy-lab/benchexec/issues/402#issuecomment-452975276): We parse the file using a YAML parser (for which certainly Javascript libraries exist), extract the relevant info, and show it as links.

To make it easier than replacing only the links inside the shown YAML text (which I guess could be tricky), we could also show the task definition file in the overlay: first with all information extracted and printed out in a consistent and structured way, and below that the raw YAML file. The first part would have the links. This would have the additional benefit that users could for example expect that input files are always shown first and properties second (in the raw YAML file, the order could be different).

[...]

PR #419, opened 2019-01-29
Use regex to insert urls into yaml-files in overlay-window.

kfriedberger:
see #402.

### DIFF
--- a/benchexec/tablegenerator/template.html
+++ b/benchexec/tablegenerator/template.html
@@ -494,6 +494,10 @@ function loadContentWrapper(event) {
   return false;
 }
 
+function dirname(path) {
+    return path.match(/.*\//);
+}
+
 function loadContent(url) {
   var contentPane = $('<pre id="content"></pre>').appendTo("#contentPane")
   function writeToContentPane(text) {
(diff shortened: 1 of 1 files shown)


PhilippWendler:
> Under the assumption that yaml-files are nice,

Where does this assumption come from? Users can write all kinds of YAML files and use them with BenchExec.

A regexp-based solution does not even work with "nice" YAML files if they specify several input files, and such task definitions are already commonly in use (e.g., in SV-COMP).

kfriedberger:
In my opinion, the regex-based solution works for 95% of all existing tables.
If the regex does not match, we will not add any url, which is exactly the situation right now. Thus the solution is a safe intermediate base until someone adds a yaml-parser-based solution.

Most users work with yaml files consisting of single C-files and single properties (all from sv-benchmark repository). Just take a look at tables from CPAchecker's buildbot (except Java) and all categories of SV-COMP (except Java) and maybe also TestComp.

I am not finishing #419, but just starting (with one step/a few lines of code that handle most of the existing applications). What needs to be done for this PR to get merged soon?

PhilippWendler:
Adding a feature that silently does not work in some circumstances, or works only partially, is highly confusing to users and a bad user experience. We do not want to disappoint BenchExec's users in such ways.

Such a hack would only be acceptable if a full solution would be impossible or too difficult to implement, but this is not the case for #402.

PhilippWendler:
This conflicts with the ongoing development in branch `react-table`, which will be merged soon, thus it does not make sense to merge this PR.

PR #420, opened 2019-01-31
Parse property files explicitly to select the type of search

MartinNowack:


PhilippWendler:
I can merge this, but don't you think that this feature would be better suited in your tool itself?

Of course we suggest that all benchmarkers use BenchExec, but if there are other people who want to benchmark your tool but cannot use BenchExec for some reason, they wouldn't be able to run your tool directly on the tasks from the SV-Benchmarks repo.

MartinNowack:
@PhilippWendler Ok, I updated the wrapper. Now there is just a simple forwarder.

MartinNowack:
@PhilippWendler Thanks a lot!

PR #428, opened 2019-05-13
Add tool-info module for Frama-C

lembergerth:
I've just written this small tool-info module for using Frama-C with benchexec and thought it might be interesting to other users.

At the moment, the module does not try to interpret the output of frama-c because of the many different possibilities of outputs that can be produced by different plugins.
The module also doesn't use a given propertyfile, as Frama-C can't interpret that (or at least I didn't have an idea of how to make it).

Frama-C allows to specify batches of commands that are executed iteratively.
These batches are separated through a group of `-then*` parameters, and each batch re-uses the results of the previous batch.
Because of this, the original input file is usually given to the first batch, and this tool-info module does this, as well.


PhilippWendler:
You could rename it to `frama-c.py`. It is not a valid Python module name, but due to the way how BenchExec imports the module, it will work.

Most of the existing tool-info modules use `-`, just `two_ls.py` does not.

lembergerth:
Alright, that looks better in benchmark definitions - I renamed it.

Issue #122, opened 2016-01-21
Assignment of non-sibling processors to a single task


martinjonas:
Hi,

when running a tool which uses multiple cpu cores, BenchExec is assigning sibling cores to a single run even though non-sibling cores are available. Is this an intended behavior?

In particular, I have a benchmark with `<benchmark cpuCores="4" threads="1" ...>` and the output of BenchExec is

```
2016-01-21 17:07:28,707 - DEBUG - List of available CPU cores is [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23].
2016-01-21 17:07:28,709 - DEBUG - Physical packages of cores are defaultdict(<class 'list'>, {0: [0, 1, 2, 3, 4, 5, 12, 13, 14, 15, 16, 17], 1: [6, 7, 8, 9, 10, 11, 18, 19, 20, 21, 22, 23]}).
2016-01-21 17:07:28,710 - DEBUG - Siblings of cores are {0: [0, 12], 1: [1, 13], 2: [2, 14], 3: [3, 15], 4: [4, 16], 5: [5, 17], 6: [6, 18], 7: [7, 19], 8: [8, 20], 9: [9, 21], 10: [10, 22], 11: [11, 23], 12: [0, 12], 13: [1, 13], 14: [2, 14], 15: [3, 15], 16: [4, 16], 17: [5, 17], 18: [6, 18], 19: [7, 19], 20: [8, 20], 21: [9, 21], 22: [10, 22], 23: [11, 23]}.
2016-01-21 17:07:28,710 - DEBUG - Going to assign at most 1 runs per package, each one using 4 cores and blocking 4 cores on 1 packages.
2016-01-21 17:07:28,711 - DEBUG - Final core assignment: [[0, 1, 12, 13]].
```

However, cores 0 and 12 are sibling virtual cores mapped to the same physical core (and so are 1 and 13). Shouldn't BenchExec choose a core assignment [[0,1,2,3]], since cores 2 and 3 are available and not siblings of any of used cores?

If this is the case, I think the problem is caused by an assignment on the line 181 of `resources.py`, where _all_ sibling cores are assigned to be used. More reasonable behavior from my point of view would be to assign only the particular core and block all of its siblings, if there is enough of physical cores available. For example, the following change fixes the problem in my case:

``` python
if core not in cores_with_siblings:
    cores.append(core)
[...]

PR #429, opened 2019-05-20
no-hyperthreading feature added

alohamora:
This pull request implements the feature requested in #122. 
The pull request implements no-hyperthreading feature while doing parallel benchmarking to ensure deterministic results.
I have added a command-line argument `--no-hyperthreading` and the feature can be used just by using the following command:
`benchexec "file_name.xml" --no-hyperthreading`

### DIFF
--- a/benchexec/resources.py
+++ b/benchexec/resources.py
@@ -40,7 +40,7 @@ __all__ = [
            'get_cpu_package_for_core',
            ]
 
-def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups, use_hyperthreading, coreSet=None):
+def get_cpu_cores_per_run(coreLimit, num_of_threads, use_hyperthreading, my_cgroups, coreSet=None):
     """
     Calculate an assignment of the available CPU cores to a number
     of parallel benchmark executions such that each run gets its own cores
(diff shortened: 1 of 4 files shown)


benchexec/resources.py
```diff
@@ -40,7 +40,7 @@
            'get_cpu_package_for_core',
            ]
 
-def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups, coreSet=None):
+def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups, nohyperthreading, coreSet=None):
```
PhilippWendler: Usually it is better to have variables named in a "positive" way, e.g., `hyperthreading` or `use_hyperthreading` instead of `nohyperthreading` (with negated meaning, of course). This makes it easier to understand the code.

The command-line parameter is not affected by this, because here the name depends on what the default is.

benchexec/resources.py
```diff
@@ -99,9 +99,83 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, my_cgroups, coreSet=None):
         logging.debug("Siblings of cores are %s.", siblings_of_core)
     except ValueError as e:
         sys.exit("Could not read CPU information from kernel: {0}".format(e))
-
+    if nohyperthreading:
+        return _get_cpu_cores_per_run_no_hyperthreading(coreLimit, num_of_threads, allCpus, cores_of_package, siblings_of_core)
     return _get_cpu_cores_per_run0(coreLimit, num_of_threads, allCpus, cores_of_package, siblings_of_core)
 
+def _get_cpu_cores_per_run_no_hyperthreading(coreLimit, num_of_threads, allCpus, cores_of_package, siblings_of_core):
```
PhilippWendler: This method duplicates a lot of work and code of the existing method. This can become problematic, for example if we find a bug in one of the methods, we need to remember to check if the bug also exists in the other method.

Is there a way to reuse more code? For example, one could add a parameter to the existing method, although we need to make sure to not overcomplicate the method. Or can we refactor out parts of the method and reuse them?

Another idea would be to just remove the set of all sibling cores from `allCpus` and call the original method, this should also work, right?

benchexec/resources.py
alohamora: Changed variable name to `use_hyperthreading`

benchexec/resources.py
alohamora: @PhilippWendler I have refactored the code as per your suggestion and did the preprocessing to remove all sibling cores from `allCpus` and `cores_of_package` and `siblings_of_core`.
After the preprocessing the core allocation algorithm becomes identical for both `hyperthreading` and `no-hyperthreading` mode so the function does not get complicated.
I did the preprocessing the `_get_cpu_cores_per_run0` function only as we can test this function separately without reading from the machine architecture.

alohamora:
> Could you also please add unit tests? We have existing tests in `test_core_assignment.py`, where several classes exist that emulate various machines, and tests attempt to assign possible and impossible numbers of runs.

I will add the tests asap.

benchexec/resources.py
```diff
@@ -118,6 +118,24 @@ def _get_cpu_cores_per_run0(coreLimit, num_of_threads, allCpus, cores_of_package
     if coreLimit * num_of_threads > len(allCpus):
         sys.exit("Cannot run {0} benchmarks in parallel with {1} CPU cores each, only {2} CPU cores available. Please reduce the number of threads to {3}.".format(num_of_threads, coreLimit, len(allCpus), len(allCpus) // coreLimit))
 
+    if not use_hyperthreading:
+        logging.debug("Running in no-hyperthreading mode")
```
PhilippWendler: I would suggest to move the logging to the end of the if statement, and then let it print something like `Running in no-hyperthreading mode, avoiding the use of CPU cores x, y, z.`

That might be useful for users.

alohamora: Refined the debug statement and moved it at the end of the if block

benchexec/test_core_assignment.py
```diff
@@ -37,13 +38,13 @@ def setUpClass(cls):
         cls.longMessage = True
         logging.disable(logging.CRITICAL)
 
-    def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
-        result = _get_cpu_cores_per_run0(coreLimit, num_of_threads, *self.machine())
+    def assertValid(self, coreLimit, num_of_threads, with_hyperthreading, expectedResult=None):
```
PhilippWendler: Would it make sense to use `self.use_ht` instead of adding the parameter `with_hyperthreading`? That would mean we would not need to add a parameter to all existing tests.

Or is there some difference between the attribute and the parameter, e.g., are they set differently at some point?

alohamora: Yes using `with_hyperthreading` with `self.use_ht` is  redundant actually. I shall fix them asap.

Issue #51, opened 2015-10-23
Improve core assignment with multiple NUMA regions per socket


PhilippWendler:
If a machine has multiple NUMA regions per socket, BenchExec should try to distribute the runs across all NUMA regions of each socket. Currently it first uses the cores that belong to one NUMA region, then those of the second region etc.

PhilippWendler:
This is also a problem if several runs should be executed with few cores but much memory, such that for example the cores for all runs are taken from the first NUMA region, but not enough memory is available in that region.

alohamora:
To handle multiple nodes per socket we can alter the `cores_per_package` to `cores_per_region`  in `resources.py`.
https://github.com/sosy-lab/benchexec/blob/0775a3223d47971b7bcecbdd60edc0af6be7e4b4/benchexec/resources.py#L88
Therefore instead of finding physical packages of core, we find memory regions for each core
```
memory_regions = []
for core in allCpus:
    coreDir = '/sys/devices/system/cpu/cpu{0}/'.format(core)
    memory_regions.append(_get_memory_banks_listed_in_dir(coreDir))
``` 
And correspondingly we build a `cores_per_region` mapping instead of `cores_per_package`
https://github.com/sosy-lab/benchexec/blob/0775a3223d47971b7bcecbdd60edc0af6be7e4b4/benchexec/resources.py#L90-L91

```
for core, region in zip(allCpus, memory_regions):
    cores_of_region[region].append(core)
```
Now using `cores_per_region` shall not affect the distribution for systems with single node per socket(or package) and for systems with multiple NUMA regions the algorithm shall split the socket into different groups each with their own different NUMA region.
This shall improve core assignment as runs shall be split accross memory regions instead of packages, thus ensuring that all the memory regions are used equally.

PhilippWendler:
Sounds good. Please make sure that the code would still work for systems that have multiple packages per NUMA node (we do not know such systems, but they could exist). In this case, actual packages should be used, of course (we will always use whatever is more fine granular: NUMA nodes or packages).

PR #431, opened 2019-06-07
Improve core assignment with multiple NUMA nodes per socket

alohamora:
This is in reference to #51 
The core assignment method has been updated in the pull request.
Now, benchexec first creates a mapping from memory regions to core and physical packages to core.
Then the more fine grained mapping among the two is chosen for better core assignment.

### DIFF
--- a/benchexec/resources.py
+++ b/benchexec/resources.py
@@ -84,7 +84,7 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, use_hyperthreading, my_cgro
 
         logging.debug("List of available CPU cores is %s.", allCpus)
 
-        # read mapping of core to memory memory_region
+        # read mapping of core to memory region
         memory_regions = []
         for core in allCpus:
             coreDir = '/sys/devices/system/cpu/cpu{0}/'.format(core)
@@ -94,6 +94,16 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, use_hyperthreading, my_cgro
             cores_of_memory_region[memory_region].append(core)
         logging.debug("Memory regions of cores are %s.", cores_of_memory_region)
 
+        # read mapping of core to CPU ("physical package")
+        physical_packages = [get_cpu_package_for_core(core) for core in allCpus]
+        cores_of_package = collections.defaultdict(list)
+        for core, package in zip(allCpus, physical_packages):
+            cores_of_package[package].append(core)
+        logging.debug("Physical packages of cores are %s.", cores_of_package)
+
+        # select the more fine grained division among memory regions and physical package
+        cores_of_unit = (cores_of_memory_region if (len(cores_of_memory_region) >= len(cores_of_package)) else cores_of_package)
+
         # read hyper-threading information (sibling cores sharing the same physical core)
         siblings_of_core = {}
         for core in allCpus:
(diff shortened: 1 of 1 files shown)


alohamora:
Sure I will add tests by tonight.

alohamora:
@PhilippWendler , I have made changes in the function `get_cpu_cores_per_run` function only. So I am not able to decide how to add tests for the same as the function deals with reading system information only. Moreover are also no existing tests for the function.

benchexec/resources.py
```diff
@@ -157,64 +171,64 @@ def _get_cpu_cores_per_run0(coreLimit, num_of_threads, use_hyperthreading, allCp
...
+        sys.exit("Cannot split runs over multiple CPUs and at the same time assign multiple runs to the same CPU. Please reduce the number of threads to {0}.".format(unit_count // units_per_run))
 
-    runs_per_package = int(math.ceil(num_of_threads / package_count))
-    assert packages_per_run == 1 or runs_per_package == 1
-    if packages_per_run == 1 and runs_per_package * coreLimit > package_size:
-        sys.exit("Cannot run {} benchmarks with {} cores on {} CPUs with {} cores, because runs would need to be split across multiple CPUs. Please reduce the number of threads.".format(num_of_threads, coreLimit, package_count, package_size))
+    runs_per_unit = int(math.ceil(num_of_threads / unit_count))
+    assert units_per_run == 1 or runs_per_unit == 1
+    if units_per_run == 1 and runs_per_unit * coreLimit > unit_size:
+        sys.exit("Cannot run {} benchmarks with {} cores on {} CPUs with {} cores, because runs would need to be split across multiple CPUs. Please reduce the number of threads.".format(num_of_threads, coreLimit, unit_count, unit_size))
```
PhilippWendler: For each occurrence of `CPUs` in this message, please write `CPUs/memory regions` instead. Otherwise it will be confusing for users.

benchexec/resources.py
```diff
@@ -84,13 +84,26 @@ def get_cpu_cores_per_run(coreLimit, num_of_threads, use_hyperthreading, my_cgro
...
+
         # read mapping of core to CPU ("physical package")
         physical_packages = [get_cpu_package_for_core(core) for core in allCpus]
         cores_of_package = collections.defaultdict(list)
         for core, package in zip(allCpus, physical_packages):
             cores_of_package[package].append(core)
         logging.debug("Physical packages of cores are %s.", cores_of_package)
 
+        # select the more fine grained division among memory regions and physical package
+        cores_of_unit = (cores_of_memory_region if (len(cores_of_memory_region) >= len(cores_of_package)) else cores_of_package)
```
PhilippWendler: Could you please add a debug log statement here, if `cores_of_memory_region` is used?

I guess rewriting into an `if ...: ... else:` construct is then more clear.

benchexec/resources.py
alohamora: I have replaced the same, and also replaced the term `unit` to `CPU/memory regions` in other debug statements for better understanding.

alohamora:
@PhilippWendler 
First of  all sorry for the late reply.

> I am wondering why it uses only the first memory region for each core here?

This is because if a core is associated with multiple memory nodes and all the memory regions are considered, the `cores_of_memory_region` shall contain the same core in multiple memory regions and thus the core allocation algorithm shall fail.

> Furthermore, couldn't the code skip assigning `memory_regions` completely and immediately build `cores_of_memory_region`?

Yes, if no memory_regions is found for a core, we shall skip building `cores_of_memory_regions` and thus by the logic in `get_cpu_cores_per_run`
https://github.com/sosy-lab/benchexec/blob/ce141dea0c37d30565f3247c46da26b0a0a94b83/benchexec/resources.py#L112

It shall select the `cores_of_package` as the basis for core allocation.

I am currently working on this and shall drop a pull request asap.

PR #432, opened 2019-06-12
CLI for pqos added

alohamora:
An inital version for CLI for pqos library is added.
Currently the CLI allows the user to check support of OS and kernel interface in their interface

### DIFF
--- a/pqos/bin/pqos_wrapper
+++ b/pqos/bin/pqos_wrapper
@@ -1,0 +1,9 @@
+#!/usr/bin/env python3
+import os
+import sys
+sys.dont_write_bytecode = True # prevent creation of .pyc files
+sys.path.insert(0, '..')
+
+import pqos_wrapper.main as wrapper
+
+sys.exit(wrapper.PqosWrapper.main())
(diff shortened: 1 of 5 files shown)


alohamora:
> I wonder if it makes sense to have this wrapper in the BenchExec repository, or if we should create a separate project (like we did for cpu-energy-meter)?

Yes, I also think that using a different repository would be different as it is a completely different CLI.
Shall I create a new repository by the name `pqos-wrapper`? 
Or can you suggest a better name for the same.   
> 
> Could you please add a license to the source files? Is Apache 2.0 ok for you?

Yes, I shall add the Apache 2.0 license to the source files

PhilippWendler:
> > I wonder if it makes sense to have this wrapper in the BenchExec repository, or if we should create a separate project (like we did for cpu-energy-meter)?
> 
> Yes, I also think that using a different repository would be different as it is a completely different CLI.
> Shall I create a new repository by the name `pqos-wrapper`?
> Or can you suggest a better name for the same.

The name is fine. I suggest I create the repo such that it is under the `sosy-lab` org. But if you agree, I would create it on GitLab, because I find GitLab more powerful and with much better CI.

alohamora:
Sure, we can use Gitlab for the project.

PhilippWendler:
Moved to https://gitlab.com/sosy-lab/software/pqos-wrapper

Issue #227, opened 2017-02-19
Use Intel Cache Allocation Technology


PhilippWendler:
Some Intel Xeons support the [Cache Allocation Technology (CAT)](https://software.intel.com/en-us/articles/introduction-to-cache-allocation-technology), which allows controlling which CPU cores may use which parts of the (L3) cache. It should be possible to use this for avoiding the influence of cache performance for parallel benchmarks on the same CPU.

We should investigate whether this is indeed helpful and whether support can be added to BenchExec. There is a [tool from Intel](https://github.com/01org/intel-cmt-cat) ([documentation](https://github.com/01org/intel-cmt-cat/wiki/FAQ)) that we can use for investigation.

Support for this was also [added to Linux 4.10](https://kernelnewbies.org/Linux_4.10#head-ee17602bfd51f652da963422e401f36b5e2aa7b3) ([documentation](https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt), [article](https://lwn.net/Articles/694800/), [slides](https://events.linuxfoundation.org/sites/events/files/slides/cat8.pdf)). If we add support for CAT to BenchExec, we should probably use this kernel feature instead of directly accessing the CPU via its MSRs.

PhilippWendler:
More information and references for CAT are in [this blog post](https://danluu.com/intel-cat/).

PhilippWendler:
Since Linux 4.12, this feature also supports allocating memory bandwidth, which we should also use.

Pac23:
@PhilippWendler interested in this for gsoc 2019 the prerequisite states it says "maybe C" would be required,i am super comfortable with python(one of the requirements),worked with C but not with hardware level C.So basically my questions is what is the level of C that would be involved.

PhilippWendler:
Probably not too difficult. The goal is actually to have all in Python, and only if necessary use C. Soo far we managed to never need C, but I cannot guarantee it for this project idea.

Pac23:
> Probably not too difficult. The goal is actually to have all in Python, and only if necessary use C. Soo far we managed to never need C, but I cannot guarantee it for this project idea.

So after some digging it seems lots of C would be required,also doing this in Python in my opinion would not be the most efficient or the fastest way to do it.

miscellaneously:
Hi @PhilippWendler ,
[...]

PR #434, opened 2019-06-28
Cache allocation

alohamora:
Added pqos module to interact with pqos_wrapper cli for l3 cache allocation and isolation of cache between threads.
#227

### DIFF
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -114,6 +114,8 @@ def execute_benchmark(benchmark, output_handler):
     coreAssignment = None  # cores per run
     memoryAssignment = None  # memory banks per run
     cpu_packages = None
+    pqos = Pqos() # The pqos class instance for cache allocation
+
     if CORELIMIT in benchmark.rlimits:
         if not my_cgroups.require_subsystem(cgroups.CPUSET):
             sys.exit(
@@ -126,7 +128,6 @@ def execute_benchmark(benchmark, output_handler):
             my_cgroups,
             benchmark.config.coreset,
         )
-        pqos = Pqos()
         if pqos.cli_exists:
             pqos.allocate_l3ca(coreAssignment)
         memoryAssignment = get_memory_banks_per_run(coreAssignment, my_cgroups)
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -114,7 +114,7 @@ def execute_benchmark(benchmark, output_handler):
     coreAssignment = None  # cores per run
     memoryAssignment = None  # memory banks per run
     cpu_packages = None
-    pqos = Pqos() # The pqos class instance for cache allocation
+    pqos = Pqos()  # The pqos class instance for cache allocation
 
     if CORELIMIT in benchmark.rlimits:
         if not my_cgroups.require_subsystem(cgroups.CPUSET):
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -128,8 +128,7 @@ def execute_benchmark(benchmark, output_handler):
             my_cgroups,
             benchmark.config.coreset,
         )
-        if pqos.cli_exists:
-            pqos.allocate_l3ca(coreAssignment)
+        pqos.allocate_l3ca(coreAssignment)
         memoryAssignment = get_memory_banks_per_run(coreAssignment, my_cgroups)
         cpu_packages = set(
             get_cpu_package_for_core(core)
@@ -265,8 +264,7 @@ def execute_benchmark(benchmark, output_handler):
             "System has swapped during benchmarking. "
             "Benchmark results are unreliable!"
         )
-    if pqos.reset_required and pqos.cli_exists:
-        pqos.reset_resources()
+    pqos.reset_resources()
     output_handler.output_after_benchmark(STOPPED_BY_INTERRUPT)
 
     return 0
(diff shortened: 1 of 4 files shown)


benchexec/localexecution.py
```diff
@@ -260,7 +265,8 @@ def execute_benchmark(benchmark, output_handler):
             "System has swapped during benchmarking. "
             "Benchmark results are unreliable!"
         )
-
+    if pqos.reset_required and pqos.cli_exists:
```
PhilippWendler: I would move the check for `reset_required` into the `Pqos` class (better encapsulation).

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        success = True
+        args_list = list(args)
+        args_list.insert(0, self.CMD)
+        try:
+            ret = json.loads(check_output(args_list, stderr=STDOUT))
+            logging.debug(ret[function]["message"])
+        except CalledProcessError as e:
+            ret = json.loads(e.output)
+            logging.warning(ret["message"])
```
PhilippWendler: Are these log messages understandable on their own for the user, or should we prefix something like "Could not set cache allocation: ..."?

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class Pqos(object):
+    """
+        The Pqos class defines methods to interact with pqos_wrapper cli.
+    """
+
+    CMD = "pqos_wrapper"
+    DEBUG = 0
+    WARNING = 1
```
PhilippWendler: What is this?

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+            ret.append("[" + ",".join(str(core) for core in benchmark) + "]")
+        return "[" + ",".join(ret) + "]"
+
+    def allocate_l3ca(self, core_assignment):
+        """
+            This method checks if L3CAT is available and calls pqos_wrapper to
+            allocate equal cache to each thread.
+        """
+        self.check_capacity("l3ca")
+        if self.cap == True:
```
PhilippWendler:
```suggestion
        if self.cap:
```

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+        ):
+            self.cli_exists = True
+
+    def execute_command(self, function, *args):
+        """
+            Execute a given pqos_wrapper command and log the output
+        """
+        success = True
+        args_list = list(args)
+        args_list.insert(0, self.CMD)
```
PhilippWendler:
```suggestion
        args_list = [self.CMD] + list(args)
```
This is easier to understand.

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+            find_executable("pqos_wrapper", exitOnError=False, use_current_dir=False)
+            is not None
+        ):
+            self.cli_exists = True
+
+    def execute_command(self, function, *args):
+        """
+            Execute a given pqos_wrapper command and log the output
+        """
+        success = True
```
PhilippWendler: I would suggest to use direct `return` statements instead of a variable for the result value in trivial cases like this.

benchexec/pqos.py
```diff
@@ -0,0 +1,100 @@
...
+        args_list = list(args)
+        args_list.insert(0, self.CMD)
+        try:
+            ret = json.loads(check_output(args_list, stderr=STDOUT))
+            logging.debug(ret[function]["message"])
+        except CalledProcessError as e:
+            ret = json.loads(e.output)
```
PhilippWendler: I guess this could fail, right? For example if some unexpected error occurs and the tool was not started, `output` might be empty or no valid Json. Better have some fallback for this?

benchexec/pqos.py
alohamora: Sorry, this was used by me for some testing...i forgot to remove it in the pull request

benchexec/pqos.py
alohamora: The error messages describe the function error e.g "Unable to set COS 0 for core 0".
So I think we could prefix "Could not set cache allocation.." for each error message for better understanding

benchexec/localexecution.py
alohamora: I have set the field `reset_required` in the Pqos class itself. Are you saying to move the above two functions in `reset_resources` function itself?

alohamora:
I have added a `suppress_warning` option in `execute_command` to prevent multiple warnings when a command fails and then resources are reset.

benchexec/pqos.py
alohamora: Yes, this could fail if incorrect cli format is passed, I have added a try and except block for the same and logged an `unable to execute command` warning.

benchexec/localexecution.py
PhilippWendler: At last the check for `reset_required`, yes, because this seems like internal-only knowledge of that class. But `cli_exists` could be moved as well.

alohamora:
@PhilippWendler  where should I add the documentation for the above implemented feature?

PhilippWendler:
>  where should I add the documentation for the above implemented feature?

I guess `doc/resources.md` would be the best place.

alohamora:
Added a `check_for_errors` function for detailed error analysis about the failed execution of a `pqos_wrapper` command

PR #456, opened 2019-10-18
Add Tool Info for JDart

mmuesly:
We want to integrate JDart into the BenchExec framework.
Currently, JDart doesn't support a fixed Version system that is extractable from the binary.  Therefore, I left the Version String empty.

### DIFF
--- a/benchexec/tools/jdart.py
+++ b/benchexec/tools/jdart.py
@@ -1,0 +1,71 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+Copyright (C) 2007-2018  Dirk Beyer
+All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import os
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for JDart modified by TU Dortmund
+    (https://github.com/tudo-aqua/jdart).
+    """
+
+    REQUIRED_PATHS = ["jconstraints-extensions",
+                      "lib",
+                      "z3",
+                      "basic.jpf",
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Do you want to participate in Test-Comp or SV-COMP? Note that for these competitions it is required to have a version in the tool that is extracted by this tool-info module.

mmuesly:
Yes, we are planing to participate in SV-COMP with this tools.

It would be nice, if this requirement might be documented in the [rule section](https://sv-comp.sosy-lab.org/2020/rules.php) or somewhere else in the future. At the moment, it is written in various places that returning a version is desired, but optional.

I will rework the PR once the tool is adapted.

mmuesly:
I updated the PR and made the artifact able to provide a version on request.

Earlier in the day, I had a discussion with Dirk Beyer on JVM provisioning, in case I want to use a non-default JVM.
At the end, I decided to include the JVM version that is suitable in the artifact. After reading up on the discussion in #453, my understanding is that the REQUIRED_PATHS variable needs to list all files on the top-level explicitly and every subfolder that should be included completely in the  execution environment. I therefore adapted the REQUIRED_PATHS to match our artifact with this interpretation of the code.
Could you shortly confirm this assumption? Is there an easy way to test that all files end up at the right place in this distribution step?

Thanks!

PhilippWendler:
Your assumption is correct. You can test it using [these instructions](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#testing-the-tool-integration), it will output the paths that it will use.

In the distribution step, everything listed as part of these paths is copied to a different location, so make sure that you do not use any absolute paths anywhere, but the relative location of the paths will stay the same.

PhilippWendler:
> It would be nice, if this requirement might be documented in the [rule section](https://sv-comp.sosy-lab.org/2020/rules.php) or somewhere else in the future. At the moment, it is written in various places that returning a version is desired, but optional.

@dbeyer

dbeyer:
Thanks for the hint about the version reporting.  I thought every tool has this anyway. Now added to the rules page.

dbeyer:
@PhilippWendler Is this ready to be merged?

PhilippWendler:
Yes, though all paths below `jdk8u222-b10/...` in `REQUIRED_PATHS` are redundant, and I think @mmuesly wanted to test it first.

mmuesly:
I tested the tool integration as described. In the program files output section, the expanded path for 'jdk8u222-b10' is listed but not all subfiles.

I trust @PhilippWendler that this is sufficient as he knows more about the tool than I do. Hence, I simplified the ´REQUIRED:PATHS´and squashed the commits. From my point of view, this should be ready to merge.

Thank you for the comments and support!

PhilippWendler:
If a directory is listed, everything in it is copied, yes.

PR #457, opened 2019-10-19
Add tool-info module for Tracer-X

rasoolmaghareh:
…Tracer-X is built on top of KLEE. The configuration/setting and running of the programs are similar to KLEE. The difference is that our tool Tracer-X  extends KLEE with interpolation for pruning of the Symbolic Execution Tree.

### DIFF
--- a/benchexec/tracerx.py
+++ b/benchexec/tracerx.py
@@ -89,11 +89,11 @@ class Tool(benchexec.tools.template.BaseTool):
         """
         for line in output:
             if line.startswith("KLEE: ERROR: "):
-                if line.find("ASSERTION FAIL:") != -1:
+                if "ASSERTION FAIL:" in line:
                     return result.RESULT_FALSE_REACH
-                elif line.find("memory error: out of bound pointer") != -1:
+                elif "memory error: out of bound pointer" in line:
                     return result.RESULT_FALSE_DEREF
-                elif line.find("overflow") != -1:
+                elif "overflow" in line:
                     return result.RESULT_FALSE_OVERFLOW
                 else:
                     return "ERROR ({0})".format(returncode)
@@ -106,6 +106,6 @@ class Tool(benchexec.tools.template.BaseTool):
         # stop after the first line, that contains the searched text
         for line in lines:
             if line.startswith("KLEE: done: ") and line.find(identifier + " = ") != -1:
-                startPosition = line.rfind("=") + 2
-                return line[startPosition:].strip()
+                splittedLine = line.split(" = ")
+                return splittedLine[1].strip()
         return None
--- a/benchexec/tracerx.py
+++ b/benchexec/tracerx.py
@@ -1,0 +1,111 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+Copyright (C) 2007-2015  Dirk Beyer
+All rights reserved.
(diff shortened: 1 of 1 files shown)


dbeyer:
@rasoolmaghareh If it is similar to KLEE, how about inheriting from the KLEE tool-info module?

rasoolmaghareh:
@dbeyer We have some extra command line options compared to KLEE.

PhilippWendler:
This is not a problem, you can inherit from the Klee module and overwrite those functions that you need (e.g., `cmdline`).

dbeyer:
Or merge and then refactor later? (The merge of the archive is blocked by this merge.)

rasoolmaghareh:
I am not sure if I correctly understand your suggestion. So the suggestion is that we change our archive to use the klee.py? The file tracer.py that we have pushed here is made quite similar to the KLEE file

dbeyer:
You cannot use klee.py if you need a different tracer.py.

But if you only have extra command-line options, then you can specify those
options in the benchmark definition.  For example, like here:
https://github.com/sosy-lab/sv-comp/blob/98c7ff9942ffa029478fc19901df598c1240ed53/benchmark-defs/cpa-seq.xml#L9

rasoolmaghareh:
I think in general the amount of difference with KLEE might be more than just command-line options. We have different executable file and also different libraries might be used. 

Based on this may I ask if you can accept this PR and I will rerun the CI on gitlab?

dbeyer:
I think it can be merged as is, but I will leave the merge click for Philipp, who should also have a look.

benchexec/tracerx.py
```diff
@@ -0,0 +1,111 @@
...
+        This method always needs to be overridden.
+        If the tool gave a result, this method needs to return one of the
+        benchexec.result.RESULT_* strings.
+        Otherwise an arbitrary string can be returned that will be shown to the user
+        and should give some indication of the failure reason
+        (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
+        """
+        for line in output:
+            if line.startswith("KLEE: ERROR: "):
+                if line.find("ASSERTION FAIL:") != -1:
```
PhilippWendler: Instead of all the calls to `find`, just do `if "foo" in line`.

benchexec/tracerx.py
```diff
@@ -0,0 +1,111 @@
...
+                return result.RESULT_DONE
+        return result.RESULT_UNKNOWN
+
+    def get_value_from_output(self, lines, identifier):
+        # search for the text in output and get its value,
+        # stop after the first line, that contains the searched text
+        for line in lines:
+            if line.startswith("KLEE: done: ") and line.find(identifier + " = ") != -1:
+                startPosition = line.rfind("=") + 2
+                return line[startPosition:].strip()
```
PhilippWendler: Code like this would crash if there is for example a truncated line in the output. I guess it would make sense to make it more robust, e.g., using `split` on `" = "`.

benchexec/tracerx.py
rasoolmaghareh: fixed.

benchexec/tracerx.py
rasoolmaghareh: fixed.

dbeyer:
Note commit #0fd24a7.

PR #465, opened 2019-10-21
Tool-info module for Legion

gernst:


benchexec/tools/legion.py
```diff
@@ -0,0 +1,23 @@
+import benchexec.result as result
```
PhilippWendler: Please add a license header (we strongly prefer Apache 2.0 with a copyright assignment to Dirk, if possible, as this would simplify future project maintenance).

gernst:
Thanks!

PR #468, opened 2019-10-22
tools: update Symbiotic

mchalupa:
New version of Symbiotic uses LLVM 8


mchalupa:
I hit a problem with this PR, I'll create a new one once I solve it.

mchalupa:
Problem solved. This PR is fine, so I just re-opened it.

PR #469, opened 2019-10-22
Update gacal.py

bquiring:
I'm not sure why "use-stderr=True" is needed, but I believe it is, referencing [https://github.com/sosy-lab/benchexec/blob/master/benchexec/tools/predatorhp.py]

### DIFF
--- a/benchexec/tools/gacal.py
+++ b/benchexec/tools/gacal.py
@@ -34,7 +34,7 @@ class Tool(benchexec.tools.template.BaseTool):
         return "GACAL"
 
     def version(self, executable):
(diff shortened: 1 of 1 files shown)


PhilippWendler:
> I'm not sure why "use-stderr=True" is needed, but I believe it is, referencing [https://github.com/sosy-lab/benchexec/blob/master/benchexec/tools/predatorhp.py]

It is needed if and only if your tool outputs its version on stderr and not on stdout. Is this the case?

bquiring:
Yes, I believe so.

PR #470, opened 2019-10-24
Add tool-info module for coastal

jacogeld:


benchexec/tools/coastal.py
```diff
@@ -0,0 +1,54 @@
...
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for COASTAL
+    (http://www.cs.sun.ac.za/coastal/).
+    """
+
+    REQUIRED_PATHS = ["coastal-sv-comp"]
```
PhilippWendler: Is this really the only required file for executing your tool? If yes, you can remove this line, because the default implementation of `program_files` adds the executable anyway. If not, please add all other required directories and files.

jacogeld:
It is not!  I'm fixing it now.

--
Jaco

PR #473, opened 2019-11-01
Refactor of react-table

DennisSimon:
Hi @PhilippWendler 

I read through the `react-table` branch and found a couple of places that i thought
could be more readable/maintainable through some refactoring. 

The biggest changes here are : 
* refactor of `LinkOverlay` functions to use `async/await` instead of Promise chains
* rewrite of filter logic in `utils`
* addition of some utility functions + tests

The rest are smaller refactorings here and there as well as some typo fixes. 

Looking forward to your feedback!


PhilippWendler:
Thank you very much, great work!

PR #477, opened 2019-11-01
Rewrite of HTML tables with React

### LINKED ISSUE #123: Lazy-load JavaScript libraries where possible

### LINKED ISSUE #130: table-generator: add sorting feature to columns
I think this feature would be really beneficial.
I know there's a FF plugin, but I (and many others) use Chrome, and existing table-sorting plugins for Chrome do not play nicely with Benchexec.

### LINKED ISSUE #195: Load zip.js from CDN instead of sosy-lab.org
All Javascript libraries except zip.js are loaded from the JSDelivr CDN, because the latter has an older version of zip.js with missing files. As soon as an updated version is available, we want to switch to the CDN (also because using a different domain currently breaks the `--offline` flag).

This depends on https://github.com/gildas-lormeau/zip.js/issues/137#issuecomment-173612814 and probably https://github.com/gildas-lormeau/zip.js/issues/145.

### LINKED ISSUE #214: Make "Filter Rows" dialog in tables apply all changes at the end
Currently each click in the "Filter Rows" dialog in a table produced by table-generator immediately updates the table, which can take some time for large tables. This can be inconvenient, for example when changing existing filters: the user has to first click on "reset" and then select a new filter, and has to wait twice until the browser has rendered the table.

It would be better if there were no table updates while this dialog is open, and only when it is closed the table is updated once with the final visibility state.

### LINKED ISSUE #237: Investigate use of DataTables library
The [DataTables](https://datatables.net/) library seems to be a powerful JavaScript library for tables, providing a lot of features that we have currently implemented ourselves or which would be nice to have:
- [row highlighting](https://datatables.net/examples/api/highlight.html)
- [column hiding](https://datatables.net/examples/api/show_hide.html) with [automatic updating of colspan values](https://datatables.net/examples/advanced_init/complex_header.html)
- [scrolling with fixed header](https://datatables.net/examples/basic_init/scroll_y_dynamic.html)
- [pagination](https://datatables.net/examples/basic_init/alt_pagination.html) with [deferred rendering](https://datatables.net/examples/ajax/defer_render.html) (could be faster for large tables)
- [callbacks for live updating of footer values](https://datatables.net/examples/advanced_init/footer_callback.html) (cf. #125)
- [sorting](https://datatables.net/examples/basic_init/table_sorting.html) including [multi-column sorting](https://datatables.net/examples/basic_init/multi_col_sort.html) (cf. #130)
- [row filters](https://datatables.net/examples/api/multi_filter_select.html)

We should probably use this library in the HTML tables.

### LINKED ISSUE #296: Manual selection of header rows in HTML tables
In some cases, like for [SV-COMP](https://sv-comp.sosy-lab.org/2018/results/results-verified/META_ReachSafety.table.html), the header of HTML tables is still large even after clicking "Shrink Header". However, other rows in the header would be sufficient for the user to distinguish the columns. Thus, it could be useful for the user to be able to choose manually which header rows to show (could be integrated in the "Filter Rows" dialog, e.g., with a checkbox for each header row).

### LINKED ISSUE #297: Compute meaningful row(s) to show in shrinked header of HTML tables
Currently, the "shrinked" header of HTML tables always shows the row "Run set" in addition to the column titles. This is often useful, but not always, e.g., in [regression tests](https://sv-comp.sosy-lab.org/2018/results/results-verified/META_ReachSafety.table.html). Of course, #296 would help, but even better would be an automatic computation which header row(s) to show such that the user can distinguish between the columns. We could define a list of preferred header rows, and then find the first row from this list that has a different value in all cells. If none is present, we could just fall back to the current behavior, or show more than one row.

### LINKED ISSUE #375: header in local HTML-table without loading local Javascript not scrolling scroll left-right.
I know this case is weird, but I got a html-table for some tool via mail and opening it is not nice.
The table was created in local mode (or maybe downloaded from sosy-lab-server including all Javascript sources), such that all internal references to Javascript files are pointing to local files, which were not part of the mail :-(

When opening this local html-table, no Javascript-files are available and some features are missing. I am ok with that, because at least the table is readable.
However, all Javascript-related buttons are shown (and useless) and the header is floating (yes, that works partially). My problem is that the header does not move from riht to left when scrolling from left to right in larger tables. I can not read the right half of the header.

The wanted solution is 
- (important for me:) to deactivate the floating header if some important Javascript files are missing, and
- (important for users in general:) to deactivate all buttons if Javascript cannot be loaded.

### LINKED ISSUE #423: Table-generator: Overview level
Currently, for a given set of benchmark-results (XML files),
table-generator generates one huge table.

The large tables are nice to have a look at details,
however, for getting a quick overview, they are too slow to load.

It would be nice to have, in addition, an overview table (dashboard view) generated
that presents a coarser view on the benchmark results.

First proposal: Omit all results rows, and most of the header, and print only the bottom statistics.
One statistics block for each column (benchmark tool or configuration) and for each vertical block as well (if the input table definition has a `union`).

Either always written (like the csv currently) or only if an extra parameter is given.

PhilippWendler:
This provides a complete rewrite of the HTML tables that `table-generator` produces. Instead of statically creating a HTML `<table>`, we now insert the data as JSON and use [React](https://reactjs.org/) to render a table dynamically. Furthermore, all the spaghetti code for interactive features like plots is now completely replaced by a proper architecture using React components, and there exists infrastructure to test this.

Some new features like better row filters and improvements have also been added based on a user study. But the main advantage is that it is now possible to open large tables in browsers without long waiting periods.

The studies and the rewrite were done by @bschor and a report about this is available as her [Bachelor's thesis](https://www.sosy-lab.org/research/bsc/2019.Bschor.Modern_Architecture_and_Improved_UI_for_Tables_of_BenchExec.pdf).

Users who whish to continue using the old tables can pass `--static-table` to `table-generator`, but we expect this to be removed in BenchExec 3.0.

Fixes #130 (sorting added)
Fixes #423 (summary tab with only header and footer of table added)
Closes #123 (all Javascript is now bundled in the HTML file)
Closes #195 (all Javascript is now bundled in the HTML file)
Closes #214 (filtering is now longer in an overlay, and much faster anyway)
Closes #237 ([react-table](https://www.npmjs.com/package/react-table) is used)
Closes #296 (main table has only small header, full header is shown on summary tab)
Closes #297 (main table has only small header, full header is shown on summary tab)
Closes #375 (no longer applicable)


dbeyer:
Awesome!  Thanks a lot @PhilippWendler and @laurabschor ! Can't wait using this for the competitions!

Issue #438, opened 2019-07-12
Drop Support for Python  2 and 3.4?


PhilippWendler:
Python 2 will be [out of support in 2020](https://pythonclock.org/), so we also drop support for it (already now only `runexec` supports Python 2, not `benchexec`).

Python 3.4 is quite old, too (already > 5 years) and 3.5 was released 4 years ago. All currently supported Ubuntu versions provide Python 3.5 or newer. For Debian it is similar, the only version that has only Python 3.4 is Jessie, which is currently oldoldstable, i.e., regular support ended in 2018 and even the extended security support will end in mid of 2020. Thus we can probably drop support for Python 3.4 at the same time.

If this would be a problem for you, please respond here.

PR #498, opened 2019-11-08
Add tool-info module for new validator: metaval

MartinSpiessl:
metaval works by transforming the validation problem into a
reachability problem and then transparently delegates it to
a classical verifier. For now, there is support for validating
correctness witnesses with ESBMC and CPAchecker, but this will
later be extended to other verifiers with compatible license.

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -54,29 +54,27 @@ class Tool(benchexec.tools.template.BaseTool):
         (knownargs, options) = parser.parse_known_args(options)
         verifierName = knownargs.metaval.lower()
         witnessName = knownargs.witness
-        self.lock.acquire()
-        if not hasattr(self, "wrappedTool"):
-            self.wrappedTool = __import__(
-                "benchexec.tools." + verifierName, fromlist=["Tool"]
-            ).Tool()
-        else:
-            if not verifierName == self.wrappedTool.name().tolower():
-                exit("metaval is called with mixed wrapped tools")
-        self.lock.release()
+        with self.lock:
+            if not hasattr(self, "wrappedTool"):
+                self.wrappedTool = __import__(
+                    "benchexec.tools." + verifierName, fromlist=["Tool"]
+                ).Tool()
+            else:
+                if not verifierName == self.wrappedTool.name().tolower():
+                    exit("metaval is called with mixed wrapped tools")
 
         if hasattr(self, "wrappedTool"):
-            self.lock.acquire()
-            oldcwd = os.getcwd()
-            os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[verifierName]))
-            wrappedOptions = self.wrappedTool.cmdline(
-                self.wrappedTool.executable(),
-                options,
-                [os.path.relpath(os.path.join(oldcwd, "output/ARG.c"))],
-                os.path.relpath(os.path.join(oldcwd, propertyfile)),
-                rlimits,
-            )
-            os.chdir(oldcwd)
-            self.lock.release()
+            with self.lock:
+                oldcwd = os.getcwd()
+                os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[verifierName]))
+                wrappedOptions = self.wrappedTool.cmdline(
+                    self.wrappedTool.executable(),
+                    options,
+                    [os.path.relpath(os.path.join(oldcwd, "output/ARG.c"))],
+                    os.path.relpath(os.path.join(oldcwd, propertyfile)),
+                    rlimits,
+                )
+                os.chdir(oldcwd)
             return [
                 executable,
                 "--verifier",
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -1,0 +1,85 @@
+"""
+BenchExec is a framework for reliable benchmarking.
+This file is part of BenchExec.
+
+Copyright (C) 2007-2019  Dirk Beyer
+All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+import os
+
+
+class Tool(benchexec.tools.template.BaseTool):
+
+    REQUIRED_PATHS = ["CPAchecker", "esbmc"]
+    TOOL_TO_PATH_MAP = {"cpachecker": ["CPAchecker"], "esbmc": ["esbmc"]}
+
+    def executable(self):
+        return util.find_executable("metaval.sh")
+
+    def name(self):
+        return "metaval"
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        if not hasattr(self, "wrappedTool"):
+            return "METAVAL ERROR"
+        return self.wrappedTool.determine_result(
+            returncode, returnsignal, output, isTimeout
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        verifierOption = options.index("--metaval")
+        self.verifierName = options[verifierOption + 1].lower()
+        del options[verifierOption : verifierOption + 2]
+        witnessOption = options.index("--witness")
+        self.witnessName = options[witnessOption + 1]
+        del options[witnessOption : witnessOption + 2]
+        if self.verifierName == "cpachecker":
+            import benchexec.tools.cpachecker as cpachecker
+
+            self.wrappedTool = cpachecker.Tool()
+        elif self.verifierName == "esbmc":
+            import benchexec.tools.esbmc as esbmc
+
+            self.wrappedTool = esbmc.Tool()
+
+        if hasattr(self, "wrappedTool"):
+            os.environ["PATH"] += os.pathsep + os.path.join(
+                os.curdir, *self.TOOL_TO_PATH_MAP[self.verifierName]
+            )
+            return (
+                [executable]
+                + ["--witness"]
+                + [self.witnessName]
+                + tasks
+                + ["--"]
+                + self.wrappedTool.cmdline(
+                    self.wrappedTool.executable(),
+                    options,
+                    ["output/ARG.c",],
+                    propertyfile,
+                    rlimits,
+                )
+            )
+        else:
+            sys.exit("ERROR: Could not find wrapped tool")
+
+    def version(self, executable):
+        stdout = self._version_from_tool(executable, "--version")
+        metavalVersion = next((l.strip() for l in stdout.splitlines())).split()[2]
+        return metavalVersion
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -54,6 +54,7 @@ class Tool(benchexec.tools.template.BaseTool):
         (knownargs, options) = parser.parse_known_args(options)
         verifierName = knownargs.metaval.lower()
         witnessName = knownargs.witness
+        self.lock.acquire()
         if not hasattr(self, "wrappedTool"):
             self.wrappedTool = __import__(
                 "benchexec.tools." + verifierName, fromlist=["Tool"]
@@ -61,6 +62,7 @@ class Tool(benchexec.tools.template.BaseTool):
         else:
             if not verifierName == self.wrappedTool.name().tolower():
                 exit("metaval is called with mixed wrapped tools")
+        self.lock.release()
 
         if hasattr(self, "wrappedTool"):
             self.lock.acquire()
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -44,9 +44,9 @@ class Tool(benchexec.tools.template.BaseTool):
         )
 
     def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
-        parser = argparse.ArgumentParser(add_help=False)
-        parser.add_argument("--witness")
-        parser.add_argument("--metaval")
+        parser = argparse.ArgumentParser(add_help = False, usage = argparse.SUPPRESS)
+        parser.add_argument("--witness", required = True)
+        parser.add_argument("--metaval", required = True)
         (knownargs, options) = parser.parse_known_args(options)
         self.verifierName = knownargs.metaval.lower()
         self.witnessName = knownargs.witness
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -44,17 +44,23 @@ class Tool(benchexec.tools.template.BaseTool):
         )
 
     def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
-        parser = argparse.ArgumentParser(add_help = False, usage = argparse.SUPPRESS)
-        parser.add_argument("--witness", required = True)
-        parser.add_argument("--metaval", required = True)
+        parser = argparse.ArgumentParser(add_help=False, usage=argparse.SUPPRESS)
+        parser.add_argument("--witness", required=True)
+        parser.add_argument("--metaval", required=True)
         (knownargs, options) = parser.parse_known_args(options)
-        self.verifierName = knownargs.metaval.lower()
-        self.witnessName = knownargs.witness
-        self.wrappedTool = __import__("benchexec.tools." + self.verifierName, fromlist=["Tool"]).Tool()
+        verifierName = knownargs.metaval.lower()
+        witnessName = knownargs.witness
+        if not hasattr(self, "wrappedTool"):
+            self.wrappedTool = __import__(
+                "benchexec.tools." + verifierName, fromlist=["Tool"]
+            ).Tool()
+        else:
+            if not verifierName == self.wrappedTool.name().tolower():
+                exit("metaval is called with mixed wrapped tools")
 
         if hasattr(self, "wrappedTool"):
             oldcwd = os.getcwd()
-            os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[self.verifierName]))
+            os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[verifierName]))
             wrappedOptions = self.wrappedTool.cmdline(
                 self.wrappedTool.executable(),
                 options,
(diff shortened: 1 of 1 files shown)


benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+                    rlimits,
+                )
+            )
+        else:
+            sys.exit("ERROR: Could not find wrapped tool")
+
+    def version(self, executable):
+        stdout = self._version_from_tool(executable, "--version")
+        metavalVersion = next((l.strip() for l in stdout.splitlines())).split()[2]
+        print("###############" + metavalVersion)
```
PhilippWendler:
```suggestion
```

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+            import benchexec.tools.cpachecker as cpachecker
+
+            self.wrappedTool = cpachecker.Tool()
+        elif self.verifierName == "esbmc":
+            import benchexec.tools.esbmc as esbmc
+
+            self.wrappedTool = esbmc.Tool()
+
+        if hasattr(self, "wrappedTool"):
+            import os
```
PhilippWendler:
```suggestion
```
Redundant

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        if not hasattr(self, "wrappedTool"):
+            return "METAVAL ERROR"
+        return self.wrappedTool.determine_result(
+            returncode, returnsignal, output, isTimeout
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        verifierOption = options.index("--metaval")
```
PhilippWendler: Wouldn't some proper option parsing make this easier to understand?
Either using `argparse`, or by going through the list with an iterator and branching on the current option. Instead of modifying `options` a new list could be build.

The current solution for example crashes if a user uses `--metaval` without argument.

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+import benchexec.tools.cpachecker as cpachecker
```
PhilippWendler: Why is this imported eagerly here?

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+
+
+class Tool(benchexec.tools.template.BaseTool):
+
+    REQUIRED_PATHS = ["CPAchecker", "esbmc"]
+    TOOL_TO_PATH_MAP = {"cpachecker": ["CPAchecker", "scripts"], "esbmc": ["esbmc"]}
```
PhilippWendler: Why is `scripts` here? This should be specific knowledge of the wrapped tool-info module, shouldn't it?

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        verifierOption = options.index("--metaval")
+        self.verifierName = options[verifierOption + 1].lower()
+        del options[verifierOption : verifierOption + 2]
+        witnessOption = options.index("--witness")
+        self.witnessName = options[witnessOption + 1]
+        del options[witnessOption : witnessOption + 2]
+        if self.verifierName == "cpachecker":
```
PhilippWendler: Why not do this in a generic way without per-verifier branching?

Using `import` will fail with tool-info modules that have a `-` in their name.

One could even reuse the import logic from BenchExec (`benchexec.model.load_tool_info`).

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,89 @@
...
+            self.wrappedTool = esbmc.Tool()
+
+        if hasattr(self, "wrappedTool"):
+            import os
+
+            os.environ["PATH"] += os.pathsep + os.path.join(
+                os.curdir, *self.TOOL_TO_PATH_MAP[self.verifierName]
+            )
+            return (
+                [executable]
```
PhilippWendler: Why not code like `[executable, "--witness", self.witnessName, ...`?

benchexec/tools/metaval.py
MartinSpiessl: Yeah, I remembered this myself a while after I pushed, thx.

benchexec/tools/metaval.py
MartinSpiessl: Because of current time constraints, there is not really time to make everything super nice. I would prefer to do iterative code improvements later. It might also happen that the option structure gets changed completely in the near future, nothing set in stone (yet).

benchexec/tools/metaval.py
MartinSpiessl: because i based my file on the tool-info module of cpachecker and forgot to remove it :sweat_smile: . Thx for the hint.

benchexec/tools/metaval.py
MartinSpiessl: because finding the executable via the utils method did not work without it for some reason. I have no time to debug this *now*. Might do it later

benchexec/tools/metaval.py
MartinSpiessl: I do not fully understand what you mean. I will have a look at  `benchexec.model.load_tool_info` once I have time.

benchexec/tools/metaval.py
MartinSpiessl: Because there will also be lists with more than one element later (`self.wrappedTool.cmdline(...)`). I could also use list deconstruction or something similar. It did organically grow, I have to admit it now looks rather ugly. i will improve on that in a later version

benchexec/tools/metaval.py
MartinSpiessl: Ok, I retried it without the "scripts", it somehow suddenly works without, as I would have expected, so it can be removed. It is however not knowledge specific to the wrapped tool-info module. The wrapped tool is in a sub-folder in metaval's directory, and I control the name of that folder. So instead of "CPAchecker" I could call it "CPAchecker_1.8" , have multiple such directories/versions etc.

benchexec/tools/metaval.py
PhilippWendler: I would suggest `[executable, "--witness", self.witnessName] + tasks + ["--"] + self...` as easy improvement.

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,85 @@
...
+                    propertyfile,
+                    rlimits,
+                )
+            )
+        else:
+            sys.exit("ERROR: Could not find wrapped tool")
+
+    def version(self, executable):
+        stdout = self._version_from_tool(executable, "--version")
+        metavalVersion = next((l.strip() for l in stdout.splitlines())).split()[2]
```
PhilippWendler:
```suggestion
        metavalVersion = next(stdout.splitlines()).strip().split()[2]
```

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,85 @@
...
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+import os
+
+
+class Tool(benchexec.tools.template.BaseTool):
+
+    REQUIRED_PATHS = ["CPAchecker", "esbmc"]
```
PhilippWendler:
```suggestion
    REQUIRED_PATHS = util.flatten(TOOL_TO_PATH_MAP.values())
```
Would avoid duplication of required paths (order of fields needs to be swapped, of course).

Or even simplify `TOOL_TO_PATH_MAP` by making its values single strings instead of lists and remove the `flatten`.

benchexec/tools/metaval.py
MartinSpiessl: This will not work, as `splitlines()` does not return an iterator.

benchexec/tools/metaval.py
MartinSpiessl: > Ok, I retried it without the "scripts", it somehow suddenly works without,

Looks like I retried this in the wrong benchexec folder on my PC. The `CPAchecker/scripts` was necessary because though the tool info module of CPAchecker has `scripts/cpa.sh` as fallback for searching the executable, this is only tried in the current working directory. But the current working directory for my tool is not the CPAchecker folder, it is one above:
```
/metaval
├── CPAchecker
├── esbmc
├── LICENSE
└── metaval.sh
```

I changed the tool mechanics to enter the wrapped tool's folder now before resolving everything, such that i can fully delegate everything to the wrapped tool-info module.

benchexec/tools/metaval.py
PhilippWendler: Then `stdout.splitlines()[0]`

MartinSpiessl: Yes, that is already pushed :smile:

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,82 @@
...
+            return "METAVAL ERROR"
+        return self.wrappedTool.determine_result(
+            returncode, returnsignal, output, isTimeout
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument("--witness")
+        parser.add_argument("--metaval")
+        (knownargs, options) = parser.parse_known_args(options)
```
PhilippWendler: Thanks for the improvements.

Some error handling would still be good. Are these options required? `add_argument` has a parameter `required`.

benchexec/tools/metaval.py
MartinSpiessl: Using `benchexec.model.load_tool_info` does not work in all circumstances, because it is overwritten by the [Monkey-patch™](https://github.com/sosy-lab/cpachecker/blob/7bf51638177f593549e98fc1985752af12dd3903/scripts/benchmark.py#L192) in CPAcheckers `benchmark.py`.
I used the mechanism from `benchexec.model.load_tool_info` to instantiate the tool.

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,82 @@
...
+        self.verifierName = knownargs.metaval.lower()
+        self.witnessName = knownargs.witness
+        self.wrappedTool = __import__("benchexec.tools." + self.verifierName, fromlist=["Tool"]).Tool()
+
+        if hasattr(self, "wrappedTool"):
+            oldcwd = os.getcwd()
+            os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[self.verifierName]))
```
PhilippWendler: Hm, unfortunately, this won't work. For example, if two threads use this tool-info module at the same time, they will overwrite their respective changes to the current directory. (Even if this was not a problem, the second `chdir` would need to be in a `finally` block.)

Maybe the best is to consider how to make CPAchecker's tool-info module find `cpa.sh` in your case. Manipulating `os.environ["PATH"]` would be an option, but has the same problem. Some changes to CPAchecker's tool-info module or `util.find_executable()` maybe?

A last possibility would be to add an optional parameter with a base dir to CPAchecker's `executable()` method. Metaval's module would then attempt to call the method with the parameter and fall back to calling it without. But maybe there are better options.

MartinSpiessl: > Manipulating os.environ["PATH"] would be an option, but has the same problem

Yes, I did this before. That was why I added the "scripts" in the `TOOL_TO_PATH_MAP`.
What about locking and thus making this transaction atomic?

> A last possibility would be to add an optional parameter with a base dir to CPAchecker's executable() method

The problem is not just CPAchecker, potentially every tool ...

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument("--witness")
+        parser.add_argument("--metaval")
+        (knownargs, options) = parser.parse_known_args(options)
+        self.verifierName = knownargs.metaval.lower()
+        self.witnessName = knownargs.witness
+        self.wrappedTool = __import__("benchexec.tools." + self.verifierName, fromlist=["Tool"]).Tool()
```
PhilippWendler: I just noticed that these values are stored in fields. This is problematic, because the next call to `cmdline` will immediately overwrite these values. So everything that is not necessary outside of this method must not be stored in a field.

For `wrappedTool`, I suggest (1) setting it only if it is not set (guarding against race conditions, of course) and (2) if it is set, checking its name against the specified name. The latter would fix the problem that if a user specifies two different validators for some task sets, currently `determine_result` would be called on the wrong module.

benchexec/tools/metaval.py
MartinSpiessl: By the way:
> For example, if two threads use this tool-info module at the same time, they will overwrite their respective changes to the current directory.

Why isn't it designed in a way that the two threads use different tool instances?

PhilippWendler: > What about locking and thus making this transaction atomic?

I guess that could work. The lock should cover only the smallest amount of possible code (calling `executable()`, not `cmdline()`). It can be combined with the lock that covers the loading of the tool-info module, if the result of `executable()` is also stored in a field.

But did you actually test this? The result of doing `chdir` will be that the paths returned by `executable()` and `cmdline()` are relative to that directory, and this shouldn't work for the actual run, which is executed from a different directory. You could specify a working directory for the run (in the respective method), but this would make Metaval more difficult for users to execute manually, I guess?

> Why isn't it designed in a way that the two threads use different tool instances?

Why should it be? Tool-info modules are supposed to be stateless anyway. And it just fits naturally into BenchExec's architecture that way.

benchexec/tools/metaval.py
MartinSpiessl: I changed it accordingly

benchexec/tools/metaval.py
```diff
@@ -44,17 +48,26 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
+        parser = argparse.ArgumentParser(add_help=False, usage=argparse.SUPPRESS)
+        parser.add_argument("--witness", required=True)
+        parser.add_argument("--metaval", required=True)
         (knownargs, options) = parser.parse_known_args(options)
-        self.verifierName = knownargs.metaval.lower()
-        self.witnessName = knownargs.witness
-        self.wrappedTool = __import__("benchexec.tools." + self.verifierName, fromlist=["Tool"]).Tool()
+        verifierName = knownargs.metaval.lower()
+        witnessName = knownargs.witness
+        self.lock.acquire()
```
PhilippWendler: Please use `with` for such things.

benchexec/tools/metaval.py
MartinSpiessl: > But did you actually test this? The result of doing chdir will be that the paths returned by executable() and cmdline() are relative to that directory, and this shouldn't work for the actual run, which is executed from a different directory.

Yes, I tested this and it works nicely. This is why `--verifier` is passed to the tool, such that it can figure out relative to which folder the paths are specified. This way everything is identical to as if I would call the tool directly. Some tools like esbmc do not support being called from outside of their folder (`esbmc-wrapper.py` hardcoded `the path to the executable as "./esbmc"` for example)

> You could specify a working directory for the run (in the respective method), but this would make Metaval more difficult for users to execute manually, I guess?

I guess this is essentially what the `--verifier` option now does (though the resolving of toolname to verifier dir is done by the tool, as to not burden the user with that choice).

> Why should it be? Tool-info modules are supposed to be stateless anyway. And it just fits naturally into BenchExec's architecture that way.

I am not talking about the module, but the instance of the subclass of the class `Tool`

benchexec/tools/metaval.py
MartinSpiessl: Thanks, done!

benchexec/tools/metaval.py
MartinSpiessl: > tool-info modules are supposed to be stateless anyway.

Hmm, then instead of having a state in `self.wrappedTool` for `determine_result`, I could put the tool name at the start of the ouput and read this in again once determine_result is called :thinking:.

Would that be an improvement in your opinion?

PhilippWendler: As long as you do not require different validators per task, the existing solution is ok for me. (It is not mutable state, just lazily initialized constant state.) For a better general-purpose solution there is #166.

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,94 @@
...
+        (knownargs, options) = parser.parse_known_args(options)
+        verifierName = knownargs.metaval.lower()
+        witnessName = knownargs.witness
+        with self.lock:
+            if not hasattr(self, "wrappedTool"):
+                self.wrappedTool = __import__(
+                    "benchexec.tools." + verifierName, fromlist=["Tool"]
+                ).Tool()
+            else:
+                if not verifierName == self.wrappedTool.name().tolower():
```
PhilippWendler: There is no guarantee that the name of the tool-info module equals `Tool.name().tolower()`. Indeed, it is already not always the case, for example for Ultimate-based tools. The `name()` function is supposed to return a human-readable name, which includes spaces in many cases.

Either store the module name, or try `inspect.getmodule()` on the tool instance or its class.

benchexec/tools/metaval.py
```diff
@@ -0,0 +1,94 @@
...
+                oldcwd = os.getcwd()
+                os.chdir(os.path.join(oldcwd, self.TOOL_TO_PATH_MAP[verifierName]))
+                wrappedOptions = self.wrappedTool.cmdline(
+                    self.wrappedTool.executable(),
+                    options,
+                    [os.path.relpath(os.path.join(oldcwd, "output/ARG.c"))],
+                    os.path.relpath(os.path.join(oldcwd, propertyfile)),
+                    rlimits,
+                )
+                os.chdir(oldcwd)
```
PhilippWendler: Please still use a finally block (`cmdline` could of course throw an exception).

benchexec/tools/metaval.py
MartinSpiessl: it is easier to store the name, done

benchexec/tools/metaval.py
PhilippWendler: Sorry for being unclear, I specifically meant the second `chdir` call.

MartinSpiessl: Oh, that makes sense! I did not find a try-with-finally construct in python (I guess they don't have this, which is a bit odd) which is similar to the one in Java, so I added a try-finally block inside the with block.

MartinSpiessl:
Sure, done.  
Btw, python 3.4 has reached its EOL on 2019-03-16 if I am not mistaken. I would rather be able to use new features than having support for deprecated legacy versions.

PhilippWendler:
> Btw, python 3.4 has reached its EOL on 2019-03-16 if I am not mistaken. I would rather be able to use new features than having support for deprecated legacy versions.

Python 3.4 was supported in an Ubuntu LTS until a few months ago and is still supported in Debian oldoldstable, so we didn't want to immediately remove support for it. People might want to use BenchExec on machines like HPC clusters which often tend to run older distributions. But dropping Python 3.4 is already planned for January (#438).

Issue #513, opened 2019-11-17
Table does not resize vertically on window-size change


MartinSpiessl:
When I zoom out in the table tab (e.g. by pressing ctrl+- multiple times) I expect the table to fill out at least the same percentage of my browser window.
Currently, the number of rows visible seems to be fixed, so the table shrinks until almost my whole screen is empty.
![tables](https://user-images.githubusercontent.com/33026673/69012800-26cf4780-097a-11ea-8870-67cd3336f3a7.jpg)

MartinSpiessl:
I just realized I tool a bad table as example, because it has only a few entries. The problem also exists if you take a table with >100 entries, i.e., of there are more rows than can be displayed.

PhilippWendler:
Also happens when resizing the window. A workaround is to reload the page.

I do not know how to solve this unfortunately.

PR #517, opened 2019-11-18
#513: make table resize

DennisSimon:
Hi @PhilippWendler 

here is a PR mainly for https://github.com/sosy-lab/benchexec/issues/513 .
I have added a CSS rule containing percentages instead of hard coded pixel sizes to deal with
browser zooms, as there is no event that handles them (they don't trigger a `resize` event).

I also ran `npm audit fix` as there was a medium and severe vulnerability in the dependencies.

The last thing i did was editing the `start` script so it triggers the generation of `dependencies.json` before the local dev server starts, as we would miss the file otherwise. 

Looking forward to your feedback!


PhilippWendler:
Hm, that doesn't work for me because now the horizontal scroll bar in the table is missing.

PhilippWendler:
I added the two other fixes, though, thanks for bringing this up.

DennisSimon:
Hi Philipp, 

can you describe a case where this is happening? 
I altered the data.json a little bit to add more columns and for me, a scroll bar is rendered. 
It is also rendered when zooming in far enough. 
table_1
table-2

The only case where the scrollbar is hidden, is if the height of the window is less than the minimal
height of the table, in which case the bar will be rendered at the bottom and you would need to scroll
down to see it.

table-3

This is observable on the current master branch as well though.

Would it be expected behaviour for the scroll bar to be in a fixed position?

PhilippWendler:
Indeed, the scroll bar should always be visible and it is for me:
grafik

DennisSimon:
Interesting, I wonder if this is browser specific in some way. I will look at it further this evening.

marto97:
@PhilippWendler now scroll bar is now always visible. 

Looking forward to your feedback!

PhilippWendler:
I am not sure we want a solution that overrides styles from the browser and does things like forcing scrollbar width to a specific value (which could look weird if the user's other scrollbars are all different).

Could you please look from scratch at #513 and think how this could be solved without such a workaround? Assinging #517 to you was mostly done as a way to show you the previous attempt.

marto97:
Normally only adding `overflow-y: scroll;` should work as I see in: [stackoverflow](https://stackoverflow.com/questions/24700184/how-do-i-force-a-vertical-scrollbar-to-appear) or [css-tricks](https://css-tricks.com/snippets/css/force-vertical-scrollbar/), but unfortunately this way doesn't fix the issue.

Another way I think it could fix the issue, is by reading which browser the user is using and then forcing the same scrollbar style. Is it okay?

PhilippWendler:
No, we should definitely not check make any assumptions about how scrollbars of a browser look.

marto97:
Unfortunately I couldn't find another way to fix this issue.

PhilippWendler:
@marto97 Does this mean that you did not find a way to fix the problem with this PR, or that you think that there is no other solution for fixing #513? If the former, please ignore this PR and work on #513 from scratch.

PR #521, opened 2019-11-20
update BRICK tool-info

brick-tool-dev:
add environment variable

### DIFF
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -67,3 +67,6 @@ class Tool(benchexec.tools.template.BaseTool):
                 break
 
         return status
+
+    def environment(self, executable):
+        return {"additionalEnv": {"LD_LIBRARY_PATH": "./lib"}}


PhilippWendler:
While it is possible to do what is suggested in this PR, this would mean that executing your tool manually would be tricky for users. If they just use a command line as copied from the tool logs, it will fail, probably with an error message that is not helpful.

Don't you think it would make your tool more user-friendly if it set this environment variable internally, e.g., in a wrapper script?

brick-tool-dev:
Hi,
actually this environment variable is needed before trying to set it internally.
If I don't add LD_LIBRARY_PATH, it can't run to set it.

Can I merge it first and find a better solution later?

dbeyer:
It would be nicer to the users (and me) if you had a wrapper script that sets the variable and then starts the actual tool.

In the log files, the first line is the command line and it is convenient for reproducing results to be able to simply copy & paste the command line.

brick-tool-dev:
Hi, I have written a wrapper script and uploaded a new zip to [archives-2020](https://gitlab.com/sosy-lab/sv-comp/archives-2020/merge_requests/95).
Can I ask a new run to check whether it can run properly?
A time limit of 60s will be enough to see whether the script can work.

PhilippWendler:
Is this still relevant?

dbeyer:
According to the latest benchmark of BRICK in the context of SV-COMP,
using the new archive, the new wrapper script works well.

@PhilippWendler I suggest to close this issue.

@brick-tool-dev Could you please confirm?

brick-tool-dev:
Hi, the warpper script works properly. So it's nice to close this pr. Thanks so much!

PR #522, opened 2019-11-21
Improve and extend metaval tool-info module

MartinSpiessl:
There was a bug in the delegation of `determine_result`
to the wrapped tool. Some tools like Symbiotic determine the version
again in this step and need access to the binary again. If we do not
chdir into the wrapped tool's folder, the binary will not be found.

This PR also adds some configuration options. The first one is
needed for running older versions of wrapped tools which need Java 8
and therefore a custom entry in the PATH variable. The second one
will enable the tool user to ensure that only a certain type of
witness is checked.

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -28,7 +28,14 @@ import threading
 
 class Tool(benchexec.tools.template.BaseTool):
 
-    TOOL_TO_PATH_MAP = {"cpachecker": "CPAchecker", "esbmc": "esbmc"}
+    TOOL_TO_PATH_MAP = {
+        "cpachecker-metaval": "CPAchecker",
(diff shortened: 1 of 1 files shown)


PhilippWendler:
The code needed refactoring. Because the nicest way to write the code uses somewhat advanced Python features, I took the liberty of adding the code myself.

But I did not test it, so please do this.

Furthermore, while you are at it:
- There is a lone `exit` in `cmdline` that needs to be a `sys.exit`.
- The tool-info module does not have documentation. Please add a docstring with at least name and URL of tool (as in the other modules).

MartinSpiessl:
Thanks for the welcome change regarding the context, I had the same change on my ToDo list for after the competition.

I added documentation, changes `exit` to `sys.exit` and ran some integration tests, everything is still working.

Issue #515, opened 2019-11-18
Easier way to load development environment with other tables


PhilippWendler:
Our setup for development of the new HTML tables has the command `npm start` that starts a web server with an example table in development mode (non-minified build, debug mode of React, live reload after corresponding files change). This is nice and useful, but by default always uses the same [small table](https://github.com/sosy-lab/benchexec/blob/master/benchexec/tablegenerator/react-table/src/data/data.json). Furthermore, the code that is responsible for this is part of the main code, not part of the test code or the development scripts (but at least it is just 3 lines).

It is already possible to use the development mode with all of our [test tables](https://github.com/sosy-lab/benchexec/tree/master/benchexec/tablegenerator/test_integration/expected) if one replaces `data.json` with  a hardlink to one of the test tables (the `.html` files, which actually contain pure JSON), but this could be improved:

- Provide a way to pass the desired test table as parameter to `npm start`, without need for hardlinks or symlinks.
- Provide a way to specify an arbitrary existing HTML table as source of the data and not only files that contain only the pure JSON. Everything except the JSON data should then be ignored.

DennisSimon:
This would definitely be a huge dev experience improvement.
Will look into creating something for this this evening.

PR #523, opened 2019-11-21
#515: optional custom data

DennisSimon:
Hi @PhilippWendler,

i created a solution for https://github.com/sosy-lab/benchexec/issues/515. 

A developer is now able to pass a path to a test file as an argument to the start script.
`npm run start [<./relative/path/to/test/file>]`
The script then copies the file into `<root>/src/data`, so it's accessible by webpack. 
If no argument is passed, the default `data.json` will be used. 
I also aliased the path to whichever test file is used as `@data` for readability.

Apart from this, i also fixed the `start` and `test` scripts in the `package.json` by wrapping the
first expression in parenthesis, as code execution would stop if one condition of the logical 
`or` connection is satisfied. 

Lastly, tests were breaking because of the inline usage of `imports-loader`, as jest tests
do not get precompiled by webpack. 
I configured the module resolver to only return a string if it matches `imports-loader?.*`.
We are able to extend this into a full-fledged mock if required in the future.

As always, looking forward to your feedback :)

### DIFF
--- a/benchexec/tablegenerator/react-table/package.json
+++ b/benchexec/tablegenerator/react-table/package.json
@@ -19,9 +19,9 @@
     "react-vis": "^1.11.7"
   },
   "scripts": {
-    "start": "(test -f src/data/dependencies.json ||node scripts/dependencies.js) && node scripts/start.js",
+    "start": "test -f src/data/dependencies.json ||node scripts/dependencies.js && node scripts/start.js",
     "build": "node scripts/dependencies.js && node scripts/build.js",
-    "test": "(test -f src/data/dependencies.json || node scripts/dependencies.js) && node scripts/test.js"
+    "test": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/test.js"
   },
   "eslintConfig": {
     "extends": [
@@ -70,7 +70,6 @@
     "moduleNameMapper": {
       "^react-native$": "react-native-web",
       "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy",
-      "imports-loader?.*": "<rootDir>/src/tests/mocks/importsLoader.js",
       "@data": "<rootDir>/src/data/data.json"
     },
     "moduleFileExtensions": [
--- a/benchexec/tablegenerator/react-table/package.json
+++ b/benchexec/tablegenerator/react-table/package.json
@@ -19,9 +19,9 @@
     "react-vis": "^1.11.7"
   },
   "scripts": {
-    "start": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/start.js",
+    "start": "(test -f src/data/dependencies.json ||node scripts/dependencies.js) && node scripts/start.js",
     "build": "node scripts/dependencies.js && node scripts/build.js",
-    "test": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/test.js"
+    "test": "(test -f src/data/dependencies.json || node scripts/dependencies.js) && node scripts/test.js"
   },
   "eslintConfig": {
     "extends": [
--- a/benchexec/tablegenerator/react-table/package.json
+++ b/benchexec/tablegenerator/react-table/package.json
@@ -69,7 +69,9 @@
     "modulePaths": [],
     "moduleNameMapper": {
       "^react-native$": "react-native-web",
-      "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy"
+      "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy",
+      "imports-loader?.*": "<rootDir>/src/tests/mocks/importsLoader.js",
+      "@data": "<rootDir>/src/data/data.json"
     },
     "moduleFileExtensions": [
       "web.js",
--- a/benchexec/tablegenerator/react-table/scripts/start.js
+++ b/benchexec/tablegenerator/react-table/scripts/start.js
@@ -48,10 +48,8 @@ if (dataParam) {
   fs.writeFileSync(path.resolve("./src/data/custom-data.json"), content);
 }
 
-const DATA =
-  (dataParam && "./src/data/custom-data.json") || "./src/data/data.json";
-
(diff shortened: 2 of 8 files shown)


benchexec/tablegenerator/react-table/package.json
```diff
@@ -69,7 +69,9 @@
     "modulePaths": [],
     "moduleNameMapper": {
       "^react-native$": "react-native-web",
-      "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy"
+      "^.+\\.module\\.(css|sass|scss)$": "identity-obj-proxy",
+      "imports-loader?.*": "<rootDir>/src/tests/mocks/importsLoader.js",
```
PhilippWendler: You said that tests were breaking because of `imports-loader`, but in 9137bf4 I specifically added the line `"^.*/src/vendor/.*$": "<rootDir>/config/jest/fileTransform.js"` above to keep them working, and they do work for me and in the CI.

In which circumstances does it not work?

benchexec/tablegenerator/react-table/scripts/start.js
```diff
@@ -40,6 +41,18 @@ const createDevServerConfig = require("../config/webpackDevServer.config");
 const useYarn = fs.existsSync(paths.yarnLockFile);
 const isInteractive = process.stdout.isTTY;
 
+const dataParam = process.argv[2];
+
+if (dataParam) {
+  const content = fs.readFileSync(path.resolve(__dirname, "../", dataParam));
+  fs.writeFileSync(path.resolve("./src/data/custom-data.json"), content);
```
PhilippWendler: Copying the file is a nice and simple solution, I like that.

I guess we could use [copyFileSync](https://nodejs.org/api/fs.html#fs_fs_copyfilesync_src_dest_flags)?

benchexec/tablegenerator/react-table/scripts/start.js
```diff
@@ -40,6 +41,18 @@ const createDevServerConfig = require("../config/webpackDevServer.config");
...
+}
+
+const DATA =
+  (dataParam && "./src/data/custom-data.json") || "./src/data/data.json";
```
PhilippWendler: A ternary operator is probably more obvious in such a case. And the value could be directly assigned to `process.env.DATA`.

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -21,7 +21,7 @@ import { faQuestionCircle } from "@fortawesome/free-solid-svg-icons";
 
 //example data for development
 if (process.env.NODE_ENV !== "production") {
-  window.data = require("../data/data.json");
+  window.data = require("@data");
```
PhilippWendler: The alias is certainly not a bad idea, but I wonder whether it is overkill here. To understand what file is loaded, one has to first look at the alias definition, and then at the environment variable, so a double indirection.

Can we use either the alias or the environment variable instead of both?

benchexec/tablegenerator/react-table/package.json
```diff
@@ -19,9 +19,9 @@
     "react-vis": "^1.11.7"
   },
   "scripts": {
-    "start": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/start.js",
+    "start": "(test -f src/data/dependencies.json ||node scripts/dependencies.js) && node scripts/start.js",
```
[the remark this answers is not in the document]
DennisSimon: Javascripts binary operators return early and will not evaluate unnecessary expressions: 
`A && B` will not evaluate `B` if `A` is falsy, as the whole expression will be false anyway. 
`A || B` will not evaluate `B` if `A` is truthy, as it is deemed unnecessary (`(true || false) === true`).

One easy way to test this is like this:
```
>node

const a = () => {
    console.log('a');
    return true;
}

const b = () => {
    console.log('b');
    return false;
}

b() || a() // will print 'b' and 'a'
a() || b() // will only print 'a'

```
Thats exactly the case that happened to me when `dependencies.json` was present.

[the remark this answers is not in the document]
PhilippWendler: But this is shell code, not JavaScript code. Of course both `||` and `&&` are also short-circuiting in shell, but the operator precedence seems different. In my shell, `false || echo 1 && echo 2` prints both numbers and `true || echo 1 && echo 2` prints `2`, just as expected and as we need it here. In which environment did it not work for you?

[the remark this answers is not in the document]
DennisSimon: You are completely right, i think the only thing short-circuting in this case was my brain :D

benchexec/tablegenerator/react-table/package.json
DennisSimon: I cleaned my environment and forced a npm reinstall. After that it worked. 
I think something was ccorrupted on my end for whatever reason. :/

benchexec/tablegenerator/react-table/scripts/start.js
DennisSimon: Nice catch, will refactor this part

PhilippWendler:
Sorry for the small conflict that exists now. Solving it is easy, just apply the changes you made to `Overview.js` to `App.js` instead.

I decided to do de1fa9f66dbd3e80b1b3258b346fe722be7f889f now although this PR is open because it might actually help. Now the only test that needs `data.json` is  `App.test.js`, all other tests do not even attempt to load it. So we could also load the data in `App.test.js` if this simplifies things in other places.

benchexec/tablegenerator/react-table/src/components/Overview.js
DennisSimon: I went with a dynamic bind of `@data` in the webpack config, as we are unable to require modules from a env variable

PhilippWendler:
Thanks for the changes, looks good. Could you please cleanup the PR such that it contains only the relevant changes and not those where we noticed they are unnecessary?

PhilippWendler:
Thanks, I squash-merged in 415c1ef.

PR #527, opened 2019-11-25
Fix REQUIRED_PATHS for LibKluzzer

hoangmle:
@dbeyer I had a problem with the last two pre-runs that a directory from the zip archive was missing. Now I realize that possibly only paths specified in REQUIRED_PATHS will be copied for each single run. Can you please merge this and schedule a new pre-run for LibKluzzer using the current zip archive? Thanks!


dbeyer:
Your assumption is correct. Only the required paths are copied.

PR #540, opened 2019-12-12
Add script to transform old-style tasks to task definition files

lembergerth:
This PR adds the python script that we use to transform old-style C-file tasks to the new yaml task definitions.
This may be interesting for other users that still use the old format.

The script takes as arguments C files and/or .set files that contain a glob-list of C files.
The script renames the C files with '_true-PROPERTY' and "_false-PROPERTY"
verdicts in filenames to not contain these verdicts anymore.
If two files handled in the same call to this script have a same new
name, the script adds a suffix '-1', '-2', etc. to the new names to avoid
collisions. If the name with the suffix also already exists, the counter is increased.

Next to each renamed task, a yml-file with the same name as the new C file
is created. The yml file links to the renamed file and lists all
properties that were previously encoded in the file name.
This yml-file is the new way to define tasks,
cf. https://github.com/sosy-lab/benchexec/blob/master/doc/benchexec.md#task-definition-files

### DIFF
--- a/contrib/create_yaml_files.py
+++ b/contrib/create_yaml_files.py
@@ -1,0 +1,274 @@
+#!/usr/bin/env python3
+
+import glob
+import os
+import re
+import argparse
+from typing import Tuple
+
+CHANGE_FILE_NAME = True
+"""Whether to shorten file names by removing the verdicts, if possible without collisions."""
+AVOID_COLLISIONS_ACROSS_DIRECTORIES = True
+"""Whether to avoid same file names in different directories."""
+
+NAME_TO_PROP_AND_SUBPROP = {
+    "unreach-call": ("unreach-call.prp", None),
+    "cover-error": ("coverage-error-call.prp", None),
+    "unreach-label": ("ALL.prp", None),
+    "termination": ("termination.prp", None),
+    "no-overflow": ("no-overflow.prp", None),
+    "valid-memcleanup": ("valid-memcleanup.prp", None),
+    "valid-memsafety": ("valid-memsafety.prp", None),
+    "valid-deref": ("valid-memsafety.prp", "valid-deref"),
+    "valid-free": ("valid-memsafety.prp", "valid-free"),
+    "valid-memtrack": ("valid-memsafety.prp", "valid-memtrack",),
+    "def-behavior": ("def-behavior.prp", None),
+}
+
+CANDIDATE_REGEX = re.compile(r".*\.(c|i)")
+
+DUMMY_SET = "__NO_SET__"
+"""Dummy set name used for C files given on the command-line."""
+
+
+def _get_prop(property_file, property_dir, task_dir):
+    return os.path.relpath(os.path.join(property_dir, property_file), task_dir)
+
+
+def handle_c(task_file) -> Tuple[str, dict]:
+    """ Create yml task definition for the given file.
+    Return a tuple of a recommended new task name and the yml info as dictionary.
+
+    """
+    properties = list()
+    name_pcs_dot = task_file.split(".")
+    new_name_pcs_dot = list()
+    for pd in name_pcs_dot:
+        name_pcs = pd.split("_")
+        new_name_pcs = list()
+        for p in name_pcs:
+            offset = 0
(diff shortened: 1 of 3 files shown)


contrib/create_yaml_files.py
```diff
@@ -0,0 +1,274 @@
...
+                prefix_len = 0
+            prefix = task[:prefix_len]
+            globbed_tasks = glob.glob(prefix + glob_suffix)
+            globbed_tasks = [t for t in globbed_tasks if t.endswith(".yml")]
+            assert len(globbed_tasks) > 0
+
+            globbed_tasks = glob.glob(prefix + glob_suffix)
+            globbed_tasks = [t for t in globbed_tasks if t.endswith(".yml")]
+            if not task_set.endswith("testable.set") and (
+                not USE_SUFFIX_WILDCARDS or len(globbed_tasks) > 3 or prefix[-1] == "/"
```
PhilippWendler: `USE_SUFFIX_WILDCARDS` is an undefined identifier.

contrib/create_yaml_files.py
```diff
@@ -0,0 +1,274 @@
...
+            os.rename(old_name, curr_task)
+            if old_name[-1] == "i":
+                # *.i -> *.c
+                if os.path.exists(old_name[:-1] + "c"):
+                    old_c = old_name[:-1] + "c"
+                # *.c.i -> *.c
+                elif os.path.exists(old_name[:-2]):
+                    old_c = old_name[:-2]
+                # ldv-memsafety/memleaks*.i -> ldv-memsafety/memleaks-notpreprocessed/memleaks*.c
+                elif old_name.startswith("ldv-memsafety/memleaks"):
```
PhilippWendler: This could be removed?

contrib/create_yaml_files.py
```diff
@@ -0,0 +1,274 @@
...
+                        + old_name.split("/")[-1][:-1]
+                        + "c"
+                    )
+                else:
+                    old_c = None
+                if old_c:
+                    assert old_c not in all_tasks
+                    curr_task_name = curr_task.split("/")[-1]
+                    new_c_name = (
+                        os.path.dirname(old_c) + "/" + curr_task_name[:-1] + "c"
```
PhilippWendler: I suggest to use the standard Python functions for path handling instead of string manipulation:
- `os.path.basename` instead of `split("/")[-1]`, `os.path.split` if both `dirname` and `basename` are needed
- `os.path.join` instead of string concatenation

(of course also at other locations)

contrib/create_yaml_files.py
```diff
@@ -0,0 +1,274 @@
...
+
+import glob
+import os
+import re
+import argparse
+from typing import Tuple
+
+CHANGE_FILE_NAME = True
+"""Whether to shorten file names by removing the verdicts, if possible without collisions."""
+AVOID_COLLISIONS_ACROSS_DIRECTORIES = True
```
PhilippWendler: These two look like they should be arguments instead of constants?

contrib/create_yaml_files.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/env python3
+
```
PhilippWendler: Could you please add a license?

contrib/create_yaml_files.py
lembergerth: Sorry about that. I removed it.

contrib/create_yaml_files.py
lembergerth: You're right, that was specific to the conversions in the sv-benchmarks repository. I removed the elif branch.

contrib/create_yaml_files.py
lembergerth: Yes, of course. I didn't find any other locations were this happened, though, so sorry if I missed some.

contrib/create_yaml_files.py
lembergerth: Done.

contrib/create_yaml_files.py
lembergerth: I felt that they should be changed that infrequently that I didn't make the effort to transform them, but it's done now.

PR #547, opened 2020-01-27
fix import of "util" for contrib-script.

kfriedberger:
Without this fix, we get an exception:
<pre>
$> contrib/plots/quantile-generator.py
  File "benchexec/contrib/plots/quantile-generator.py", line 34, in <module>
    Util = tablegenerator.Util
AttributeError: module 'benchexec.tablegenerator' has no attribute 'Util'
</pre>


PhilippWendler:
Thanks, I overlooked this script. I added a different fix because importing an import of a different module should not be done anyway.

PR #548, opened 2020-01-28
Fix layout of table for selecting columns

### LINKED ISSUE #543: The table for selecting columns does not correctly show missing columns
I have two task-sets with different data-columns (one task-set has only a column `A`, the other task-set has columns `A` and `B`). Combining them side-by-side in one html-table works. However, the dialog (table) for selecting viewed columns does not show this correctly. Clicking seems to work properly, e.g., hiding and viewing a column (and also a complete set of identical-named columns) works as expected, and all columns are available. Just the view is invalid.

Example: see column `#reached`/`host`/`#blocks`/`#hits` in the appended screenshot.

Table_Bug_BDD

marto97:
Fixes #543.


benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +53,39 @@ export default class SelectColumn extends React.Component {
   };
 
   renderColumns = index => {
-    return this.state.list[index].columns.map(column => (
+    this.listTable = [];
+    let idxColumn = [];
+    this.state.list.forEach(tool => {
+      this.selectable.forEach(headerRow => {
+        let length = tool.columns.length;
+        tool.columns.forEach(function(column, idx, i) {
```
PhilippWendler: This loops over `tool.columns`, but only does something in one loop iteration (either where the matching column is or at the end). All other loop iterations are empty, so this seems like an unnecessary loop. Wouldn't it be easier to just check if `tool.columns` contains a matching `column` with an `if`?

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +53,39 @@ export default class SelectColumn extends React.Component {
...
+          if (headerRow.display_title === column.display_title) {
+            idxColumn.push({
+              display_title: column.display_title,
+              isVisible: column.isVisible,
+              key: idx * i.max_width
+            });
+          } else if (
+            idx === length - 1 &&
+            idxColumn
+              .map(elem => elem.display_title.includes(headerRow.display_title))
```
PhilippWendler: `.map().includes()` can be simplified with `.some()`.

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +53,39 @@ export default class SelectColumn extends React.Component {
...
+              isVisible: false,
+              key: idx * headerRow.max_width
+            });
+          }
+        });
+      });
+      this.listTable.push({ columns: idxColumn });
+      idxColumn = [];
+    });
+    return this.listTable[index].columns.map(column => (
```
PhilippWendler: The code creates `listTable` incrementally just for immediately looping over this array once and never using it again. Wouldn't it be easier to merge this and just directly create the `<td>` elements?

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +53,39 @@ export default class SelectColumn extends React.Component {
...
+    let idxColumn = [];
+    this.state.list.forEach(tool => {
+      this.selectable.forEach(headerRow => {
+        let length = tool.columns.length;
+        tool.columns.forEach(function(column, idx, i) {
+          if (headerRow.display_title === column.display_title) {
+            idxColumn.push({
+              display_title: column.display_title,
+              isVisible: column.isVisible,
+              key: idx * i.max_width
```
PhilippWendler: Wouldn't the index of the current column in `headerRow´ be an easier key?

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -18,6 +18,7 @@ export default class SelectColumn extends React.Component {
       list: [...this.props.tools]
     };
     this.selectable = [];
+    this.listTable = [];
```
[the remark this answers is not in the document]
marto97: Done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: Done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: Done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: Done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +52,32 @@ export default class SelectColumn extends React.Component {
...
+    const columns = this.state.list[index].columns;
+    this.selectable.forEach(function(headerRow, idxHeader) {
+      const found = columns.some(
+        el => el.display_title === headerRow.display_title
+      );
+      if (found) {
+        idxRow.push({
+          display_title: headerRow.display_title,
+          isVisible: columns.find(
+            el => el.display_title === headerRow.display_title
```
PhilippWendler: The `columns.some()` and `columns.find()` with the same callback can be simplified. Just use `find` to get the respective column and replace `if (found)` with `if (column !== undefined)`.

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,10 +52,32 @@ export default class SelectColumn extends React.Component {
   };
 
   renderColumns = index => {
-    return this.state.list[index].columns.map(column => (
+    const idxRow = [];
```
PhilippWendler: I still don't understand why the array `idxRow` is used. It is only filled and then immediately used for iterating over it. Why not get rid of it?

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: Done

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
marto97: `idxRow` is removed. I used `self.handleSelecion` instead of `this.handleSelecion` because it leads to an error. Now empty cells are empty `<td>` with keys.

marto97:
`idxRow` is removed. I used `self.handleSelecion` instead of `this.handleSelecion` because it leads to an error. Now empty cells are empty `<td>` with keys.

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -52,24 +52,35 @@ export default class SelectColumn extends React.Component {
...
-            id={index + "--" + column.display_title}
-            name={index + "--" + column.display_title}
-            type="checkbox"
-            checked={column.isVisible}
-            onChange={this.handleSelecion}
-          ></input>
-        </label>
-      </td>
-    ));
+    const self = this;
```
PhilippWendler: Is there a specific reason for having `self` instead of just using an arrow function that does not bind `this`?

marto97: It was leading to an error, that's why I used `self`, but using an arrow function seems to be a better solution.

marto97:
I think code is good now. If yes, can I push it in master?

PhilippWendler:
I did it myself now, because when the JavaScript code is changed, we always also need to recreate the bundled JS files. And in this case of a small change with some back and forth in the commits a squash merge is better.

Thanks for implementing this!

PR #549, opened 2020-02-09
Add integration test for SelectColumn.js of react-table

marto97:


marto97:
I created the file `SelectColumn.test.js`, which is similar to the `Summary.test.js`. But when I run all test with command `npm test`, it leads to:

> Test Suites: 4 failed, 2 passed, 6 total
> Tests:       25 passed, 25 total
> Snapshots:   2 passed, 2 total
> Time:        34.033s
> Ran all test suites.

and the error from all 4 failed is:  

> Test suite failed to run 
> Cannot find module 'imports-loader?this=>window!./zip.js' from 'index.js'

I am not sure, if this should be the desired result.

PhilippWendler:
> > Test suite failed to run
> > Cannot find module 'imports-loader?this=>window!./zip.js' from 'index.js'
> 
> I am not sure, if this should be the desired result.

Of course this is not the desired result. But [it works in the CI](https://gitlab.com/sosy-lab/software/benchexec/-/jobs/430732341) and on my machine, so it seems there is some issue with the setup on your machine that you would need to sort out.

The changes from this PR, however, do not work currently: [CI log](https://gitlab.com/sosy-lab/software/benchexec/-/jobs/430678448)

PhilippWendler:
Any updates?

marto97:
> Any updates?

Unfortunately for Summary.test.js, ReactTable.test.js and App.test.js I still have the same error:

> Test suite failed to run
> Cannot find module 'imports-loader?this=>window!./zip.js' from 'index.js'

Error on line 15:

> 15 | require("imports-loader?this=>window!./zip.js");

>   at Resolver.resolveModule (C:\UserData\z004055n\Documents\LMU../../../../../../../../../benchexec/benchexec/tablegenerator/react-table/node_modules/jest-runtime/node_modules/jest-resolve/build/index.js:259:17)
>       at Object.require (C:\UserData\z004055n\Documents\LMU../../../../../../../../../benchexec/benchexec/tablegenerator/react-table/src/vendor/zip.js/index.js:15:1)

I reinstalled everything, tried also with ` npm install imports-loader` but still the same issue on my machine. Do you have any advice how could I fix it?

PhilippWendler:
> > 15 | require("imports-loader?this=>window!./zip.js");
> 
> > at Resolver.resolveModule (C:\UserData\z004055n\Documents\LMU../../../../../../../../../benchexec/benchexec/tablegenerator/react-table/node_modules/jest-runtime/node_modules/jest-resolve/build/index.js:259:17)
> > at Object.require (C:\UserData\z004055n\Documents\LMU../../../../../../../../../benchexec/benchexec/tablegenerator/react-table/src/vendor/zip.js/index.js:15:1)
> 
> I reinstalled everything, tried also with ` npm install imports-loader` but still the same issue on my machine. Do you have any advice how could I fix it?

`imports-loader` is declared as dependency in `package.json` anyway. But it is not even necessary for these tests, because the file `src/vendor/zip.js/index.js`, which contains the reference to `imports-loader` must not be loaded in tests because loading it will always fail with just this error. We use a `transform` declaration for Jest to avoid that any file from `src/vendor/` is loaded at all in tests:
https://github.com/sosy-lab/benchexec/blob/52c20acb10c536bcc644cc6245efc69fba0298a2/benchexec/tablegenerator/react-table/package.json#L60

So I guess you need to find out why in your case the transformation is not applied and the file is loaded. Maybe it has to do something with Windows.

Apart from that, the same error as [here](https://github.com/reactjs/react-modal/issues/133) occurs now in the tests and needs to be fixed.

PhilippWendler:
There exists already an element with `id="root"` in the DOM in the full table, just not during tests. So adding this element to `SelectColumn` breaks the real table and cannot be done. Could you please find a different solution?

marto97:
> There exists already an element with `id="root"` in the DOM in the full table, just not during tests. So adding this element to `SelectColumn` breaks the real table and cannot be done. Could you please find a different solution?

I used  `  if (process.env.NODE_ENV !== "test") {
      ReactModal.setAppElement("#root");
    }` to fix the problem.

`SelectColumnt.test.js ` still has two issues to pass. First is  `   TypeError: this.props.tools is not iterable`:
```
      16 |     this.state = {
      17 |       deselect: true,
    > 18 |       list: [...this.props.tools]
         |                            ^
      19 |     };
      20 |     this.selectable = [];
      21 |   }
```
I could fix it using curly brackets  `{...this.props.tools}` and then with a function to convert it in an Array:

>  objToArray = obj => {
>     const result = Object.keys(obj).map(function(key) {
>       return obj[key]; 
>   }); 
>   this.setState({list: result});
>   }

But maybe there is a better solution to `iterable` problem.

The second error is `TypeError: parentInstance.children.indexOf is not a function`
```
      102 |           .create(<Overview data={data} />)
      103 |           .getInstance();
    > 104 |         const component = renderer.create(component_func(overview));
          |                                    ^
      105 | 
      106 |         expect(component).toMatchSnapshot();
      107 |       });
```

Which I did't find a way to fix it. Should tests be with Jest and snapshots?

PhilippWendler:
> > There exists already an element with `id="root"` in the DOM in the full table, just not during tests. So adding this element to `SelectColumn` breaks the real table and cannot be done. Could you please find a different solution?
> 
> I used ` if (process.env.NODE_ENV !== "test") { ReactModal.setAppElement("#root"); }` to fix the problem.

Is there a way to fix this without needing the check for test mode in production code? That would be nicer.

> `SelectColumnt.test.js ` still has two issues to pass. First is: ` TypeError: this.props.tools is not iterable`:

> The second error is `TypeError: parentInstance.children.indexOf is not a function`

I guess the test setup is not correct and certain fields are not probably initialized? Can you fix this?

marto97:
I created Snapshot Testing with Jest for `App.js` which includes also `SelectColumn.js`. Seems it works as it should be. If it's okay - this way I will implement this code to the main `App.test.js` in `src` folder.

PhilippWendler:
Testing `App.js` this way tests much more than just `SelectColumn`, which makes the tests difficult to understand and debug.

Can't you for example provide a `#root` element in the test instead?

marto97:
Now it is testing only SelectColumn.test.js. I will check it tomorrow again but I think everything is fine now.

PhilippWendler:
`SelectColumn.js` still contains the new condition, you added, is this necessary or just an oversight?

What does setting `ReactDOM.createPortal` in the test do? Such code needs to be documented.

Wouldn't it be easier and more obvious to just add a `<div id="#root">` element to the code that is rendered in the test?

marto97:
> `SelectColumn.js` still contains the new condition, you added, is this necessary or just an oversight?
> 
> What does setting `ReactDOM.createPortal` in the test do? Such code needs to be documented.
> 
> Wouldn't it be easier and more obvious to just add a `<div id="#root">` element to the code that is rendered in the test?

Without condition in `SelectColumn.js` and just adding  a `<div id="#root">` element to the code that is rendered in the test returns an error:

> react-modal: No elements were found for selector #root.

Another way to fix `#root` problem without condition is using `ReactModal.setAppElement(document.getElementById("#root"));` instead of `    ReactModal.setAppElement("#root");` and `ariaHideApp={false}` to avoid getting a warning. In this case I will not need `ReactModal.setAppElement(document.createElement("div"));` in `SelectColumn.test.js` as well.

`ReactDOM.createPortal` provide a way to render children into a DOM node that exists outside the hierarchy of the DOM component. Without `ReactDOM.createPortal` the test returns TypeError:

>   TypeError: parentInstance.children.indexOf is not a function

PhilippWendler:
Ok, thanks for the explanation. Last question: shouldn't `document.getElementById` be called with `root` instead of `#root` as parameter?

And did you test it in a browser that it still works there?

marto97:
Hi Philipp,

thanks for quick response. Yes, it should be`root` instead of `#root` as in [react-modal documentation](https://reactcommunity.org/react-modal/accessibility/).

I tested it in Safari and Chrome browser and it works without errors or new warnings in the console.

Issue #483, opened 2019-11-02
Hash routing for react table


PhilippWendler:
Our react table consists of several tabs, and it would be great if we could link directly to each of them.

For example, a link `table.html#table` should directly open the table tab, and `table.html#quantile-walltime` or something similar could directly open the tab with the quantile plot and configure it to show a walltime plot.

If this is done together with #126, we could use links like `table.html#diff` for directly showing only differences and avoid the need to statically generate diff tables in most cases.

PR #550, opened 2020-02-16
#483: implement hash routing

DennisSimon:
Hi @PhilippWendler 

Here is my PR for the implementation of hash routing as suggested in #483 . 

This PR removes the `react-tabs` dependency and implements `react-router` for routing. 

Added functionality: 

- Navigate to any page via hash-routing (e.g. `localhost/#/table`) 
- Every change to default configurations of graphs will be reflected in the URL using query parameters. 
This URL is also used to configure the graphs with the settings provided in the query parameters on load. 

Looking forward to  your feedback :)


DennisSimon:
Hi @PhilippWendler 
I rebased to master and incorporated your comments into the code. 
About the Hash Routing not working on Chromium:  I don't know too much about that sadly, it should work and it does work on Chrome. So it's likely something that is browser specific and is hopefully fixed soon.

PhilippWendler:
I noticed that when I open a plot, copy the URL and then open it in a new tab, not all settings that are in the URL are applied. It seems to affect at least the plot type (direct/quantile) and the correct/all results toggle (for both scatter and quantile plots), for reference here are some URLs for testing:
- http://127.0.0.1:3000/#/quantile?column=walltime&quantile=false&linear=true&correct=false
- http://127.0.0.1:3000/#/scatter?toolX=0&columnX=walltime&toolY=0&columnY=memUsage&line=5&linear=true&correct=false

Could you please check that all URL settings are applied (maybe I forgot to test some)?

It is ok for now if some settings are not stored in the URL (e.g., all the layout settings of the table tab), but those that are stored in the URL should be applied on load, otherwise this is confusing for experienced users.

DennisSimon:
> I noticed that when I open a plot, copy the URL and then open it in a new tab, not all settings that are in the URL are applied. It seems to affect at least the plot type (direct/quantile) and the correct/all results toggle (for both scatter and quantile plots), for reference here are some URLs for testing:
> 

They should be applied on load. 
I will take a look

PhilippWendler:
Now the defaults are wrong again?

Issue #532, opened 2019-11-29
Close button of log overlay should stay visible while scrolling


PhilippWendler:
In the HTML tables, there is an overlay that shows tool logs with an "X" as close button in the upper right corner. If the log is too large and the users scrolls (either horizontally or vertically), the close button is moved out of the view. Users then have to scroll back or use escape or click outside the overlay to close it.

It would be better if the button stays where it is while scrolling.

Currently, the button is marked as `position: absolute` since e28b2b74a9e157df172cfabe30bc148fafcf0565 because `position: fixed` causes other problems. `position: sticky` should be what we want, but so far I did not get it working for horizontal scrolling.

PR #551, opened 2020-02-16
532: Fix scroll behaviour in LinkOverlay

DennisSimon:
Hi @PhilippWendler ,

i stumbled over #532 when going through the issues and took a quick look. 

The easiest solution i came up with is moving the `overflow: scroll;` property to the `<pre>` block. 
This way, the close icon is not affected by any scroll movements at all.


PhilippWendler:
@marto97 Could you please take over #532? You might be able to use this PR as base. The open question is discussed in my review: https://github.com/sosy-lab/benchexec/pull/551#pullrequestreview-359537079

marto97:
@PhilippWendler How could I get a log which is too large, so I can see the problem when user scrolls?

PhilippWendler:
Whatever is most convenient for you, for example you could:
- reduce the size of browser window,
- add dummy text via browser developer tools (make sure to add enough text for horizontal+vertical scrolling),
- temporarily add some code that adds dummy text whenever a log is opened, or
- use our [demo table](https://sosy-lab.github.io/benchexec/example-table/svcomp-simple-cbmc-cpachecker.table.html) (the logs that appear for the right status column are the largest).

marto97:
It could fix it, by using `window.addEventListener("scroll", function);` in `SelectColumn.js`. This will change the position of "X" close button dynamically while scrolling. The problem is that "X" close button is in an `overlay`, and somehow doesn't register events in the `overlay`.

PhilippWendler:
@marto97 With 1c6a2f4ba6335835313c7cbf5486d08a0277a555 the `X` appears in the upper right corner of the whole window for me, not in the modal (Firefox and Chromium):
image
So it seems there is still something wrong?

marto97:
@PhilippWendler Now with 555436fbf748d34ef262c2966764730d938ea876 the `X` close button should be always in the modal.

PhilippWendler:
But now we are back to the situation from https://github.com/sosy-lab/benchexec/pull/551#pullrequestreview-359537079, right (overlap with scrollbar)?

marto97:
Now overlap with scrollbar is fixed. If it's necessary I could move `X` close button on the left side in the modal.

PhilippWendler:
> Now overlap with scrollbar is fixed.

No, for people with large scroll bars there will still be an overlap (actually, I see an overlap although my scroll bar is not particularly large), whereas for people with small scroll bars the X will be awkwardly far away from the border.

As explained in https://github.com/sosy-lab/benchexec/pull/551#pullrequestreview-359537079, we do not want to make any assumptions about how wide scroll bars of users are, because this varies depending on OS and browser. We want to position the X a constant amount of space away from the scroll bar (just like it is in current master) but with proper scrolling behavior. Is there a way to achieve this?

marto97:
Now `X` is positioned in a constant amount of space away from the scroll bar. 

Function `getScrollBarWidth()`checks the width of the scroll bar. And I used `componentDidMount()` with `setTimeout()`to wait for Modal script to be loaded first.

PhilippWendler:
That seems like a good idea!

However, for me the X is now always to the left:
image
(in both Firefox and Chromium)

Is there still a bug?

marto97:
Did you used b3b344bc8a589dd9af5e14beb06626b911ff7620 commit for all tests? I tried on MacOS: Chrome, Firefox, Safari and Chromium and on Windows 10: Firefox and Chrome (all browsers with latest stable versions) and the `X` close button is always on the right side of the modal.

Screenshot 2020-03-10 at 09 11 05

PhilippWendler:
Ah, now I see, you tested with the modal for selecting columns. I tested with the modal for showing log files (click on one of the cells in the "status" column on the "Table" tab).

Can you please copy over your solution for that modal as well? (Of course avoiding duplicate code).

And I am not sure which of the changes in this PR actually still need to be made. It might make sense to start from scratch on a new branch and change only what is needed, or revert the unneeded changes from this branch.

PhilippWendler:
Replaced by #555

PR #552, opened 2020-02-16
Better codestyle enforcement

DennisSimon:
Hi @PhilippWendler ,

this is a quality of life improvement as well as a better way of enforcing code style in the project. 

I have added a couple of things: 
- eslint and prettier configuration files
- `lint` script to check codestyle on all files in scope. This also now gets run in the gitlab linting step
- `lint:fix` script to add a way to let eslint fix all files in scope, this also runs prettier on all stylesheets and json files
- `husky` hook to automatically fix staged `.js`, `.json`, `.css` and `.scss` files and check for linting errors
- create `dependency.json` in gitlab to prevent eslint from throwing because of missing file.

The  commits b000fb0 and c83b539 include the results of a run of `lint:fix` as well as manual fixes of all remaining warnings and errors.

Looking forward to your feedback :)


.gitlab-ci.yml
```diff
@@ -43,37 +43,37 @@ stages:
 unit-tests:python-2.7:
   <<: *unit-tests
   variables:
-    PYTHON_VERSION: '2.7'
+    PYTHON_VERSION: "2.7"
```
PhilippWendler: Where do these formatting changes to the Yaml file come from? Is there now a linter for that? (Where?)

.gitlab-ci.yml
```diff
@@ -145,15 +143,15 @@ javascript-eslint:
   before_script:
     - cd benchexec/tablegenerator/react-table
     - npm install
+    - npm run dependencies
```
PhilippWendler: This should be run automatically if it is necessary, just like we do it for the other build jobs.

benchexec/tablegenerator/react-table/.eslintrc
```diff
@@ -0,0 +1,16 @@
+{
```
PhilippWendler: There is already an `eslintConfig` in `package.json`. We should not have two places where this is configured, so I suggest to move this to `package.json`.

benchexec/tablegenerator/react-table/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
PhilippWendler: The point of having an opinionated code formatter is to have no need for thinking about and discussing its configuration. In fact, a good opinionated code formatter would not even provide options at all (e.g., google-java-format). So I would prefer to keep the default config unless there are really strong arguments.

A line length of 160 is way too much anyway, line lengths should be kept to 80-100 characters. But the default line length is better than others.

benchexec/tablegenerator/react-table/package.json
```diff
@@ -19,9 +19,41 @@
...
+    "dependencies": "node scripts/dependencies.js",
     "start": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/start.js",
     "build": "node scripts/dependencies.js && node scripts/build.js",
-    "test": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/test.js"
+    "test": "test -f src/data/dependencies.json || node scripts/dependencies.js && node scripts/test.js",
+    "lint:fix:js": "prettier-eslint --write $(pwd)'/**/*.js'",
+    "lint:fix:css": "prettier --write $(pwd)'/**/*.css'",
+    "lint:fix:scss": "prettier --write $(pwd)'/**/*.scss'",
+    "lint:fix:json": "prettier --write $(pwd)'/**/*.json'",
+    "lint:fix": "npm run lint:fix:js && npm run lint:fix:css && npm run lint:fix:scss && npm run lint:fix:json",
```
PhilippWendler: I guess we could simplify this into a single rule? I see no need for ever calling prettier only on a subset of files.

(Same for rule `lint` as well, of course.)

benchexec/tablegenerator/react-table/package.json
```diff
@@ -19,9 +19,41 @@
...
+    "lint:scss": "prettier --check $(pwd)'/**/*.scss'",
+    "lint:json": "prettier --check $(pwd)'/**/*.json'",
+    "lint": "npm run lint:js && npm run lint:css && npm run lint:scss && npm run lint:json"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
```
PhilippWendler: This partially duplicates the above rules, why? Also, it seems inconsistent, e.g., it runs `git add` only for `*.js`, and it uses `eslint-friendly-formatter` that is not referenced anywhere else.

benchexec/tablegenerator/react-table/package.json
```diff
@@ -123,12 +157,14 @@
     "file-loader": "^4.2.0",
     "fs-extra": "^8.1.0",
     "html-webpack-plugin": "^4.0.0-beta.8",
+    "husky": "^4.2.3",
```
PhilippWendler: Isn't `husky` more a development tool that one would install locally on your machine (like an editor) instead of per project?

PhilippWendler (commented):
Thank you, improvements like this are very much appreciated.

I added a first round of questions, could you please address them and then do a rebase? Because of the line-length change there are a lot of changes in the remainder of the files, so it will be easier to review the rest afterwards.

Also, I wonder thy changes like this introduce changes in the bundled files?

benchexec/tablegenerator/react-table/package.json
DennisSimon: Not really as husky is a tool for git hooks and therefore must/should be present whenever we run git commands. 
The configuration for said hooks are defined in the ´package.json´ and might be husky version specific. 
It is therefore recommended to install husky as a dev dependency to enforce its presence and version.

benchexec/tablegenerator/react-table/package.json
DennisSimon: Totally agree, cleaned it up.

benchexec/tablegenerator/react-table/package.json
DennisSimon: I also cleaned this up a bit.
Note that this `lint-staged` part is not a script but a configuration for the `lint-staged` package.
`eslint-friendly-formatter` is a terminal friendly wrapper for eslint. 
When you try to verify code with eslint in the console, you will see that it spams the terminal. This tool gets rid of that and makes it easier to find mistakes.

benchexec/tablegenerator/react-table/.eslintrc
DennisSimon: Will do

.gitlab-ci.yml
DennisSimon: I think they automatically got applied when saving the file because the prettier plugin of code kicked in.

.gitlab-ci.yml
DennisSimon: I added a specific script to the package.json that does that

DennisSimon:
I addressed the review comments and rebased to master. 
I think the built files changed because i changed some variable names from snake_case to camelCase.

.gitlab-ci.yml
```diff
@@ -146,8 +146,7 @@ javascript-eslint:
     - cd benchexec/tablegenerator/react-table
     - npm install
   script:
-    - npx eslint src --max-warnings 0
```
DennisSimon: Accidentally removed this as well. 
Do you think we should move this to the script?
Will change after I finish work today

benchexec/tablegenerator/react-table/package.json
```diff
@@ -23,8 +23,23 @@
   },
   "scripts": {
     "start": "(test -f src/data/dependencies.json || node scripts/dependencies.js) && node scripts/start.js",
+    "dependencies": "node scripts/dependencies.js",
     "build": "node scripts/dependencies.js && node scripts/build.js",
-    "test": "(test -f src/data/dependencies.json || node scripts/dependencies.js) && node scripts/test.js"
+    "test": "(test -f src/data/dependencies.json || node scripts/dependencies.js) && node scripts/test.js",
+    "lint:fix": "prettier-eslint --write $(pwd)'/**/*.js' && prettier --write $(pwd)'/**/*.+(js|css|scss|json)'",
```
DennisSimon: Forgot to remove `js` from prettier

benchexec/tablegenerator/react-table/package.json
```diff
@@ -33,7 +48,25 @@
       "prettier/react"
     ],
     "rules": {
-      "prettier/prettier": "warn"
+      "prettier/prettier": "warn",
+      "no-console": "off",
+      "no-underscore-dangle": "off",
```
PhilippWendler: Why are some rules turned off here? We currently have no warnings, so it should not be necessary to disable any?

benchexec/tablegenerator/react-table/package.json
PhilippWendler: Hm, I am not yet convinced about the value of this. In the current config, it only runs a subset of our checks, so developers still need to remember to run other checks manually. But running more checks automatically would soon start to annoy developers if it takes too much time everytime they want to create a temporary commit etc. Furthermore, the JS project is only a part of the repo, and running JS checks on commits that do not touch the JS code is useless.

So I suggest we drop this and developers who like git hooks can of course just configure them locally on their own without much disadvantage.

benchexec/tablegenerator/react-table/src/App.js
```diff
@@ -10,11 +10,13 @@ import Overview from "./components/Overview";
 
 if (process.env.NODE_ENV !== "production") {
   // load example data for development
+  // eslint-disable-next-line global-require, import/no-unresolved
```
PhilippWendler: Why do we now need more disabling of warnings?

benchexec/tablegenerator/react-table/src/components/LinkOverlay.js
```diff
@@ -28,11 +28,10 @@ export default class LinkOverlay extends React.Component {
   //    (this fails for ZIPs on the local disk).
   // 3) Try loading url from within ZIP archive without Range header.
   loadContent = async url => {
-    console.log("load content", url);
```
PhilippWendler: The logging of this method has been quite useful in various circumstances, please do not remove it.

benchexec/tablegenerator/react-table/src/components/QuantilePlot.js
```diff
@@ -205,7 +212,7 @@ export default class QuantilePlot extends React.Component {
           const task = this.state.selection;
           const data = this[task + i];
           const id = getRunSetName(tool);
-          this.lineCount++;
+          this.lineCount += 1;
```
PhilippWendler: What is the purpose of changes like this?

benchexec/tablegenerator/react-table/src/components/QuantilePlot.js
```diff
@@ -291,11 +301,11 @@ export default class QuantilePlot extends React.Component {
     );
     const type = this.state.isValue ? this.possibleValues[index].type : null;
 
+    const interpolation = this.state.linear ? "linear" : "log";
```
PhilippWendler: This does not have anything to do with interpolation.

PhilippWendler (changes requested):
While a lot of these changes make sense, there is too much mixed up here. Lots of formatting changes mixed with renamings and actual code changes and commits that revert changes of previous commits in the same PR are not good. Furthermore, it is not possible to see which config changes led to which code changes, because the latter have been committed separately.

To make this PR mergeable there should be a clean sequence of commits, each of them changing one thing (like adding a new linting config) together with all the changes that result from it (like code formatting or rebuilding the bundled files). Code improvements like renamings that are done manually should not be mixed with other changes, but done separately before other changes.

benchexec/tablegenerator/react-table/src/components/QuantilePlot.js
DennisSimon: That change seems to have happened because of this eslint rule:
eslint.org/docs/rules/no-plusplus

benchexec/tablegenerator/react-table/package.json
DennisSimon: In the current configuration it would run `lint:fix` and then would add the linted files to the commit stage again. 
This would happen automatically for the extensions specified in the `lint:fix` script (namely css, scss, json, js). 
It would also be possible to configure the hook in way where it would only run on files in the `react-tables` subfolder. 
This was mainly an idea to make sure that the ci doesn't fail because of a missing trailing comma or a missing space after open brackets. 
I do however strongly agree on the fact that the JavaScript part of this project is not the main focus and therefore may have implications on people not working with this (sub)module. 
I will remove the hook, and clean up the whole branch.

DennisSimon:
Moved to #556.

PR #555, opened 2020-03-11
Fix close button in modal

marto97:
I tried to implement  `getScrollBarWidth()` in `Overview.js` and use the value of scroll bar width as `props` for `SelectColumn.js` and  `LinkOverlay.js` but it returns error, because function `getScrollBarWidth()` works only when modal is opened. 

So I implemented `getScrollBarWidth()` in both `SelectColumn.js` and  `LinkOverlay.js` and seems it works good.


benchexec/tablegenerator/react-table/src/components/LinkOverlay.js
```diff
@@ -148,6 +155,13 @@ export default class LinkOverlay extends React.Component {
     }
   };
 
+  getScrollBarWidth() {
+    const modalBox =
+      window.document.body.lastElementChild.lastElementChild.lastElementChild;
```
PhilippWendler: This looks error-prone and not very robust. Is there no better way to get the desired DOM element, e.g., via an id?

marto97: Now it gets desired DOM element via class name, because there is no id in the desired DOM element.

PR #557, opened 2020-03-16
Ultimate tool info module

skanav:


danieldietsch:
@PhilippWendler : The CI fail of flake8 points to code that was not changed by this PR. Do you know anything about that?

benchexec/tools/ultimate.py
```diff
@@ -76,11 +76,12 @@ def __init__(self):
     @functools.lru_cache()
     def executable(self):
         exe = util.find_executable("Ultimate.py")
-        for (dirpath, dirnames, filenames) in os.walk(exe):
-            if "Ultimate" in filenames and "plugins" in dirnames:
-                return exe
-            break
```
danieldietsch: The break here is important because it limits ``os.walk`` to the top level

benchexec/tools/ultimate.py
```diff
@@ -76,11 +76,12 @@ def __init__(self):
     @functools.lru_cache()
     def executable(self):
         exe = util.find_executable("Ultimate.py")
-        for (dirpath, dirnames, filenames) in os.walk(exe):
```
danieldietsch: I am not sure why this does not work for you in the first place. 

The ``None`` check seems necessary, but changing the parameter to ``os.path.dirname`` seems not.

danieldietsch (changes requested):
I would just add the ``None`` check , i.e., the ``if exe:`` part. 
The removal of ``break`` is wrong. 
The change to ``os.path.dirname`` seems not necessary.

benchexec/tools/ultimate.py
PhilippWendler: `exe` contains a path to a file, but `os.walk` needs a path to a directory, how can this work?

PhilippWendler:
> The CI fail of flake8 points to code that was not changed by this PR. Do you know anything about that?

Just ignore that, it is fixed in master already.

benchexec/tools/ultimate.py
skanav: os.walk takes a directory, not a file.  `exe` points to the file.

danieldietsch: I see, you are right. Good catch!

danieldietsch (approved):

PR #560, opened 2020-03-22
Fix #545: Grey-out legend entries in quantile plot, if data is hidden

### LINKED ISSUE #545: Grey-out legend entries in quantile plot, if data is hidden.
In our old tables, a legend entry in the quantile plot was hidden (via grey-out) after clicking on it. This feature seems to be missing in the newer tables. It made the legend more useful, if several entries are hidden.

marto97:


PhilippWendler:
This seems quite hacky and error-prone, both accessing the DOM elements and reading the current style. Please check whether there is a solution by properly adding the attributes directly to the items when rendering them.

marto97:
Now grey out entries in quantile plot works by adding the attributes directly to the items when rendering them. I used `disabled` attribute which is from `DiscreteColorLegend`.

PR #566, opened 2020-04-13
make table resize when zoom out in the table

### LINKED ISSUE #513: Table does not resize vertically on window-size change

marto97:
Fixes #513, replaces #517.


PhilippWendler:
Now my horizontal scroll bar is hidden behind the footer bar with the pagination, and if my browser window is vertically small, I have two vertical scroll bars (one for the whole window and one for the table part). Can you please look into this?

PR #567, opened 2020-04-15
Make links for file names when showing task-definition files

### LINKED ISSUE #402: clickable filenames if yaml-file is shown as overlay in html-table

PhilippWendler:
Fixes #402.


benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
```diff
@@ -0,0 +1,114 @@
...
+        contentPart.match(
+          "^" + this.state.fileTag + "(?:.)+" + this.state.fileTag + "$",
+        )
+      ) {
+        contentPart = contentPart.replace(
+          new RegExp(this.state.fileTag, "g"),
+          "",
+        );
+        jsxContentPart = (
+          <span
```
PhilippWendler, line 98: Is there a specific reason for not using `<a>` tags?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/TaskDefinitionViewer.test.js.snap
```diff
@@ -0,0 +1,684 @@
...
+
+exports[`Render TaskDefinitionViewer for input_files_brackets.yml 1`] = `
+<pre
+  className="link-overlay-text"
+>
+  <span>
+    format_version: "1.0"
+
+input_files:
+  [
```
PhilippWendler, line 51: Content like `input_files: [...]` is split across multiple lines, is it possible to avoid this?

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
```diff
@@ -0,0 +1,114 @@
...
+            onClick={() => this.props.loadNewFile(contentPart)}
+            className="link-overlay-file-link"
+            key={contentPart}
+          >
+            {contentPart}
+          </span>
+        );
+      } else {
+        jsxContentPart = <span key={contentPart}>{contentPart}</span>;
```
PhilippWendler, line 107: Wouldn't it be possible to skip the attribute-less `<span>`?
```suggestion
        jsxContentPart = contentPart;
```

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
```diff
@@ -0,0 +1,114 @@
...
+      this.state.fileTag +
+      this.state.splitterTag
+    );
+  };
+
+  render() {
+    const contentBySplitter = this.state.content.split(this.state.splitterTag);
+    const jsxContent = [];
+
+    contentBySplitter.forEach(contentPart => {
```
PhilippWendler, line 85: Instead of `forEach` with a mutable `jsxContent`, I suggest to simply use `map`. And `jsxContentPart` could then be replaced with directly returning the result.

benchexec/tablegenerator/react-table/src/components/LinkOverlay.js
```diff
@@ -7,35 +7,61 @@
...
-  componentDidMount() {
-    setTimeout(() => {
-      this.getScrollBarWidth();
-    }, 100);
+  isYAMLFile(filePath) {
+    return filePath.endsWith(".yml");
   }
 
+  loadNewFile = relativeURL => {
+    let newURL = path.join(this.props.link, "../" + relativeURL);
```
PhilippWendler, line 39:
```suggestion
    const newURL = path.join(this.props.link, "../" + relativeURL);
```

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
```diff
@@ -0,0 +1,114 @@
...
+      this.state.splitterTag +
+      this.state.fileTag +
+      fileName +
+      this.state.fileTag +
+      this.state.splitterTag
+    );
+  };
+
+  render() {
+    const contentBySplitter = this.state.content.split(this.state.splitterTag);
```
DennisSimon, line 82: I don't see the benefit of modifying the `yamlObj` in `prepareTextForRendering` and wrapping the filename in tags only to split them again here. 
Is there an argument against building an array of objects in `prepareTextForRendering` instead of modifying the mutable properties and then calling `toString` ?
I think we could build objects of this structure 
```
[
   {
      inputFiles: [...],
      propertyFiles: [...],
   },...
]
```
and then iterate through both arrays in here. 
Or is there something that i am missing?

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
DennisSimon: I think the issue here is that we will get a warning if we render a list of items without giving each item a key.

PhilippWendler: Even if those items without a key are just plain strings?

DennisSimon: Hmmm good point. 
I rarely render a list of plain strings in react and just assumed that there is some internal wrapping into JSX objects going on. 
I tested it in a code-sandbox and can verify that it does not care about plain strings not having keys. 
Disregard my previous comment.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
lachnerm: There was during the development but it doesn't make a difference at this point. `<a>` tags are more suitable here though, so I will change it.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
lachnerm: The point of using the `<span>` tag here was indeed solely for giving a key to the items. I've also just tested changing them to plain strings and didn't notice any unwanted behavior or warnings, so there should be no problem removing those tags.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
lachnerm: The reason I implemented it like this was simplicity. It should be possible to build an array like you suggested, however the library that is used stores the yaml file in a pretty complex object. To keep the formation (like spaces or empty lines, brackets vs. new lines, multiline strings etc.) of the yaml file, there will be a lot of code necessary which the `toString()` method automatically does for us here.

Also the current code works with yaml files that aren't of the standard for the task definition files by just ignoring everything except for the important parts. If we rebuild the yaml file manually, we will have to take all possibilities into consideration to also rebuild the (in this case) not important parts correctly.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/TaskDefinitionViewer.test.js.snap
lachnerm: Hm, this seems to be related to the line length of the list. When adding the splitter tags around the input files, the line exceeds a certain amount which results in the parser splitting it across multiple lines like in your example above.

This occurs every time the amount is exceeded, so even without the splitter tags there will be the same behavior if the names of the input files are too long or there are too many in a single line.

Unfortunately I couldn't find a way to disable this behavior in the lib that is used. I guess there probably won't be a way to avoid this without a very hacky workaround.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
lachnerm: I just noticed that this actually does bring a problem with it.
ESLint shows the following warning:

> The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid

 In this case a `href` attribute is not what we want though. Therefore I would suggest keeping the `<span>` tags.

PhilippWendler: I see. But this raises an interesting other question: Is the current solution accessible, i.e., is it usable with a keyboard (using <kbd>Tab</kbd> to focus the "link" and <kbd>Enter</kbd> to activate it)? If no, we should fix that. At least we should probably provide an ARIA role.

lachnerm: No, the current solution is not accessible via the keyboard.
How about simply using `<button>` tags for the links (and also the "Back to task definition" link)? The button style can easily be overriden by a few css attributes. Of course we could also add ARIA markup, but if we can use a native html element that already works I wouldn't see an argument against using it.

PhilippWendler: Yes, that seems to make sense.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
DennisSimon: I see. That makes sense. 
One more question: Is the splitter tag needed? Wouldn't it work with the file tag alone?

lachnerm: Hm, the reason why I used both tags is that the splitter tag is removed from the string by the `split()` method. The file tag is then used to identify the elements of the array that are supposed to be linked.
It would be possible to use only one tag and leave them in the array by using regex with the `split()` method, but this would result in an array like this:
`[..., "input_files: ", "<fileTag>", "inputFileName", "<fileTag>", ...]`
So then additional code will be necessary to only link the strings with preceding and following `<fileTag>` elements in the array and not to render the `<fileTag>` strings. I don't think this will have much benefit for the code quality.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.js
lachnerm: The links could easily be changed to `<button>` tags, however there were some problems with the style of the "Back to task definition" button due to the standard vertical alignment of the `<button>` tag. It seems that the only way to affect the vertical alignment of buttons is with hacky workarounds, so I instead made it behave like a button with some markup.

Also the focus is changed to the body element of the page when new content is loaded into the overlay, so Tab had to be pressed two times to access the button. To avoid this, I added some code that focuses the modal container when updating.

Any way, all links are now accessible via keyboard.

PhilippWendler: Thanks! Does it maybe work if setting `display: inline`? But if not, we can just keep it as it is.

lachnerm: No, unfortunately altering the `display` value doesn't work.

lachnerm:
> 👍
> 
> Only required thing left seems to be the tests with HTML content in the YAML file.

Sorry, I missed that part in your comment above. I have now added such a test.

PR #571, opened 2020-04-23
Fix position of close button in select columns view

marto97:
Hi @PhilippWendler 

I copied the solution for X close button from `LinkOverlay.js`.

I updated dependencies with `npm update` and `npx browserslist@latest --update-db`  because when I run `npm run build` I get error: `Browserslist: caniuse-lite is outdated. Please run next command npm update`. Now it seems everything is okay.


PhilippWendler:
The message about `Browserslist` is just a warning, you can safely ignore it. And anyway we would need to do it in a separate commit.

Could you please provide a commit with only the desired change (which I think looks good)? Probably the easiest way is `git reset --hard origin/master`, redo the change, and `git push --force`.

marto97:
Okay I will push a separate commit with only the desired change.

marto97:
@PhilippWendler last commit have `package-lock.json` and  `package.json` reverted as in `master` branch.

PhilippWendler:
But the bundles have not changed, so they still contain the updated libraries.

marto97:
Oh sorry, I will fix it now.

PR #575, opened 2020-05-02
added tool infos for cond test tools.

skanav:
Added tool info modules for the conditional testing tools.


skanav:
not ready for review yet.

benchexec/tools/condtesttools.py
```diff
@@ -0,0 +1,46 @@
...
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        return (
+            [executable]
+            + options
+            + ["--spec"]
+            + [propertyfile or "None"]
+            + [f for f in tasks]
+        )
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
```
PhilippWendler, line 45: Overriding this method is not necessary because its default implementation is the same.

benchexec/tools/condtesttools.py
```diff
@@ -0,0 +1,46 @@
...
+limitations under the License.
+"""
+import benchexec.tools.template
+import benchexec.result as result
+import benchexec.util as util
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    An abstract class for the conditonal testing tools.
```
PhilippWendler, line 27: Please add the URL here as well, because people will not now what "the conditional testing tools" are (and btw., these are "conditional-testing tools").

benchexec/tools/test-criterion-instrumentor.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    This tool instruments the test criteria in the C file, such that a verifier
+    can be used on it.
+    url: https://gitlab.com/sosy-lab/software/conditional-testing
+    """
+
+    _exec_path = "bin/instrumenter/instrumenter"
+
+    def name(self):
+        return "condtest-instrumenter"
```
PhilippWendler, line 33: This can be a human readable name, e.g. "CondTest Instrumenter".

benchexec/tools/condtesttools.py
skanav: done

PR #576, opened 2020-05-13
Make it easier to select link-text in tables

lembergerth:
When looking at benchexec tables, I often want to copy a task's name, e.g.
to grep for it in the run logs or access the result files of that task.
At the moment, this is very difficult, because the task names are links of type "block"
that take up the full cell. Firefox does not allow me to click+select to select them.

This makes it impossible to select the text in those links
if there is no space around the links to start clicking (and there isn't).

This small PR makes it possible to copy text from table cells that contain links
by changing the display style of links in tables from 'block' to 'inline-block'.
This way, the links do not take up the full cell and it is
possible to select the text with click+select.
Also add a small margin to the links so that there is always
a little space to start the click.

Downside: Not the full cell is a link anymore,
but only a reasonable space around the actual link text.


PhilippWendler:
Hm, I can understand the reason, of course, but this makes the UI of the tables more inconsistent, because other cells are also fully clickable.

I actually do manage to select the task name in the current tables because there are a few pixels on each side of the cell (the padding of the `div` around the `a` tag) where you can start selecting.

You can also get the task name from right-clicking and choosing "Copy link address". And in many cases you can copy relevant parts of the task name from the log output.

Is this sufficient for you?

If not, would there be alternative solutions for getting the task name into the clipboard?

PhilippWendler:
If you need this often, you could also install this [browser extension](https://addons.mozilla.org/de/firefox/addon/copy-link-text-webextension/). That would be even more convenient because it avoids the need to manually select anything (which is always more tedious than just clicking).

lembergerth:
> I actually do manage to select the task name

Ah ok, thanks for that tip! If I start at the very left of the browser window and make sure not to leave the browser window when clicking, I can also manage. It sometimes takes multiple tries, though, and I doubt that the common user will get that idea (I didn't).

> You can also get the task name from right-clicking and choosing "Copy link address"

That works like a charm! I'll just use that then, thanks. For the table cells, this behavior seems to be the same as described by the extension you recommend.

I would argue that not being able to just click+select the text is more of a usability issue than the inconsistency to the clickable cell-headers this PR introduces. But since I'm very happy with the right-click solution, feel free to close that PR.

Thanks again!

Issue #481, opened 2019-11-02
Multi-select column filter for status columns in react table


PhilippWendler:
The filter feature for status columns in the react table currently allows to select exactly one status or category value to be used as filter. We could extend this to allow selecting several different values, for example, in order to allow showing lines with certain errors, but without showing all errors.

The UI could look like the file filter of GitHub's PRs:
![Screenshot_2019-11-02 sosy-lab benchexec](https://user-images.githubusercontent.com/2545335/68068659-5ade2200-fd57-11e9-8b01-46340c51c45f.png)

kfriedberger:
I would like to have this feature again, as in the older tables. It is quite helpful when filtering for related statuses like `TIMEOUT`/`TIMEOUT(exception)`/`TIMEOUT(assertion)`.

PhilippWendler:
Of course. For now I suggest to filter on category and then sort by status.

PhilippWendler:
A first improvement would be to have negated filters (e.g., `not correct` as category). This would be easy to implement if we find a UI solution that works with the current drop-down field and is still understandable. Any suggestions?

lembergerth:
How about clicking once on an item selects it, and clicking twice negates
it? A third click could disable it again.
The current selection could be signaled by showing a filled or checked
checkbox for the first, and a minus-sign in the checkbox for the second.

PhilippWendler:
This is not possible in a standard drop-down field I guess.

PR #581, opened 2020-06-05
New filter UI and functionality

DennisSimon:
Hi @PhilippWendler , 

finally i can open the PR for the new filter UI. 

The major changes in this PR are as follows: 

* A filter slide-out that can be opened by clicking the filter summary in the top right corner. This UI is available from everywhere in the application
* As filtering up until now was done solely in the `ReactTable` component, i added custom filter functions that can filter the dataset without needing to communicate with the `ReactTable` component. The functions added are quite performant and could be used to implement filtering in a future `ReactTable` upgrade to `v7`.
* While trying to squeeze out some more performance i also refactored the `ReactTable` component to use hooks. This refactor could be used for the transition to `v7` as well. 
* Using the new UI filters now support the selection of multiple status and categorical values. multiselect
* Selecting numerical values can now be done using a easy-to-use range slider, thus improving usability range_slider

 

The filter state is shared between the new UI and the UI of the Table component.


benchexec/tablegenerator/react-table/src/App.js
```diff
@@ -6,6 +6,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import "./App.scss";
+import 'rc-slider/assets/index.css';
```
PhilippWendler, line 9: Isn't it enough to import this in the respective component?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -194,6 +423,27 @@ const setParam = (param) => {
 
 const stringAsBoolean = (str) => str === "true";
 
+const deepEquals = (a, b) => {
```
PhilippWendler, line 426: We already include the [deep-equal](https://github.com/substack/node-deep-equal) library, can we reuse it here?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -7,6 +7,83 @@
 
 import React from "react";
 
+const getFilterableData = ({ tools, rows }) => {
```
PhilippWendler, line 10: I guess due to the amount of code it is now the time to move all filter-specific utils to a separate file like `utils/filters.js`. General methods like `omit` can stay here.

Could you also please add a short (JavaDoc-style) comment on each utility method that you add?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterBox.js
```diff
@@ -0,0 +1,110 @@
+import React from "react";
```
PhilippWendler, line 1: Several files miss the license header.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterBox.js
```diff
@@ -0,0 +1,110 @@
...
+    newFilters[toolIdx] = newFilters[toolIdx] || [];
+    newFilters[toolIdx][columnIdx] = data;
+    this.setState({ filters: newFilters });
+    this.sendFilters(newFilters);
+  }
+
+  render() {
+    return (
+      <div
+        className={`filterBox ${this.props.visible ? "" : "filterBox--hidden"}`}
```
PhilippWendler, line 84: You can use [classnames](https://github.com/JedWatson/classnames) to simplify this.

PhilippWendler:
@kfriedberger @lembergerth Because I know you like #481 you can have a look at this PR if you are interested. Just checkout `new_filters` and run `bin/table-generator`. Note that this is still a first draft and we have some known open issues as listed in my review, but more feedback is of course always welcome.

kfriedberger:
I like the new feature a lot and it should really be added. :-)

Comments on the feature:
- the new feature is quite hidden: You need to click on the "Showing N of M tasks"-button. Ok, once you know it, it is simple to use.
- floats are shown very precise, with 15 digits after the comma. This breaks the layout, the fields go beyond the right border of my screen.
- adding a new filter requires 2 click: first select the name of the column, then click on the checkmark. Can this be reduced to one click?
- how can I remove a filter?
- when all columns are used as filters, there is an empty dropdown for new filters. Can we avoid this?
- inserting a value that is out of range (on the other side of min/max) for cputime or memory seems to lock the field. This is bad for entering a walltime of "0.123", because inserting the initial "0" already blocks.
- when filtering for status, I am especially bad at aiming for the small clickable rectangle in front of the status. I would like to click on the full name of the status.

Bug:
I have two runsets with each having columns for `status` and `cputime` in one large table. Adding a filter for the `first cputime` (without selecting any value!) and then adding a filter for the `second status` works, but as soon as I select any status, it removes the `first cputime` filter and sets the `status` filter for the first runset instead of the second runset.
The appended sceenshots differ in just one click on status `error`.

before click | after click
:-------------------------:|:-------------------------:
 |

kfriedberger:
Maybe some of my comments overlap with Philipp review, which I did not read before. :-)

lembergerth:
Really nice, thank you!

It also took me a while to find the filter menu - thanks @kfriedberger for the hint.

Adding to the existing comments:

The filter-menu is quite wide and obstructs the view of all columns on the right - on a table with two runsets and default columns shown, the 5 most-right columns are obstructed:

s0

vs.

s1

This is only a slight issue, I just wanted to note that this obstructs some information that people *may* want to see for constructing their filters.

Possible ideas: (a) reduce the width of the table by the width of the filter menu, so that the horizontal scroll-bar is fully accessible and can be used to scroll to the columns of importance. or (b) make the filter-menu slightly transparent - may only benefit people with good vision, though.


Additional feature request for the far future: Allow filters that define relations between columns. E.g.: I want to filter all rows for which cputime A is less than cputime B.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.js
```diff
@@ -0,0 +1,307 @@
...
+    const { filter, editable, availableFilters } = this.props;
+    console.log("FilterCard rendering", { filter });
+    const filterAddSelection = () => (
+      <>
+        <select
+          class="filter-selection"
+          onChange={(e) => this.setState({ idx: e.target.value, active: true })}
+        >
+          {availableFilters.map(({ idx, display_title }) => (
+            <option value={idx}>{display_title}</option>
```
lachnerm, line 95:
```suggestion
            <option key={idx} value={idx}>{display_title}</option>
```

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.js
```diff
@@ -0,0 +1,307 @@
...
+    };
+  }
+
+  render() {
+    const { filter, editable, availableFilters } = this.props;
+    console.log("FilterCard rendering", { filter });
+    const filterAddSelection = () => (
+      <>
+        <select
+          class="filter-selection"
```
lachnerm, line 91:
```suggestion
          className="filter-selection"
```

benchexec/tablegenerator/react-table/src/tests/__snapshots__/TaskDefinitionViewer.test.js.snap
```diff
@@ -2,756 +2,611 @@
 
 exports[`Render TaskDefinitionViewer for empty_file.yml 1`] = `
 <pre
-  className="link-overlay-text"
```
DennisSimon, line 5: These changes come from the added json setup declarations in the `package.json`.
These extend the jest snapshot serializer before tests were run. Before this change, the extension of `expect` only happened on import of `tests/utils.js::test_snapshot_of`.

DennisSimon: All snapshot changes in commit ae5fd37 can be retraced to this change. 
(Apart from the changes in `FilterCard.test.js.snapshot`, as i forgot to update snapshots after fixing sliders and removing the checkmark

DennisSimon:
Hello @PhilippWendler @kfriedberger @lembergerth @lachnerm , 

first of all: Thanks for your thorough review of the usability and functionality of the feature. 
From your comments i have compiled a list of actions of which i have completed most of. 
For some actions i still would like to request your opinions. 

- [x] Remove the extra step needed to add filters
- [x] Don't remove filters automatically when values have been set/reset to initial value
- [x] Fix validation of numerical input fields on filters containing the sliider element
- [x] Move css import of slider element to respective component
- [x] Add missing license headers
- [x] Fix filter logic to disjunct multiple selected values in the same Filter card and conjunct all Filter cards of a runset. 
- [x] Add an icon button to manually remove filters

My open questions would be about the following
- [ ] Add extra button to reset filters (
**Question:** where should this button be best located to not disrupt the users workflow?)

- [ ] Filter sidebar is overlapping and thus hiding parts of the table: (
**Question:** If we make the sidebar smaller, we will still have some columns that will be overlapped. Making the filter sidebar semi-transparent will likely result in improving complexity of the ui as we will be mixing different elements. This would likely result in the user having a harder time to concentrate on following the filter flow. 
**Suggestion:** we could make the sidebar behave more "solid" and let it push the right side of the screen towards the left. This way we will always see the whole table.) 

- [ ] Make it easier to understand on how to access the filter menu. (
**Question:** We could either create a new button and place it next to the filter info or change the label of the filter info button to something like `"Showing X of Y (Open filter menu)"`

- [ ] Round digits to a more manageable number. (
**Question:**A lot of you have experienced very precise numbers. Should we introduce a hard limit on the precision? This would result in some super small numbers not being filterable via the slider)`

I'm looking forward to your feedback and your suggestions! :)

lembergerth:
Hi @DennisSimon ,

>  we could make the sidebar behave more "solid" and let it push the right side of the screen towards the left. This way we will always see the whole table.)

I agree. That's what I described with my suggestion (a) :-)

DennisSimon:
@PhilippWendler 
I have included all the changes that we talked about the last sessions. This PR is ready for another review.

DennisSimon:
@PhilippWendler This should include all open points. 
The PR should be ready to go through a usability test again

PhilippWendler:
Thanks!

I added code that adds the names of the individual parts of the task id, they are now available as `task_id_names` list in the `head` part of the data (`tableHeader` in `Overview`). But it was not immediately clear to me what to return from `getTaskIdParts`, could you please adjust this?

PhilippWendler:
Looks good! Just found one bug: If I select all status values in a filter and exactly one category value, the filter drop down in the table will show one status value (`ERROR`) instead of the category value. And if I select all category values and exactly one status value, the drop down shows one category value (`correct`).

PhilippWendler:
And could you please make the vertical scrolling of the filter overlay only apply to its lower part, such that the header is always visible?

Applying `top: 0px; position: sticky` to `filterBox--header` is similar to what I want, but a better solution would be to wrap all the lower parts in a new `<div>` with `overflow-y: scroll` such that the position of the scroll bar indicates what is scrollable.

PhilippWendler:
@kfriedberger @lembergerth If you would like to have a look at this again, it is now almost ready and close to being merged. Only the overly precise numbers and [a small visual bug](https://github.com/sosy-lab/benchexec/pull/581#issuecomment-664781732) remain. So any feedback would we welcome.

lembergerth:
Looks really good!

I like both the filter icon and the immediate tooltip.
I think people can easily find the functionality this way.
And I'm especially happy about the functionality to filter by expected verdict, I was looking for this functionality for a paper last week 😁 

Does it make sense to replace the current textfields for `Property` and `expected verdict` by a dropdown with the possibilities?

When dragging the value-sliders, it is really easy to select the surrounding text. You just have to drag slightly above or below the slider rail. (happens in Firefox, Windows)
This behavior can probably be annoying, but I'm not sure whether it's easy to fix.

grafik

These small suggestions aside, I think it's great to see that we get such a comprehensive and well-working filter function in benchexec, thanks a lot! :-)

PhilippWendler:
Thanks for your feedback!

> And I'm especially happy about the functionality to filter by expected verdict, I was looking for this functionality for a paper last week :grin:

Note that you can already do this with the filter in the left-most column. It is just more visible now, and now you can have specific filters for each of task name, property, and expected verdict.

> Does it make sense to replace the current textfields for `Property` and `expected verdict` by a dropdown with the possibilities?

Maybe even checkboxes. But I think we can iterate this later.

> When dragging the value-sliders, it is really easy to select the surrounding text. You just have to drag slightly above or below the slider rail. (happens in Firefox, Windows)
> This behavior can probably be annoying, but I'm not sure whether it's easy to fix.

@DennisSimon?

DennisSimon:
> Maybe even checkboxes. But I think we can iterate this later.

Yes, this is definitely possible. This was actually one of my earlier approaches after which i reverted in favor of free text search.
Implementing this should be not too much work but, as you suggested, i also believe we should d this in a future iteration.

> > When dragging the value-sliders, it is really easy to select the surrounding text. You just have to drag slightly above or below the slider rail. (happens in Firefox, Windows)
> > This behavior can probably be annoying, but I'm not sure whether it's easy to fix.
> 
> @DennisSimon?

Right, i have noticed that behaviour as well. I have tried to solve this by stopping propagation of click events to parents of the slider. I sadly wasn't successful.  After reading the comment i did some more research and found another possible solution involving css: 
```
    -webkit-touch-callout: none;
    -webkit-user-select: none;
    -khtml-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
```
Early tests look promising. 

I will work on the issues mentioned here this evening and hope that i can push fixes for them today.

DennisSimon:
@PhilippWendler , @kfriedberger , @lembergerth and everyone else that was involved: 
Thank you very much for the thorough reviews and great feedback! It really has been of great help. :smile:

DennisSimon:
@PhilippWendler The input fields should now have a more sensible step size and the issue with the unwanted selection of text should be gone. Could you verify this on your end?

DennisSimon:
@PhilippWendler The latest commit should also have resolved the visual bug in the drop down fields of the ReactTable component that you have reported.

DennisSimon:
@PhilippWendler Lastly, i have made the filters for task id sticky and made the other filter cards scrollable. 
Currently a scrollbar is visible inside the filter box, i have left some commented-out css rules in case this is unwanted behaviour and i'm unavailable tomorrow.

PhilippWendler:
Thanks! I did not notice an improvement regarding the numbers in the numeric filters, though?

And a few final minor nitpicks:
- `filter-card--container` should get some `margin-bottom`, otherwise the lowest "Add filter" box is too near to the bottom of the window.
- I would slightly prefer the task filter to be part of `filter-card--container` and scroll out of view. This leaves more space when scrolling, and it actually looks a little bit weird if the other filter cards are scrolled up towards the task filter.
- The name of the task filter should be "task filter" (with appropriate case, of course). In its box it is still named "Task name filter", and in the text that appears in the filter cell of the table it is still called "task id filter".

DennisSimon:
@PhilippWendler The rounding issue should now be fixed. I managed to use the rounding helper that i created for the benchexec frontend. 
I noticed one thing that might be related: 
Currently the data sets that get created by benchexec do not include the `number_of_significant_digits` fields anymore. 
Without this information we are unable to round to a correct value. I needed to manually add the fields back for my tests.
Could you verify this?

PhilippWendler:
When I run `npm run test`, I get [lots of console messages](https://gitlab.com/sosy-lab/software/benchexec/-/jobs/665396241), and one error about missing unique keys:
```                                                                                                                         
    console.error node_modules/react/cjs/react.development.js:315                                                                  
      Warning: Each child in a list should have a unique "key" prop.                                                               
                                                                                                                                   
      Check the render method of `FilterCard`. See https://fb.me/react-warning-keys for more information.                          
          in option (at FilterCard.js:134)                                                                                         
          in FilterCard (at FilterCard.test.js:14)                                                                                                                                                    
```
Could you please check this?

DennisSimon:
Good Morning @PhilippWendler , 

i have removed the warnings from the test and rewritten the number formatter. 
Could you verify that they are now working as expected?

PhilippWendler:
This seems to work well, except for one bug: I have a table where the lowest CPU-time values are 1.85, 1.87, and 1.95. If I add a CPU-time filter, it shows this:
grafik
However, when I hover over the left-most end of the slider, the tooltip shows the value 1.9. And when I changed the right end of the slider, the left end adjusts itself automatically to 1.9:
grafik
This also actually changes the filter as shown in the filter cell (content is `1.9:47.4`), and the two rows with values below 1.9 are indeed hidden.

The status and category filters are currently broken. After adding such a filter and clicking on one of the checkboxes, none of the checkboxes are selected. Clicking on a checkbox again results in a filter that hides everything.

The tests still show some debug messages, can they be removed or hidden? Otherwise this clutters the output and hides the status messages from the test runner.

PhilippWendler:
Everything seems fixed now, thanks!

Unfortunately, I just noticed a related potential confusing behavior. In my table, the lowest value (shown as `1.85`) is actually something like `1.849`. Adding a filter like `1.85:50` hides such a value. (This is already the existing behavior.) We could change this, but I think it is more effort if we do not want to round each value during each filter run, and it is not a big problem. With the new UI, it becomes misleading, though: If I create a numeric filter and change only the right end of the slider, I expect all small values to be shown in the table. However, the actual filter always has both a minimum and maximum (like `1.85:50`) and thus can hide some table entries from the other end. This also occurs if I manually add a filter like `:50` and then change only the right end of the slider. This can lead people to draw wrong conclusions from a table, thus I would like to fix this before merging it.

Is it possible for the slider to create filters with half-open intervals like `2:` and `:50` if one end is still at its minimum/maximum?

DennisSimon:
@PhilippWendler I have removed unnecessary console.log statements and wrapped helpful ones in environment checks so they don't spam the console.

DennisSimon:
@PhilippWendler I have implemented the suggested behavior. Could you verify that it works as expected?

PR #582, opened 2020-06-15
565 add column selection of tables to url state

### LINKED ISSUE #565: Add column selection of HTML tables to URL state
The HTML tables have state that defines which columns are currently shown. Sometimes it is nice to be able to link to tables with only a subset of columns shown (for example, when linking to regression tables in a bug report irrelevant columns can be hidden from view).

To enable this, we need to add the definition of visible columns to the state that is managed as part of the URL from #550.

lachnerm:
When selecting columns to be hidden in the table tab, the hidden columns 
are now stored in the URL as parameters, which enables sending links 
with predefined subsets of the columns that will be shown. Accordingly 
the table checks the URL for the parameter on opening/updating and hides 
the defined columns.  If there is no parameter defined, but the data 
object stores columns that are supposed to be hidden, the table 
automatically creates the URL parameter with the hidden columns as 
values. Fixes #565.


benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -97,6 +100,50 @@ export default class Table extends React.Component {
...
+    window.removeEventListener("popstate", this.updateTools, false);
+  }
+
+  updateTools = () => {
+    this.setState((prevState) => ({
+      tools: this.adjustToUrlParams(prevState.tools),
+    }));
+  };
+
+  adjustToUrlParams(tools) {
```
PhilippWendler, line 118: Because column visibility influences all tabs, this should maybe be in a different file?

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -173,6 +173,7 @@ export default class Overview extends React.Component {
                     filtered={this.state.filtered}
                     toggleLinkOverlay={this.toggleLinkOverlay}
                     changeTab={this.changeTab}
+                    hasAccessToUrlParams={!this.state.showSelectColumns}
```
PhilippWendler, line 176: I am not sure why this is necessary? In general, more comments would be good, in particular for cases like this one.

lachnerm:
> The `display_title` of columns is not unique, so this currently breaks if there are more than one column with the same name.
>
Is it safe to assume then that the `title` attribute is unique? Or is there any other attribute that identifies the columns?
> 
> The hiding of columns has influence on the plot tabs, but when I change tabs the parameter `hidden` is removed from the URL. I think it should always stay in the URL (except if it is empty).
> 
I thought the parameters were only supposed to be added for the Table tab, but in this case I will change the current implementation to work for all parts of the table.
> Currently there are two ways how `SelectColumns` influences the visibility of the columns: one way is the old one, where it directly sets the boolean flag of the column objects, and one way is the new one via `setParam`. Because of the hidden side effects, the former is not nice anyway, and now that it is no longer necessary I think it should be removed.
>
Alright, I will do so.

benchexec/tablegenerator/react-table/src/components/Overview.js
lachnerm: This was necessary due to the `SelectColumn` and `ReactTable` components both having a popstate event-listener while operating on the same page. This resulted in unwanted side effects, so this flag was for deactivating the popstate event in the `ReactTable` component while the `SelectColumn` is open. However this solution will not be necessary any more after removing the `isVisible` flag of the columns object any way. I will check where to add commentary for the following changes though.

lachnerm:
And one more question: Is it necessary to differentiate between the columns of each tool? So is it supposed to be possible for a column to be hidden in one tool while being shown in another? This would add more complexity to the feature though and may result in some large URLs.

PhilippWendler:
> > The `display_title` of columns is not unique, so this currently breaks if there are more than one column with the same name.
> 
> Is it safe to assume then that the `title` attribute is unique? Or is there any other attribute that identifies the columns?

AFAIR there is no guaranteed unique identifier for each column, except its index. A combination of run set (tool) name and column name is unique in practice, but rather long. But you could have a look at how other components in the table identify columns.

> And one more question: Is it necessary to differentiate between the columns of each tool? So is it supposed to be possible for a column to be hidden in one tool while being shown in another? This would add more complexity to the feature though and may result in some large URLs.

Yes, absolutely. The existing UI supports this.

lachnerm:
I've found a working solution for the distinction of columns specifically being hidden for individual runsets as well was for all runsets (and all combinations of those of course). However, it's still not 100% clear to me how to handle columns with the same display name.

Anway it seems that currently the `SelectColumn` Modal does not support differentiating between columns with the same `display_title` property. I've manually added such a column to the test data and got the following result (from the current master branch):

example

I've also thoroughly checked the `SelectColumn` class itself as there were a lot of modifications necessary for the URL solution - however it does only seem to care about the `display_title` property currently.
If there is nothing against it, I would also modify the component to be able to differentiate between columns with the same `display_title`. It's necessary to have the URL solution working and I've already made a lot of changes to the file anyway.

Now since there is no unique identifier for the column so far, I've resorted to using the index of the column to be able to differentiate between columns with the same title (and to make the URLs a little bit shorter). This works fine within one runset, however there is a small problem when working with multiple runsets. Assuming the `SelectColumn` component can correctly display multiple same-titled columns, there could happen something like this:

example2

We have two columns with the same title `cputime` in the first runset while the second runset only contains one column called `cputime`. Is there now still any way to be sure which of the two `cputime`s the column of the second runset should be assigned to? I don't really know what the practical examples of same-titled columns look like, so I'm not sure what's the best way to handle cases like these.

PhilippWendler:
There are no practical examples of same-titled columns :-)

Rethinking about it, I guess we can avoid adding extra handling for this, though of course if our new solution naturally supports it this is also fine.

Using the index as column identifier is probably the best way. It is not as nice if users want to manually edit the URL, but this is probably rare.

lachnerm:
Alright, as stated in the commit it is now possible to hide single columns in every runset via the URL parameters.

I have not fully added the support for displaying same titled columns in the `SelectColumn` component as this would afford a lot of additional checks to be made and there is still the problem with assigning the columns over multiple runsets as I mentioned before.

However, I've made some smaller changes so the component will not break any more if there are multiple same titled columns. For this to work, the `SelectColumn` component displays only the columns with different titles, but **toggling them between hidden/visible will affect all same-titled columns**. So for example for the previously mentioned case, there is only one column called `cputime` that can be toggled in the `SelectColumn` component, but both columns that have this title will be affected by the toggling. Respectively, both columns will be added to the URL as hidden.
This was the easiest and still somehow intuitive solution I found for this problem - and I think regarding that there are no practical examples of this case anyway, I would suggest to leave it at that. 

I will add the URL parameter feature to all other components of the HTML table next.

PhilippWendler:
What you explain sounds good. I will have a look at it.

> I will add the URL parameter feature to all other components of the HTML table next.

Don't touch the filters in this regard, because will be done as part of more work on filters related to #581.

Apart from this I think we have all features covered?

lachnerm:
> Don't touch the filters in this regard, because will be done as part of more work on filters related to #581.
>
Alright, I will refrain from changing the file in this case. I'm not completely up to date which changes will be done there, but I want to mention that I intend to completely remove the `isVisible` property for tools and columns of the data object. This would be instead managed by a `hiddenCols` object that is created for each component with the data from the URL. But I think there should be no problem to integrate this solution into the changes done from the other issue.

> Apart from this I think we have all features covered?
>
Yes, I think apart from the changes for the other components everything should be covered (except for possible changes to my code so far of course).

PhilippWendler:
> I intend to completely remove the `isVisible` property for tools and columns of the data object. This would be instead managed by a `hiddenCols` object that is created for each component with the data from the URL. But I think there should be no problem to integrate this solution into the changes done from the other issue.

Filtering is not related to hidden columns, so this should be fine.

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
```diff
@@ -131,74 +175,103 @@ export default class SelectColumn extends React.Component {
...
+      hiddenCols,
+    }));
   };
 
-  handleSelectColumns = ({ target }) => {
-    const value = target.type === "checkbox" ? target.checked : target.value;
-    const list = [...this.state.list];
+  // Toggles all columns of the runset with the id of the target
+  toggleToolHidden = ({ target }) => {
+    let toolIdx = parseInt(target.name.split("--")[1]);
```
PhilippWendler, line 197: Wouldn't it be cleaner to add the index as parameter to `toggleToolHidden` and use a lambda that captures the value for the event handler instead of a direct method reference?

The less string parsing the better.

lachnerm:
> I tried it out (compiled bundle from [c9816d8](https://github.com/sosy-lab/benchexec/commit/c9816d8b7d07c805de608b3ab259474cbf6760be)) and it seems to work fine for the Table tab, but not on other tabs?
> 
> It still looses the parameter when switching tabs, and using the select column UI on the first tab adds the parameters to the URL, but does not affect the summary table.
> 
> And even if I manually append the parameters to the URL while being on a plot tab this does not hide the column there.

Yes, this is what I was referring to with
> I will add the URL parameter feature to all other components of the HTML table next.
>
Sorry if that was not entirely clear before. I have mainly worked on the `SelectColumn` component and all possible cases on how the URL parameters can be affected there. I will add this feature to the other tabs next - however this should not take too long as I should be able to implement it pretty analogous to the `ReactTable` component.

PhilippWendler:
Oh, I see. I thought you mean you will add state storage in URL to all other components.

benchexec/tablegenerator/react-table/src/components/SelectColumn.js
lachnerm: Yes, absolutely. I picked up the parsing of the html tag's name from the existing code, but providing the data directly via lambdas is definitely cleaner. I've applied your suggestion here and at other similiar parts of the code.

PhilippWendler:
Can you have a look at the test failure and then ping me when it is ready for review, please?

lachnerm:
@PhilippWendler 
The feature is now fully implemented. I've underestimated the amount of changes and the extent of error handling that was necessary a little bit, so it took longer than I expected - I'm sorry about that. However, I think everything should be working now as expected. In case there are any bugs left that I have missed, I will fix them of course.

I have fully removed the `isVisible` property of the `tools` object the `Overview` component shares between all parts of the table. Instead, the hiding of runsets and columns is now fully regulated via the URL parameters. These parameters are shared between components of the table and can be manually modified by changing the URL or by adjusting them in the `SelectColumn` modal. The parameters are of the following form:

- `hidden=x,y,z` with `x`, `y` and `z` as an index of a runset and the parameter taking at least one index
- `hiddenT=x,y,z` with `T` as an index of a runset and `x`, `y` and `z` as an index of columns of the runset `T`  and the parameter taking at least one index

All indexes that refer to a runset or column that is not defined in the data object will be ignored. Also all kinds of parameters/typos that do not exactly match this pattern will be ignored. If a parameter of the form `hiddenT` is defined to hide all columns of the runset `T`, the runset will respectively also be treated as hidden. Conversely, if a runset is hidden via the `hidden` parameter, all columns of this runset will also be treated as hidden.

In addition there were several bugs in the `SelectColumn` and `QuantilePlot` components that were directly related to the visibility of columns. As I was changing the whole setup of how columns are hidden with this feature any way, I also fixed those bugs on the way. These bugs mainly resulted in the application crashing when columns/runsets or certain combinations of those were hidden, but should not appear any more. Also I've added error handling in the `QuantilePlot` for possible unknown values of the URL parameters.

lachnerm:
I'm glad to hear that!

The changes to the `vendor.min.js` happened because I accidentally built the project one time on my windows machine instead of the docker image. They disappeared again after building the project with docker instead.

Issue #256, opened 2017-09-07
More flexible interface for tool-info module


PhilippWendler:
Currently it is hard to extend the interface between `benchexec` and the tool-info module(s) in some places. For example, we cannot add parameters to the `cmdline` methods for multiple property files, more options, etc.

We can add versioning to the API by letting the tool-info module declare which version of the API it supports, but this can lead to much code in `benchexec` for handling many options. Thus, I think the best way is to change the interface now and introduce parameter objects for e.g. `cmdline`. These parameter objects can than get new parameters without problems. For backwards compatibility with the old modules, we probably still need some version flag, though.

This issue is a prerequisite for solving #166 and #217.

PhilippWendler:
Some more information:

The new solution should be backwards compatible with existing tool-info modules, because there may exist an unknown number of tool-info modules outside of this repository.

The new solution should also provide a clear way how to extend it in the future with additional information, such that we do not have similar compatibility problems again in the future.

If possible, it would of course also be nice if new tool-info modules would be backwards-compatible with previous versions of BenchExec, but this is not a strict requirement. However, I think it should at least be possible to avoid any conflicts between old and new tool-info modules, such that (if a user wishes) they could write a single tool-info module that supports both the old and the new API. This could mean for example to not reuse existing method names.

The naive solution of adding new parameters to existing methods would not work, because this would fail with existing tool-info modules (their methods expect a fixed list of parameters).

However, our [documentation](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#writing-a-tool-info-module) requires that the `Tool` classes of tool-info modules inherit from `benchexec.tools.template.BaseTool`. So we can use this class to create backwards compatibility. One solution might be to add new methods and provide adapter methods that delegate to the old methods in `BaseTool`.

I still do not have a good idea yet how the new solution would actually look like such that it is easily extensible in the future. Not sure what best practices for Python callback APIs are.

skanav:
The following is without any consideration for the backward compatibility. That has to be discussed after we agree on the changes.

The use cases I could gather from the above issues are:
* allow to pass arbitrarily named arguments
* allow access to the arguments in the determine result function

Additionally, I would like the tool info module to differentiate between options and arguments.

This is my draft proposal:
1. Put the options in the constructor. Even consider getting the `executable` in the constructor and making it a field.
2. Change the signature of `cmdline` to either of the two:
[...]

PR #587, opened 2020-07-23
Toolinfo coveriteam

skanav:
Tool info modules for CoVeriTeam. Contains 2 files:
1) Generic CoVeriTeam TI
2) A verifier or a validator based on CoVeriTeam.

### DIFF
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -57,16 +57,16 @@ class Tool(coveriteam.Tool):
         # We expect one tasks and a propertyfile.
         if not tasks:
             raise UnsupportedFeatureException(
-                "Can't execute CoVeriTeam-Verifier-Validator:"
+                "Can't execute CoVeriTeam-Verifier-Validator: "
                 "Input program is missing."
             )
         if len(tasks) > 1:
             raise UnsupportedFeatureException(
-                "Can't execute CoVeriTeam-Verifier-Validator:"
+                "Can't execute CoVeriTeam-Verifier-Validator: "
                 "Too many input files to analyze"
             )
         if not propertyfile:
             raise UnsupportedFeatureException(
-                "Can't execute CoVeriTeam-Verifier-Validator:"
+                "Can't execute CoVeriTeam-Verifier-Validator: "
                 "Specification is missing."
             )
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -1,0 +1,32 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.coveriteam as coveriteam
+
+
+class Tool(coveriteam.Tool):
+    """
+    Tool info for a verifier or a validator based on
+    CoVeriTeam: a Configurable Software-Verification Platform.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+    """
+
+    def name(self):
+        return "Verifier (or Validator) Based on CoVeriTeam"
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        """
+        Prepare command for the example coveriteam program for a verifier.
+        """
+
+        spec = (
+            ["--input", "spec_path=" + propertyfile] if propertyfile is not None else []
+        )
+        # We don't support more than one tasks at the moment.
+        prog = ["--input", "prog_path=" + tasks[0]]
+        additional_options = prog + spec
+        return [executable] + options + additional_options
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -7,6 +7,7 @@
 
 import benchexec.tools.coveriteam as coveriteam
 import benchexec.result as result
+import sys
 
 
 class Tool(coveriteam.Tool):
@@ -23,10 +24,14 @@ class Tool(coveriteam.Tool):
         """
         Prepare command for the example coveriteam program for a verifier.
         """
-        spec = (
-            ["--input", "spec_path=" + propertyfile] if propertyfile is not None else []
-        )
-        # We don't support more than one tasks at the moment.
+        # We expect one tasks and a propertyfile.
+        if len(tasks) != 1 or not propertyfile:
+            sys.exit(
+                "Can't execute CoVeriTeam-Verifier-Validator."
+                "Either propertyfile is missing or the number of tasks is not 1."
+            )
+
+        spec = ["--input", "spec_path=" + propertyfile]
         prog = ["--input", "prog_path=" + tasks[0]]
         additional_options = prog + spec
 
@@ -34,9 +39,9 @@ class Tool(coveriteam.Tool):
 
     def determine_result(self, returncode, returnsignal, output, isTimeout):
         """
-        This function will be useful in case of a verifier and a validator.
         It assumes that any verifier or validator implemented in CoVeriTeam
-        will print out the produced aftifacts in the end.
+        will print out the produced aftifacts.
+        If more than one dict is printed, the last one is taken.
         """
         res = {}
         for line in output:
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import benchexec.tools.coveriteam as coveriteam
+import benchexec.result as result
 
 
 class Tool(coveriteam.Tool):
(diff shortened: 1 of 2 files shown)


benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Tool(coveriteam.Tool):
+    """
+    Tool info for a verifier or a validator based on
+    CoVeriTeam: a Configurable Software-Verification Platform.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+    """
+
+    def name(self):
+        return "Verifier (or Validator) Based on CoVeriTeam"
```
PhilippWendler, line 19: This seems more like a description instead of a name. And it is very long and will probably disturb users when printed in the column headers of tables. Why not just `CoVeriTeam`?

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for CoVeriTeam: On-Demand Composition of Cooperative Verification Systems.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+
+    This class has 2 purposes:
+        1. to serve as an abstract class for specific coveriteam programs like verifiers, validators, etc.
+        2. to serve as the tool info module for any generic coveritea program.
```
PhilippWendler, line 20: If this is not only intended to be an abstract class, it needs `name()`.

But later on there are comments that further contradict this. So what is actually the intention?

Wouldn't it make much more sense to have just one module for CoVeriTeam? Otherwise there will probably be lots of modules, BenchExec needs to be changed every time CoVeriTeam gets a new use case, and the names of these are not really nice as can be seen already.

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        """
+        Prepare command for the example coveriteam program for a verifier.
+        """
+
+        spec = (
+            ["--input", "spec_path=" + propertyfile] if propertyfile is not None else []
+        )
+        # We don't support more than one tasks at the moment.
```
PhilippWendler, line 29: Then it should fail instead of silently swallowing all but the first input file.

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    # TODO: I am not sure about the following folders:
+    # 1. examples and config: should be included or not? It can also be dealt with the required files tag in the behchdef.
```
PhilippWendler, line 24: The paths returned by `program_files` should list those files and directories that are required for executing the tool. A directory called `examples` suggest that it should not be necessary for executing the tool (if it is, I would strongly suggest to refactor the directory layout in order to not confuse users).

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+    # 2. tools and toolinfocache: these are cache folders. Isn't it better just to wrap them in one folder called cache?
```
PhilippWendler, line 25: This seems out of place, it is a question for CoVeriTeam and not related to the tool-info module?

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        super().__init__()
+
+    def executable(self):
+        return util.find_executable("bin/coveriteam")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def program_files(self, executable):
+        return [executable] + self.REQUIRED_PATHS
```
PhilippWendler, line 47: This cannot be correct, because it would work only if the user happens to be executing BenchExec from some specific directory. Please make sure that this method always returns paths that are relative to the current working directory. (Why not use the default implementation of this method?)

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        "bin",
+        "lib",
+        "examples",
+        "config",
+        "tools",
+        "toolinfocache",
+    ]
+
+    def __init__(self):
+        super().__init__()
```
PhilippWendler, line 38: redundant

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        CoVeriTeam in general, wouldn't be used on its own.
+        Instead, we expect a tool to inherit from this class and that be executed.
+        That tool should define its own cmdline method.
+        """
+        return [executable] + options
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        """
+        This function will be useful in case of a verifier and a validator.
```
PhilippWendler, line 59: If indeed the split into individual tool-info modules is intended, why is code with this comment in the base module?

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        It assumes that any verifier or validator implemented in CoVeriTeam
+        will print out the produced aftifacts in the end.
+        """
+
+        # Assumption: the result dict is printed last
```
PhilippWendler, line 64: I would suggest to make a less string assumption.

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        s = output[-1].rstrip().strip("{}")
```
PhilippWendler, line 65: This seems relatively dangerous, because it removes all braces from the front and back of the string, not just one pair.

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,72 @@
...
+        # Reconstruct the dict from the printed string. Simple literal_eval does not work.
```
PhilippWendler, line 66: All the parsing will fail in relatively arbitrary ways if the output is not as expected, e.g., due to a timeout. So please make this more robust.

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,32 @@
...
+    CoVeriTeam: a Configurable Software-Verification Platform.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+    """
+
+    def name(self):
+        return "Verifier (or Validator) Based on CoVeriTeam"
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        """
+        Prepare command for the example coveriteam program for a verifier.
```
PhilippWendler, line 23: The class description claims that this tool-info module is also for validators. Where is the command line for them?

skanav: This function produces command for both.
as with any parameter which is not program or property: it is in the options.

PhilippWendler: Then the comment needs to be updated, otherwise it is highly confusing. Also, why does it mention an example program?

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,57 @@
...
+    def name(self):
+        return "CoVeriTeam-Verifier-Validator"
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        """
+        Prepare command for the example coveriteam program for a verifier.
+        """
+        # We expect one tasks and a propertyfile.
+        if len(tasks) != 1 or not propertyfile:
+            sys.exit(
```
PhilippWendler, line 29: Please split this into two conditions and provide separate error messages. Otherwise this is bad UI if we know precisely what is wrong but let the user guess.

Instead of `sys.exit`, please raise `UnsupportedFeatureException` and please refer to "input files", not "tasks" in the error message.

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,57 @@
...
+        It assumes that any verifier or validator implemented in CoVeriTeam
+        will print out the produced aftifacts.
+        If more than one dict is printed, the last one is taken.
+        """
+        res = {}
+        for line in output:
+            line = line.strip()
+            if "verdict" in line:
+                s = line.rstrip().strip("{}")
+                # TODO find a better way to do it.
```
PhilippWendler, line 51: This looks like a poor man's attempt at parsing JSON, but will break for example as soon as a comma is contained in a value. Why not use a proper JSON parser?

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,56 @@
...
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        """
+        In case when someone wants to run a generic CoVeriTeam program.
+        In this case the caller has to set the inputs in the options field,
+        because we do not know what would be the inputs to a generic CoVeriTeam program.
+        """
+        return [executable] + options
```
PhilippWendler, line 56: Is it really intended that the names of the input files are completely omitted? To me it seems that this would make this tool-info module completely unusable.

PhilippWendler:
As this is not finished yet, let's move it to the next release together with #256? Or do you need it urgently?

skanav:
the comment for `cmdline` in coveriteam.py is still under consideration

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -0,0 +1,72 @@
...
+                        return d.get("verdict", result.RESULT_ERROR)
+                except SyntaxError:
+                    pass
+        return result.RESULT_ERROR
+
+    def check_inputs(self, tasks, propertyfile):
+        # We expect one tasks and a propertyfile.
+        if not tasks:
+            raise UnsupportedFeatureException(
+                "Can't execute CoVeriTeam-Verifier-Validator:"
```
PhilippWendler, line 60: There is a space missing after the colon in all these error messages.

benchexec/tools/coveriteam.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for CoVeriTeam: On-Demand Composition of Cooperative Verification Systems.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+
+    This class has 2 purposes:
+        1. to serve as an abstract class for specific coveriteam programs like verifiers, validators, etc.
+        2. to serve as the tool info module for any generic coveriteam program.
```
PhilippWendler, line 19: Is this still true now that `cmdline` has been removed? There is a default `cmdline` implementation, so it could be, I am just asking for confirmation.

skanav: yes

PhilippWendler (approved):

Issue #516, opened 2019-11-18
Score-based quantile plots in HTML table


PhilippWendler:
SV-COMP uses score-based quantile plots for presenting results, which have some advantages over reqular quantile plots. We could add these as a variant on the tab with the quantile plot (and probably make it the default if all tasks in the table have a score).

Explanation of score-based quantile plots is for example in our [paper, Section 7.8](https://www.sosy-lab.org/research/pub/2019-STTT.Reliable_Benchmarking_Requirements_and_Solutions.pdf).

PhilippWendler:
This plot type would be available only if there are scores in the table. If all runs have scores, I see no reason against making it the default, but maybe we should not make it the default if only some runs have scores? Or would that be too confusing?

In any case, if the score-based quantile plot is selected, the (future) drop down for Results (all/correct only) needs to be disabled and should probably show a blank value (because for score-based quantile plots the choice of visible results is defined). We could also hide it, but I guess showing it in disabled state is less confusing.

lachnerm:
I have started to implement a solution for this new plot type. However, there are still a few things I'm missing or not sure about how to handle them:

- Where can I see whether there is a scoring scheme defined for the runset and where can I find it? Or do we use a fixed scoring scheme?
- In case it's not already defined in the scheme: Besides `correct` & `incorrect` I have also encountered the categories `wrong` & `missing` for some results. How are those handled with regard to the score-based plot?
- Is the option `Score-based Quantile Plot` supposed to be shown (possibly disabled) even if there are no runsets that support scoring?
- In case a column is selected with the score-based plot type, are runsets that contain this column and don't support scoring supposed to be hidden? Is this combination even possible in practice?

PhilippWendler:
>  * Where can I see whether there is a scoring scheme defined for the runset and where can I find it? Or do we use a fixed scoring scheme?

Currently, there is only one supported scoring scheme, the one from [SV-COMP](https://sv-comp.sosy-lab.org/2020/rules.php#scores). But I think we should not need to implement this information in the HTML table, and instead just read the scores from the JSON data.

>  * In case it's not already defined in the scheme: Besides `correct` & `incorrect` I have also encountered the categories `wrong` & `missing` for some results. How are those handled with regard to the score-based plot?

`missing` means that no information about correctness of the result is available, and thus also no score. `incorrect` should not occur, but `wrong` means the same.

[...]

PR #588, opened 2020-07-27
Redesign UI of buttons and legend for plots in HTML tables

PhilippWendler:
The current UI for changing the plot settings is suboptimal. The most important point is that the buttons show only which kind of alternative setting would be available, they do not indicate the current state. This means for example that a new user will not see that the plot by default hides incorrect and error results until after they have happened to click twice on the button for changing the selection of results. This can be highly confusing and should be changed.

Furthermore, we want to add score-based quantile plots (cf. #516) and then we have three kinds of plots, so toggling between them with a two-state button would not work anymore anyway.

Using drop-down fields for plot settings solves both problems.

Some part of the UI for changing plot settings is above the plot and some part below. This roughly corresponds to "what does the plot show" vs. "how is it shown", but again the button for changing the selection of results does not respect this, and neither does the legend (which also allows to change what is shown). If all UI elements would be in a single place, the user would also have to move the mouse less. Above the plot seems like the best place because most other UI elements of our tables are also at the top.


lachnerm:
> I am really impressed with how well the sizing of the drop downs and the adjustment to the window width work now. Great job!

I am glad that the solution is working fine!
> 
> * Add tooltip to "Results" drop down to mention that the filter is applied always in addition to what is selected here.

I have added a tooltip. 
> * The plot library leaves much whitespace to the left of the plots. Is it possible to remove this?

The whitespace is some hard coded margin that was probably put there so there will be space for the labels of the Y-Axis. There doesn't seem to be a way for the plot to be rendered with a margin that is automatically as wide as the widest label of the axis. Also this would make the axis of the plot and the whole plot itself jump around when switching between selections, so I can't really think of a better solution.

Additionally I have added an optgroup for the Aux. lines dropdown serving as a header. The options of the dropdown were also changed from `{value}` to `c = {value}`.

PR #592, opened 2020-08-13
Redesign of API of tool-info modules

PhilippWendler:
This is intended to solve #256 and follows what was discussed in https://github.com/sosy-lab/benchexec/issues/256#issuecomment-665650327.

The first three commits of this PR add a new version of the tool-info API (by adding a new base class `BaseTool2`) and the necessary adjustments in BenchExec such as a wrapper that makes old tool-info modules (based on `BaseTool`) continue to work.

Commit c3a9565ea0a8ddad74e4b56d4fa88a0312b95fb2 provides one API improvement that can now be made for `BaseTool2` and that was not possible before because it would have broken backwards compatibility. This can be seen as an example for showing that our adapter approach works and how it looks like in practice, and as example for how I intend the future API to look like (e.g., using named tuples as immutable domain-specific parameter objects).

The plan is to continue making such changes to `BaseTool2` (e.g., add a `Task` object, implement #166, and provide the info from #589) until this is merged and released, and after that time make only backwards compatible changes to `BaseTool2`. If new incompatible changes are needed at some later point in the future, a new class `BaseTool3` together with an adapter for old tool-info modules would be added.

I changed my opinion on the removal of the `executable()` method (which was suggested in https://github.com/sosy-lab/benchexec/issues/256#issuecomment-664175379), though. Currently, `benchexec` takes the path returned from `executable()` and makes it relative to what `working_directory()` returns. If we remove the former method and let the tool-info module internally add the executable to the command line, this would also needed to be implemented in the tool-info module, where it can be easily forgotten. So the `executable()` method feels like the better solution.
@skanav: Do you agree?

I used a manual approach for adding the adapter at runtime (i.e., "if instance of `BaseTool`, wrap with adapter class") instead of decorators because we cannot change existing tool-info modules. I also plan to continue with this style and do not use decorators for the new API version (which were suggested in https://github.com/sosy-lab/benchexec/issues/256#issuecomment-665539388) because we want to use a base class like `BaseTool2` anyway in order to provide default implementations and utility methods to tool-info modules. Requiring them to inherit from `BaseTool2` and add a decorator like `@API2` would just be redundant and possibly inconsistent. The code that adds the wrapper manually is only a few lines anyway and centralized in a single place (in `benchexec.tooladapter.adapt_to_current_version()`).

As suggested in https://github.com/sosy-lab/benchexec/issues/256#issuecomment-665502721, I made `BaseTool2` an abstract base class as defined by the `abc` module. This allows us to declare methods as abstract (needs to be overwritten) and to define subclasses of `BaseTool2` that do not actually inherit anything from `BaseTool2`. The latter is useful for special subclasses like our `ContainerizedTool` class and our new adapter class `Tool1To2`.
@skanav: In https://github.com/sosy-lab/benchexec/issues/256#issuecomment-665519459 you argued against using `abc` but did not provide any concrete arguments. Could you please have a look at how it is used here and whether you like it? If not, what would you suggest?

There is also an open naming question: Current code of BenchExec uses (relatively consistently) the spellings `cpu_cores` and `cputime`, though now where both are used as fields of the new class `ResourceLimits` this feels a little bit inconsistent. The spelling as two words (`cpu_cores`) seems more correct, but specifically `cputime` is spread wide throughout the code base and even appears in the HTML tables etc. So I am not sure which of these spellings should be adopted or whether we should just keep the existing spellings.


benchexec/tools/template.py
```diff
@@ -22,17 +35,386 @@ class UnsupportedFeatureException(benchexec.BenchExecException):
...
+    @abstractmethod
+    def executable(self):
+        """
+        Find the path to the executable file that will get executed.
+        This method always needs to be overridden,
+        and most implementations will look similar to this one.
+        The path returned should be relative to the current directory.
+        @return a string pointing to an executable file
+        """
+        return util.find_executable("tool")
```
lembergerth, line 83: Subclasses of `BaseTool2` can call `super().executable()`, which wouldn't make sense.

It may be more safe to replace the method definition with `raise NotImplementedError()`, but I like that an example is given.

Similar reasoning for the other abstract methods of this class.

benchexec/tools/template.py
```diff
@@ -22,17 +35,386 @@ class UnsupportedFeatureException(benchexec.BenchExecException):
...
+            cputime_hard=None,
+            walltime=None,
+            memory=None,
+            cpu_cores=None,
+        ):
+            return super().__new__(
+                cls, cputime, cputime_hard, walltime, memory, cpu_cores
+            )
+
+        pass
```
lembergerth, line 402: What's that `pass` for?

benchexec/tools/template.py
```diff
@@ -22,17 +35,386 @@ class UnsupportedFeatureException(benchexec.BenchExecException):
...
+
+
+class BaseTool(object):
```
lembergerth, line 405: can be simplified to `class BaseTool:`

benchexec/tooladapter.py
```diff
@@ -0,0 +1,88 @@
...
+
+import inspect
+
+from benchexec.tools.template import BaseTool, BaseTool2
+
+import benchexec.model
+
+
+CURRENT_BASETOOL = BaseTool2
+"""Alias for the latest base-tool class in benchexec.tools.template"""
```
lembergerth, line 18: Developers of tool-info modules may be tempted to use this constant as type.
I would explicitly discourage the use of this constant in tool-info modules, and tell users in the pydoc that they should use the concrete `BaseToolX` type

benchexec/model.py
```diff
@@ -93,16 +95,51 @@ def load_task_definition_file(task_def_file):
     if not task_def:
         raise BenchExecException("Invalid task definition: empty file " + task_def_file)
 
-    if str(task_def.get("format_version")) not in ["0.1", "1.0"]:
+    format_version = str(task_def.get("format_version"))
+    if format_version not in ["0.1", "1.0", "1.1"]:
```
lembergerth, line 99: The supported format versions could be refactored into a constant at the file start, to make the supported versions easy to find

benchexec/tooladapter.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+# We do not let Tool1To2 actually inherit from BaseTool2 because we do not want to
+# inherit any default implementations, but we still declare it as a subclass.
+@BaseTool2.register
+class Tool1To2(object):
```
lembergerth, line 24: Can be simplified to `class Tool1To2:`

benchexec/tooladapter.py
```diff
@@ -0,0 +1,88 @@
...
+            rlimits_dict,
+        )
+
+
+def adapt_to_current_version(tool):
+    """
+    Given an instance of a tool-info module's class, return an instance that conforms to
+    the current API. Might be either the same or a different instance.
+    """
+    if isinstance(tool, BaseTool2):
```
lembergerth, line 80: How about using `CURRENT_BASETOOL` here? I think that's closer to the intended semantics, and avoids future modifications of the used type here.

```python
if isinstance(tool, CURRENT_BASETOOL):
```

lembergerth (changes requested):
Very nice, thank you!
I only have minor comments.

benchexec/tools/template.py
```diff
@@ -22,17 +35,386 @@ class UnsupportedFeatureException(benchexec.BenchExecException):
     pass
 
 
-class BaseTool(object):
+class BaseTool2(object, metaclass=ABCMeta):
```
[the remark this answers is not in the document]
PhilippWendler: I did this on purpose to make it easier to understand. When inheriting from `ABC`, it is not clear that this entails (there could be inherited methods, for example), whereas with the explicit metaclass it is clear that only the metaclass is changed.

benchexec/tools/template.py
```diff
@@ -22,17 +35,386 @@ class UnsupportedFeatureException(benchexec.BenchExecException):
...
     usually determine_result(), cmdline(), and version(),
     and maybe working_directory() and get_value_from_output(), too).
-    The class for each specific tool need to be named "Tool".
-    For more information, please refer to
-    https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md
+
+    In special circumstances, it can make sense to not inherit from this class.
+    In such cases the tool-info module's class needs to implement all the methods
+    defined here and BaseTool2.register needs to be called with the respective class
+    (cf. documentation of abc.ABCMeta.register) to declare compatibility with BaseTool2.
```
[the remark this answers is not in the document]
PhilippWendler: This would be fine, and also calling `register(Tool)` afterwards in the module source. I prefer not to add too much details here, though, in order to not confuse and distract from the regular method. This info is also not specific to tool-info modules, it is standard `ABCMeta` stuff.

benchexec/tools/template.py
PhilippWendler: I prefer to not change the code of the old code to keep the diff minimal.

benchexec/tooladapter.py
PhilippWendler: Hm, isn't it enough that this is in a completely different module that tool-info modules should not even import, and developers of tool-info modules should not even know about typically?

We could add such a general comment to `benchexec.tooladapter`.

benchexec/tooladapter.py
PhilippWendler: I would prefer to keep the current code. This way, there is one branch per API version, and the content of this branch changes depending on how the API changes. When a new API version is added, a new branch is added.

It is also more fail-safe: If we forget to change this method when a new API version is added, it will immediately crash if a new tool-info module is tried out. If we adopt your suggestion, it would fail only if someone tests an old tool-info module.

benchexec/tools/template.py
PhilippWendler: This would be possible, but I guess it is not necessary. Calling `super().executable()` is not something that would be done automatically, and it would also immediately fail and be detected, right?

benchexec/tools/template.py
PhilippWendler: It was necessary before I added `__new__` and I forgot to remove it.

benchexec/tools/template.py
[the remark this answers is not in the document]
lembergerth: I'd argue it's easier to understand if there's less cognitive load. `ABC` should be just as well known as `ABCMeta`, so I don't think there's a risk of people fearing that there are any methods inherited.

benchexec/tools/template.py
lembergerth: I agree - I only mentioned it because it already is part of the git diff. But I get your point.

benchexec/tools/template.py
lembergerth: But why go with the less safe way, if there's a more safe (and more expressive) alternative?

The only argument I see is having a 'working' example.
But if you put the code example in the comments, IDE refactoring will still find it in case of method renames or method moves.

benchexec/tooladapter.py
lembergerth: Probably. A clarifying comment can still be added later, if it actually is used wrong.

benchexec/tools/template.py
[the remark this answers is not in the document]
PhilippWendler: My thought was that most developers of tool-info modules are unfamiliar with all of `abc`, but using `metaclass` instead of inheritance indicates to them that there is less (nothing) they need to care and know about.

benchexec/tools/template.py
PhilippWendler: Hm, my thought was to have
```
raise NotImplementedError()
return util.find_executable("tool")
```
but this looked ugly and confusing.

A comment is maybe a better idea.

lembergerth: We could even test the examples with [`doctest`](https://docs.python.org/3.5/library/doctest.html).
Unfortunately, doctest seems to require these ugly `>>>` to recognize code examples..

lembergerth (approved):

MartinSpiessl (commented):
It looks fine to me, but I cannot provide a exhaustive review without spending considerable time on understanding what the code does. I am not familiar enough yet with python abc, annotations or named tuples.

> So I am not sure which of these spellings should be adopted or whether we should just keep the existing spellings.

I would just keep the existing spellings. It is just a minor inconsistency, nothing which will hinder readability.

I had a look at the `Tool1To2` class as part of the review you requested for me on Pr #602 and liked how this was implemented. The problems I found were specific to PR #602 and as far as this PR #592 goes everything looks fine with the tool adapter.

PhilippWendler:
Current state: most of this was implemented and several improvements to the API have been made, including #602 (fixing #593) and #604 (fixing #166). Some more docs will be added and then this will be merged and released.

So if anybody has any further comments, it would be good to get them as soon as possible.

PR #596, opened 2020-08-20
Add score-based Quantile plot to the HTML table

### LINKED ISSUE #516: Score-based quantile plots in HTML table

lachnerm:
Adds the score-based Quantile Plot to the HTML table. The plot option is only shown when there are runsets that support a scoring scheme and disables the Results dropdown. The option is set as default, in case all runsets have scores.

Fixes #516.


benchexec/tablegenerator/react-table/src/utils/plot.js
```diff
@@ -15,22 +15,38 @@ import React from "react";
...
+        disabled={isDisabled}
       >
         {Object.values(options).map((option) => (
           <option value={option} key={option} name={option + " " + name}>
             {option}
           </option>
         ))}
+        {isDisabled ? (
+          <option value="disabled" name="disabled">
+            -----
```
PhilippWendler, line 45: Maybe use a single [long dash](https://en.wikipedia.org/wiki/Dash#Common_dashes_and_Unicode_characters), like the two-em dash?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -20,6 +20,7 @@ const prepareTableData = ({ head, tools, rows, stats, props }) => {
         ...column,
         colIdx: idx,
       })),
+      scoreBased: rows[0].results[idx].score !== undefined,
```
PhilippWendler, line 23: Doesn't this just check the first row? Could happen that we do not have a score there but in other rows.

benchexec/tablegenerator/react-table/src/components/QuantilePlot.js
```diff
@@ -232,67 +307,75 @@ export default class QuantilePlot extends React.Component {
...
+        arrayY = this.sortArray(arrayY, colTitle);
       }
     }
 
     this.hasInvalidLog = false;
     const newArray = [];
-
-    arrayY.forEach((el, i) => {
-      const value = el[0];
+    let xPositionCounter = scoreOfIncorrectResults;
```
PhilippWendler, line 358: I guess it would be easier to understand if the code is refactored and `xPositionCounter` is always used for quantile plots? I.e., initialize with 0 or `scoreOfIncorrectResults` and then increment by 1 or score? Then we can get rid of `i` and do not need the nested `+=` as side effect.

PhilippWendler (commented):
Thanks, this looks good!

The computation appears correct. In particular, I verified for an example table that the first data point has `x = (sumOfIncorrectScores + fastestCorrectScore)` and that the last data point has the same `x` value as is given for the score on the summary tab. (These are rules that are always valid for these kind of plot.)

What would be a good way to test this? Currently, we do not have snapshot tests for plots, and I am not sure whether we could add them?

Maybe have some kind of snapshot tests, but instead of HTML just dump a JSON data structure with all data points (i.e., what `renderAll` computes)? Is this easily possible? Or do you have a better idea?

Showing a score-based quantile plot is intended for specific columns (like `cputime`). Showing it for all columns of a single run set is not wrong, but of little use (it will be almost the same as the regular run set, just with a different x axis and potentially a somewhat different shape of the plots). Do you have a good idea for how we could handle this in the UI? Maybe disable the run set part of the "Selection" drop-down and add a tooltip that tells why? If we do not find something good, we can leave it enabled.

Could you please also replace the tooltip of the "Results" drop-down with something like "Score-based quantile plots always show correct results offset by the score of wrong results. Any defined filters will still be applied.".

And could you please make the "Score-based Quantile Plot" be the first entry of the "Plot" drop-down? It is good if the default is the first, and it should be next to "Quantile Plot".

benchexec/tablegenerator/react-table/src/utils/utils.js
lachnerm: Yes, I was assuming that given a score exists, it will be defined in all rows. I was trying to avoid to use `Array.some(...)` because in case there are no scores (which will be most of the time currently), all rows will unnecessarily be iterated through all the time. I guess there's no way around it then, right?

PhilippWendler: Yes.

Will this be checked each time a plot is opened? Then we could cache it.

PhilippWendler (approved):
Thanks!

Could you please merge master into the branch and updated the built files such that I can merge it?

lachnerm:
> What would be a good way to test this? Currently, we do not have snapshot tests for plots, and I am not sure whether we could add them?
> 
I've tried creating some snapshots tests for the generated HTML but there seems to be a problem with the `react-vis` package  together with the `react-test-renderer`, more precisely the `FlexibleXYPlot` we are using for flexible widths and heights of the plot. Every single test throws a similar error like this:
`TypeError: Cannot read property 'offsetHeight' of null
          at ResultClass._this._onResize (...\benchexec\tablegenerator\react-table\node_modules\react-vis\dist\make-vis-flexible.js:160:45)`
The same problem was already mentioned [here](https://github.com/uber/react-vis/issues/713), however I didn't find a solution. I didn't find anything about `makeVisFlexible` actually being deprecated either and it's still being used when creating a `FlexibleXYPlot`, which is also still defined in the [docs](https://github.com/uber/react-vis/blob/master/docs/flexible-plots.md). Switching to a "normal" `XYPlot` fixes this by the way, but then we are back to having to calculate the width and height by hand - and I don't think we should change the actual code just for a test to work any way.
> Maybe have some kind of snapshot tests, but instead of HTML just dump a JSON data structure with all data points (i.e., what `renderAll` computes)? Is this easily possible? Or do you have a better idea?
> 
Something like this should be possible. We could render a `QuantilePlot` instance which will contain the data points for all entries (which are computed in `renderAll`) and compare those data points to a predefined snapshot. We would need some rerendering in between for changing the current selection of the plot to force the plot to render the corresponding data points each time, but I think it's possible. This won't test the component as comprehensive as a snapshot of the full HTML, but it should suffice for ensuring that the score calculation is done correctly. Of course this would also avoid above errors.
In case we decide for a solution like this: Should I also add the same test for the other 2 plot types?

> Showing a score-based quantile plot is intended for specific columns (like `cputime`). Showing it for all columns of a single run set is not wrong, but of little use (it will be almost the same as the regular run set, just with a different x axis and potentially a somewhat different shape of the plots). Do you have a good idea for how we could handle this in the UI? Maybe disable the run set part of the "Selection" drop-down and add a tooltip that tells why? If we do not find something good, we can leave it enabled.
> 
In this case I would suggest simply disabling the options of the dropdown and adding the mentioned tooltip. Completely hiding them could be a bit confusing and I don't think there are any real advantages over disabling them. Currently, the runset options will be disabled if the runset itself doesn't have any columns that offer scores, so I'd just need to change the code to always disable them for score-based plots instead.

benchexec/tablegenerator/react-table/src/utils/utils.js
lachnerm: No, this is only done one time on the initial render of the `Overview` component.
I've applied the change.

PhilippWendler:
> Switching to a "normal" `XYPlot` fixes this by the way, but then we are back to having to calculate the width and height by hand - and I don't think we should change the actual code just for a test to work any way.

Hm, maybe we could render a `XYPlot` in tests only? I.e., refactor this out into a component that can be based either on `FlexibleXYPlot` or `XYPlot`, and then instantiate it differently. Or maybe this problem goes away if we put the `FlexibleXYPlot` into some container with a fixed size in the test?
 
> > Maybe have some kind of snapshot tests, but instead of HTML just dump a JSON data structure with all data points (i.e., what `renderAll` computes)? Is this easily possible? Or do you have a better idea?
> 
> Something like this should be possible. We could render a `QuantilePlot` instance which will contain the data points for all entries (which are computed in `renderAll`) and compare those data points to a predefined snapshot.

You mean a dummy `QantilePlot` that just produces some HTML containing all data points? Yes, that would be possible.

> In case we decide for a solution like this: Should I also add the same test for the other 2 plot types?

Yes, why not? The score-based plot is the most complex, but in the beginning we did have some problems with sorting and filtering for all kinds of plot, so regression tests would be good to have. There should also be some test with a status column.

> > Showing a score-based quantile plot is intended for specific columns (like `cputime`). Showing it for all columns of a single run set is not wrong, but of little use (it will be almost the same as the regular run set, just with a different x axis and potentially a somewhat different shape of the plots). Do you have a good idea for how we could handle this in the UI? Maybe disable the run set part of the "Selection" drop-down and add a tooltip that tells why? If we do not find something good, we can leave it enabled.
> 
> In this case I would suggest simply disabling the options of the dropdown and adding the mentioned tooltip. Completely hiding them could be a bit confusing and I don't think there are any real advantages over disabling them. Currently, the runset options will be disabled if the runset itself doesn't have any columns that offer scores, so I'd just need to change the code to always disable them for score-based plots instead.

Ok, let's try this out.

lachnerm:
> > Switching to a "normal" `XYPlot` fixes this by the way, but then we are back to having to calculate the width and height by hand - and I don't think we should change the actual code just for a test to work any way.
> 
> Hm, maybe we could render a `XYPlot` in tests only? I.e., refactor this out into a component that can be based either on `FlexibleXYPlot` or `XYPlot`, and then instantiate it differently.

Yes, I've tried this out and it seems to be working. This would need a few properties to be added to the `QuantilePlot` only necessary for the testing, but I'd actually say that's still reasonable as it's only minor changes. I'd prefer this solution then if there's nothing else against it - it's short and simple. The dimensions of the plot shouldn't matter any way for the tests I think.
> Or maybe this problem goes away if we put the `FlexibleXYPlot` into some container with a fixed size in the test?

No, this doesn't work. The problem is that the test-renderer cannot find the container itself with the used functions, not that the container doesn't have a computed width or height.
> 
> > > Maybe have some kind of snapshot tests, but instead of HTML just dump a JSON data structure with all data points (i.e., what `renderAll` computes)? Is this easily possible? Or do you have a better idea?
> > 
> > 
> > Something like this should be possible. We could render a `QuantilePlot` instance which will contain the data points for all entries (which are computed in `renderAll`) and compare those data points to a predefined snapshot.
> 
> You mean a dummy `QantilePlot` that just produces some HTML containing all data points? Yes, that would be possible.

Yes exactly, however I'd now prefer above solution. This would of course also include the comparison of the data points by default.
> 
> > In case we decide for a solution like this: Should I also add the same test for the other 2 plot types?
> 
> Yes, why not? The score-based plot is the most complex, but in the beginning we did have some problems with sorting and filtering for all kinds of plot, so regression tests would be good to have. There should also be some test with a status column.
> 
Alright, I'll also add tests for this then.

PR #600, opened 2020-09-10
Scatter and Quantile plot tests

lachnerm:
Adds tests for both the Scatter and the Quantile Plot.

Scatter Plot tests include cases with selections of the same runset and of different runsets as well as selections of different types. All cases are tested with all results and only correct results. Only the `big-table.diff.html` file is used as test data to limit the size of the snapshot file due to the exponentially increasing number of test cases.

Quantile Plot tests include cases for a quantile, direct and score-based quantile plot for all available selections. All cases are tested with all results and only correct results. In addition to the `big-table.diff.html`, the `rows-with-scores.html` file is used as test data too to have data for a score-based quantile plot.


PhilippWendler (approved):
Great!

I only force-pushed in order to remove the huge `QuantilePlot.test.js.snap` from the earlier commits of this PR from the git history.

PR #601, opened 2020-09-15
Tablegenerator tests on Windows

lachnerm:
Adds AppVeyor to the project to run the tablegenerator tests on Windows. Additionally some changes were made to the tablegenerator to successfully run the test on Windows.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,25 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python35"
+    - PYTHON: "C:\\Python36"
+    - PYTHON: "C:\\Python37"
+    - PYTHON: "C:\\Python38"
+
+build: off
+
+install:
+  - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+  # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
+  # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
+  - python -m pip install lxml
+  - python -m pip install --user .
+
+test_script:
+  - python setup.py nosetests --test benchexec.tablegenerator
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -797,7 +797,10 @@ class RunResult(object):
                 )
             )
             path_in_zip = urllib.parse.unquote(
-                os.path.relpath(url_parts.path, os.path.dirname(log_zip_path))
+                # os.path.relpath creates os-dependant paths, but windows separators can produce errors with zipfile lib
+                util.fix_path_if_on_windows(
+                    os.path.relpath(url_parts.path, os.path.dirname(log_zip_path))
+                )
             )
             if log_zip_url.startswith("file:///") and not log_zip_path.startswith("/"):
                 # Replace file:/// with file: for relative paths,
@@ -1245,7 +1248,8 @@ def create_tables(
     # get common folder of sourcefiles
     # os.path.commonprefix can return a partial path component (does not truncate on /)
     common_prefix = os.path.commonprefix([r.id.name for r in rows])
-    common_prefix = common_prefix[: common_prefix.rfind("/") + 1]
+    separator = "/" if "://" in common_prefix else os.sep
+    common_prefix = common_prefix[: common_prefix.rfind(separator) + 1]
     for row in rows:
         Row.set_relative_path(row, common_prefix, outputPath)
 
@@ -1323,7 +1327,8 @@ def write_csv_table(
         name, values, value_repetitions=itertools.repeat(1)  # noqa: B008
     ):
         if any(values):
-            out.write(name)
+            # name may contain paths, so standardize the output across OSs
+            out.write(util.fix_path_if_on_windows(name))
             for i in range(num_id_columns):  # noqa: B007
                 out.write(sep)
             for value, count in zip(values, value_repetitions):
@@ -1349,7 +1354,8 @@ def write_csv_table(
     )
 
     for row in rows:
-        out.write(row.short_filename)
+        # row.short_filename may contain paths, so standardize the output across OSs
+        out.write(util.fix_path_if_on_windows(row.short_filename))
         for row_id, is_relevant in zip(row.id[1:], relevant_id_columns[1:]):
             if is_relevant:
                 out.write(sep)
--- a/benchexec/tablegenerator/test_integration/__init__.py
+++ b/benchexec/tablegenerator/test_integration/__init__.py
@@ -15,13 +15,14 @@ import unittest
 
 import benchexec
 import benchexec.util
+import benchexec.tablegenerator.util
 
 sys.dont_write_bytecode = True  # prevent creation of .pyc files
 
 here = os.path.relpath(os.path.dirname(__file__))
 base_dir = os.path.join(here, "..", "..", "..")
 bin_dir = os.path.join(base_dir, "bin")
-tablegenerator = os.path.join(bin_dir, "table-generator")
+tablegenerator = [sys.executable, os.path.join(bin_dir, "table-generator")]
 
 # Set to True to let tests overwrite the expected result with the actual result
 # instead of letting them fail.
@@ -73,7 +74,7 @@ class TableGeneratorIntegrationTests(unittest.TestCase):
         output_path=None,
     ):
         output = self.run_cmd(
-            *[tablegenerator] + list(args) + ["--outputpath", output_path or self.tmp]
+            *tablegenerator + list(args) + ["--outputpath", output_path or self.tmp]
         )
         generated_files = {os.path.join(self.tmp, x) for x in os.listdir(self.tmp)}
 
@@ -159,6 +160,8 @@ class TableGeneratorIntegrationTests(unittest.TestCase):
         if expected_counts:
             # output of table-generator should end with statistics about regressions
             counts = output[output.find("REGRESSIONS") :].strip()
+            # Normalize line endings to avoid problems between OSs as Windows uses \r\n while Unix uses \n
+            counts = benchexec.tablegenerator.util.normalize_line_endings(counts)
             self.assertMultiLineEqual(expected_counts, counts)
         else:
             self.assertNotIn("REGRESSIONS", output)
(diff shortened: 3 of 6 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,20 @@
...
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+environment:
+  matrix:
+  - TOXENV: py35
+  - TOXENV: py36
+  - TOXENV: py37
```
PhilippWendler, line 12: Is there no Python 3.8 there?

Also, this duplicates information from `tox.ini`, is this necessary?

benchexec/tablegenerator/__init__.py
```diff
@@ -1245,7 +1249,14 @@ def create_tables(
     # get common folder of sourcefiles
     # os.path.commonprefix can return a partial path component (does not truncate on /)
     common_prefix = os.path.commonprefix([r.id.name for r in rows])
-    common_prefix = common_prefix[: common_prefix.rfind("/") + 1]
+    separator = (
```
PhilippWendler, line 1252: There is `os.sep` on Python, which already provides the separator. So we would just need to check for URLs (which could also be different from HTTP, I think we usually check for `://` to detect URLs).

benchexec/tablegenerator/test_integration/__init__.py
```diff
@@ -12,16 +12,22 @@
...
+import benchexec.tablegenerator.util
 
 sys.dont_write_bytecode = True  # prevent creation of .pyc files
 
 here = os.path.relpath(os.path.dirname(__file__))
 base_dir = os.path.join(here, "..", "..", "..")
 bin_dir = os.path.join(base_dir, "bin")
-tablegenerator = os.path.join(bin_dir, "table-generator")
+tablegenerator = [os.path.join(bin_dir, "table-generator")]
+line_ending = "\n"
```
PhilippWendler, line 27: `os.linesep` can be used instead. But instead of changing all the constants, I would normalize the tool output before comparing it against the expected value.

tox.ini
```diff
@@ -0,0 +1,13 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: I would suggest to merge this file with `pyproject.toml` ([documentation](https://tox.readthedocs.io/en/latest/example/basic.html#pyproject-toml-tox-legacy-ini)) to avoid clutter.

tox.ini
```diff
@@ -0,0 +1,13 @@
...
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[tox]
+envlist = py{35,36,37}
+
+[testenv]
+description = Unit tests
```
PhilippWendler, line 12: Having this definition here is somewhat dangerous, it could lead people to think that they should primarily use tox to run the tests, and then they execute only the table-generator tests.

Is tox actually necessary for testing on Travis? Wouldn't it be possible to directly execute `nosetests` from `.appveyor.yml`? If not, can we make tox execute all our tests by default, and pass an argument from `.appveyor.yml` to run only the table-generator tests?

.appveyor.yml
```diff
@@ -0,0 +1,20 @@
...
+
+build: off
+
+install:
+  - pip install tox
```
PhilippWendler, line 17: This seems to use Python 2 to install tox. Is this really the best way to run Python tests on Travis?

tox.ini
lachnerm: Tox doesn't seem to be necessary, however [AppVeyor uses tox in their documentation on Python testing](https://www.appveyor.com/docs/lang/python/), so I took the same approach. 
I've tried executing the tests directly from `.appveyor.yml`, which (after some adjustments) [seems to be working now](https://ci.appveyor.com/project/PhilippWendler/benchexec). For Python 3.8, this requires the `--traverse-namespace` option, otherwise the tests won't be found (like for example [here](https://ci.appveyor.com/project/PhilippWendler/benchexec/builds/35291867/job/arux4n6s62rmjo05) - the same problem occurs locally on my Windows machine without this option too). I've added this option in the `setup.cfg` file as it should make no difference for lower Python versions and like this there won't be any more problems regarding this topic in the future. If there is something against it, of course I can also remove it from there and add it in the `.appveyor.yml` file instead.

I suppose we will then keep the setup without tox, right? If so, I will remove the tox file from the project.

PhilippWendler: The option seems fine. In general it is a good idea to add the reason for having a non-obvious option (basically what you explained above, maybe shortened) either in the respective commit message or as a comment next to it, otherwise it will be hard to find later on.

Yes, I see no reason for using tox.

.appveyor.yml
```diff
@@ -7,14 +7,19 @@
...
 
 build: off
 
 install:
-  - pip install tox
+  - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+  - python -m pip install -U pip
+  - python -m easy_install -U setuptools
+  - python -m pip install -U pyyaml
+  - python -m pip install -U lxml
```
PhilippWendler, line 22: Summarize this into a single `pip` call? And is it really necessary to install pip and setuptools? After all, you are already using pip to install pip... :-)

benchexec/tablegenerator/test_integration/__init__.py
```diff
@@ -165,6 +164,12 @@ def expected_diff_file_name(ending):
         if expected_counts:
             # output of table-generator should end with statistics about regressions
             counts = output[output.find("REGRESSIONS") :].strip()
+            # Add \r before all \n on Windows if it's not there already
+            if platform.system() == "Windows":
+                line_feed_without_carriage = re.compile("(?<!\\r)\\n")
+                expected_counts = re.sub(
+                    line_feed_without_carriage, "\\r\\n", expected_counts
+                )
```
PhilippWendler, line 172: Would it be possible to simplify this to something like `counts = counts.replace('\r\n', '\n')`?

I would suggest to normalize the actual output to `\n` (more common) and we can do this unconditionally, no need for a platform check.

benchexec/tablegenerator/test_integration/__init__.py
```diff
@@ -12,16 +12,21 @@
...
 
 sys.dont_write_bytecode = True  # prevent creation of .pyc files
 
 here = os.path.relpath(os.path.dirname(__file__))
 base_dir = os.path.join(here, "..", "..", "..")
 bin_dir = os.path.join(base_dir, "bin")
-tablegenerator = os.path.join(bin_dir, "table-generator")
+tablegenerator = [os.path.join(bin_dir, "table-generator")]
+if platform.system() == "Windows":
+    tablegenerator = [sys.executable] + tablegenerator
```
PhilippWendler, line 29: I would suggest to drop the platform check. Using `sys.executable` actually seems like a good idea for Linux as well because of virtual envs.

benchexec/tablegenerator/test_integration/__init__.py
PhilippWendler: Basically call the utility function that you already have, just simplify its content and rename it to `normalize_line_endings`.

lachnerm: Yes, replacing the line endings in this direction actually seems to make more sense and is simpler. I've applied the changes.

Should we also do the same for the `fix_path_if_on_windows` function, i.e. always call the `replace` function regardless of the current OS? No problems occured when I tried it out. Would that make sense?

PhilippWendler: `\` is an allowed file-name character on Linux, so here we should not do this.

.appveyor.yml
PhilippWendler: Something that came to my mind only now (sorry!): Can't we just let pip install all dependencies of BenchExec? On Travis and GitLab CI we do this with `pip install --user .`. Would of course be nice if we would not need to maintain the dependencies in several places. At least currently we don't have dependencies that would look problematic on Windows.

Btw.: `lxml` is only necessary for tests, but not for table-generator tests, so it could even be excluded. But if we adopt my latest suggestion, then it won't hurt if `lxml` is installed nevertheless.

lachnerm: I added them in the beginning because they seemed to solve some error messages that were occuring (installing pip was used to update it to the newest version). Those error messages seem to be taken care of now any way though, so I could remove them and summarize the other two calls into a single one as you suggested.

When using the suggested command with the `--user` flag however, the same problem occurs that already occured when the install command for `lxml` was missing, so [the build fails](https://ci.appveyor.com/project/PhilippWendler/benchexec/builds/35339936). Strangely enough it only fails for Python 3.5 and 3.8, which was also the case [when not installing `lxml`](https://ci.appveyor.com/project/PhilippWendler/benchexec/builds/35291790). I actually can't really think of how this is happening, do you have any ideas? As already stated above, explicitly giving the calls for installing `lxml` fixes the problem.

PhilippWendler: Note that in both cases, the actual testing phase with `setup.py nosetests` fails, because whenever `setup.py` is executed it ensures that its `setup_requires` dependencies are installed, and ours list `lxml`. `setup.py` uses `easy_install` to install any missing dependencies, which is an old and somewhat deprecation installation utility. `easy_install` downloads the source package of `lxml` and this fails because it bundles C code, which would need to be compiled into a binary library with a compiler, which is missing.

For Python 3.6 and 3.7 it seems that `lxml` is already installed in the Python environment by Travis and so not installation attempt is made.

When you install `lxml` explicitly with `pip`, you can see that `pip` downloads a binary package of `lxml` (`lxml-4.5.2-cp38-cp38-win32.whl`) instead and installation of this works without the need for a compiler.

I think we should probably modernize our `setup.py` somewhat, though I do not have a good overview over modern best practices for Python packages.

An easy solution is to to just explicitly install `lxml` with `pip` before doing `pip install .` (we could even add `--ignore-installed` to avoid any redundant installation if the package already exists).

For the same reason (preferring binary packages provided by Ubuntu instead of packages installed by `easy_install` or `pip`) our GitLab tests run in Docker containers with `lxml` preinstalled.

lachnerm: I see, that makes sense. 
So I guess we should handle the updating of the `setup.py` in a separate issue? Those changes would probably exceed the initial intention of the PR to run the tablegenerator tests on Windows platforms.

So for now, I've added the installation of `lxml` before the call of `pip install .`. However, the [`--ignore-installed` flag](https://pip.pypa.io/en/stable/reference/pip_install/#options) seems to overwrite the installed package instead of simply not installing it if it's already installed. The standard behavior of `pip` is to leave already installed package alone if I understand correctly. For some reason, [`lxml` is still installed for Python 3.6](https://ci.appveyor.com/project/PhilippWendler/benchexec/builds/35358438/job/o0pw6ar8db51p6xu) and 3.7. Could it be possible that the directories of the preinstalled and newly installed versions are different? And should we do something about it? Installing the package, even if it's possible that it's already preinstalled somewhere, doesn't seem like a big deal to me - especially regarding our intention to modernize the `setup.py` anyway and remove the implicit installation afterwards again.

Any way, as soon as the review of the PR is finished, I'd start rebasing the branch to fix and squash the commit history a bit, in case nothing speaks against this.

PhilippWendler: Thanks!

> So I guess we should handle the updating of the `setup.py` in a separate issue?

That was just background info for you, I don't think we need to handle this anytime soon.

> However, the [`--ignore-installed` flag](https://pip.pypa.io/en/stable/reference/pip_install/#options) seems to overwrite the installed package instead of simply not installing it if it's already installed.

Sorry, I read the docs wrong :-)

> And should we do something about it? Installing the package, even if it's possible that it's already preinstalled somewhere, doesn't seem like a big deal to me

Exactly. Would have been just a minor optimization, only worth it if it is easy.

So for me this is ready.

PR #602, opened 2020-09-17
Add parameter --tool-directory to benchexec

PhilippWendler:
This allows to specify the directory of the benchmarked tool easily without having to resort to workarounds like changing PATH.

It needs changes to the tool-info API, so it can only be done on top of #592 and will work only with new or upgraded tool-info modules.

### DIFF
--- a/benchexec/tooladapter.py
+++ b/benchexec/tooladapter.py
@@ -13,7 +13,7 @@ This is an internal module for BenchExec and not to be used by tool-info modules
 
 import inspect
 
-from benchexec.tools.template import BaseTool, BaseTool2
+from benchexec.tools.template import BaseTool, BaseTool2, ToolNotFoundException
 
 import benchexec.model
 
@@ -31,7 +31,6 @@ class Tool1To2:
     """
 
     _FORWARDED_METHODS = [
-        "executable",
         "program_files",
         "version",
         "name",
@@ -50,6 +49,22 @@ class Tool1To2:
             setattr(self, method, getattr(wrapped, method))
 
         self.__doc__ = inspect.getdoc(wrapped)
+
+    def executable(self, tool_locator):
+        if tool_locator.tool_directory:
+            raise ToolNotFoundException(
+                "Tool-info module for {} does not support parameter --tool-directory. "
+                "Instead, you can add the tool to PATH, "
+                "execute benchexec from the tool directory, "
+                "or upgrade the tool-info module.".format(self.name())
+            )
+
+        assert tool_locator.use_path and tool_locator.use_current
+        # This is the behavior that old tool-info modules are expected to have.
+        try:
+            return self._wrapped.executable()
+        except SystemExit as e:
+            raise ToolNotFoundException(str(e)) from e
 
     def cmdline(self, executable, options, task, rlimits):
         rlimits_dict = {}
@@ -90,3 +105,14 @@ def adapt_to_current_version(tool):
             "{} is not a subclass of one of the expected base classes "
             "in benchexec.tools.template".format(tool.__class__)
         )
+
+
+def create_tool_locator(config):
+    """
+    Create an instance of ToolLocator with the standard behavior based on the given
+    command-line options.
+    """
+    if config.tool_directory:
+        return CURRENT_BASETOOL.ToolLocator(tool_directory=config.tool_directory)
+    else:
+        return CURRENT_BASETOOL.ToolLocator(use_path=True, use_current=True)
--- a/benchexec/benchexec.py
+++ b/benchexec/benchexec.py
@@ -126,6 +126,14 @@ class BenchExec(object):
         )
 
         parser.add_argument(
+            "--tool-directory",
+            help="Use benchmarked tool from given directory "
+            "instead of looking in PATH and the current directory",
+            metavar="DIR",
+            type=util.non_empty_str,
+        )
+
+        parser.add_argument(
             "-n",
             "--name",
             dest="name",
--- a/benchexec/localexecution.py
+++ b/benchexec/localexecution.py
@@ -20,6 +20,7 @@ from benchexec import resources
 from benchexec.runexecutor import RunExecutor
 from benchexec.pqos import Pqos
 from benchexec import systeminfo
+from benchexec import tooladapter
 from benchexec import util
 from benchexec.intel_cpu_energy import EnergyMeasurement
 
@@ -43,7 +44,8 @@ def init(config, benchmark):
             )
     config.containerargs["use_namespaces"] = config.container
 
-    benchmark.executable = benchmark.tool.executable()
+    tool_locator = tooladapter.create_tool_locator(config)
+    benchmark.executable = benchmark.tool.executable(tool_locator)
     benchmark.tool_version = benchmark.tool.version(benchmark.executable)
 
 
--- a/benchexec/test_tool_info.py
+++ b/benchexec/test_tool_info.py
@@ -17,6 +17,7 @@ import yaml
 import benchexec
 import benchexec.benchexec
 from benchexec import model
+from benchexec import tooladapter
 from benchexec import util
 from benchexec.tooladapter import CURRENT_BASETOOL
 import benchexec.tools.template
@@ -116,13 +117,13 @@ def log_if_unsupported(msg):
         )
 
 
-def print_tool_info(tool):
+def print_tool_info(tool, tool_locator):
     """Print standard info from tool-info module"""
     print_multiline_text("Documentation of tool module", inspect.getdoc(tool))
 
     print_value("Name of tool", tool.name())
 
-    executable = tool.executable()
+    executable = tool.executable(tool_locator)
     print_value("Executable", executable)
     if not os.path.isabs(executable):
         print_value("Executable (absolute path)", os.path.abspath(executable))
@@ -137,6 +138,11 @@ def print_tool_info(tool):
         logging.warning("Designated executable does not exist.")
     elif not os.access(executable, os.X_OK):
         logging.warning("Designated executable is not marked as executable.")
+    if tool_locator.tool_directory:
+        abs_directory = os.path.abspath(tool_locator.tool_directory)
+        abs_executable = os.path.abspath(executable)
+        if not os.path.commonpath((abs_directory, abs_executable)) == abs_directory:
+            logging.warning("Executable is not within specified tool directory.")
 
     try:
         print_value("Version", tool.version(executable))
@@ -328,6 +334,12 @@ def main(argv=None):
     )
     parser.add_argument("tool", metavar="TOOL", help="name of tool info to test")
     parser.add_argument(
+        "--tool-directory",
+        help="look for tool in given directory",
+        metavar="DIR",
+        type=benchexec.util.non_empty_str,
+    )
+    parser.add_argument(
         "--tool-output",
         metavar="OUTPUT_FILE",
         nargs="+",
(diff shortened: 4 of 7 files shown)


PhilippWendler:
@masp @skanav I am especially interested in reviews of the API part of this (i.e., `ToolLocator`) because the user side is probably pretty clear. For example, would this work for cases where you call the `executable()` method of another tool-info module? One goal of this would be to avoid the need for changing the current directory in use cases like MetaVal's tool-info module.

benchexec/tooladapter.py
```diff
@@ -90,3 +105,14 @@ def adapt_to_current_version(tool):
             "{} is not a subclass of one of the expected base classes "
             "in benchexec.tools.template".format(tool.__class__)
         )
+
+
+def create_tool_locator(config):
```
MartinSpiessl, line 110: This is inconvenient for me. Where should I take a config object (argparse.Namespace) from? In MetaVal I already have the tool directory and need to create a `ToolLocator`. Since you probably do not want me to import and instantiate`ToolLocator` directly  a method like `create_tool_locator_from_tool_directory(tooldirectory)` would be handy.

I created a `types.SimpleNamespace` for now and got it working. This is a deja vu, I had to do this in the past. The dependency on `argparse.Namespace` is inconvenient and frustrating. 

The problem now is that none of the other tool-info modules that MetaVal wraps currently support the `--tool-directory`, so I get the following warnings:

```
22:20:33   array-tiling/skippedu.yml                       2020-09-19 22:20:33,974 - CRITICAL - Tool-info module for CPAchecker does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   array-examples/sanfoundry_43_ground.yml         2020-09-19 22:20:33,978 - CRITICAL - Tool-info module for CPAchecker does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   memsafety-ext3/getNumbers3.yml                  2020-09-19 22:20:33,985 - CRITICAL - Tool-info module for symbiotic does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   memsafety-ext3/derefAfterFree1.yml              2020-09-19 22:20:33,988 - CRITICAL - Tool-info module for symbiotic does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   termination-crafted/2Nested-2.yml               2020-09-19 22:20:33,999 - CRITICAL - Tool-info module for ULTIMATE Automizer does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   termination-crafted/TelAviv-Amir-Minimum.yml    2020-09-19 22:20:34,002 - CRITICAL - Tool-info module for ULTIMATE Automizer does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
22:20:33   bitvector/jain_1-1.yml                          2020-09-19 22:20:34,006 - CRITICAL - Tool-info module for ULTIMATE Automizer does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.
```

My changes to the MetaVal tool-info module would be as follows:
https://github.com/MartinSpiessl/benchexec/commit/41ed5f722426cc760be09fd17f03254221ca922e

But for `determine_result` we are also changing into the tool directory, and I cannot yet get rid of this with the changes proposed in this PR because there is no `determine_result` method in `Tool1To2` that accepts a `ToolLocator`

MartinSpiessl (changes requested):
I tried it out and had some difficulties, which I described in my comment in the code. There you will also find a link to what I changed in the MetaVal tool-info module in order to test this out.

skanav:
This feature can only be used with newer formats of tool info module. The challenges (also mentioned by @MartinSpiessl)  are:  i) I cannot use it with existing TI modules, and 2) `determine_result` might crash

I would also like to know something else also. At the moment if we have to execute a tool from another directory, we can simply add its path to `PATH`. Would adding `--tool-directory` to `PATH` caused some issue? 

Wouldn't it solve the problem we are trying to address?

benchexec/tooladapter.py
PhilippWendler: Ok, there is a misunderstanding that needs to be cleared up.

The code in `benchexec.tooladapter` is internal, not guaranteed to be a stable API, and should not be used by tool-info modules (cf. module doc in `benchexec.tooladapter`). Tool-info modules should use only stuff in `benchexec.tools.template`.

This means that `create_tool_locator` is irrelevant (of course I would not design a public API to look like this!).

Another thing is that yes, a tool directory can only be supported for tool-info modules with the new API. If it were possible to do it otherwise, we would have added this feature long ago and #592 would probably not be necessary.

For `determine_result`, why do you consider changing the current working directory to be necessary? And what would `determine_result` do with a `ToolLocator`? `determine_result` should not be searching for a tool's executable, the only method that should do this is `executable`.

PhilippWendler:
> I would also like to know something else also. At the moment if we have to execute a tool from another directory, we can simply add its path to `PATH`. Would adding `--tool-directory` to `PATH` caused some issue?
> 
> Wouldn't it solve the problem we are trying to address?

I am not sure I understand. Of course users can change `PATH` if they prefer instead of using `--tool-directory` as before, but it was requested by some people to make this more convenient.

In practice both ways also behave differently in some cases:
- With `PATH`, one needs to add the directory of the tool's executable. With `--tool-directory` one can either use the directory of the tool's executable or the tool's base directory (if the tool-info module is written properly). This can be easier to use and more intuitive if the executable is in a subdirectory like `bin`.
- With `--tool-directory`, it is guaranteed that the tool from that directory is benchmarked (if the tool-info module is written properly). With `PATH` if could happen that different instances/versions of the tool are all in directories listed in `PATH`, so one needs to make sure about the correct order of directories in `PATH`. And one needs to remember that `PATH` is checked before the current directory by `benchexec`.

benchexec/tooladapter.py
MartinSpiessl: > The code in benchexec.tooladapter is internal, not guaranteed to be a stable API

I see, since I probably do not want to translate tools myself anyway in order to be backwards compatible, I guess just checking whether the tool module I created is version 1 or 2 might be enough here. How do I check which version the `Tool` object is? Is there a convenient helper method already or do I need to is `isInstance` or so?

> Tool-info modules should use only stuff in benchexec.tools.template

Suggestion: document this requirement explicitly there (in the docs of `BaseTool{,2}`).

> For determine_result, why do you consider changing the current working directory to be necessary? 

You tell me:thinking:! We must have built this into the tool-info module of MetaVal for a reason. I cannot remember, but probably there is a tool that actually depends on this. Though I couldn't find one now. I could imagine that a tool-info module wants to support different versions of its tool with different output formats, and for this it would have to call `version` in order to get the version. The tool could of course just output the version string upon tool execution and identify it this way, but why do the hassle if there is already a `version` method? It does not explicitly state in the documentation of `determine_result` or `version` that `version` may not be used inside of `determine_result` or only be used inside of certain methods (if inside any at all?).

OK, I guess `version` is called with the `executable`, and that one we would be able to find with `--tool-directory` already. Still some tools might expect to be called from within their tool directory (terrible design but it is not ours to judge the tool author's design), so `_version_from_tool` might fail. There is no requirement that states that a tool that is integrated in BenchExec needs to be able to be called from outside its tool directory, so we cannot just assume this is not (or will never) be) the case. I guess with `Metaval` we just made it bullet proof, such that a wrapped tool could never possibly find anything that differs from the case where the tool is run directly.

PhilippWendler: > How do I check which version the `Tool` object is? Is there a convenient helper method already or do I need to is `isInstance` or so?

A helper method would not be more convenient than `isinstance(tool, BaseTool2)`.

> > For determine_result, why do you consider changing the current working directory to be necessary?
> 
> You tell me:thinking:! We must have built this into the tool-info module of MetaVal for a reason. I cannot remember, but probably there is a tool that actually depends on this. Though I couldn't find one now. I could imagine that a tool-info module wants to support different versions of its tool with different output formats, and for this it would have to call `version` in order to get the version. The tool could of course just output the version string upon tool execution and identify it this way, but why do the hassle if there is already a `version` method? It does not explicitly state in the documentation of `determine_result` or `version` that `version` may not be used inside of `determine_result` or only be used inside of certain methods (if inside any at all?).

Ok, `version` might be used in `determine_result`, yes. But I would think that it then makes more sense to store the version in a field instead of retrieving it in every `determine_result` call (which is typically called thousands of times).

And MetaVal would not need to handle this use case, because tool-info modules that conform to the new API also do not have a way to find the tool in `determine_result` when executed outside of MetaVal, so they need to adopt the above strategy anyway.

> Still some tools might expect to be called from within their tool directory (terrible design but it is not ours to judge the tool author's design), so `_version_from_tool` might fail. There is no requirement that states that a tool that is integrated in BenchExec needs to be able to be called from outside its tool directory, so we cannot just assume this is not (or will never) be) the case.

This is fully correct. However, if this is the case, then the tool-info module needs to account for that, so it (for example) needs to ensure the appropriate working directory when executing the tool for getting its version (note that changing the working directory is not necessary for this because nobody is forced to use `_version_from_tool`).

And again this is something that needs to be done by the tool-info module anyway to be usable outside of MetaVal, so MetaVal's tool-info module does not need to handle it.

MartinSpiessl: > Ok, version might be used in determine_result, yes. But I would think that it then makes more sense to store the version in a field instead of retrieving it in every determine_result call (which is typically called thousands of times). 

Isn't that discouraged because BenchExec can call a tool-info module concurrently. OK for the version I guess this will always be the same, but still in `MetaVal/CoveriTeam` for example we might have different versions of the same tool. But I guess then we would also have different `Tool` instances.

> This is fully correct. However, if this is the case, then the tool-info module needs to account for that, so it (for example) needs to ensure the appropriate working directory when executing the tool for getting its version (note that changing the working directory is not necessary for this because nobody is forced to use _version_from_tool).

I did not consider this, that makes perfect sense, thanks! So then it is the tool-info module's responsibility to detect that the executable that is passed to `version` is actually not in the current directory and that it has to change directory if necessary.

> And again this is something that needs to be done by the tool-info module anyway to be usable outside of MetaVal, so MetaVal's tool-info module does not need to handle it.

Cool, then I think this should work. That means we could already remove the directory changing around `determine_result` in the `MetaVal` tool-info module, because if something breaks I will just blame the wrapped tool-info module for not considering that the executable might be outside the current working directory.

MartinSpiessl (approved):

benchexec/tooladapter.py
PhilippWendler: > > Ok, version might be used in determine_result, yes. But I would think that it then makes more sense to store the version in a field instead of retrieving it in every determine_result call (which is typically called thousands of times).
> 
> Isn't that discouraged because BenchExec can call a tool-info module concurrently. OK for the version I guess this will always be the same, but still in `MetaVal/CoveriTeam` for example we might have different versions of the same tool. But I guess then we would also have different `Tool` instances.

You are right that one needs to be careful because some methods of the tool-info module can be called concurrently, for example cf. #166, but here it is safe and I think I will document this.

Would you update your prototype of the tool-info module with what we discussed and learned today to see if it now works nicely or if we missed something else?

MartinSpiessl: Yes, I can do this today in the evening. However there is not yet a tool-info module of the new format, so we cannot exactly see the logic unfold for tools that are based on the new base class.

MartinSpiessl: As promised:

https://github.com/MartinSpiessl/benchexec/commit/ff451ae3101e41eb440101cef3e0563a015f9fbc

Fun fact: I rediscovered which tool called their executable during `determine_result`, it is symbiotic, with the following call:
https://github.com/sosy-lab/benchexec/blob/d0a7f03de547491d35481ce71e96003f38c557f4/benchexec/tools/symbiotic.py#L109

It will break `MetaVal`-support  for `Symbiotic` in case `Symbiotic` updates to `BaseTool2` without fixing this first, since for tools that are based on `BaseTool2` my commit above does not switch directories anymore.

PhilippWendler: As I already said, a tool-info module like Symbiotic's need to fix that anyway when they switch to `BaseTool2` because it simply would not work without MetaVal as well (and they cannot even call `executable` because no instance of `ToolLocator` is available).

MartinSpiessl: Yes, I merely added that we know now that at least Symbiotic is such a tool, and that I am not planning on supporting this behavior in the future since it needs to be fixed by Symbiotic.

PR #604, opened 2020-09-24
Improve signature of determine_result and provide run options

PhilippWendler:
As requested in #166, some tool-info modules need the run command line in order to be able to properly parse the tool's output. This adds it together with a general improvement of what info is available to `determine_result`, like exit code and termination reason.

Differently from the discussion in #166 we provide not the `options` list that the `cmdline` method receives, but the full command line, similar to what is returned from `cmdline`. I think this is more helpful, and it should work equally well for the use cases mentioned in #166.

So far no other results like measurements are added because I see no need for that and it will be easy to add in the future (no breaking change due to parameter object).

The method `get_value_from_output` still does not get this information because it was never requested. But we pass the more convenient `RunOutput` object to it instead of a dumb list of strings for the output because we have it anyway.

@MartinSpiessl You requested this feature for MetaVal, is this implementation sufficient for you?

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -1072,7 +1072,7 @@ class Run(object):
         assert (
             self.runSet.benchmark.executable is not None
         ), "executor needs to set tool executable"
-        return cmdline_for_run(
+        self._cmdline = cmdline_for_run(
             self.runSet.benchmark.tool,
             self.runSet.benchmark.executable,
             self.options,
@@ -1082,6 +1082,7 @@ class Run(object):
             self.task_options,
             self.runSet.benchmark.rlimits,
         )
+        return self._cmdline
 
     def set_result(self, values, visible_columns={}):
         """Set the result of this run.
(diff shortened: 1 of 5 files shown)


MartinSpiessl (approved):
This has been less of an issue for MetaVal ever since we resorted to just output the needed information in stdout and read it back in from the tool output. Our use case/current solution is this:
```
verifierDir = None
        regex = re.compile("verifier used in MetaVal is (.*)")
        for line in output[:20]:
            match = regex.match(line)
            if match is not None:
                verifierDir = match.group(1)
                break
```

Now with the solution here I could instead apply a regex search to the command line or - even better - parse the command line with a new argument parser. Because we use `--` to mark where the options for MetaVal end and where the options for the wrapped tool begin, we would just add this utility method to our module:

```
_get_wrapped_verifier(self, cmdline)
    parser = argparse.ArgumentParser(add_help=False, usage=argparse.SUPPRESS)
    parser.add_argument("--verifier", required=True)
    (result,unprocessed) = parser.parse_known_args(cmdline)
    return result.verifier if verifier in result else None
```

This could maybe be generalized (`_get_option(self,cmdline,optionname)`) and provided by BenchExec via `template.py` if other tools need this as well.

Apart from this suggestion I see no actual problems with this PR :+1: 

(I only reviewed the changes in `template.py` because this is the only thing a tool-info module author should actually care about, so I just trust the rest of the changes to make the new API there work as intended.)

PhilippWendler:
Thanks. Utility methods can be added later if useful.

PR #606, opened 2020-09-25
using same name for both coveriteam TI modules

skanav:
minor change - using same name for both coveriteam TI modules


PhilippWendler (commented):
Technically fine, of course, and I cannot decide whether this is confusing for users because I do not know the usage scenarios.

skanav:
ready to merge?

PR #608, opened 2020-09-26
Fix behavior of method Task.input_files_or_identifier

kfriedberger:
I did not yet further test this fix. :-) Seems to work so far.


PhilippWendler:
The fix was correct, thanks, but the comment was both wrong (identifier cannot be a list) and not understandable without knowing what code there was before. So it needed to be removed.

kfriedberger:
Ok, valid point. You can just squash the commits then.
No need to have multiple single-line-changing commits.

PR #609, opened 2020-09-28
Fix missing folder while writing header to log

mmuesly:
I tried to use BenchExec for running some experiments, when I encountered the following error trace:
```
  File "../benchexec/bin/benchexec", line 27, in <module>
    sys.exit(benchexec.benchexec.main())
  File "../benchexec/bin/../benchexec/benchexec.py", line 478, in main
    sys.exit(benchexec.start(argv or sys.argv))
  File "../benchexec/bin/../benchexec/benchexec.py", line 73, in start
    rc = self.execute_benchmark(arg)
  File "../benchexec/bin/../benchexec/benchexec.py", line 343, in execute_benchmark
    output_handler = OutputHandler(
  File "../benchexec/bin/../benchexec/outputhandler.py", line 98, in __init__
    self.write_header_to_log(sysinfo)
  File "../benchexec/bin/../benchexec/outputhandler.py", line 307, in write_header_to_log
    self.txt_file = filewriter.FileWriter(txt_file_name, self.description)
  File "../benchexec/bin/../benchexec/filewriter.py", line 20, in __init__
    self._file = open(filename, "w")
FileNotFoundError: [Errno 2] No such file or directory:
```

It seems like the output handler writes to a folder before its creation. From my point of view, safeguarding the write command with the creation of parents folders, if they do not exist yet, seems a valid fix.

### DIFF
--- a/benchexec/filewriter.py
+++ b/benchexec/filewriter.py
@@ -4,6 +4,7 @@
 # SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
 #
 # SPDX-License-Identifier: Apache-2.0
+from pathlib import Path
 
 
 class FileWriter(object):
@@ -16,7 +17,7 @@ class FileWriter(object):
         The constructor of FileWriter creates the file.
         If the file exist, it will be OVERWRITTEN without a message!
         """
-
+        Path(filename).parents[0].mkdir(parents=True, exist_ok=True)
         self._file = open(filename, "w")
         self._pos = None
         self.append(content)


PhilippWendler:
Thanks for reporting this! Can you please give me an example command line that triggers this? I cannot reproduce this even if I specify `-o non-existing-folder/`. Furthermore, directly before the `write_header_to_log` call that is in the stacktrace the directory is created:
https://github.com/sosy-lab/benchexec/blob/master/benchexec/outputhandler.py#L95-L98

mmuesly:
Hi Philipp,

I see your point, but I am just describing, what I observe at my end.

I have build some custom task definitions and used BenchExec to run some experiments related to [jConstraints](https://github.com/tudo-aqua/jconstraints/tree/develop) on an Ubuntu machine. 

My command line call is `../benchexec/bin/benchexec --hidden-dir /home --numOfThreads 8 benchexec-defs/QF_LIA-short-cvc4.xml`

Eventually, this is related to some race condition resulting from the numerous threads and do not occur in a single threaded mode?

PhilippWendler:
Weird. The header is written before there are any other threads, and anyway, the `makedirs` call is definitely on the same thread as `write_header_to_log`.

Maybe you could add a `print(benchmark.log_folder)` to the code I linked above and a `print(filename)` to `FileWriter`? I would really like to understand the root cause instead of just fixing a symptom.

mmuesly:
The `print(benchmark.log_folder)` results in `results/QF_LIA-short-cvc4.2020-09-28_16-03-09.logfiles/`

And the filename in FileWriter is: `results/QF_LIA-short-cvc4.2020-09-28_16-03-09.results.smt/.short-qf_lia-cvc4.txt`

If this mismatch in the folder names is due to a configuration issue, I might change my run configuration, but once this step is passed, BenchExec works as expected with my configurations. But it is a clear mismatch in the folder name.


My git log is the following:
```
d13b2123 (HEAD -> jConstraintsRunner, origin/jConstraintsRunner) Add RESULT_SAT and RESULT_UNSAT
56114fb2 Fix missing folder while writing header to log
4158aa18 Fix typo
d02b486a jconstraints runner
355dcf92 (sosy/master) Fix behavior of method Task.input_files_or_identifier
10cb23ee Prepare version number for next development cycle.
76acd0af (tag: 3.3) Release 3.3
```

Where I adapted the commit as you requested:

```
commit 56114fb26502279b177347c76f62e865ad39fb5d
Author: Malte Mues <mail.mues@gmail.com>
Date:   Mon Sep 28 12:26:08 2020 +0200

    Fix missing folder while writing header to log

diff --git a/benchexec/filewriter.py b/benchexec/filewriter.py
index dee5ab56..4c6dc225 100644
--- a/benchexec/filewriter.py
+++ b/benchexec/filewriter.py
@@ -4,7 +4,7 @@
 # SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
 #
-
+from pathlib import Path
 
 class FileWriter(object):
     """
@@ -16,6 +16,9 @@ class FileWriter(object):
         The constructor of FileWriter creates the file.
         If the file exist, it will be OVERWRITTEN without a message!
         """
+        # Path(filename).parents[0].mkdir(parents=True, exist_ok=True)
+
+        print(filename)
 
         self._file = open(filename, "w")
         self._pos = None
diff --git a/benchexec/outputhandler.py b/benchexec/outputhandler.py
index be0c1b7a..b29d139e 100644
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -93,6 +93,8 @@ class OutputHandler(object):
 
         # create folder for file-specific log-files.
         os.makedirs(benchmark.log_folder, exist_ok=True)
+        print(benchmark.log_folder)
+
 
         self.store_header_in_xml(version, memlimit, timelimit, corelimit)
         self.write_header_to_log(sysinfo)
```

The other three commits ahead of current master are only concerned with my tool runner required for jConstraints and do not touch any files in BenchExec.

PhilippWendler:
> The `print(benchmark.log_folder)` results in `results/QF_LIA-short-cvc4.2020-09-28_16-03-09.logfiles/`
> 
> And the filename in FileWriter is: `results/QF_LIA-short-cvc4.2020-09-28_16-03-09.results.smt/.short-qf_lia-cvc4.txt`

Thanks, this clears it up. You have something like `<rundefinition name="smt/.short-qf_lia-cvc4">` in your `benchexec-defs/QF_LIA-short-cvc4.xml`.

Having a `/` in names of run definitions is something that was never considered. I would actually tend to forbid this, because while it seems to work at first glance with your fix, there are several problems with this:
- Result files will have unexpected names, for example, in your case `.short-qf_lia-cvc4.txt` (a hidden file).
- Result files of `benchexec` will be spread inconsistently over several directories.
- `table-generator` will not be able to find log files if given such results, and also log links in HTML tables are broken.
- Run-definition names like `a/../../b` would create even further problems, possibly even leading to unexpected file overwrites.
Most of these problems would persist even if we add a more extensive fix for this problem.

Would it be a problem for you to avoid `/` in names? Then I would rather add some checks that return a proper error message (and the same for names of `<tasks>` etc. as well).

mmuesly:
I am happy to avoid the `/`in the names.

If you could add an appropriate error message in the configuration parsing part of the code along with a validity check, this might help to avoid this problem in the future.

I have not realized that this might cause an error, but it works now. Thank you for the support!

PhilippWendler:
Definitely! I created #610 for this. Again thanks for bringing this up!

PR #612, opened 2020-09-29
Add tool info for Korn

gernst:
For more information, see https://github.com/gernst/korn

### DIFF
--- a/benchexec/tools/korn.py
+++ b/benchexec/tools/korn.py
@@ -24,7 +24,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         "eld.jar",
     ]
 
-    def executable(self,tool_locator):
+    def executable(self, tool_locator):
         return tool_locator.find_executable("run")
 
     def version(self, executable):
--- a/benchexec/tools/korn.py
+++ b/benchexec/tools/korn.py
@@ -12,7 +12,7 @@ import benchexec.result as result
 import benchexec.tools.template
 
 
-class Tool(benchexec.tools.template.BaseTool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
     Tool info for Korn, a software verifier based on Horn-clauses.
     URL: https://github.com/gernst/korn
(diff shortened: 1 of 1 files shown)


benchexec/tools/korn.py
```diff
@@ -0,0 +1,52 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+import re
```
PhilippWendler, line 9: Unused?

benchexec/tools/korn.py
```diff
@@ -0,0 +1,52 @@
...
+import benchexec.util as util
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool):
```
PhilippWendler, line 15: Would you mind switching to our new `BaseTool2` API? ([migration guide](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#migrating-tool-info-modules-to-new-api)) Should be only little effort and could for example make `determine_result` even easier.

benchexec/tools/korn.py
gernst: removed

benchexec/tools/korn.py
gernst: thanks, I've migrated it to the new API

benchexec/tools/korn.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def name(self):
+        return "Korn"
+
+    def determine_result(self, run):
+        """
+        This is literally the output from the underlying CHC solver
+        """
+
+        for line in run.output:
```
PhilippWendler, line 41: This could be simplified, `strip()` is probably not necessary anymore except if there is whitespace before or after the result on the line. And you could even do `if "unsat" in run.output:`. But current code also works.

PhilippWendler (approved):

benchexec/tools/korn.py
gernst: right, but `"sat" in run.output` would neet to follow the test for `"unsat"` and that seems not very robust against changing the order of statements ;) the `strip()` was in fact necessary when I ran the experiments on Monday

gernst:
thanks for merging!

PR #615, opened 2020-10-02
Add tool info for the witness linter.

SvenUmbricht:
This should probably not be merged before https://github.com/sosy-lab/sv-witnesses/pull/16.

### DIFF
--- a/benchexec/tools/witnesslint.py
+++ b/benchexec/tools/witnesslint.py
@@ -21,8 +21,14 @@ class Tool(benchexec.tools.template.BaseTool2):
     def name(self):
         return "witnesslint"
 
+    def version(self, executable):
+        version_string = self._version_from_tool(executable)
+        return version_string.partition("version")[2].strip().split(" ")[0]
+
     def determine_result(self, run):
         if run.exit_code.value == 0:
             return result.RESULT_TRUE_PROP
-        else:
+        elif run.exit_code.value == 1:
             return result.RESULT_FALSE_PROP
+        else:
+            return result.RESULT_ERROR
--- a/benchexec/tools/witnesslint.py
+++ b/benchexec/tools/witnesslint.py
@@ -1,0 +1,21 @@
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for the witness checker (witnesslint)
+    (https://github.com/sosy-lab/sv-witnesses)
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("witnesslint.py")
+
+    def name(self):
+        return "witnesslint"
+
+    def determine_result(self, run):
+        if run.exit_code.value == 0:
+            return result.RESULT_TRUE_PROP
+        else:
+            return result.RESULT_FALSE_PROP
--- a/benchexec/tools/witnesslint.py
+++ b/benchexec/tools/witnesslint.py
@@ -1,3 +1,10 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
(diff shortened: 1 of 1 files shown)


benchexec/tools/witnesslint.py
```diff
@@ -0,0 +1,21 @@
+import benchexec.result as result
```
PhilippWendler, line 1: A license and copyright declaration is needed. In your case, as you are an employee of SoSy-Lab, our standard header applies, you can just copy it from any of the BenchExec files.

benchexec/tools/witnesslint.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Tool info for the witness checker (witnesslint)
+    (https://github.com/sosy-lab/sv-witnesses)
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("witnesslint.py")
+
+    def name(self):
+        return "witnesslint"
```
PhilippWendler, line 15: If the tool has a nicer human-readable name, you can return that here. Or keep it as it is.

benchexec/tools/witnesslint.py
```diff
@@ -0,0 +1,21 @@
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
```
PhilippWendler, line 5: This module does not override `version`, which means that users will not see in the result tables which tool version was used. This can be confusing and hinders reproducibility.

benchexec/tools/witnesslint.py
```diff
@@ -0,0 +1,21 @@
...
+        return tool_locator.find_executable("witnesslint.py")
+
+    def name(self):
+        return "witnesslint"
+
+    def determine_result(self, run):
+        if run.exit_code.value == 0:
+            return result.RESULT_TRUE_PROP
+        else:
+            return result.RESULT_FALSE_PROP
```
PhilippWendler, line 21: This always returns `RESULT_FALSE_PROP`, even in cases where the tool crashes etc. Is there a way to distinguish these cases, e.g., is there a specific return code for `false`? If yes, checking for it here would make the tool-info more useful.

PhilippWendler (changes requested):

PhilippWendler:
Note that this can be merged even before your tool is ready or merged somewhere else. However, the interface of your tool should be stable (e.g., how it is called and what it returns) such that the tool-info module won't need changes soon. Please mark the PR as draft or not accordingly.

benchexec/tools/witnesslint.py
SvenUmbricht: Added.

benchexec/tools/witnesslint.py
SvenUmbricht: There's no other name for the tool. Should be fine.

benchexec/tools/witnesslint.py
SvenUmbricht: The linter can now provide version info via the `--version` flag -> added `version` to tool-info module.

benchexec/tools/witnesslint.py
SvenUmbricht: The exit code of the linter now reflects whether the given witness is faulty or whether an internal error occurred -> adjusted `determine_result` accordingly.

PhilippWendler (commented):
Looks good.

Will the "main name" under which this tool is advertised be witnesslint or witnesslinter? If the former, I suggest to adjust the comment and the result of `name()`.

SvenUmbricht:
The tool is a linter for witnesses but the tool name is "witnesslint". I had to rename the executable to "witnesslinter" to avoid clashes with the package name, but if that makes it too confusing an `__init__.py`  file in the witnesslint package should work, too.

PhilippWendler:
Not sure about what you suggest as alternative, but I assume this means that tool name and script name will stay as they are? Then I will merge this.

SvenUmbricht:
Yes, I would propose to leave the name as-is.

PhilippWendler (approved):

Issue #125, opened 2016-02-09
Update statistics in HTML tables when filtering rows


PhilippWendler:
The statistics in the footer of the HTML tables should get automatically recomputed after the row filter was changed. For this the raw data values need to be added to the HTML, possibly as additional `data` attributes to the table cell tags.

The statistics computation inside the table-generator should still stay such that the statistics are also available if JavaScript is blocked in the browser, and to not delay loading the table further.

PhilippWendler:
There is the question how to show the filtered statistics, there are two possibilities:

1. If a filter is defined, *replace everything* that is shown in the statistics with statistics over the filtered data, and replace the text `total results` with something like `filtered tasks`. The row `local summary` would be hidden in this case (because it is not possible to show filtered data in it.)
    *Advantage*: Consistent with other tabs (table view, plots) which also have this behavior of using only the filtered data.
    *Disadvantage*: Potentially unexpected for long-time users of BenchExec that expect statistics to be unfiltered (could confuse filtered with non-filtered statistics). Only one kind of statistics (filtered/unfiltered) shown at a given time.
2. If a filter is defined, insert an *additional row* with the statistics data over the filtered data (without breakdown into correct/incorrect etc.) and keep all the other rows as they are:
    ![grafik](https://user-images.githubusercontent.com/2545335/152316960-df0e9a1f-8d2d-4f66-b7cf-97fa536553a7.png)
    *Advantage*: Does not confuse long-time users, shows both filtered and unfiltered statistics at the same time.
    *Disadvantage*: No breakdown of filtered statistics into correct/incorrect etc.

[Previously](https://github.com/sosy-lab/benchexec/pull/616#pullrequestreview-638682810) I was in favor of option 2, but now I am not so sure anymore. Option 1 does seem more consistent and intuitive overall (disregarding the old behavior of the statistics table), so maybe it is the better long-term solution.

Note that the other arguments about which statistics are visible are not so critical: For Option 1, users can easily toggle between filtered and unfiltered statistics by deleting the filters (2 clicks) and resurrecting them (back button). There has also never been a request for seeing unfiltered plots while filters are defined. For Option 2, users can manually request the desired statistics with additional filters, though if you want to get them for many run sets, this can involve some clicking.

Could you please give feedback?
Vote :rocket: for Option 1 and :tada: for Option 2. Of course, also feel free to reply with a comment.

dbeyer:
I am in favor of option 1.

[...]

PR #616, opened 2020-10-05
Async stat recalculation

DennisSimon:
Hi @PhilippWendler , 

i can finally open the PR for the automated calculation of stats. 

This PR includes a Controller for pooling and dispatching of WebWorkers, a Stat calculation WebWorker, 
a number formatter that performs the same type of formatting as in the table-generator python code. 

Over the course of the day i will push some commits that will clean some artifacts that are left over from 
the development process. 

Regarding the updates in the snapshot, i would suggest we can go over them in our call today? 

I will add some more details to this PR over the course of the day


benchexec/tablegenerator/react-table/.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
PhilippWendler, line 1: Is this related to the changes here? If not, I suggest to remove it (and add it separately if desired).

benchexec/tablegenerator/react-table/src/utils/stats.js
```diff
@@ -0,0 +1,184 @@
+import { isNil, NumberFormatterBuilder } from "./utils";
```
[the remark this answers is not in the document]
DennisSimon: Will update it

benchexec/tablegenerator/react-table/src/tests/Summary.test.js
```diff
@@ -8,17 +8,31 @@
...
-    selectColumn={overview.toggleSelectColumns}
-    stats={overview.stats}
-    prepareTableValues={overview.prepareTableValues}
-    changeTab={overview.changeTab}
-    hiddenCols={overview.state.hiddenCols}
-  />
-));
+// mock uniqid to have consistent names
+// https://stackoverflow.com/a/44538270/396730
+jest.mock("uniqid", () => (i) => i + "uniqid");
```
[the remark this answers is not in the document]
DennisSimon: Awesome, i will remove it then

benchexec/tablegenerator/react-table/.vscode/launch.json
DennisSimon: I added it at the same time as the new `test:debug` script in the package.json. 
I can commit them to master separately though if you would prefer that.

benchexec/tablegenerator/react-table/src/components/Summary.js
```diff
@@ -138,17 +155,101 @@ export default class Summary extends React.Component {
...
+      return row;
+    });
+
+    this.setState({
+      stats: transformed,
+      statUpdateCycle: this.state.statUpdateCycle + 1,
+    });
+  }
+
+  async updateStats(oldStats) {
```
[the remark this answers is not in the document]
DennisSimon: Yeah i was planning on doing that anyway. 

The structure of the whole flow was very volatile so i wanted to wait for a more stable state before i add that. 
I will go through the flow start to end to add detailed explanations about each steps and input/outputs

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -8673,7 +8673,7 @@ exports[`Render Summary for multi-table-only-columns.table.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"212"}
-         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.79869280115049655695465610"
+         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.7986928011505"
```
DennisSimon, line 8676: As no number of significant digits was given the value was rounded to javascripts default precision of 15 digits. 
@PhilippWendler is this an issue?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -8849,7 +8849,7 @@ exports[`Render Summary for multi-table-only-columns.table.html 1`] = `
...
```
DennisSimon, line 8852: As no number of significant digits was given the value was rounded to javascripts default precision of 15 digits.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -10463,7 +10463,7 @@ exports[`Render Summary for multi-table-with-columns.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"2.90"}
-         title="avg: 1.45, max: 1.49, median: 1.45, min: 1.41, stdev: 0.041"
+         title="avg: 1.45, max: 1.49, median: 1.45, min: 1.41, stdev: 0.0410"
```
DennisSimon, line 10466: The column has three significant digits, that's why the 0 is included

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -10570,7 +10570,7 @@ exports[`Render Summary for multi-table-with-columns.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={".036"}
-         title="avg: 0.012, max: 0.015, median: 0.015, min: 0.006, stdev: 0.00424"
+         title="avg: 0.0120, max: 0.015, median: 0.015, min: 0.006, stdev: 0.00424"
```
DennisSimon, line 10573: The column has three significant digits, that's why the 0 is included. 
(max, median and min don't have a trailing zeroes as they were taken directly from the dataset. 
Should they also be padded with zeroes then @PhilippWendler  ?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -10639,7 +10639,7 @@ exports[`Render Summary for multi-table-with-columns.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={".331"}
-         title="avg: 0.166, max: 0.167, median: 0.166, min: 0.164, stdev: 0.0015"
+         title="avg: 0.166, max: 0.167, median: 0.166, min: 0.164, stdev: 0.00150"
```
DennisSimon, line 10642: The column has three significant digits, that's why the 0 is included

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -11890,7 +11890,7 @@ exports[`Render Summary for multi-table-with-columns.table.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={"5.71"}
+         dangerouslySetInnerHTML={"5.72"}
```
DennisSimon, line 11893: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -11910,7 +11910,7 @@ exports[`Render Summary for multi-table-with-columns.table.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"212"}
-         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.79869280115049655695465610"
+         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.7986928011505"
```
DennisSimon, line 11913: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -12086,7 +12086,7 @@ exports[`Render Summary for multi-table-with-columns.table.html 1`] = `
...
```
DennisSimon, line 12089: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -13719,7 +13719,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.diff.html 1`] = `
...
```
DennisSimon, line 13722: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -13699,7 +13699,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.diff.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={"5.71"}
+         dangerouslySetInnerHTML={"5.72"}
```
DennisSimon, line 13702: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -13895,7 +13895,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"212"}
-         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.79869280115049655695465610"
+         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.7986928011505"
```
DennisSimon, line 13898: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -15127,7 +15127,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.table.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={"5.71"}
+         dangerouslySetInnerHTML={"5.72"}
```
DennisSimon, line 15130: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -15147,7 +15147,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.table.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"212"}
-         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.79869280115049655695465610"
+         title="avg: 53, max: 212, median: 0, min: 0, stdev: 91.7986928011505"
```
DennisSimon, line 15150: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -15323,7 +15323,7 @@ exports[`Render Summary for multi-table-with-diff-over-column.table.html 1`] = `
...
```
DennisSimon, line 15326: See comments of multi-table-only-columns.table.html

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -19752,7 +19752,7 @@ exports[`Render Summary for nan_and_inf.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"NaN"}
-         title="avg: NaN, max: Inf, median: 1, min: -Inf, stdev: NaN"
+         title="avg: NaN, max: Inf, median: NaN, min: -Inf, stdev: NaN"
```
DennisSimon, line 19755: Snippet from `statistics.py` 
```
        if min_value == -inf and max_value == +inf:
            values_sum = nan
            mean = nan
            stdev = nan
```

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -19790,7 +19790,7 @@ exports[`Render Summary for nan_and_inf.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"-Inf"}
-         title="avg: -Inf, max: 0, median: 0, min: -Inf, stdev: Inf"
+         title="avg: -Inf, max: 0, median: -Inf, min: -Inf, stdev: NaN"
```
DennisSimon, line 19793: Snippet from `statistics.py`
```
        elif min_value == -inf:
            values_sum = -inf
            mean = -inf
            stdev = inf
```

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -24669,8 +24671,9 @@ exports[`Render Summary for rows-with-scores.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={0}
-        />
+        >
+         -
```
DennisSimon, line 24675: @PhilippWendler  is the expected behaviour to show `-` or `0` ?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -8653,7 +8653,7 @@ exports[`Render Summary for multi-table-only-columns.table.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={"5.71"}
+         dangerouslySetInnerHTML={"5.72"}
```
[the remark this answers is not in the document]
PhilippWendler: Seems like a bug in the original rounding code, we have to investigate.

PhilippWendler:
So, I finally have an update. The Python code for number rounding and formatting got improved in #654 and there are several output changes (fixes) due to this.

One fix (b8fddc2) is for the incorrect rounding that you discovered, so the Python code now produces the same result as your code.

One improvement (34ad7b7) is for the alignment of values in the summary table, which is now computed from the values in that table instead of the values from the main table. I think your code already does that and so some differences should go away due to this.

One improvement (5a53125) is for the rounding of avg and stdev in the tooltips of columns with type `count` and when no significant digits are specified. Then we now round to 2 decimal(!) digits. Probably your code does not do this yet, but I guess it should be feasible to add this. (Other cases are unchanged: when significant digits are given, we round using them, and for type `measure` we also always round according to significant digits.)

One fix (445edd9) avoids truncation of very small values. This occured only in test that I just added as well, so we have to see whether your code already handles this correctly or needs fixes.

There were also some fixes regarding trailing zeros, but I am not sure whether these resolve the differences that you saw.

For full details, check the PR description of #654 and the individual commit messages. I made sure to update tests along with code so you can see which commits actually lead to output changes.

Next steps would be to merge master into your branch, regenerate all snapshots, and go through the remaining differences again.

DennisSimon:
@PhilippWendler Awesome news!
 I have merged master and will start implementing the required changes over the cause of the week.

PhilippWendler:
Another change to the statistics was made in bd8e09b: sums over lists without any values now produce `null` (rendered as dash), not 0 anymore.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -3424,8 +3425,9 @@ exports[`Render Summary for big-table.diff.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={0}
-        />
+        >
+         -
```
DennisSimon, line 3429: @PhilippWendler Theses were generated in the status columns. The python code still displays a "0" instead of a dash when it doesn't have any values in the status columns. I believe this is a bug? Are they handled separately in the python code?

PhilippWendler: No, this should stay as it is. The status column *counts* the rows for the respective category, and the count over an empty list is of course 0. The other columns *sum* the values in the rows of the respective category, and there we can decide do have no result for a sum of an empty list to distinguish from the case where the sum happens to be 0. For the status column there are no such two cases that could be distinguished.

DennisSimon: Makes sense, i will adjust the code accordingly.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -10878,9 +10876,8 @@ exports[`Render Summary for multi-table.table.html 1`] = `
        >
         <div
          className="cell"
-        >
-         -
```
DennisSimon, line 10882: These are status cells that are in `local summary`. I believe they should also be `0` as the rows are being counted, right

DennisSimon: @PhilippWendler  :point_up:

PhilippWendler: No, `0` would actually be wrong here, because the local summary is shown only if there are such statistics, and of course these do not exist for only an empty set of rows. The number here should actually be always equal to the total number of results, but we don't show it because it would be redundant.

DennisSimon: Well, we definitely have rows in the local summary in this case. Do they maybe not have any status in the local summary? 
Here is a screenshot for context (the `0` in status comes from my current changes) 
Screenshot from 2020-12-11 09-11-56

DennisSimon: Yes, i took a step back and read through the code again. 
The issue as missing differentiation between `undefined` and `null`. 
`undefined` represents a missing result caused by no available data, and usually gets represented by a `0` for status values. 
`null` gets passed on by the stats table and makes the code default to `-`

PhilippWendler: Ok, but I would suggest to not use `undefined` *and* `null`, this gets highly confusing. The code that counts the rows per category for the status column should simply return 0.

DennisSimon: Yeah, this is caused by the way that the stats are prepared. The categorization is done on the fly and buckets (like `correct-true` and so on) will only be created if we have lines matching this criteria. This is why some cells in the stats table will receive a `undefined` value if there are no matching lines for this category. 
Actually, `local summary` is the only edge-case where we need to differentiate between these two values as this is the only place where we retrieve values directly from the `stats` object of the data-set.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -7305,7 +7305,7 @@ exports[`Render Summary for multi-table.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"134"}
-         title="avg: 134, max: 134, median: 134, min: 134, stdev: 0"
+         title="avg: 134, max: 134, median: 134, min: 134, stdev: 0.00"
```
DennisSimon, line 7308: @PhilippWendler quick question. As `stdev` is a calculated value, should we format it to match the number of significant digits?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -14714,7 +14714,7 @@ exports[`Render Summary for multi-table-only-columns.diff.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"2.90"}
-         title="avg: 1.45, max: 1.49, median: 1.45, min: 1.41, stdev: 0.041"
+         title="avg: 1.45, max: 1.49, median: 1.45, min: 1.41, stdev: 0.0410"
```
DennisSimon, line 14717: @PhilippWendler Also here: The last zero is truncated even though we have three significant digits, is this expected behaviour?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: Yes, except for columns of type `count` without significant digits.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: This is probably the result of never having more than 3 decimal digits in the input in this column?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -59000,7 +59000,7 @@ exports[`Render Summary for simple-table-with-scaling.table.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"11000000000"}
-         title="avg: 2200000000, max: 2300000000, median: 2200000000, min: 2100000000, stdev: 72000000"
+         title="avg: 2200000000, max: 2300000000, median: 2200000000, min: 2100000000, stdev: 63000000"
```
DennisSimon, line 59003: @PhilippWendler I'm a bit confused about this change. Let's go over this and the other topics in our call later

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: This is now fixed on `master`, though the actual result is now `0.000`.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: No, it was a bug, this is now fixed on `master`.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: The correct (non-rounded) raw values are now added to the JSON on `master`, so this should be fixed after merging.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -14820,8 +14820,8 @@ exports[`Render Summary for multi-table-only-columns.diff.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={".036"}
-         title="avg: 0.012, max: 0.015, median: 0.015, min: 0.006, stdev: 0.00424"
+         dangerouslySetInnerHTML={".036&#x2007;"}
+         title="avg: 0.0120, max: 0.015, median: 0.015, min: 0.006, stdev: 0.00424"
```
DennisSimon, line 14824: As `avg` is calculated, should we fill missing significant digits with `0` here?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -15025,7 +15025,7 @@ exports[`Render Summary for multi-table-only-columns.diff.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={".029"}
+         dangerouslySetInnerHTML={".0290"}
```
DennisSimon, line 15028: Does `sum` count as a calculated value and should be padded?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -19882,8 +19882,8 @@ exports[`Render Summary for multi-table-only-columns.table.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={".000"}
-         title="avg: 0.000, max: 0.000, median: 0.000, min: 0.000, stdev: 0.000"
+         dangerouslySetInnerHTML={"&#x2008;&#x2007;&#x2007;&#x2007;&#x2007;"}
```
DennisSimon, line 19885: A sum of `0` values should be represented as `0`, right?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: It would be better to not do this.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: Similarly to `avg`, it would be better to not add more digits than in the inputs.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: Is the question about `0` vs. `.000`? Would depend on the circumstances. For type measure, the latter would also be natural.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
DennisSimon: Urrgghhh go home javascript, you are drunk
```
> 0.018+0.011
0.028999999999999998

```

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -52612,7 +52612,7 @@ exports[`Render Summary for rows-with-scores.html 1`] = `
        >
         <div
          className="cell"
-         dangerouslySetInnerHTML={".128&#x2007;&#x2007;"}
```
DennisSimon, line 52615: Seems like we are padding with whitespaces to match the maximum length of the values from the "values table" here.
I believe we should determine the width of the values using the displayed `sum` values of the corresponding column of the statistics table.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
```diff
@@ -53502,7 +53502,7 @@ exports[`Render Summary for rows-with-scores.html 1`] = `
         <div
          className="cell"
          dangerouslySetInnerHTML={"0&#x2008;&#x2007;&#x2007;&#x2007;&#x2007;"}
-         title="avg: 0, max: 0, median: 0, min: 0, stdev: 0"
+         title="avg: 0, max: 0, median: 0, min: 0, stdev: 0.000"
```
DennisSimon, line 53505: Sorry for re-iterating over this again but could you outline the rules for padding stdev again? 
We only do it for some column types, right?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: Does this occur only for `rows-with-scores.html`? Then I suspect it is just because this manually created file is wrong.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: If significant digits are given, we always use them. If there are none given, and this is a measure column, we use a default of 3 significant digits. If it is a count column, we use a default of 2 decimal digits.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
DennisSimon: Hi Philipp, thanks for the response. 
The issue appears in `rows-with-scores.html` and `test.2015-03-03_1613.results.predicateAnalysis.custom-score.html`. 
Does the latter one also have a manually created part?

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
DennisSimon: Okay, gotcha. In this case, the column has a `number_of_significant_digits` of 3, so i understand the new result `0.000` would be correct?
Again, this (and all other currently failing snapshots) appear in `rows-with-scores.html` and `test.2015-03-03_1613.results.predicateAnalysis.custom-score.html´, so i don't know how much manual creation might have influenced this result.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: `test.2015-03-03_1613.results.predicateAnalysis.custom-score.html` was not created manually, but there was a (different) bug leading to incorrect alignment, and I fixed it in a930f1f0. So please merge and then check again. For `rows-with-scores.html`, just fix the JSON in it manually.

benchexec/tablegenerator/react-table/src/tests/__snapshots__/Summary.test.js.snap
PhilippWendler: Are you sure this happens with `test.2015-03-03_1613.results.predicateAnalysis.custom-score.html`? I saw only `0.000` as stdev in it when I looked at it in my browser. If there are wrong values, I guess the easiest way to show me would be a comment for the respective line on GitHub.

For `rows-with-scores.html`, as usual please just fix the JSON.

DennisSimon:
The Summary snapshot is now equal to the one on master.

lachnerm:
As stated in #712, this PR will also fix #672.

DennisSimon:
I ran some benchmarking on a table with ~11500 tasks. Calculation takes ~750ms on average for all tasks and all rendered statistics.
When the user filters data, the underlying dataset will decrease and statistics calculation will complete faster.

DennisSimon:
Calculation times are sub-second and only increase slightly for bigger tables:
execution_time_stats
I will address the conflicts today and then suggest we can finally merge this PR :)

benchexec/tablegenerator/react-table/.vscode/launch.json
PhilippWendler: Seems my upvote was not noticed. Yes, please, separate these changes.

benchexec/tablegenerator/react-table/src/workers/dataUrls.js
```diff
@@ -0,0 +1,11 @@
+// This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: This file is checked in, but is generated and should only be necessary during the build process. Can we simply avoid checking it in, like we do with for example `dependencies.json`?

This would also mean we do not need to care about a proper license header.

If it is not checked in, we should maybe move it to the `build/` directory.

benchexec/tablegenerator/react-table/src/workers/scripts/stats.worker.js
```diff
@@ -0,0 +1,392 @@
...
+    const valuesHavePosInf = Number(total.max) === Infinity;
+    values.stdev = calculateStdev(
+      valuesHaveNegInf,
+      valuesHavePosInf,
+      values.variance,
+      values.items.length,
+    );
+
+    for (const [key, val] of Object.entries(values)) {
+      if (key === "meta") {
```
PhilippWendler, line 360: What is the meaning of `meta` here? It is not a common term in BenchExec, and it did not find documentation. Because the term is so vague we might also want to rename it (depending on what it is used for).

benchexec/tablegenerator/react-table/src/workers/scripts/stats.worker.js
```diff
@@ -0,0 +1,392 @@
+// This file is part of BenchExec, a framework for reliable benchmarking:
+// https://github.com/sosy-lab/benchexec
+//
+// SPDX-FileCopyrightText: 2019-2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// COPY OF utils.js, as imports will not work here
```
PhilippWendler, line 8: Hm, this seems error prone. At the very least, there must be a similar comment at the original location. But can we do better? For example, move common code to a separate file and include both in the worker URI (the code for this already seems to support an arbitrary number of files)?

PhilippWendler (commented):
Thanks! I agree that we can move forward with just using the JS computed statistics. Before merging, I would like to discuss a few things:

### Filtered Statistics

Right now, filtering rows just updates the statistics transparently. This is nice and consistent with the rest of the application, but might be unexpected for experienced BenchExec users. Furthermore, it hides the total statistics. I see several options:
1. Never update the standard statistics, just add one statistics row "filtered" (without breaking it further down into "correct"/"wrong" etc. because people who use filters to calculate the desired statistics could add a filter for correct results if desired). This is the least confusing for existing users, but might not be the best way going forward.
2. Rename "total" to "filtered" when a filter is applied and hide row "local summary" (because filters don't apply to it). This is also consistent, and showing the total statistics while a filter is used is not important, because users can delete filters with two clicks (and restore them with one click using the back button).
3. Like option 2, but additionally hide the breakdown rows for "correct"/"wrong" etc. As mentioned above, these are not so important when using filters, and removing them would create a bigger visual distinction between filtered and standard view.

### Calculation Info
From the code it was not obvious to me what the user will see in the time period while the statistics are being calculated. Could you please comment on this?

### Python Statistics Code
The statistics that are added to the data JSON from Python are not used at all anymore, correct? So we should either remove the respective Python code or use it for integration tests that continue to check whether both statistics implementations compute the same output.

If we decide to remove it, we should do it as part of this PR. If not, the test infrastructure could be added later. But the decision might depend on how complex it would be to add such tests. Maybe we add a flag to `table-generator` to optionally keep adding the statistics, and then we write JS tests that compare them? Would that be a lot of effort?

If we decide to keep the JS rendering code for the precomputed statistics in addition to the Python code for computing them, we might be able to use snapshot tests to do this kind of double check?

DennisSimon:
Hi @PhilippWendler , sorry for the late update. I will finally have some to work on this on the weekend.

DennisSimon:
I have updated the code to only display filtered statistics in a separate row in the statistics table. 
I will work on the rest over the week. 

Regarding the removal of the python statistics: I agree, the only issue that i see is that we can not calculate the local summary in the application. Maybe we should keep at least that information in json?

PhilippWendler:
> I have updated the code to only display filtered statistics in a separate row in the statistics table.
> I will work on the rest over the week.
> 
> Regarding the removal of the python statistics: I agree, the only issue that i see is that we can not calculate the local summary in the application. Maybe we should keep at least that information in json?

That information should be kept, yes. But there is actually no statistical information in this row, just some measured numbers.

Still unsure whether we should remove the Python code, though.

benchexec/tablegenerator/react-table/src/workers/dataUrls.js
DennisSimon: I agree, as the file is created on every build, it can be removed. 
I've implemented the automated creation of the license header anyways.

benchexec/tablegenerator/react-table/src/workers/scripts/stats.worker.js
DennisSimon: The idea of moving the code into the common section of the dataURIs (at least thats how i interpret the comment) is an interesting idea. 
I am however not sure if this will work, as the dataURI should be loaded as a worker without any context/closures as it is basically just text. 

I see two options: Either we apply the same comment in the areas that were copied over into worker files or we extend the dataURI creation script to inject worker-global code into the dataURIs while building.

One issue that i see with injecting worker-global code into the dataURIs is that we might clash with functions/variables defined in the worker because we are suddenly shadowing objectsthat were not previously present. We could try to mitigate this issue with some clever usage of JSDoc but, depending on the IDE and the used extensions, this might still result in problems. 

What are your thoughts on that? 
For now i will simply add another comment in the utils script

benchexec/tablegenerator/react-table/src/workers/scripts/stats.worker.js
DennisSimon: `meta` stores meta-information (like `maxDecimals` and `type`) of the current bucket, which will be used by the number formatter. 
I will add additional documentation.

DennisSimon:
> > Regarding the removal of the python statistics: I agree, the only issue that i see is that we can not calculate the local summary in the application. Maybe we should keep at least that information in json?
> 
> That information should be kept, yes. But there is actually no statistical information in this row, just some measured numbers.
> 
> Still unsure whether we should remove the Python code, though.

So if the python code doesn't add too much overhead then my feeling is that we can leave it in (at least for now), as it can be used to display the initial values. 
While it does bloat the data.json additionally a bit, the stored size of the statistics table is relatively small compared to the main table.
I don't really have any strong opinions on that topic though, i think both options are fine.
Your thoughts?

benchexec/tablegenerator/react-table/src/workers/scripts/stats.worker.js
PhilippWendler: I understand. I might have suggested simply concatenating global stuff with the worker code, but silent shadowing seems like a bad thing to debug. So I don't have a better solution than adding comments.

PhilippWendler:
Thanks for the commits! My current opinion is to keep the Python code until we notice (if ever) that it causes more trouble than it is worth. I have asked @lachnerm to think about snapshot tests that check both the numbers calculated by the Python code as well as the numbers calculated by the JS code against the same snapshot to make sure they stay in sync.

lachnerm:
> I have asked @lachnerm to think about snapshot tests that check both the numbers calculated by the Python code as well as the numbers calculated by the JS code against the same snapshot to make sure they stay in sync.
>
I've been working on this currently, but there is a problem.
I've set up new tests that create the Summary tab for both the Python and JS stats and check whether the corresponding snapshots are identical. To do so, one time a `onStatsReady` function is passed and one time there isn't. This leads to the JS stat computation in the first case and to presenting the old Python stats in the second one. However, for the former, there is an additional row called `filtered` added to the table, which is skipped if the JS stat computation is not executed, i.e. the snapshots for both cases can't be identical. 
So either the code needs to be refactored such that the `filtered` row is added independent of which stats are being used, or we need another way to test this than using snapshots. What should we do about this?

PhilippWendler:
The "filtered" row should only be shown in the stats if a filter has been applied. I guess this would solve this? If not, simply add an internal parameter to have filtered statistics nor not.

lachnerm:
> The "filtered" row should only be shown in the stats if a filter has been applied. I guess this would solve this? If not, simply add an internal parameter to have filtered statistics nor not.

I see, this should already solve the problem, yes. However, currently the "filtered" row seems to be shown by default, so we'd need to fix this first. Is @DennisSimon going to take care of this, or should I fix it?

DennisSimon:
Hi @lachnerm , @PhilippWendler . I have fixed the code so that the "filtered" row is only displayed if a filter is set.

lachnerm:
> Hi @lachnerm , @PhilippWendler . I have fixed the code so that the "filtered" row is only displayed if a filter is set.

Yes, looks good! I'm still having a bit of troubles with the snapshot tests and I noticed something that I'd like to confirm first.

> ### Python Statistics Code
> The statistics that are added to the data JSON from Python are not used at all anymore, correct? So we should either remove the respective Python code or use it for integration tests that continue to check whether both statistics implementations compute the same output.
> 

If I understand correctly, the `stats` part of the `data.json` shouldn't be used for the table any more and all necessary data is instead calculated by the JS code. Does this account for the initial render as well?
I've noticed that when I make manual changes to the values in the `stats` part of our dev data, they are shown in the table. On the other hand, if I make manual changes to the values calculated via JS (I assume this is the `res` variable in the `updateStats` function of the `StatisticsTable` component, right?), they are never shown in the table. Is this the expected behavior?
Maybe I'm also just changing the values at the wrong location in our code, since I don't have a full overview of all changes of this PR yet. If so, where would be the correct location to manually meddle with the values shown in the table? (For testing purposes only of course).

PhilippWendler:
> > ### Python Statistics Code
> > The statistics that are added to the data JSON from Python are not used at all anymore, correct? So we should either remove the respective Python code or use it for integration tests that continue to check whether both statistics implementations compute the same output.
> 
> If I understand correctly, the `stats` part of the `data.json` shouldn't be used for the table any more and all necessary data is instead calculated by the JS code. Does this account for the initial render as well?

That has changed, and the new JS code should only calculate the filtered row (except in the snapshot tests, where it should compute everything).

As long as we keep the Python code for statistics, we can also use its output save a little bit of time for users.

lachnerm:
Alright, I've finally finished the snapshot tests. 
I wasn't able to find a good and working solution to use one snapshot for two tests. Therefore, I had to change the setup a bit and used 3 tests for each test file:

- Snapshot test for the `StatisticsTable` using the python-computed stats
- Snapshot test for the `StatisticsTable` using the js-computed stats
- Test that compares the above two components after being JSON serialized

The third test doesn't really give a helpful error on what exactly is different, so in this case the other two snapshot tests will be quite helpful. It's still possible to directly identify which computation (or which values) is different from before due to the snapshots.

Since the current JS code doesn't compute all stats any more, I was reusing some code from a previous commit. It would be helpful if @DennisSimon could take a look on whether this makes sense, since I don't really have a good overview of the whole code for this computation. Everything appears to work fine, but feel free to make any changes if it makes sense.

There is one test file for which the tests fail, namely `smt.diff.html`. I've used the data in this file to generate a sample table and I receive the same error message there as well, so I assume it's not the tests, but some bug in the `StatisticsTable` component (or in the data of the file itself). I haven't investigated further though, since I didn't want to meddle with the code too much in case @DennisSimon is currently working on something as well.

DennisSimon:
Hi @lachnerm , sorry for the super late response.
I am stuck with exams and work right now. 
I've seen that you already added some stuff. Is there still some things that I should have a look at? I can have a look at the bug with the StatisticsTable this weekend

lachnerm:
> Hi @lachnerm , sorry for the super late response.
> I am stuck with exams and work right now.
> I've seen that you already added some stuff. Is there still some things that I should have a look at? I can have a look at the bug with the StatisticsTable this weekend

Sure, no problem, don't worry. I took care of the bug already. I also merged the current master into the branch and fixed all pipeline problems as well as the merge conflicts. @PhilippWendler would there be anything left from your side?

PhilippWendler:
Thank you very much!

Could it be that there is a bug that prevents the row with filtered statistics to be missing if the "local summary" row exists? At least that seems to correlate in my tests. For example, in the local development version I could not see the filtered statistics (but in some other statistics it worked nicely!).

We also need to indent the name "filtered" with `&nbsp;&nbsp;&nbsp;&nbsp;`, otherwise it look as if "correct results" is a sub-statistic of "filtered".

@DennisSimon Could you have a last look as well if everything is ok and whether you agree with @lachnerm's work? That would be great.

PhilippWendler:
Update: According to the discussion in #125, it is better to just show filtered statistics if filters are used, and not an additional row. So this part of the implementation was changed back.

I have also merged `main` into the branch (no important conflicts luckily), refactored a little bit, and added support for the category `correct-unconfirmed`. Now I think is is finished.

PhilippWendler:
Thank you @DennisSimon for this! It works really well.

PR #617, opened 2020-10-05
Scatter plot swap axis

lachnerm:
Adds an icon to the Scatter Plot between the selections for the X-Axis and Y-Axis that enables switching the values of both axes.


PhilippWendler:
Could you please add something to our test utilities that automatically reduces the amount of text present in the snapshots for icons? Currently the full SVG path is present for every occurrence. I guess just removing the content of every SVG or at least those with `svg-inline--fa` would be fine. If you want to exclude more than that, that would also be fine for me (the rendering of this icon is not something important to test).

PhilippWendler (approved):
Nice, exactly as I wished!

lachnerm:
> Could you please add something to our test utilities that automatically reduces the amount of text present in the snapshots for icons? Currently the full SVG path is present for every occurrence. I guess just removing the content of every SVG or at least those with `svg-inline--fa` would be fine. If you want to exclude more than that, that would also be fine for me (the rendering of this icon is not something important to test).

I've mocked the `@fortawesome/react-fontawesome` module to return only an `i` tag for `FontAwesomeIcons` when running tests. So instead of the whole `svg`, we will only have an `i` tag with the attribute `className="fa {iconName}"`. I think this is the best solution to resolve the problem.

PhilippWendler:
Good idea, also nice for the other snapshot tests actually.

`__mocks__` will be automatically applied in tests? We could move our `uniqid` mock there, I guess?

PR #618, opened 2020-10-05
Add statistical analysis to the scatter plot

### LINKED ISSUE #594: Add statistical information to scatter plots in HTML tables
I would like to request the following feature:

Sometimes it is hard to determine from the scatter plots between two configurations which is best. Adding some statistical information like the fit parameters + confidence intervals from a linear regression would be useful for me.

PhilippWendler:
This adds a line with linear regression to the scatter plot.

Fixes #594


PhilippWendler:
@MartinSpiessl Because you requested this feature, would you have a look at it and tell @lachnerm if there is something to improve?

benchexec/tablegenerator/react-table/src/components/ScatterPlot.js
```diff
@@ -37,6 +44,11 @@ const resultsOptions = {
   correct: "Correct only",
 };
 
+const regressionOptions = {
+  hidden: "Hidden",
```
PhilippWendler, line 48: I would prefer something like `None`. Actually, with just enabled/disable I would prefer a check box, but as far as I understood we could trivially add non-linear options, so let's do that and keep the drop down.

PhilippWendler:
ping @MartinSpiessl Would you be able to review this in the next weeks?

MartinSpiessl:
Sure, I will have a look on Wednesday

MartinSpiessl:
Looks nice!
I would have two minor nitpicks:
- In the help information that is showed when hovering over "Regression", it would be nice if the actual regression method used would be mentioned. It is probably ordinary least squares (OLS)
- Ideally also the units for the fit parameters are given. I think the units are available to BenchExec in many cases, so adding them helps to avoid confusion.

lachnerm:
I've updated the branch with the current master and did some adjustments:
> I would prefer something like `None`. Actually, with just enabled/disable I would prefer a check box, but as far as I understood we could trivially add non-linear options, so let's do that and keep the drop down.
>
I've renamed the option to `None`. And yes, further regression types would be possible to add, so I've gone with your suggestion and kept the dropdown.
> * In the help information that is showed when hovering over "Regression", it would be nice if the actual regression method used would be mentioned. It is probably ordinary least squares (OLS)
>
This is now shown at the very top of the tooltip.

> * Ideally also the units for the fit parameters are given. I think the units are available to BenchExec in many cases, so adding them helps to avoid confusion.
>
I wasn't sure where to put them, so I've added the units behind the `Predictor variable` and `Response variable` in the tooltip. If you prefer them somewhere else, I can change the location of course.

Furthermore, the linear regression is now automatically deactivated and changed to `None` if one of the selections is not numerical (as otherwise, the linear regression wouldn't make sense in the first place). When the user tries to select `Linear` from the dropdown while one of the selections is non-numerical, a small alert is shown. Alternatively, I could also deactivate the whole dropdown and grey out the UI, if that's preferred. Of course, I could also implement any other kind of notification etc.

Lastly, I've done some minor other changes that don't really change the functionality. There was a bug where the values were slightly different than before due to roundings for the regression, but this is fixed now as well.

PhilippWendler:
Nice, thanks!

Looks good for me, @MartinSpiessl what do you think?

@lachnerm Is there anything from your side that is still open for this PR?

Could we add some tests? Maybe be need to separate the plot rendering and the regression analysis more for this?

lachnerm:
From my side, there wouldn't be anything that we'd need to add to the PR. Only some additional tests would make sense.
Would it suffice to use snapshot tests in this case? This would cover both the regression line and the confidence intervals. I could also add some further small tests that compare the regression results (i.e. the inferred function). Both could be added to the current tests for the scatter plot.

PhilippWendler:
Snapshot tests would be better than nothing, but if it is easy, just having a set of tests for the inferred function would also be good (and if tests fail, these are probably easier to debug).

lachnerm:
During the implementation of the tests, I noticed some edge cases where bugs appeared, which I think I should have taken care of by now.
I've added snapshot tests for the linear regression plot for all column pairs (for the single test file we are using for the scatter plot tests). I didn't add tests for the other dropdown selections (only correct/all results, linear/logarithmic scaling, aux. lines), as these should be independent from the regression computation and would only blow up the snapshot file. Would that be fine?

I've thought about how to test the inferred function, but couldn't really find a good solution. The problem is that there is a lot of computation going on within the `ScatterPlot` component on which data points are shown within the plot, which is quite essential for the regression as well, so I'd have to access this processed data from the tests as well. However, there would be no meaning in computing the regression with this data from the tests and compare the results with these from the `ScatterPlot` component, as the exact same is done within our code as well. So basically this would just test whether the same function returns the same output for the same input - which, needless to say, won't really help much.
I've thought about using another framework to infer the function and compare the two results, however, here we would more or less just test the framework instead of our code and we might find some failing tests due to slight differences in the frameworks themselves (rounding etc.)
Would you have any other idea on which tests might be helpful? Otherwise, I'd suggest to stick with the snapshot tests, even if it's a bit more difficult to debug.

PhilippWendler:
Sounds plausible.

I guess this is finished now? @MartinSpiessl Do you want to have a second look?

PR #619, opened 2020-10-06
Add gazer-theta tool-info

AdamZsofi:
[Gazer](https://github.com/ftsrg/gazer)
[Theta](https://github.com/ftsrg/theta)
[FMCAD Paper](https://ftsrg.mit.bme.hu/theta/publications/fmcad2017.pdf)

Note: I tried to use `BaseTool2` based on the migration guide, but I still couldn't use --tool-directory.
(`Error: Tool-info module for gazer-theta does not support parameter --tool-directory. Instead, you can add the tool to PATH, execute benchexec from the tool directory, or upgrade the tool-info module.`).
If you can tell me, what to upgrade, I'll update the tool-info.

### DIFF
--- a/benchexec/tools/gazer-theta.py
+++ b/benchexec/tools/gazer-theta.py
@@ -8,6 +8,7 @@
 import benchexec.result as result
 import benchexec.tools.template
 
+
 class Tool(benchexec.tools.template.BaseTool2):
     """
     Tool info for gazer-theta
@@ -40,7 +41,11 @@ class Tool(benchexec.tools.template.BaseTool2):
             elif "Result of gazer-theta run: TRUE" in line:
                 status = result.RESULT_TRUE_PROP
 
-        if not run.was_timeout and status == result.RESULT_UNKNOWN and run.exit_code.value != 0:
+        if (
+            not run.was_timeout
+            and status == result.RESULT_UNKNOWN
+            and run.exit_code.value != 0
+        ):
             status = result.RESULT_ERROR
 
         return status
--- a/benchexec/tools/gazer-theta.py
+++ b/benchexec/tools/gazer-theta.py
@@ -8,7 +8,6 @@
 import benchexec.result as result
 import benchexec.tools.template
 
-
 class Tool(benchexec.tools.template.BaseTool2):
     """
     Tool info for gazer-theta
@@ -17,10 +16,10 @@ class Tool(benchexec.tools.template.BaseTool2):
     https://github.com/ftsrg/theta
     """
 
-    REQUIRED_PATHS = ["gazer", "gazer/tools/gazer-bmc", "gazer/tools/gazer-theta"]
+    REQUIRED_PATHS = [".."]
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("gazer_starter.py", subdir="gazer/scripts")
+        return tool_locator.find_executable("gazer_starter.py", subdir="scripts")
 
     def name(self):
         return "gazer-theta"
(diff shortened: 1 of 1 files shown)


benchexec/tools/gazer-theta.py
```diff
@@ -0,0 +1,50 @@
...
+        return tool_locator.find_executable("gazer_starter.py", subdir="gazer/scripts")
+
+    def name(self):
+        return "gazer-theta"
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        assert len(list(task.input_files_or_identifier)) == 1
```
PhilippWendler, line 32: Does the tool support tasks that are defined without an input file? If yes, it is fine as is, otherwise I suggest to use `task.input_files` to give a nice error message.

benchexec/tools/gazer-theta.py
```diff
@@ -0,0 +1,50 @@
...
+        # possible option: --output (default value if flag isn't used: working directory)
+        return [executable] + options + list(task.input_files_or_identifier)
+
+    def determine_result(self, run):
+        status = result.RESULT_UNKNOWN
+        if run.was_timeout:
+            status = "TIMEOUT"
```
PhilippWendler, line 39: Note that this is not necessary. If you just return `UNKNOWN` by default, BenchExec will overwrite it with `TIMEOUT` anyway. The method `was_timeout` is present mostly for letting developers distinguish between a crash and tool killed due to timeout.

benchexec/tools/gazer-theta.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for gazer-theta
+    (combined tool of Gazer and Theta)
+    https://github.com/ftsrg/gazer
+    https://github.com/ftsrg/theta
+    """
+
+    REQUIRED_PATHS = ["gazer", "gazer/tools/gazer-bmc", "gazer/tools/gazer-theta"]
```
PhilippWendler, line 20: Including `gazer` implies the rest. However, note that the default implementation of `program_files` treats `REQUIRED_PATHS` as relative to the executable's directory. So you would either need to change the content of the list or overwrite `program_files` (you can still call our utility method `_program_files_from_executable` with appropriate parameters if you wish).

PhilippWendler (commented):
I have a few comments that you might want to address?

Note that `python3 -m benchexec.test_tool_info gazer-theta --tool-directory something` works for me as expected (it does not find the tool and prints a message that it was searching for it in `something/gazer/scripts`). What exactly did you try?

benchexec/tools/gazer-theta.py
AdamZsofi: If I understand right what program_files does and if these paths are relative to the executable, would `"../../gazer"` (or simply `..`) suffice? *(Although even if it works, the executable will be included twice in the list.)* If not, I'll take a look at _program_files_from_executable and program_files and overwrite the latter.

benchexec/tools/gazer-theta.py
AdamZsofi: Thanks, I checked out `Task` in the template and I think I understand the fields now.
Just so I get this right:
I still need the `assert len(list(task.input_files)) == 1`, because we don't support more than one input file, but I should use `task.input_files` in both the assert and the return statement, as in the case of no input files we would get and identifier and a length of 1 instead of a length with 0 in the assert if it stays like this.

PhilippWendler: Exactly. I will add a utility method that will let you avoid the `assert`, but it is easier to add the tool-info module as you suggest and refactor it later.

benchexec/tools/gazer-theta.py
PhilippWendler: `".."` would effectively have the same effect as `"../../gazer"` but would be more robust in case of renamings of that directory. `test_tool_info` will tell you which things are included, so you can check with that.

If something is included twice (directly or indirectly) this is not a problem.

PhilippWendler (approved):

PhilippWendler:
Could you please fix the formatting with `black` (cf. CI)?

hajduakos:
Thanks for your feedback @PhilippWendler !

PR #622, opened 2020-10-14
Refactor nitwit.py to use the new format.

blizzard4591:
I tried to adapt to the changes as presented in cpachecker and the dummy tool, but if someone with more knowledge of the framework and Python could take a quick look, that'd be much appreciated.

### DIFF
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -43,7 +43,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     def cmdline(self, executable, options, task, rlimits):
         if task.property_file:
             options = options + ["-p", task.property_file]
-        return [executable] + options + task.single_input_file
+        return [executable] + options + [task.single_input_file]
 
     def determine_result(self, run):
         """
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -63,7 +63,4 @@ class Tool(benchexec.tools.template.BaseTool2):
         ):
             status = "OUT OF MEMORY"
 
-        if not status:
-            status = result.RESULT_ERROR
-
         return status
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -58,7 +58,9 @@ class Tool(benchexec.tools.template.BaseTool2):
         else:
             status = result.RESULT_ERROR
 
-        if ("out of memory!".lower() in (line.lower() for line in run.output)) or (run.exit_code.value == 251):
+        if ("out of memory!".lower() in (line.lower() for line in run.output)) or (
+            run.exit_code.value == 251
+        ):
             status = "OUT OF MEMORY"
 
         if not status:
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -53,8 +53,6 @@ class Tool(benchexec.tools.template.BaseTool2):
         """
         if run.exit_code.value in [0, 245]:
             status = result.RESULT_FALSE_REACH
-        elif (run.exit_code.signal == 9) or run.exit_code.value == 9:
-            status = "TIMEOUT"
         elif run.exit_code.value in [4, 5, 241, 242, 243, 250]:
             status = result.RESULT_UNKNOWN
         else:
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -58,16 +58,16 @@ class Tool(benchexec.tools.template.BaseTool2):
         about result codes.
         @return: status of validator after executing a run
         """
-        if run.exit_code.signal is None and (run.exit_code.value == 0 or run.exit_code.value == 245):
+        if run.exit_code.value in [0, 245]:
             status = result.RESULT_FALSE_REACH
-        elif run.exit_code.value is None or run.exit_code.value in [-9, 9]:
+        elif (run.exit_code.signal == 9) or run.exit_code.value == 9:
             status = "TIMEOUT"
         elif run.exit_code.value in [4, 5, 241, 242, 243, 250]:
             status = result.RESULT_UNKNOWN
         else:
             status = result.RESULT_ERROR
 
-        if run.output.any_line_contains("Out of memory") or (run.exit_code.value == 251):
+        if ("out of memory!".lower() in (line.lower() for line in run.output)) or (run.exit_code.value == 251):
             status = "OUT OF MEMORY"
 
         if not status:
--- a/benchexec/tools/nitwit.py
+++ b/benchexec/tools/nitwit.py
@@ -18,7 +18,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     URL: https://github.com/moves-rwth/nitwit-validator
     """
 
-    REQUIRED_PATHS = []
+    REQUIRED_PATHS = ["bin/nitwit32", "bin/nitwit64"]
     BIN_DIR = "bin"
 
     def executable(self, tool_locator):
@@ -33,13 +33,6 @@ class Tool(benchexec.tools.template.BaseTool2):
                 "Required binary files for Nitwit not found in {0}.".format(bin_path)
             )
         return executable
-
-    def program_files(self, executable):
-        return [
-            executable,
-            os.path.join(self.BIN_DIR, "nitwit32"),
-            os.path.join(self.BIN_DIR, "nitwit64"),
-        ]
 
     def version(self, executable):
         return self._version_from_tool(executable, "--version")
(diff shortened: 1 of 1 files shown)


benchexec/tools/nitwit.py
```diff
@@ -48,30 +47,27 @@ def version(self, executable):
...
         return "Nitwit"
 
-    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
-        if propertyfile:
-            options = options + ["-p", propertyfile]
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options = options + ["-p", task.property_file]
+        return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 53: Does nitwit support tasks without an input file? Does it support tasks with more than one input file?

Depending on the answer, I would suggest to use `task.input_files` or `task.single_input_file`. This immediately fails in unsupported use cases with a nice message.

benchexec/tools/nitwit.py
```diff
@@ -48,30 +47,27 @@ def version(self, executable):
...
         """
         See README.md at https://github.com/moves-rwth/nitwit-validator for information
         about result codes.
-        @param returncode: code returned
-        @param returnsignal: signal, which terminated validator
-        @param output: the std output
         @return: status of validator after executing a run
         """
-        if returnsignal == 0 and (returncode == 0 or returncode == 245):
+        if run.exit_code.signal is None and (run.exit_code.value == 0 or run.exit_code.value == 245):
```
PhilippWendler, line 61: The check for `signal is None` is no longer neccessary, `value` is only a number if `signal` is `None`.
```suggestion
        if run.exit_code.value in [0, 245]:
```

benchexec/tools/nitwit.py
```diff
@@ -48,30 +47,27 @@ def version(self, executable):
...
             status = result.RESULT_FALSE_REACH
-        elif returncode is None or returncode in [-9, 9]:
+        elif run.exit_code.value is None or run.exit_code.value in [-9, 9]:
```
PhilippWendler, line 63: Same here, plus `value` is always a positive number.
```suggestion
        elif run.exit_code.value == 9:
```

But does nitwit actually return value 9 on timeout? Does it have an internal time limit and terminates by itself is that elapses?

benchexec/tools/nitwit.py
```diff
@@ -48,30 +47,27 @@ def version(self, executable):
...
             status = "TIMEOUT"
-        elif returncode in [4, 5, 241, 242, 243, 250]:
+        elif run.exit_code.value in [4, 5, 241, 242, 243, 250]:
             status = result.RESULT_UNKNOWN
         else:
             status = result.RESULT_ERROR
 
-        if "Out of memory" in output or returncode == 251:
+        if run.output.any_line_contains("Out of memory") or (run.exit_code.value == 251):
```
PhilippWendler, line 70: Here, the previous code was actually buggy and the condition `"Out of memory" in output` was never fulfilled because `output` contained only strings with an included line separator.

Is the intention to check that there is a line in the output that contains only `Out of memory` and nothing else except for the line separator? Then the condition would be `"Out of memory" in run.output` (the new API improves this use case and makes it possible to write this).

Or is the intention to check that the string appears somewhere in all of the output, even in the middle of some line somewhere? Then the new code is correct.

PhilippWendler:
I now noticed that the existing code in `program_files` is wrong, it works only if Nitwit's directory is the current directory, but this is not guaranteed.

The easiest way to correct this is to set `REQUIRED_PATHS = ["bin/nitwit32", "bin/nitwit64"]` and remove `program_files` completely to use the default implementation, which uses `REQUIRED_PATHS` and correctly treats the content of that list as relative to the directory of the executable.

If this is done, one could also simplify the check in `executable` to check all files in `REQUIRED_PATHS` instead of hardcoding the file names there again.

benchexec/tools/nitwit.py
blizzard4591: I think this should have been a signal check for SIGKILL. Afaik we do not return 9 anywhere.

benchexec/tools/nitwit.py
blizzard4591: Thanks for the clarification, only single input files are supported. Fixed.

benchexec/tools/nitwit.py
blizzard4591: Thanks for the clarification!

benchexec/tools/nitwit.py
```diff
@@ -50,24 +43,24 @@ def name(self):
     def cmdline(self, executable, options, task, rlimits):
         if task.property_file:
             options = options + ["-p", task.property_file]
-        return [executable] + options + list(task.input_files_or_identifier)
+        return [executable] + options + task.single_input_file
```
PhilippWendler, line 46:
```suggestion
        return [executable] + options + [task.single_input_file]
```
Otherwise there is an error.

I suggest to always test the tool-info module with `python3 -m benchexec.test_tool_info nitwit --tool-directory ...`.

benchexec/tools/nitwit.py
```diff
@@ -50,24 +43,24 @@ def name(self):
...
         """
         See README.md at https://github.com/moves-rwth/nitwit-validator for information
         about result codes.
         @return: status of validator after executing a run
         """
-        if run.exit_code.signal is None and (run.exit_code.value == 0 or run.exit_code.value == 245):
+        if run.exit_code.value in [0, 245]:
             status = result.RESULT_FALSE_REACH
-        elif run.exit_code.value is None or run.exit_code.value in [-9, 9]:
+        elif (run.exit_code.signal == 9) or run.exit_code.value == 9:
```
PhilippWendler, line 56: As you said in your comment that Nitwit never returns exit code 9, the second part seems unnecessary. If the process is killed by a signal, `value` won't be set.

But a check for "SIGKILL => TIMEOUT" is not necessary at all in a tool-info module. If there is a timeout, BenchExec will mark it as such automatically. In fact, the current code will only hide other crashes with signal 9 behind a `TIMEOUT` status, which I guess is not intended. So I would suggest to remove this branch completely.

benchexec/tools/nitwit.py
```diff
@@ -50,24 +43,24 @@ def name(self):
...
         elif run.exit_code.value in [4, 5, 241, 242, 243, 250]:
             status = result.RESULT_UNKNOWN
         else:
             status = result.RESULT_ERROR
 
-        if run.output.any_line_contains("Out of memory") or (run.exit_code.value == 251):
+        if ("out of memory!".lower() in (line.lower() for line in run.output)) or (run.exit_code.value == 251):
             status = "OUT OF MEMORY"
 
         if not status:
```
PhilippWendler, line 66: I just noticed that this is dead code, because `status` will always be set already.

benchexec/tools/nitwit.py
```diff
@@ -50,24 +43,24 @@ def name(self):
...
-        elif run.exit_code.value is None or run.exit_code.value in [-9, 9]:
+        elif (run.exit_code.signal == 9) or run.exit_code.value == 9:
             status = "TIMEOUT"
         elif run.exit_code.value in [4, 5, 241, 242, 243, 250]:
             status = result.RESULT_UNKNOWN
         else:
             status = result.RESULT_ERROR
 
-        if run.output.any_line_contains("Out of memory") or (run.exit_code.value == 251):
+        if ("out of memory!".lower() in (line.lower() for line in run.output)) or (run.exit_code.value == 251):
```
PhilippWendler, line 63: Please reformat with `black` or apply the [diff](https://travis-ci.org/github/sosy-lab/benchexec/jobs/735977162#L945) from CI manually, this line is too long.

benchexec/tools/nitwit.py
blizzard4591: Why is it dead? Won't it override the status if it matches?

benchexec/tools/nitwit.py
blizzard4591: Removed.

benchexec/tools/nitwit.py
PhilippWendler: Maybe that was unclear, the branch `if not status:` is dead AFAIS.

benchexec/tools/nitwit.py
blizzard4591: Done.

benchexec/tools/nitwit.py
blizzard4591: Ah yes, of course. Deleted.

PhilippWendler (approved):

PR #623, opened 2020-10-15
Update veriabs.py

priyankadarke:
As per the new implementation.

### DIFF
--- a/benchexec/tools/veriabs.py
+++ b/benchexec/tools/veriabs.py
@@ -32,7 +32,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     ]
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("scripts/veriabs")
+        return tool_locator.find_executable("veriabs", subdir="scripts")
 
     def version(self, executable):
         return self._version_from_tool(executable)
--- a/benchexec/tools/veriabs.py
+++ b/benchexec/tools/veriabs.py
@@ -10,7 +10,7 @@ import benchexec.tools.template
 import benchexec.result as result
 
 
-class Tool(benchexec.tools.template.BaseTool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
     VeriAbs
     """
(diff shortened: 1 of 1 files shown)


benchexec/tools/veriabs.py
```diff
@@ -31,8 +31,8 @@ class Tool(benchexec.tools.template.BaseTool):
         "supportFiles",
     ]
 
-    def executable(self):
-        return util.find_executable("scripts/veriabs")
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("scripts/veriabs")
```
PhilippWendler, line 35:
```suggestion
        return tool_locator.find_executable("veriabs", subdir="scripts")
```
This will have better behavior if a user decides to add VeriAbs to their `PATH`.

benchexec/tools/veriabs.py
```diff
@@ -45,9 +45,9 @@ def program_files(self, executable):
     def name(self):
         return "VeriAbs"
 
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        if propertyfile:
-            options = options + ["--property-file", propertyfile]
+    def cmdline(self, executable, options, tasks, rlimits):
```
PhilippWendler, line 48:
```suggestion
    def cmdline(self, executable, options, task, rlimits):
```
I suggest to rename this in order to not cause confusion.

benchexec/tools/veriabs.py
```diff
@@ -45,9 +45,9 @@ def program_files(self, executable):
...
+        if tasks.property_file:
+            options += ["--property-file", task.property_file]
         return [executable] + options + tasks
```
PhilippWendler, line 51: The new value of `task` is an object that cannot be appended to a list. You probably need to use `[task.single_input_file]`, `[task.input_files]`, or something else depending on what use case your tool supports (cf. the documentation of the `benchexec.tools.template.BaseTool2.Task` class).

PhilippWendler (changes requested):
The method `determine_result` also needs adjustment.

I suggest reading the [migration guide](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#migrating-tool-info-modules-to-new-api) and testing the tool-info module with `python3 -m benchexec.test_tool_info veriabs --tool-directory ...`.

benchexec/tools/veriabs.py
priyankadarke: Yes, Thanks.

PR #625, opened 2020-10-15
Add tool info for CoVeriTest

cedricrupb:


benchexec/tools/coveritest.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.cpachecker as cpachecker
+
+
+class Tool(cpachecker.Tool):
+    """Tool info for CoVeriTest."""
+
+    REQUIRED_PATHS = [
```
PhilippWendler, line 14: This duplicates the info from CPAchecker's tool-info module. I would suggest to deduplicate this, otherwise CoVeriTest's tool-info module is likely to become outdated when something in CPAchecker requires adding a new item to this list.

benchexec/tools/coveritest.py
```diff
@@ -0,0 +1,25 @@
...
+        "lib/java/runtime",
+        "lib/*.jar",
+        "lib/native/x86_64-linux",
+        "scripts",
+        "cpachecker.jar",
+        "config",
+        "resources",
```
PhilippWendler, line 21: As a CPAchecker developer, I am curious about where this path comes from? AFAIK there is nothing in current CPAchecker that would access such a directory.

cedricrupb: There is a new CoVeriTest version coming with an advanced time scheduler. The scheduler locates some resources in the resource folder. Changes to CPAchecker which would require the resources/ path were not pushed to trunk yet.

PhilippWendler (approved):
Thanks for also thinking of PeSCo!

PR #626, opened 2020-10-15
Add Goblint tool-info

sim642:


benchexec/tools/goblint.py
```diff
@@ -0,0 +1,90 @@
...
+                else:
+                    raise benchexec.tools.template.UnsupportedFeatureException(
+                        "Unsupported data_model '{}'".format(data_model)
+                    )
+
+        return [
+            executable,
+            *options,
+            *additional_options,
+            *task.input_files_or_identifier,
```
PhilippWendler, line 58: Just to make sure that this is actually intended: This means that your tool supports both tasks with no input file as well as tasks with more than one input file? If not, consider using `task.input_files` or `task.single_input_file` to automatically produce an error message.

benchexec/tools/goblint.py
```diff
@@ -0,0 +1,90 @@
...
+                        return "EXCEPTION"
+            else:
+                m = re.match(r"SV-COMP result: (.*)", line)
+                if m:
+                    status = m.group(1)
+
+        if status:
+            return status
+
+        if run.was_timeout:
```
PhilippWendler, line 85: This branch could be removed, BenchExec does that automatically.

PhilippWendler (commented):
Thanks, this looks really good. It is nice to see tool-info modules that go beyond the minimum and add support for helpful return values, for example.

I have just one question for clarification and a minor recommendation.

PhilippWendler (approved):

PR #627, opened 2020-10-15
Add support to FuSeBMC

rafaelsamenezes:
This intends to add the support for FuSeBMC https://github.com/kaled-alshmrany/FuSeBMC


benchexec/tools/FuSeBMC.py
```diff
@@ -0,0 +1,79 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool):
```
PhilippWendler, line 14: For a new tool-info module, would you mind writing it for our new, more convenient, API based on `BaseTool2`? We have a [migration guide](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#migrating-tool-info-modules-to-new-api).

benchexec/tools/FuSeBMC.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    This class serves as tool adaptor for ESBMC (http://www.esbmc.org/)
```
PhilippWendler, line 16: There is already a tool-info module for ESBMC, is this comment wrong?

benchexec/tools/FuSeBMC.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    REQUIRED_PATHS = ["esbmc", "esbmc-wrapper.py", "my_instrument"]
```
PhilippWendler, line 19: Having the executable that is returned from `executable()` in this is redundant.

benchexec/tools/FuSeBMC.py
```diff
@@ -0,0 +1,79 @@
...
+            if isTimeout:
+                status = "TIMEOUT"
+            elif output.endswith(("error", "error\n")):
+                status = "ERROR"
+
+        return status
+
+    """ helper method """
+
+    def allInText(self, words, text):
```
PhilippWendler, line 70: This helper method seems unnecessary because it is always only called with a single-element list. `if word in text` is enough instead.

benchexec/tools/FuSeBMC.py
```diff
@@ -0,0 +1,79 @@
...
+        return executableDir
+
+    def version(self, executable):
+        return self._version_from_tool(executable, "-v")
+
+    def name(self):
+        return "FuSeBMC"
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        assert len(tasks) == 1, "only one inputfile supported"
```
PhilippWendler, line 35: After changing to `BaseTool2`, this assert is no longer necessary because you can simply use `task.single_input_file`.

PhilippWendler (changes requested):

rafaelsamenezes:
@PhilippWendler when porting it I looked over some examples and saw that cpa-witness2test.py only inherits from cpachecker. FuSeBMC wrapper in a nutshell is the same as esbmc with the only real change being the REQUIRED_PATHS. May I just do that instead of porting it?

PhilippWendler:
Sure. This is recommended if you expect that any changes to ESBMC's tool-info module (e.g., improving output parsing or adding required files) will also be relevant for your tool.

PhilippWendler (commented):
Looks good, please just fix format as requested by CI.

PhilippWendler (approved):

PhilippWendler:
Unfortunately, just after I merged the PR, I noticed the name of the file, which is `FuSeBMC` and does not match our typical name of "lowercase tool-name". Is it ok for you if I rename it?

PR #633, opened 2020-10-16
Add Tool-Info Module for CMA-ES Fuzz

gernst:


benchexec/tools/cmaesfuzz.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+import re
+import benchexec.util as util
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool):
```
PhilippWendler, line 14: For a new tool-info module, would you mind writing it for our new, more convenient, API based on `BaseTool2`? We have a [migration guide](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#migrating-tool-info-modules-to-new-api).

benchexec/tools/cmaesfuzz.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def name(self):
+        return "CMA-ES Fuzz"
+
+    def get_value_from_output(self, lines, identifier):
+        for line in reversed(lines):
+            pattern = identifier
+            if pattern[-1] != ":":
+                pattern += ":"
+            match = re.match("^" + pattern + "([^(]*)", line)
```
PhilippWendler, line 44: Wouldn't it make more sense to move the construction of `pattern` out of the loop, and while we are at it also use `re.compile` to pre-compile the resulting regex?

PhilippWendler (commented):
This is in principle acceptable, but I have some questions.

Does the tool support tasks with no input file as well as tasks with more than one input file? If not, consider overwriting `cmdline` and using `task.input_files` or `task.single_input_file` in it to automatically produce an error message.

benchexec/tools/cmaesfuzz.py
gernst: Sure :)

benchexec/tools/cmaesfuzz.py
gernst: good catch

benchexec/tools/cmaesfuzz.py
```diff
@@ -0,0 +1,42 @@
...
+    REQUIRED_PATHS = [
+        "fuzzer",
+        "fuzzer.py",
+        "cma",
+        "verifiers_bytes",
+        "verifiers_real",
+    ]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("fuzzer")
```
PhilippWendler, line 30: Is `fuzzer` correct? (Most tools who have files name `tool` and `tool.py` need the wrapper script here.)

PhilippWendler (commented):
Looks good, just two questions to make sure:

Does the tool support tasks with no input file as well as tasks with more than one input file? If not, consider overwriting `cmdline` and using `task.input_files` or `task.single_input_file` in it to automatically produce an error message.

(other question in comment)

benchexec/tools/cmaesfuzz.py
gernst: Yes this is correct, the name came relatively late

gernst:
Ok I've overwritten `cmdline` hope it works (untested)

PhilippWendler (approved):

PR #636, opened 2020-10-16
Updated Verifuzz.py tool-info module

raveenkm:
BaseTool2 migration.


raveenkm:
Hi,
I tried to find why it is failing. But I am unable to figure out. It complains about the if-then-else block in the determine_result section. I re-looked at it, but I could not find any problem. Can you please help?

PhilippWendler:
No, it does snot complain about the if-then-else block, it complains about several other lines: https://travis-ci.org/github/sosy-lab/benchexec/jobs/736591164#L949

You can either apply the diff that is printed in the CI log, reformat the source code with [black](https://pypi.org/project/black/), or fix the issues (trailing whitespaces, double blank lines) manually.

benchexec/tools/verifuzz.py
```diff
@@ -41,34 +40,36 @@ def program_files(self, executable):
...
+                return result.RESULT_FALSE_TERMINATION
+            elif "FALSE(valid-deref)" in line:
+                return result.RESULT_FALSE_DEREF
+            elif "FALSE(valid-free)" in line:
+                return result.RESULT_FALSE_FREE
+            elif "FALSE(valid-memtrack)" in line:
+                return result.RESULT_FALSE_MEMTRACK
+            elif "NOT SUPPORTED" in line or "VERIFUZZ_UNKNOWN" in line:
+                return result.RESULT_UNKNOWN
         else:
```
PhilippWendler, line 74: Here is an error, you have a "for-else" block, which does not make sense. You need to remove the `else` and adjust the indentation of the following line such that it returns the error result if the loop terminates.

raveenkm: Though I am not an expert python programmer, I have picked it up in the for-else block in dummy.py (I thought it is a template), and veriabs. veriabs went through the checks

PhilippWendler: There is no `else` in `dummy.py`?

raveenkm: My mistake dummy.py has no for-else block. But veriabs has. It went through.

PhilippWendler: Oh, that is an error that I did not catch.

PhilippWendler: Ok, actually I am not sure whether it is an error or not, but it is certainly unnecessary here and a quite unusual Python syntax, so please replace it with what I suggested.

raveenkm: If I format the if block, as you mentioned, logically the result conveyed back would be in-correct. If the output produced by the tool does not contain any of the above strings (which generally the case), for the first line itself the for loop terminates by returning the result.RESULT_ERROR, which is incorrect. The result.RESULT_ERROR should be returned after the for loop termination.

PhilippWendler: That's what I suggested in my first comment. The `return result.RESULT_ERROR` should be a simple statement after the loop.

raveenkm: Done that now. and submitted

raveenkm: Failed again. Not clear what is it complaining about. I am using spyder to edit this code. It flags no syntax errors.

PhilippWendler: I have answered this here: https://github.com/sosy-lab/benchexec/pull/636#issuecomment-710837747

The current set of required changes is here: https://travis-ci.org/github/sosy-lab/benchexec/jobs/736608127#L949

PhilippWendler: If you see a CI job complaining, please always check its logs.

raveenkm:
Giving up. Need help. The CI log is not giving the right clues (it is not clear why it is "Not ok"). Syntactically there seems to be no errors. It's a matter of spaces and extra lines there rather than any syntactic issues.

PhilippWendler:
Yes, as mentioned this is an issue about trailing spaces and blank lines. I have repeatedly linked to the diff that shows the required changes, the latest instance is here: https://travis-ci.org/github/sosy-lab/benchexec/jobs/736628179#L949 Is this clear? Also, you could use black to reformat everything automatically as required.

PhilippWendler:
I reformatted it for you.

PR #641, opened 2020-10-18
remove equivalence check for property files during parsing task definitions

kfriedberger:
The method `os.path.samefile` does not work reliably on Windows when using network storage.

Using VirtualBox with shared directories,
Python seems to be not able to distinguish reliably between different property files.
I do not know the exact reason, but the following benchmark definition causes a weird behaviour:
(requires CPAchecker r35613 and a current BenchExec version)
  `$> scripts/benchmark.py @webcloud test/test-sets/integration-bam.xml`

The benchmark is expected to have 1847 tasks,
and Linux shows that number when submitting the tasks.
Windows however finds for the same benchmark 2070 tasks,
because it ignores seme invalid property matches.
I have even seen other numbers, between 1847 and 2070.
The bug seems to depend on the provided inodes for those property files.
Windows provides identical data via `os.stat`, even if the files are different.

I do not know, when files need to be identical, whether this is a bug
in our scripts or in Python (3.9.0 for Win64) or in VirtualBox (5.2.42 on Ubuntu 18.04).
The change in these lines fixes the bug, however it removes the equivalence check.
I do not know a case where the equivalence check would have helped.

I would like to discuss about this issue. Maybe I am overlooking something.

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -853,9 +853,7 @@ class RunSet(object):
 
             # TODO We could reduce I/O by checking absolute paths and using os.path.samestat
             # with cached stat calls.
-            if prop.filename == expanded[0] or os.path.samefile(
-                prop.filename, expanded[0]
-            ):
+            if prop.filename == expanded[0]:
                 expected_result = prop_dict.get("expected_verdict")
                 if expected_result is not None and not isinstance(
                     expected_result, bool


PhilippWendler:
> The method `os.path.samefile` does not work reliably on Windows when using network storage.
> 
> Using VirtualBox with shared directories,
> Python seems to be not able to distinguish reliably between different property files.

> I would like to discuss about this issue.

I guess that for discussing an issue a GitHub issue is probably better suited because here the discussion is inevitably tied to the current proposed change instead of to the general problem.

> The change in these lines fixes the bug, however it removes the equivalence check.
> I do not know a case where the equivalence check would have helped.

The existing code is useful because it implements exactly what we want ("if users list the same file in their benchmark definition then it will match"). Otherwise there will be cases where the same file was specified but it won't match because users referred to the property file not in exactly the same ways as it was referred to in the task-definition file, e.g., via symlinks, hard links, etc. Worse, they would silently be benchmarking only a subset of their expected tasks.

Furthermore, if it were not generally useful, changing the behavior now is not possible because it might break existing setups of users.

So this PR cannot be merged.

> The bug seems to depend on the provided inodes for those property files.
> Windows provides identical data via `os.stat`, even if the files are different.
> 
> I do not know, when files need to be identical, whether this is a bug
> in our scripts or in Python (3.9.0 for Win64) or in VirtualBox (5.2.42 on Ubuntu 18.04).

If you claim that "Windows" provides identical data via `os.stat`, how can it be a bug in BenchExec? It would either be a Windows problem or (because the method is not a Windows method but part of the Python API!) a Python problem.

So the way forward would probably be to investigate more about this behavior and then for example file a Python bug.

Alternatively you could suggest a suitable workaround for BenchExec. But note that because (a) Windows is unsupported for `benchexec` anyway, (b) this only occurs with a third-party extension to BenchExec, (c) even on Windows it only occurs for a certain niche use case (network files), and (d) adding more unexpected tasks to the benchmark set is a less critical problem than missing some tasks, if we are going to add a workaround it must not break any real use cases and it must not considerably increase the complexity of the code and its maintenance.

PR #642, opened 2020-10-19
Make links to other files in yaml files behave as normal links

### LINKED ISSUE #628: Make links to C-files clickable as default URLs
### Context
With the introduction of yml-files, we lost the chance to directly click on the filename to see the source code.
Later, the possibility was added to click on filenames in the yml-file overlay by using a `link-overlay-file-link` button.

### Missing Feature
I want to be able to click on a filename and handle it just like a default url, e.g., `open in a new tab` or `save as`... or use `mouse-middle-click` to open in a new tab.

Can we replace the button-based solution with a link-based solution?

lachnerm:
Fixes #628.


PhilippWendler (approved):
Thanks!

PR #643, opened 2020-10-19
SMACK now prints version to stdout

zvonimir:


PhilippWendler (commented):
Is it possible to write the tool-info module in a way that supports both old and new SMACK? Otherwise it will just crash if people use it with the wrong SMACK version, which is of course not helpful.

zvonimir:
Good point! I updated the pull request accordingly.

PhilippWendler (approved):

PR #644, opened 2020-10-22
Add support for Ultimate 0.2.0 and later, fix copyright header

danieldietsch:


benchexec/tools/ultimate.py
```diff
@@ -105,24 +108,21 @@ def _ultimate_version(self, executable):
...
         self.api = len(cmds)
         for cmd in cmds:
-            version = self._query_ultimate_version(cmd, self.api)
-            if version != "":
-                return version
+            for java_version, java in java_versions.items():
+                version = self._query_ultimate_version([java] + cmd, self.api)
+                if version != "":
+                    logging.debug(
+                        f"Using Java {java} for API version {self.api} of Ultimate {version}"
```
PhilippWendler, line 120: f-strings don't work on Python 3.5, which is still supported. Could you please avoid them? The logging module has lazy builtin string formatting anyway.
```suggestion
                       "Using Java %s for API version %s of Ultimate %s", java, self.api, version
```
(Also for the other log statements, of course.)

benchexec/tools/ultimate.py
```diff
@@ -139,39 +139,42 @@ def _query_ultimate_version(self, cmd, api):
...
-                "Exit code:       %s\n"
-                "Error output:    %s\n"
-                "Standard output: %s",
-                api,
-                " ".join(map(util.escape_string_shell, cmd)),
-                process.returncode,
-                util.decode_to_string(stderr),
-                stdout,
-            )
+            if logging.getLogger().getEffectiveLevel() == logging.DEBUG:
```
PhilippWendler, line 142: Hiding only a `logging.debug` statement inside this check doesn't make much sense, right?

Another question is: Does it make sense for the tool-info module to continue if getting the version does not work? Is there a chance that actual benchmarking will work in this case? Otherwise I would simply produce an error here that contains all the relevant information, no need to worry about warnings and debug messages.

If you want to continue, then why not simply have a debug log message with all the raw info first and afterwards a warning if the version cannot be determined?

benchexec/tools/ultimate.py
```diff
@@ -474,18 +482,23 @@ def get_java(self):
...
+                logging.debug(
+                    f"Found Java installation {candidate} with version {version}"
+                )
+                rtr[version] = candidate
+        if not rtr:
+            raise BenchExecException("Could not find any Java version")
+        return rtr
+
+    @staticmethod
+    def which(program):
```
PhilippWendler, line 496: Actually, there is `shutil.which`, which seems like it could do what you want?

benchexec/tools/ultimate.py
```diff
@@ -1,12 +1,12 @@
 # This file is part of BenchExec, a framework for reliable benchmarking:
 # https://github.com/sosy-lab/benchexec
 #
-# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
-# SPDX-FileCopyrightText: 2015 Daniel Dietsch
+# SPDX-FileCopyrightText: 2015-2020 Daniel Dietsch <dietsch@informatik.uni-freiburg.de>
```
PhilippWendler, line 4: Please keep the existing copyright statements. The license requires that we "must cause any modified files to carry prominent notices stating that we changed the files", which we do by adding our copyright line. And there have and probably will be changes and contributions from us to these files. Furthermore, one could consider tool-info modules derivative works of BenchExec because they are tightly coupled and based on our API definitions and are typically inspired or have code copied from the template module or other tool-info modules.

Note that such a copyright declaration *does not change* who owns the copyright at all anyway (this is defined by law) and it thus does not diminish any rights that one might have, it is just a (hopefully true) description of facts. It also does not say anything about who owns copyrights to which parts or about a quantitative relationship or whether this is an exhaustive list. In this case it is just the statement that at least you and Dirk claim to hold at least some copyright over some content of this file.

(Although it does not strictly apply in this case, also note that if you redistribute files under the Apache license, you must keep all relevant existing copyright etc. notices, otherwise it is a license violation. So one needs to be really careful when removing such notices.)

Of course, updating the line with your copyright is totally fine.

benchexec/tools/ultimate.py
```diff
@@ -105,24 +108,21 @@ def _ultimate_version(self, executable):
...
         ]
 
         self.api = len(cmds)
         for cmd in cmds:
-            version = self._query_ultimate_version(cmd, self.api)
-            if version != "":
-                return version
+            for java_version, java in java_versions.items():
+                version = self._query_ultimate_version([java] + cmd, self.api)
+                if version != "":
```
PhilippWendler, line 118:
```suggestion
                if version:
```
This would also cover cases like `None` being returned.

PhilippWendler (changes requested):

benchexec/tools/ultimate.py
danieldietsch: The old copyright statement was added in https://github.com/sosy-lab/benchexec/commit/d0f03ac52be3ba95ace4c249b6d67cf487b18c26#diff-0a9af880936ddc6aa5d95be815747dd617133804c4b2a9bc3e9fc8443d33d0ec, where Dirk Beyer was also added (in particular with a strange timeframe that goes before the inception of the file itself).

We can add any additional contributor, of course. But perhaps some kind of rule when a contribution is significant enough to change the header would be nice to have.

I leave the question whether implementing an API is derivative work to someone else. I believe not ;)

PhilippWendler: > The old copyright statement was added in [d0f03ac#diff-0a9af880936ddc6aa5d95be815747dd617133804c4b2a9bc3e9fc8443d33d0ec](https://github.com/sosy-lab/benchexec/commit/d0f03ac52be3ba95ace4c249b6d67cf487b18c26#diff-0a9af880936ddc6aa5d95be815747dd617133804c4b2a9bc3e9fc8443d33d0ec), where Dirk Beyer was also added (in particular with a strange timeframe that goes before the inception of the file itself).

Yes, when we changed the header to be machine readable we also made sure to update the information in it to be more accurate where applicable (and for example also to handle the license requirement that we must notify about changed files).

The year in the copyright declaration has no legal purpose, it serves just for documentation. For simplicity we decided to just add the same range of years (that covers BenchExec and the origin of its source from before it became BenchExec). Otherwise we would need to investigate whether there is some content in each file that might have been copied from other older sources etc., but this is clearly not worth it.

> But perhaps some kind of rule when a contribution is significant enough to change the header would be nice to have.

This is a question that we would like to avoid in general, because we are no legal experts and cannot answer this (and its answer would change depending on where you are in the world). So if someone changes a file and requires their copyright to be added, we are likely to agree.

> I leave the question whether implementing an API is derivative work to someone else. I believe not ;)

It is irrelevant here, but the Linux kernel developers and the Linux distributions would disagree with you (this is why proprietary Linux kernel modules are not bundled with Linux kernels).

benchexec/tools/ultimate.py
danieldietsch: Its just so that either the debug message or the warn message is printed. But I can combine them. 

It makes sense that it continues because it tries different Java versions for different API versions and different wrapper script versions to be downward-compatible. 

The thing is: the Ultimate >=0.2.0 family does only run with Java 11 and higher, the ones before only with Java 8 and lower. But in order to get the version, I have to execute the tool. So I try and find all Java installations and try to execute the tool with all the different APIs. The first that succeeds wins.

PhilippWendler: > Its just so that either the debug message or the warn message is printed. But I can combine them.

I understand, but I would say that having both debug and warn message in case one enables debug messages it not a problem, so the code can be kept simple.

benchexec/tools/ultimate.py
danieldietsch: Is the new copyright header ok for you?

The question of "derivative work" does indeed not apply to copyright headers. The Linux kernel does not bundle proprietary Linux kernel modules because the licenses are incompatible.

PhilippWendler: No, please don't add other people of which you don't know whether they would like to be listed or not.

(And in fact, in this case it is even wrong, I do not own any copyright on parts of BenchExec, and neither does Karlheinz.)

danieldietsch:
I dont know why travis fails. I ran ``black``  version 19.10b0 . Perhaps my latest change will fix it.

PhilippWendler:
You can see a diff of what `black` would do here: https://travis-ci.org/github/sosy-lab/benchexec/jobs/738234120#L948 (A trailing comma only makes sense on multi-line parameter lists, so it enforces this.) But as long as everything fits on one line, it is nicer to keep it that way as you did. Now it won't complain.

benchexec/tools/ultimate.py
danieldietsch: Ok, I can remove you guys and keep only Dirk.

danieldietsch: But for the record: I do not believe that Dirk has a copyright on any of the files in this commit.

benchexec/tools/ultimateautomizer.py
```diff
@@ -1,8 +1,8 @@
 # This file is part of BenchExec, a framework for reliable benchmarking:
 # https://github.com/sosy-lab/benchexec
 #
-# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
```
PhilippWendler, line 4: This copyright line is still being removed.

benchexec/tools/ultimate.py
PhilippWendler: Luckily that question does not need to be solved now. But if you think that Karlheinz and me could claim copyright over parts of the code here (as indicated by f3e78d2), then you should also think the same about Dirk, because he owns the copyright for our contributions.

danieldietsch: Fair enough.

benchexec/tools/ultimate.py
```diff
@@ -105,24 +110,24 @@ def _ultimate_version(self, executable):
...
-            ],
+            ["-Xss4m", "-jar", launcher_jar, "-data", data_dir, "--version"],
         ]
 
         self.api = len(cmds)
         for cmd in cmds:
-            version = self._query_ultimate_version(cmd, self.api)
-            if version != "":
-                return version
+            for java_version, java in java_versions.items():
```
PhilippWendler, line 118: `flake8` just told me that `java_version` is not used here. Is this intended or did you intend to check only specific combinations of commands and Java versions?

If the former:
```suggestion
            for java in java_versions.values():
```

danieldietsch: Its more for debugging purposes. I need to try all the versions because I cannot determine which one will work without running it.

PhilippWendler (approved):
Thanks!

PR #645, opened 2020-10-23
Serialize filters to enable sharing via url

DennisSimon:
Hi @PhilippWendler , 

this PR implements the serialization and embedding of filters into the URL of the window. 
This enables us to share filters via links.


PhilippWendler (commented):
This is a cool feature!

I can see that the URL changes, but when I copied the URL to a local file with filters set and tried to open that in a new tab, it did not have an effect (in Firefox). Other URL parts (like open tab) do work, though, so I guess it is not related to local files?

I noticed that an example value for a trivial feature is this:
```
[{"id":"0_status_0","value":"wrong "},{"id":"0_status_0","value":"ERROR"},{"id":"0_status_0","value":"ERROR (interpolation failed)"},{"id":"0_status_0","value":"ERROR (parsing failed)"},{"id":"0_status_0","value":"ERROR (recursion)"},{"id":"0_status_0","value":"EXCEPTION"},{"id":"0_status_0","value":"FAILED (KILLED BY SIGNAL 1)"},{"id":"0_status_0","value":"OUT OF NATIVE MEMORY (true)"},{"id":"0_status_0","value":"TIMEOUT"},{"id":"0_status_0","value":"false"},{"id":"0_status_0","value":"false(unreach-call)"},{"id":"0_status_0","value":"true"},{"id":"0_status_0","value":"unknown"},{"id":"0_status_0","value":"correct "},{"id":"0_cputime_2","value":"37:"},{"id":"id","values":[]}]
```
It lists every single status value although I filtered only for category. Furthermore, the data structure is quite redundant, wouldn't a list of values per column make more sense?

And is there any other way instead of encoding the full string as base64 (which doubles its already long length and makes it unreadable)? Something like only encoding those string parts where it is needed or so?

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -102,16 +108,28 @@ export default class Overview extends React.Component {
 
   componentDidMount() {
     window.addEventListener("popstate", this.updateHiddenCols, false);
+    window.addEventListener("popstate", this.updateFilters, false);
```
lachnerm, line 110: Not a really big deal, but I would suggest using a single handler `updateUrlParams` or something similar, that combines both `updateHiddenCols` and `updateFilters`.

lachnerm, line 110: Not a really big deal, but I would suggest using a single handler `updateUrlParams` or something similar, that combines both `updateHiddenCols` and `updateFilters`. This would make the code a little bit more structured and less redundant, as we don't need to append two event handlers that each call `setState` and initialize a rerendering on their own.

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -82,6 +87,7 @@ export default class Overview extends React.Component {
       filtered: [],
       tabIndex: 0,
       filterBoxVisible: false,
+      filter: getFilterParamsFromUrl(),
```
lachnerm, line 89: This attribute `filter` is only initialized during the first rendering of the component but isn't used any further afterwards.

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -144,6 +162,7 @@ export default class Overview extends React.Component {
   };
   filterPlotData = (filter, runFilterLogic = true) => {
     console.log({ filter });
```
lachnerm, line 157:
```suggestion
```

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -102,16 +108,28 @@ export default class Overview extends React.Component {
...
+      this.filterPlotData(this.state.filter);
+    } */
   }
 
   updateHiddenCols = () => {
     this.setState({ hiddenCols: createHiddenColsFromURL(this.state.tools) });
   };
 
+  updateFilters = () => {
+    this.setState({ filters: getFilterParamsFromUrl() });
```
lachnerm, line 123: I think there may be a small typo here. You are initializing the attribute `filter` in the constructor, but updating the attribute `filters` here in the handler - those two probably should be the same. Also this `filters` attribute here isn't used anywhere else in the code.

lachnerm (commented):
The idea of the feature is very nice, however it doesn't work for me so far. I can confirm the same in Google Chrome as already stated by @PhilippWendler. The encoded filters are set in the URL but there seems to still be a problem when trying to apply the filters that were read from the URL. This is very likely due to the new state attributes not being used anywhere else in the code (see comments).

I'm not sure on how much information for a filter needs to be stored, but would it be possible to use readable strings for the URL? Being able to see which filters are applied from the URL (like with the other URL parameters) might be nice to have if it's possible.

DennisSimon:
Hi @PhilippWendler ,
hi @lachnerm , 

thanks for your reviews. 
I believe i have now resolved any issues reported in the review comments. 
I also have greatly changed and (hopefully) enhanced the syntax used for filtering via the URL. 
The currently implementation should enhance readability and reduce unnecessary text as much as possible. 

See the attached snapshot: 
newfilters

I'm looking forward to your reviews and comments!

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -192,6 +192,355 @@ const setHashSearch = (
...
+        filter = `value(${escape(filters.value)})`;
+      }
+      if (filter !== "") {
+        columnFilters.push(`${columnFilterHeader}(${filter})`);
+      }
+    }
+    if (columnFilters.length > 0) {
+      runsetFilters.push(`${tool}(${columnFilters.join(",")})`);
+    }
+  }
```
lachnerm, line 348: Is there maybe a way to implement this part a little bit more compact? The large if-else tree makes this a little bit hard to read and possibly harder to maintain in the future. From just looking through it, I couldn't think of a better way immediately, but since you've implemented the URL filters yourself, maybe you can think of something. If not, it's also fine though.

lachnerm (commented):
Looks good! The URLs can become quite large with multiple filters, but I guess there's nothing we can do about it regarding the amount of information that needs to be stored. I've tested with different tables and everything has been working for me!

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -192,6 +192,355 @@ const setHashSearch = (
...
+      if (value.endsWith(" ")) {
+        // category value
+        const selectedCategoryValues = columnBucket.categoryValues || [];
+        selectedCategoryValues.push(value);
+        columnBucket.categoryValues = selectedCategoryValues;
+      } else {
+        // status value
+        const selectedStatusValues = columnBucket.statusValues || [];
+        selectedStatusValues.push(value);
+        columnBucket.statusValues = selectedStatusValues;
```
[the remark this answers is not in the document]
DennisSimon: I think you are making a good point here. 
The initial reason why i went for the current approach is because i wanted to make the column type (value vs id vs status) identifiable by its defined properties. Just like a couple of lines down in the else block we assign a value to the `value` property and thereby define it. 
In OOP we would properly use different implementations of a interface or abstract structure that we could then either type check in the code or invoke a type-specific member function.

Having a `statusValues` and `categoryValues` property assigned to every filter might, at least in my immediate opinion, be a little bit misleading as there might be filters without a `value` or `values` property for example. Also, we wouldn't be able to identify a status filter if all status or category values have been deselected (even though this is only an edge-case).

I do however agree that this might not be the cleanest or most obvious solution. What do you think about introducing a `type` property that we could check for in the serialization block below? That way we wouldn't need to make assumptions over the type of a filter depending on if properties exist or if they hold values but by comparing enums.

benchexec/tablegenerator/react-table/src/utils/utils.js
DennisSimon: Hmmm... i also don't see an immediate way of making this logic more compact.
I extracted the `<statusColumnFilter>` into a separate function to make the overall control-flow more obvious. 
We could also try to flatten the tree a bit but i'm not sure if that would result in more readable code. 
e.g. `if(hasStatusFilter&&hasStatusUnchecked){...}; if(hasStatusFilter&&!hasCategoryFilter){...}` 

The other if statements like `if (filter !== "")` exist to reduce the length of the resulting serialized filter so i wouldn't want to remove them. 

If you think the code might be really hard to understand for future contributors though then i will 
try to rework this section of course.

PhilippWendler: I agree with both of you :-)

Luckily, this part of the code is separate from the rest and could also be refactored later if we find a good way, there is no need to delay merging due to this.

benchexec/tablegenerator/react-table/src/utils/utils.js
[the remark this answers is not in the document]
PhilippWendler: Again I agree with both of you :-)

PhilippWendler:
Everything seems to work nicely, even opening files/URLs with attached filter values in the URL, except for the back button. I tested Firefox and Chromium with local files and URLs, and the back button changes the URL but not the actually applied filter. Could you please investigate @DennisSimon?

DennisSimon:
> Everything seems to work nicely, even opening files/URLs with attached filter values in the URL, except for the back button. I tested Firefox and Chromium with local files and URLs, and the back button changes the URL but not the actually applied filter. Could you please investigate @DennisSimon?

Hi @PhilippWendler, sure that should be an easy fix. 
I will look into that before noon today.

DennisSimon:
Hi @PhilippWendler , can you check if it now works as expected?

benchexec/tablegenerator/react-table/src/utils/utils.js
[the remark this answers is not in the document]
lachnerm: > Having a `statusValues` and `categoryValues` property assigned to every filter might, at least in my immediate opinion, be a little bit misleading as there might be filters without a `value` or `values` property for example. Also, we wouldn't be able to identify a status filter if all status or category values have been deselected (even though this is only an edge-case).
> 
I see your point. I agree that if it's possible that some filters don't have those properties, defining them for every filter wouldn't be a good implementation.
> I do however agree that this might not be the cleanest or most obvious solution. What do you think about introducing a `type` property that we could check for in the serialization block below? That way we wouldn't need to make assumptions over the type of a filter depending on if properties exist or if they hold values but by comparing enums.
>
This would also make it possible to use a single attribute like `columnBucket.value` to store the values instead of using three different ones, right? This sounds like a good solution in my opinion.

PhilippWendler:
The back button works nicely now, thanks!

I found two other smaller issues:

- When I configure a filter for id and something else (e.g., `#/table?filter=id(values(,true)),0(0*status*(category(notIn(error))))` - note how useful this feature is :-) ) and then click on the "delete all filters" trash can, only the non-id filter gets deleted. A second click on the trash can also deletes the id filter. I could not reproduce this with current master.
- When I open a table with a filter in the URL (same value as above, for example), I can first see the table rendered with all tasks, and then it is re-rendered with the filter applied. Is it possible to avoid this flash of unwanted content? Interestingly, if I configure hidden columns, even the first temporary rendering correctly shows only the desired columns.

PhilippWendler:
The two issues reported before are now fixed, thanks!

However, now the back button of the browser (and also manual editing of the URL) does not work anymore, both in Firefox and Chromium. To reproduce, open any table, set a simple filter and click "back". This has worked before as mentioned above. Could you please fix this?

DennisSimon:
Hi @PhilippWendler ,

i have re-enabled the functionality of recalculating filters when the back button has been clicked. 
This caused me to dig a bit deeper and i found a couple of issues with our current implementation of handling the back button. 

Currently the `popstate` event is called whenever we have any route change. This is due to us managing state in the URL
by modifying the `location` object. 

I have modified the URL setter functions to push to the history stack instead of modifying the location object. 
Also, to ensure maximum consistency between browsers, i subscribed directly to the routers `history` object to listen for `POP`
events.

PhilippWendler:
Thanks, now it seems to work.

I noticed that the reaction to the back button seems slightly slower now, specifically when going back from another tab to the "table" tab. Is this just my impression?

@lachnerm Could you please have a look at the new history handling? Would be good to get a second pair of eyes for this, and I do not know enough about this.

If there are no objections from @lachnerm we are ready to merge :-)

DennisSimon:
> I noticed that the reaction to the back button seems slightly slower now, specifically when going back from another tab to the "table" tab. Is this just my impression?


Hmm there is a high possibility that it might be slower. 
I noticed that we are now re-running filters on every `POP` event. 
I will add some validation so we only re-run filtering logic whenever the filter payload actually changed.

PhilippWendler:
Now hiding columns does not work anymore: Open a table, switch to table tab, hide columns - URL gets extended but nothing happens (tested in Firefox). When I switch to a URL that hides columns (e.g., when using back and forward buttons or open such a URL directly), the column hiding is applied.

PhilippWendler (approved):

lachnerm:
> @lachnerm Could you please have a look at the new history handling? Would be good to get a second pair of eyes for this, and I do not know enough about this.
> 
From looking at the changes, I think the new history handling is generally a good approach. We should however be consistent on how we handle the history changes. Currently, the `LinkOverlay`, `SelectColumn`, `ScatterPlot` and `QuantilePlot` components are all still using the `popstate` event. While we probably shouldn't run into problems regarding the first 2 components (we only use the event here to close the overlay when the back button was clicked), the latter ones use the event to write to or read from the URL. I've played around a bit and so far everything seems to work fine and maybe this will actually always work hand in hand, but I wouldn't risk anything breaking later on. Also it's a little bit confusing to have both approaches at the same time, our `setHashSearch` function does two different things depending on whether the `history` parameter is defined, while both procedures more or less result in the same thing.

So basically I'd suggest passing the `this.routerRef.current.history` object from the `Overview` to the other components too, completely remove the `popstate` event listeners and handle everything consistently, the same way we do it in the `Overview` and `LinkOverlay` component. We could do this separately too though.

And there's one additional bug that I encountered:
Hiding columns via the `SelectColumn` component works, but undoing the hiding doesn't work - at least if I try to undo all hiding. Displaying single or multiple columns works, but as soon as the changes would result in all columns being shown again, the changes aren't applied any more and the developer console shows the following error message:
`react_devtools_backend.js:2430 Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack`.
Additionally, when hiding all columns, it's not possible to display them any more. Even when selecting only a single column to be shown again, all columns stay hidden. I noticed that this can show strings like `hidden=0&hidden0=1,2,3,4,5` in the URL, which was not possible before. There was always either `hidden=0` (which means the whole runset 0 is hidden) or `hidden0=1,2,3,4,5` (which means columns 1 to 5 of runset 0 are hidden) defined. The first parameter overwrites everything from the second one, so it shouldn't be possible for both to be defined.

PhilippWendler:
Thanks! I think we can do the refactoring of the state handling later (I agree that it should be consistent), but the bug should be fixed now. @DennisSimon Do you have time to work on this through the week? I would like to make a release towards the end of the week.

DennisSimon:
> Thanks! I think we can do the refactoring of the state handling later (I agree that it should be consistent), but the bug should be fixed now. @DennisSimon Do you have time to work on this through the week? I would like to make a release towards the end of the week.

Hi @PhilippWendler i'll work on it this morning and should get it done before noon.

DennisSimon:
Hi @PhilippWendler, hi @lachnerm , 

I have addressed the bug that was report by you. 
It was caused by the newly added "keepOthers" functionality that i added in order to deal with the handler of hidden cols adding/modifying multiple query params at the same time, which were overriding the serialized filter. 
We now need to explicitely set the parameters to null or undefined to remove them (when using "keepOthers"). 

Regarding the history handling: 
That is a very good point, i think we should definitely refactor it soon. 
I was thinking about creating a higher order component that would inject listener functions into components as we need to add a little bit of set-up to deal with `history` not being available on first render. 
By moving that logic into a hoc, we would be able to keep the code dryer. 
What do you think about this approach @lachnerm?

PhilippWendler (approved):
Thanks! I am really glad about this feature!

PR #646, opened 2020-10-25
Update several tool-info modules to BaseTool2

skanav:
updated tool info modules of verification (and validation) tools from sv-comp 20 to start using data models from the task dict.
this is required for the competition after [PR 169](https://github.com/sosy-lab/sv-comp/pull/169) in benchmark definitions repo is merged

The following are still remaining: 
1. ultimate based tools - it is a bit more complicated. Should be dealt separately

### DIFF
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -45,17 +45,14 @@ class Tool(benchexec.tools.template.BaseTool2):
     def determine_result(self, run):
         status = result.RESULT_ERROR
 
-        for line in run.output._lines:
-            if line == "VERIFICATION SUCCESSFUL\n":
+        for line in run.output:
+            if line == "VERIFICATION SUCCESSFUL":
                 status = result.RESULT_TRUE_PROP
                 break
-            elif line == "VERIFICATION FAILED\n":
+            elif line == "VERIFICATION FAILED":
                 status = result.RESULT_FALSE_REACH
                 break
-            elif (
-                line == "VERIFICATION UNKNOWN\n"
-                or line == "VERIFICATION BOUNDED TRUE\n"
-            ):
+            elif line == "VERIFICATION UNKNOWN" or line == "VERIFICATION BOUNDED TRUE":
                 status = result.RESULT_UNKNOWN
                 break
 
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from benchexec.tools.sv_benchmarks_util import get_data_model_from_task
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
 
@@ -25,9 +25,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "BRICK"
 
     def cmdline(self, executable, options, task, rlimits):
-        data_model_param = get_data_model_from_task(
-            task, {"ILP32": "--32", "LP64": "--64"}
-        )
+        data_model_param = get_data_model_from_task(task, {ILP32: "--32", LP64: "--64"})
         if data_model_param and data_model_param not in options:
             options += [data_model_param]
 
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -10,7 +10,7 @@ import benchexec.tools.template
 import benchexec.result as result
 
 
-class Tool(benchexec.tools.template.BaseTool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
     Tool info for BRICK
     https://github.com/brick-tool-dev/brick-tool
@@ -18,14 +18,20 @@ class Tool(benchexec.tools.template.BaseTool):
 
     REQUIRED_PATHS = ["bin", "lib"]
 
-    def executable(self):
-        return util.find_executable("bin/brick")
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("brick", subdir="bin")
 
     def name(self):
         return "BRICK"
 
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = util.get_data_model_from_task(
+            task, {"ILP32": "--32", "LP64": "--64"}
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
+        return [executable] + options + list(task.input_files_or_identifier)
 
     def version(self, executable):
         return self._version_from_tool(executable, arg="--version")
@@ -36,10 +42,10 @@ class Tool(benchexec.tools.template.BaseTool):
             executable, paths, parent_dir=True
         )
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
         status = result.RESULT_ERROR
 
-        for line in output:
+        for line in run.output._lines:
             if line == "VERIFICATION SUCCESSFUL\n":
                 status = result.RESULT_TRUE_PROP
                 break
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import benchexec.util as util
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task
 import benchexec.tools.template
 import benchexec.result as result
 
@@ -25,7 +25,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "BRICK"
 
     def cmdline(self, executable, options, task, rlimits):
-        data_model_param = util.get_data_model_from_task(
+        data_model_param = get_data_model_from_task(
             task, {"ILP32": "--32", "LP64": "--64"}
         )
         if data_model_param and data_model_param not in options:
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -109,11 +109,10 @@ class Tool(benchexec.tools.template.BaseTool2):
         return status
 
     def determine_result(self, run):
-        returnsignal = run.exit_code.signal or 0
         returncode = run.exit_code.value or 0
         output = run.output
 
-        if returnsignal == 0 and ((returncode == 0) or (returncode == 10)):
+        if not run.exit_code.signal and ((returncode == 0) or (returncode == 10)):
             status = result.RESULT_ERROR
             if "--xml-ui" in self.options:
                 status = self.parse_XML(output, returncode, run.was_timeout)
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -48,7 +48,7 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         return [executable] + options + list(task.input_files_or_identifier)
 
-    def parse_XML(self, output, returncode, isTimeout):
+    def parse_XML(self, output, exit_code, isTimeout):
         # an empty tag cannot be parsed into a tree
         def sanitizeXML(s):
             return s.replace("<>", "<emptyTag>").replace("</>", "</emptyTag>")
@@ -78,7 +78,7 @@ class Tool(benchexec.tools.template.BaseTool2):
                     status = "INVALID OUTPUT"
 
             elif status == "FAILURE":
-                assert returncode == 10
+                assert exit_code.value == 10
                 reason = tree.find("goto_trace").find("failure").findtext("reason")
                 if not reason:
                     reason = tree.find("goto_trace").find("failure").get("reason")
@@ -88,7 +88,7 @@ class Tool(benchexec.tools.template.BaseTool2):
                     status = result.RESULT_FALSE_REACH
 
             elif status == "SUCCESS":
-                assert returncode == 0
+                assert exit_code.value == 0
                 if "--unwinding-assertions" in self.options:
                     status = result.RESULT_TRUE_PROP
                 else:
@@ -103,19 +103,18 @@ class Tool(benchexec.tools.template.BaseTool2):
             else:
                 status = "INVALID OUTPUT"
                 logging.exception(
-                    "Error parsing CBMC output for returncode %d", returncode
+                    "Error parsing CBMC output for exit_code.value %d", exit_code.value
                 )
 
         return status
 
     def determine_result(self, run):
-        returncode = run.exit_code.value or 0
         output = run.output
 
-        if not run.exit_code.signal and ((returncode == 0) or (returncode == 10)):
+        if run.exit_code.value in [0, 10]:
             status = result.RESULT_ERROR
             if "--xml-ui" in self.options:
-                status = self.parse_XML(output, returncode, run.was_timeout)
+                status = self.parse_XML(output, run.exit_code.value, run.was_timeout)
             elif len(output) > 0:
                 # SV-COMP mode
                 result_str = output[-1].strip()
@@ -138,13 +137,13 @@ class Tool(benchexec.tools.template.BaseTool2):
                 elif "UNKNOWN" in output:
                     status = result.RESULT_UNKNOWN
 
-        elif returncode == 64 and "Usage error!\n" in output:
+        elif run.exit_code.value == 64 and "Usage error!\n" in output:
             status = "INVALID ARGUMENTS"
 
-        elif returncode == 6 and "Out of memory\n" in output:
+        elif run.exit_code.value == 6 and "Out of memory\n" in output:
             status = "OUT OF MEMORY"
 
-        elif returncode == 6 and "SAT checker ran out of memory\n" in output:
+        elif run.exit_code.value == 6 and "SAT checker ran out of memory\n" in output:
             status = "OUT OF MEMORY"
 
         else:
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -8,7 +8,7 @@
 import logging
 from xml.etree import ElementTree
 
-import benchexec.util as util
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task
 import benchexec.tools.template
 import benchexec.result as result
 
@@ -38,7 +38,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         elif "--xml-ui" not in options:
             options = options + ["--xml-ui"]
 
-        data_model_param = util.get_data_model_from_task(
+        data_model_param = get_data_model_from_task(
             task, {"ILP32": "--32", "LP64": "--64"}
         )
         if data_model_param and data_model_param not in options:
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -13,7 +13,7 @@ import benchexec.tools.template
 import benchexec.result as result
 
 
-class Tool(benchexec.tools.template.BaseTool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
     Tool info for CBMC (http://www.cprover.org/cbmc/).
     It always adds --xml-ui to the command-line arguments for easier parsing of
@@ -23,8 +23,8 @@ class Tool(benchexec.tools.template.BaseTool):
 
     REQUIRED_PATHS = ["cbmc", "cbmc-binary", "goto-cc"]
 
-    def executable(self):
-        return util.find_executable("cbmc")
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("cbmc")
 
     def version(self, executable):
         return self._version_from_tool(executable)
@@ -32,15 +32,21 @@ class Tool(benchexec.tools.template.BaseTool):
     def name(self):
         return "CBMC"
 
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        if propertyfile:
-            options = options + ["--propertyfile", propertyfile]
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options = options + ["--propertyfile", task.property_file]
         elif "--xml-ui" not in options:
             options = options + ["--xml-ui"]
 
+        data_model_param = util.get_data_model_from_task(
+            task, {"ILP32": "--32", "LP64": "--64"}
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
         self.options = options
 
-        return [executable] + options + tasks
+        return [executable] + options + list(task.input_files_or_identifier)
 
     def parse_XML(self, output, returncode, isTimeout):
         # an empty tag cannot be parsed into a tree
@@ -102,12 +108,15 @@ class Tool(benchexec.tools.template.BaseTool):
 
         return status
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
+        returnsignal = run.exit_code.signal or 0
+        returncode = run.exit_code.value or 0
+        output = run.output
 
         if returnsignal == 0 and ((returncode == 0) or (returncode == 10)):
             status = result.RESULT_ERROR
             if "--xml-ui" in self.options:
-                status = self.parse_XML(output, returncode, isTimeout)
+                status = self.parse_XML(output, returncode, run.was_timeout)
             elif len(output) > 0:
                 # SV-COMP mode
                 result_str = output[-1].strip()
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -8,7 +8,7 @@
 import logging
 from xml.etree import ElementTree
 
-from benchexec.tools.sv_benchmarks_util import get_data_model_from_task
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
 
@@ -38,9 +38,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         elif "--xml-ui" not in options:
             options = options + ["--xml-ui"]
 
-        data_model_param = get_data_model_from_task(
-            task, {"ILP32": "--32", "LP64": "--64"}
-        )
+        data_model_param = get_data_model_from_task(task, {ILP32: "--32", LP64: "--64"})
         if data_model_param and data_model_param not in options:
             options += [data_model_param]
 
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -114,7 +114,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         if run.exit_code.value in [0, 10]:
             status = result.RESULT_ERROR
             if "--xml-ui" in self.options:
-                status = self.parse_XML(output, run.exit_code.value, run.was_timeout)
+                status = self.parse_XML(output, run.exit_code, run.was_timeout)
             elif len(output) > 0:
                 # SV-COMP mode
                 result_str = output[-1].strip()
@@ -137,13 +137,13 @@ class Tool(benchexec.tools.template.BaseTool2):
                 elif "UNKNOWN" in output:
                     status = result.RESULT_UNKNOWN
 
-        elif run.exit_code.value == 64 and "Usage error!\n" in output:
+        elif run.exit_code.value == 64 and "Usage error!" in output:
             status = "INVALID ARGUMENTS"
 
-        elif run.exit_code.value == 6 and "Out of memory\n" in output:
+        elif run.exit_code.value == 6 and "Out of memory" in output:
             status = "OUT OF MEMORY"
 
-        elif run.exit_code.value == 6 and "SAT checker ran out of memory\n" in output:
+        elif run.exit_code.value == 6 and "SAT checker ran out of memory" in output:
             status = "OUT OF MEMORY"
 
         else:
--- a/benchexec/tools/cseq.py
+++ b/benchexec/tools/cseq.py
@@ -29,13 +29,13 @@ class CSeqTool(benchexec.tools.template.BaseTool2):
         are either absolute or have been made relative to the designated working directory.
         @param executable: the path to the executable of the tool (typically the result of executable())
         @param options: a list of options, in the same order as given in the XML-file.
-        @param tasks: instance of class Task containg the property and input file
+        @param task: instance of class Task containg the property and input file
                             This tool info module only supports one input file.
         @param rlimits: This dictionary contains resource-limits for a run,
                         for example: time-limit, soft-time-limit, hard-time-limit, memory-limit, cpu-core-limit.
                         All entries in rlimits are optional, so check for existence before usage!
         """
-        spec = ["--spec", task.property_file] if task.property_file is not None else []
+        spec = ["--spec", task.property_file] if task.property_file else []
         return [executable] + options + spec + ["--input", task.single_input_file]
 
     def determine_result(self, run):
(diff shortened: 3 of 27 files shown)


danieldietsch:
Should I update the Ultimate tools? How soon is it needed?

benchexec/tools/brick.py
```diff
@@ -10,22 +10,28 @@
...
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = util.get_data_model_from_task(
+            task, {"ILP32": "--32", "LP64": "--64"}
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
+        return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 34: This is quite generic. Does the tool actually support tasks with more than one input file or no input file? If not, it would be much better for user experience if the tool-info module produced a proper error in such cases by using `task.input_files` or `task.single_input_file` as appropriate.

benchexec/tools/cseq.py
```diff
@@ -36,13 +36,14 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
...
         assert len(tasks) == 1, "only one inputfile supported"
         inputfile = ["--input", tasks[0]]
-        spec = ["--spec", propertyfile] if propertyfile is not None else []
+        spec = ["--spec", task.property_file] if task.property_file is not None else []
         return [executable] + options + spec + inputfile
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
-        output = "\n".join(output)
+    def determine_result(self, run):
+        output = "\n".join(run.output)
```
PhilippWendler, line 46: Please use `run.output.text` instead.

However, note that this produces different behavior than the original code: Because previously `output` contained lines with line separators, the resulting joined text would actually contain doubled line-separators (i.e., a blank line between all original lines). It is unlikely that this was intended and so the new code would actually be better, but it should be kept in mind.

Also, can the usage of `run.output.text` maybe be avoided completely? The class `RunOutput` is written specifically to provide utilities that are nice to use. For example, `if "FALSE" in output:` in this method is equivalent to `if run.output.any_line_contains("FALSE"):`.

benchexec/tools/divine.py
```diff
@@ -87,15 +97,15 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
 
-        last = output[-1]
+        last = run.output[-1]
 
-        if isTimeout:
+        if run.was_timeout:
             return "TIMEOUT"
 
-        if returncode != 0:
+        if run.exit_code.value != 0:
```
PhilippWendler, line 108: Note that this changes behavior compared to the old code. Previously , `returncode` was 0 if the tool was killed by a signal, e.g., due to a timeout, now it is `None`. Is this change intended?

skanav:
> Should I update the Ultimate tools? 

This would be great.

> How soon is it needed?

It would be needed for the preruns. So it should be done soon.

benchexec/tools/cbmc.py
```diff
@@ -102,12 +108,15 @@ def isErrorMessage(msg):
 
         return status
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
+        returnsignal = run.exit_code.signal or 0
```
PhilippWendler, line 112: The fact that previously `returnsignal` and `returncode` were 0 by default is actually quite bad API and makes the code complex to understand (because for example `returncode==0` means that the tool exited successfully, but only if `returnsignal` is not `None` at the same time). In general, handling the new values (with default `None`) is easier and less errorprone. Thus I would suggest to refactor the code to directly use the fields of `run.exit_code` as they are, preferably even by passing the `ProcessExitCode` instance around instead of individual values.

benchexec/tools/esbmc.py
```diff
@@ -31,13 +31,20 @@ def version(self, executable):
...
         assert len(tasks) == 1, "only one inputfile supported"
         inputfile = tasks[0]
-        return [executable] + ["-p", propertyfile] + options + [inputfile]
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
-        output = "\n".join(output)
+        data_model_param = util.get_data_model_from_task(
+            task, {"ILP32": "32", "LP64": "64"}
+        )
+        if data_model_param and data_model_param not in options:
```
PhilippWendler, line 42: I think this is errorprone, because the `--arch` parameter is added only if there is no item with the value `32` (or `64` respectively) in the command line. Such a value could appear as content of a different parameter, though. Furthermore, if a user adds `--arch` in their benchmark definition with a different value than what this code would produce, the parameter gets added twice.

benchexec/tools/symbiotic.py
```diff
@@ -25,19 +25,21 @@ class Tool(OldSymbiotic):
...
         This method always needs to be overridden,
         and most implementations will look similar to this one.
         The path returned should be relative to the current directory.
         """
-        exe = util.find_executable("bin/symbiotic", exitOnError=False)
-        if exe:
-            return exe
-        else:
+        # This is used in the function _version_newer_than.
+        # I don't know a better way to do it.
```
PhilippWendler, line 36: Such a comment should not be part of the code. If you have questions, please mark the PR as draft and add the questions to the PR discussion instead.

benchexec/tools/symbiotic.py
```diff
@@ -25,19 +25,21 @@ class Tool(OldSymbiotic):
...
+        self.tool_locator = tool_locator
```
PhilippWendler, line 37: `tool_locator` is probably better to not be stored, there is nothing in the API documentation that guarantees that this would work as expected. Furthermore, searching for the tool on disk for every call to `determine_result` is inefficient and not robust (disk content could change).

But the documentation of `BaseTool2` provides a suggestion what to do in this case already.

benchexec/tools/symbiotic.py
```diff
@@ -102,12 +104,15 @@ def _getPhase(self, output):
 
         return lastphase
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
-        if output is None:
+    def determine_result(self, run):
+        returnsignal = run.exit_code.signal or 0
+        returncode = run.exit_code.value or 0
+
+        if run.output is None:
```
PhilippWendler, line 111: Note that the API documentation guarantees that this is wrong. If you want to check for empty output, this needs to be done differently.

benchexec/tools/two_ls.py
```diff
@@ -42,8 +50,8 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
         elif returncode == 0:
             status = result.RESULT_TRUE_PROP
         elif returncode == 10:
-            if len(output) > 0:
-                result_str = output[-1].strip()
+            if len(run.output) > 0:
```
PhilippWendler, line 53: Should be written as `if run.output:`

PhilippWendler (changes requested):
Thanks!

I have a few comments that I added. Many of these actually apply to more than one tool-info module but I always added them in only one location, so please check for each comment in which tool-info modules it also applies.

benchexec/util.py
```diff
@@ -762,3 +762,22 @@ def check_msr():
         if all(os.access("/dev/cpu/{}/msr".format(cpu), os.W_OK) for cpu in cpu_dirs):
             res["write"] = True
     return res
+
+
+def get_data_model_from_task(task, data_models):
```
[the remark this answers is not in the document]
skanav: That's what I was wondering too. I thought adding a separate `svcomp_utils` module in `tools` a bit of overkill.

I can add a `svcomp_util` in the `tools` package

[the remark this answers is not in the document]
PhilippWendler: Maybe it makes more sense to add an abstract class for SV-COMP tools?

benchexec/tools/brick.py
skanav: I have no idea if this tool or other tools with similar changes support single or multiple files . I am not a developer of this tool.
I did the changes that were upgrade the TI module without going into tool specific details.

IMO the change you suggest should be done by the tool developers.

benchexec/tools/symbiotic.py
skanav: fixed in fb5a18e4382326ea40bdaa4ebfb99b456f6a0ebb

benchexec/tools/divine.py
skanav: `None != 0` also evaluates to true

benchexec/util.py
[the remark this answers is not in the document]
skanav: We can add an abstract class named `CompetitionTool` because this change is relevant also for `TestComp`.

This change is is relevant for tools dealing with the language `C`, not `Java`. 
Shall we call it `CompetitionToolsForC`?

Where do you want to place this class? Separate file or `template.py`?

benchexec/tools/divine.py
PhilippWendler: Exactly, this is why there is a difference to the previous behavior. The branch will now be entered if `value` is `None`, previously it was not entered if `value` was 0.

benchexec/tools/symbiotic.py
skanav: fixed in b28f883d752cc0656615bceae0bf4a5d0eeedc08

benchexec/tools/two_ls.py
skanav: fixed in 2b4a6a1fbf69d98062f433ab690e55db316abb02

benchexec/util.py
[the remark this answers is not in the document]
PhilippWendler: Hm, `CompetitionTool` or something similar would sound like this affects all kinds of competitions. In fact, its content is not even related to SV-COMP, but to SV-Benchmarks (where the `data_model` key is [defined](https://github.com/sosy-lab/sv-benchmarks#task-definitions)). So a name pointing in that direction might make sense, or something related to `C`.

We certainly should not place this class in `template.py` to make it clear it is not required by BenchExec for and tool-info modules and that it is not part of the public API. Probably we should even clearly state that out-of-tree modules should not use it and that it may change at any point in time, otherwise we would also have to version it like we do with `BaseTool2`.

PhilippWendler:
Just remembered https://github.com/sosy-lab/sv-benchmarks/issues/1125, I think it would make sense to wait for a resolution of that issue before merging this PR.

benchexec/tools/brick.py
PhilippWendler: Note that there are cases where the tool-info module already makes it clear that only some of these cases are supported (with an `assert` or maybe in comments). In this case the code should be cleaned up.

benchexec/tools/cseq.py
skanav: fixed in f86559fb27bf6844cf5515dba93b021247b2aeeb  for cseq and predatorHP.

benchexec/util.py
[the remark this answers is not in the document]
skanav: IMO it is much simpler to put it in a util file. I will do that for now.

benchexec/tools/divine.py
skanav: made it similar to others in f2d7b96e04126b075a70bf66ecc07b64b0be151e

benchexec/tools/brick.py
```diff
@@ -36,10 +42,10 @@ def program_files(self, executable):
             executable, paths, parent_dir=True
         )
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
         status = result.RESULT_ERROR
 
-        for line in output:
+        for line in run.output._lines:
```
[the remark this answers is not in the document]
skanav: fixed in 0af3679ba62ae7f83a8d5d7277ea12c9a35df353

benchexec/tools/esbmc.py
skanav: fixed in fc874d47b6d1a031f0cc7403a0ebf0706670ceea

Now using the param name for searching in optipons

benchexec/tools/cbmc.py
skanav: fixed for value in 5823bfb9ed00b07a19413ae454100020ee3a5bfa

benchexec/util.py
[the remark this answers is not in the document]
skanav: @PhilippWendler do you agree with this?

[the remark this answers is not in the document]
PhilippWendler: I would say it should have a less generic name. `toolinfo_util` sounds like an official file with officially supported API addons.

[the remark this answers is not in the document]
skanav: do you have a suggestion?

I would like to reach an agreement before making the changes.

benchexec/tools/brick.py
skanav: fixed in 6b526edc4561f9c86b55b7b52dbca0621a9c287b

benchexec/util.py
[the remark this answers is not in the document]
PhilippWendler: `sv_benchmarks_util`?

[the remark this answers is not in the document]
skanav: fixed in 14cd4fc0477b9443c6b58a4824273510afdf076e

benchexec/tools/brick.py
```diff
@@ -10,22 +10,28 @@
...
+        return tool_locator.find_executable("brick", subdir="bin")
 
     def name(self):
         return "BRICK"
 
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = util.get_data_model_from_task(
+            task, {"ILP32": "--32", "LP64": "--64"}
```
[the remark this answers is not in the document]
skanav: I think we should wait for https://github.com/sosy-lab/sv-benchmarks/issues/1125 to be resolved before handling this.

benchexec/tools/map2check.py
```diff
@@ -25,35 +25,27 @@ class Tool(benchexec.tools.template.BaseTool):
...
-        # Relative path to map2check wrapper
-        if self._get_version() == 6:
-            return util.find_executable("map2check-wrapper.sh")
-        elif self._get_version() > 6:
-            return util.find_executable("map2check-wrapper.py")
-        assert False, "Unexpected version " + self._get_version()
+    def executable(self, tool_locator):
+        try:
+            executable = tool_locator.find_executable("map2check-wrapper.sh")
+            self.__version = 6
```
PhilippWendler, line 31: I would prefer `_version`, AFAIK in Python it is uncommon to use `__field`.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -0,0 +1,28 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This module contains some useful functions related to tasks in the sv-benchmarks
```
PhilippWendler, line 9: Please add a note about lack of guarantees of stability and no use for out-of-tree modules (as we discussed).

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -0,0 +1,28 @@
...
+repository: https://github.com/sosy-lab/sv-benchmarks
+"""
+
+import benchexec.tools.template
+
+def get_data_model_from_task(task, data_models):
```
PhilippWendler, line 16: Please add documentation. Shouldn't `data_models` be renamed because it does not primarily contain the data models, but command-line parameters (or values) for data models?

benchexec/tools/symbiotic4.py
```diff
@@ -46,24 +46,30 @@ def name(self):
...
 
-    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+    def cmdline(self, executable, options, task, rlimits):
         """
         Compose the command line to execute from the name of the executable
         """
 
-        if propertyfile is not None:
-            options = options + ["--prp={0}".format(propertyfile)]
+        if task.property_file is not None:
```
PhilippWendler, line 54: Can be simplified:
```suggestion
        if task.property_file:
```
Would be nice to use this consistently.

benchexec/tools/symbiotic4.py
```diff
@@ -46,24 +46,30 @@ def name(self):
...
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
-        if isTimeout:
+        return [executable] + options + list(task.input_files_or_identifier)
+
+    def determine_result(self, run):
+        if run.was_timeout:
             return "timeout"
 
-        if output is None:
+        if run.output is None:
```
PhilippWendler, line 69: This is never satisfied (same holds true for condition in previous code, actually).

benchexec/tools/pinaka.py
```diff
@@ -5,34 +5,40 @@
...
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
 
-        return [executable] + options + tasks
+        return [executable] + options + list(task.input_files_or_identifier)
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
+        output = run.output._lines
```
PhilippWendler, line 38: Please remove all accesses to `_lines` from all of the tool-info modules.

benchexec/tools/esbmc.py
```diff
@@ -31,13 +31,19 @@ def version(self, executable):
...
+            [executable]
+            + ["-p", task.property_file]
+            + options
+            + [task.single_input_file]
+        )
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
-        output = "\n".join(output)
+    def determine_result(self, run):
+        output = "\n".join(run.output)
```
PhilippWendler, line 46: Here is another case for using `run.output.text` (without additional blank lines) or `any_line_contains` (which actually seems easy to use here).

skanav: The tool developers use a function that they have defined themselves.
Although I understand that this could be made better but in general, I refrain from code cleanup enthusiasm for the piece of code that does not belong to me.

That was the reason I have kept it so, and also some other cases like this e.g., above comment in symbiotic.

benchexec/tools/symbiotic4.py
skanav: fixed in 78f718139b1300c094474898e74de1173ee09417

benchexec/tools/esbmc.py
skanav: fixed in 4cf9ec8e54d8a90a314ac265d5cbbbdb589e8851

benchexec/tools/pinaka.py
skanav: fixed in af5bd81bdaf51faa9778931901fcbb1b5f251175

benchexec/tools/symbiotic4.py
skanav: fixed in 66eb7e734da93d400ad5e4745dbe7b45409bf451

benchexec/tools/map2check.py
skanav: fixed in f2608c929d1e7ee7f86f1fb5f2674486fe00a4cb

benchexec/tools/sv_benchmarks_util.py
skanav: fixed in 6f11a4cb9c66b01effc2b0a591ce95e368f99631

benchexec/tools/sv_benchmarks_util.py
skanav: fixed in 6f11a4cb9c66b01effc2b0a591ce95e368f99631

benchexec/tools/cbmc.py
```diff
@@ -97,17 +103,18 @@ def isErrorMessage(msg):
...
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
+        output = run.output
 
-        if returnsignal == 0 and ((returncode == 0) or (returncode == 10)):
+        if run.exit_code.value in [0, 10]:
             status = result.RESULT_ERROR
             if "--xml-ui" in self.options:
-                status = self.parse_XML(output, returncode, isTimeout)
+                status = self.parse_XML(output, run.exit_code.value, run.was_timeout)
```
PhilippWendler, line 117: I think this is buggy, right? `_prase_XML` is expected to be passed an `ProcessExitCode` instance, but gets only the return code here.

benchexec/tools/cbmc.py
```diff
@@ -130,13 +137,13 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
                 elif "UNKNOWN" in output:
                     status = result.RESULT_UNKNOWN
 
-        elif returncode == 64 and "Usage error!\n" in output:
+        elif run.exit_code.value == 64 and "Usage error!\n" in output:
```
PhilippWendler, line 140: Lines with line separators won't ever be in `run.output`, so `\n` needs to be removed from these and other strings.

benchexec/tools/divine.py
PhilippWendler: I don't think the current code is the intended way for this tool-info module (and behaves differently than before). For OOMs, it would return `ERROR - Pre-run` instead of `UNKNOWN` as before.

benchexec/tools/divine.py
```diff
@@ -86,16 +96,15 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
             return "ERROR - Pre-run"
 
         if last is None:
```
PhilippWendler, line 110: This is actually guaranteed to never be the case, we could remove the branch while we are at it.

benchexec/tools/divine4.py
```diff
@@ -86,16 +94,17 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
         and should give some indication of the failure reason
         (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
         """
+        run.exit_code.value = run.exit_code.value or 0
```
PhilippWendler, line 97: This will not work, the field should be immutable.

benchexec/tools/klee.py
```diff
@@ -45,30 +45,32 @@ def version(self, executable):
...
+        if rlimits.memory:
+            options += ["--max-memory=" + str(rlimits.memory)]
+        if rlimits.cputime_hard:
+            options += ["--max-time=" + str(rlimits.cputime_hard)]
+        if rlimits.walltime:
+            options += ["--max-walltime=" + str(rlimits.walltime)]
+        if rlimits.cputime:
+            options += ["--max-cputime-soft=" + str(rlimits.cputime)]
+        if rlimits.cputime_hard:
+            options += ["--max-cputime-hard=" + str(rlimits.cputime_hard)]
```
PhilippWendler, line 60: This changes the semantics, because the old code had an unnoticed bug. `HARDTIMELIMIT` was never present in `rlimits`, so `--max-cputime-hard` was never passed. Now it would get added and always have the same value as `--max-cputime`.

@MartinNowack What is the intention here? BenchExec only has 2 time limits (the actual one and the hard time limit), so 2 parameters would seem to be enough? In which should the main time limit be passed?

benchexec/tools/legion.py
```diff
@@ -27,11 +27,20 @@ class Tool(benchexec.tools.template.BaseTool):
...
         return "Legion"
+
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = get_data_model_from_task(
+            task, {"ILP32": "-32", "LP64": "-64"}
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
+        return super().cmdline(executable, options, task, rlimits)
```
PhilippWendler, line 46: I would suggest to simply return the value here instead of calling `super`. Having the full picture of what is used as command line in one method seems worth more than saving one line of code.

benchexec/tools/libkluzzer.py
```diff
@@ -21,11 +21,20 @@ def program_files(self, executable):
...
         return "LibKluzzer"
+
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = get_data_model_from_task(
+            task, {"ILP32": "--32", "LP64": "--64"}
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
+        return super().cmdline(executable, options, task, rlimits)
```
PhilippWendler, line 40: Same as above.

benchexec/tools/pinaka.py
```diff
@@ -5,39 +5,44 @@
...
+        return [executable] + options + list(task.input_files_or_identifier)
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
         status = ""
 
-        if returnsignal == 0 and ((returncode == 0) or (returncode == 10)):
-            if "VERIFICATION FAILED (ReachSafety)\n" in output:
+        if run.exit_code.value in [0, 10]:
+            if run.output.any_line_contains("VERIFICATION FAILED (ReachSafety)"):
```
PhilippWendler, line 41: The exact semantics of the previous code would be
```suggestion
            if "VERIFICATION FAILED (ReachSafety)" in run.output:
```
(similar for the following conditions)

benchexec/tools/tracerx.py
```diff
@@ -49,30 +49,32 @@ def version(self, executable):
...
+            options += ["--property-file=" + task.property_file]
+        if rlimits.memory:
+            options += ["--max-memory=" + str(rlimits.memory)]
+        if rlimits.cputime_hard:
+            options += ["--max-time=" + str(rlimits.cputime_hard)]
+        if rlimits.walltime:
+            options += ["--max-walltime=" + str(rlimits.walltime)]
+        if rlimits.cputime:
+            options += ["--max-cputime-soft=" + str(rlimits.cputime)]
+        if rlimits.cputime_hard:
```
PhilippWendler, line 63: This should be handled in the same was as for Klee.

PhilippWendler (changes requested):

benchexec/tools/libkluzzer.py
skanav: fixed in 4b7464cc40a620880e02a86218a0eb1d0b26ac8c

benchexec/tools/legion.py
skanav: fixed in 4b7464cc40a620880e02a86218a0eb1d0b26ac8c

benchexec/tools/divine4.py
skanav: fixed in 53d68f15f868f1dca4a307f032076a0ddc5a43c3

benchexec/tools/divine.py
skanav: fixed in 53d68f15f868f1dca4a307f032076a0ddc5a43c3

benchexec/tools/cbmc.py
skanav: fixed in a6dbfcbc3fefc99d6a7a99e9a4040b7dee88e614

benchexec/tools/cbmc.py
skanav: fixed in a6dbfcbc3fefc99d6a7a99e9a4040b7dee88e614

benchexec/tools/pinaka.py
skanav: fixed in dfb784c34463564e65652465cd7bdde773d1d174

benchexec/tools/divine4.py
```diff
@@ -86,16 +94,15 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
 
-        last = output[-1]
+        last = run.output[-1]
 
-        if isTimeout:
+        if run.was_timeout:
             return "TIMEOUT"
 
-        if returncode != 0:
+        if run.exit_code.value != 0:
```
PhilippWendler, line 105: It seems this needs the same handling as for `divine`.

benchexec/tools/brick.py
```diff
@@ -10,22 +10,28 @@
...
+        )
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
```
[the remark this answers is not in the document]
MartinSpiessl: I find `options += [data_model_param]` way more readable than `options.append(data_model_param)`

MartinSpiessl:
> Just remembered sosy-lab/sv-benchmarks#1125, I think it would make sense to wait for a resolution of that issue before merging this PR.

Didn't seem like there is active work on sosy-lab/sv-benchmarks#1125, but I actually depend on this PR (doesn't the whole competition infrastructure need this?), so I would welcome if this could get merged quickly.

benchexec/tools/divine4.py
skanav: fixed in 59886567a1e060133a94b09ebc336b10a5336361

benchexec/tools/klee.py
skanav: I had a look at the `klee` wrapper submitted to sv-comp. It seems we can live with this, or even remove. Wouldn't matter much. 
This is how it is used:

```python
    time = 0
    if args.max_cputime_soft: # Use soft timeout if available
        time = args.max_cputime_soft
    elif args.max_cputime_hard:
        time = args.max_cputime_hard
    elif args.max_walltime:
        time = args.max_walltime
    elif args.max_time:
        time = args.max_time
```

PhilippWendler: Good idea for looking at the script!
So they only have one time-limit internally. This should then be set to BenchExec's main time limit (previously named "soft"), not the "hard" time limit, and actually this is also what happened with the previous code.

For the new code, we could simplify this completely to
```
        if rlimits.cputime:
            options += ["--max-time=" + str(rlimits.cputime_hard)]
```
`rlimits.cputime` will always be set if there is some time limit in BenchExec, and it will be set to the value that is relevant for the tool. Passing the other values is irrelevant if the script does not use them anyway I would say.

(Of course, `--max-memory` should be kept.)

@MartinNowack Do you agree?

skanav: I would say
```
        if rlimits.cputime:
            options += ["--max-cputime-soft=" + str(rlimits.cputime)]
```
because this is also what has the higher priority in the wrapper code for klee.

skanav: @PhilippWendler wouldn't BenchExec anyways kill the process at hard time limit?

PhilippWendler: Would also be fine for me, @MartinNowack would be best to decide this. If he does not respond, we can still do this.

PhilippWendler: > wouldn't BenchExec anyways kill the process at hard time limit?

Yes, this is what happens.

skanav: fixed in aaeae7166b8469eebf828b91ee9f6fe768e1afc7

if @MartinNowack prefers some other flag, it can be changed later.

benchexec/tools/brick.py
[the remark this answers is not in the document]
skanav: @PhilippWendler now I wonder if we should wait for https://github.com/sosy-lab/sv-benchmarks/issues/1125 to be resolved, or just go ahead with this and incorporate the changes later.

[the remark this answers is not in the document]
PhilippWendler: I would say that refactoring the constants out into the utility module should be done, because it might make adjusting the tool-info modules to the result of sosy-lab/sv-benchmarks#1125 easier.

But we could avoid blocking this PR on sosy-lab/sv-benchmarks#1125 if there is no progress there.

MartinNowack:
From my side, changes for KLEE look good.

PhilippWendler:
@MartinNowack  Thanks for confirmation!

skanav:
@PhilippWendler do we need something more here?

PhilippWendler (approved):

PhilippWendler:
I had just hoped that we get some response for sosy-lab/sv-benchmarks#1125, but as discussed let's merge it now.

PhilippWendler:
@skanav I just [noticed](https://gitlab.com/sosy-lab/sv-comp/archives-2021/-/jobs/826667408) that this broke some other tool-info modules that inherit from modules changes in this PR. At least JBMC is affected. Would it be possible for you to have a look at which tool-info modules inherit from the ones changed here and adjust them (this should typically be small changes only)?

PR #647, opened 2020-10-26
Add tool-info module for frama-c-sv

MartinSpiessl:
We want to execute Frama-C as part of SV-COMP.
~~As such the tool-info
module needs to be updated. The previous version was never really
useful because previous efforts of running Frama-C in a portable
fashion where unsuccessful (due to hardcoded absolute paths).
Now we have a different approach where we rely on Frama-C being
installed on the benchmark system (from the official package sources)
and just call that executable.~~
For that we develop a wrapper around Frama-C named `frama-c-sv`.
We rely on Frama-C being installed on the
benchmark system and just call that executable.

The wrapper itself will:
- add some config selection based on the property
- add ACSL annotations where necessary (e.g. to mark reach_error)
- add a harness for the `__VERIFIER*` functions such that Frama-C
  can understand their SV-COMP-specific meaning.
- also support output of minimal witnesses.

~~and is developed at <https://gitlab.com/sosy-lab/software/frama-c-svcomp> (currently still private, will change soon).~~
and is developed at <https://gitlab.com/sosy-lab/software/frama-c-sv> (currently still private, will change soon).

~~If we want to have this as a separate tool-info module, we can
also do this, I have no strong feelings here.
@lembergerth would this rewrite be OK for you?~~
We decided to make it a separate tool-info module named `frama-c-sv`

### DIFF
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -5,7 +5,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import benchexec.util as util
 import benchexec.tools.template
 
 
@@ -29,7 +28,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return Tool._TOOL_NAME
 
     def cmdline(self, executable, options, task, rlimits):
-        cmd = [executable, "--program"] + list(task.input_files_or_identifier)
+        cmd = [executable, "--program"] + list(task.input_files)
         if task.property_file:
             cmd += ["--property", task.property_file]
         return cmd
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -16,7 +16,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     """
 
     REQUIRED_PATHS = ["."]
-    _TOOL_NAME = "frama-c-sv.py"
+    _TOOL_NAME = "frama-c-sv"
     _SCRIPT_NAME = _TOOL_NAME + ".py"
 
     def executable(self, _):
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -1,0 +1,43 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.util as util
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for Frama-C with a wrapper for usage in the SVCOMP.
+    URL: https://gitlab.com/sosy-lab/software/frama-c-sv
+    """
+
+    REQUIRED_PATHS = ["."]
+    _TOOL_NAME = "frama-c-sv.py"
+    _SCRIPT_NAME = _TOOL_NAME + ".py"
+
+    def executable(self, _):
+        return util.find_executable(_SCRIPT_NAME)
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return _TOOL_NAME
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [_SCRIPT_NAME, "--program"] + list(task.input_files_or_identifier)
+        if task.property_file:
+            cmd += ["--property", task.property_file]
+        return cmd
+
+    def determine_result(self, run):
+        return run.output[-1].split(":", maxsplit=2)[-1]
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -19,22 +19,17 @@ class Tool(benchexec.tools.template.BaseTool2):
     _TOOL_NAME = "frama-c-sv"
     _SCRIPT_NAME = _TOOL_NAME + ".py"
 
-    def executable(self, _):
-        return util.find_executable(_SCRIPT_NAME)
-
-    def program_files(self, executable):
-        return self._program_files_from_executable(
-            executable, self.REQUIRED_PATHS, parent_dir=True
-        )
+    def executable(self, tool_locator):
+        return tool_locator.find_executable(Tool._SCRIPT_NAME)
 
     def version(self, executable):
         return self._version_from_tool(executable)
 
     def name(self):
-        return _TOOL_NAME
+        return Tool._TOOL_NAME
 
     def cmdline(self, executable, options, task, rlimits):
-        cmd = [_SCRIPT_NAME, "--program"] + list(task.input_files_or_identifier)
+        cmd = [executable, "--program"] + list(task.input_files_or_identifier)
         if task.property_file:
             cmd += ["--property", task.property_file]
         return cmd


PhilippWendler:
I would say that if this is a third-party wrapper for Frama-C, the tool-info module needs to be separate, otherwise it would cause problems for every other user of BenchExec + Frama-C who happens to not use your wrapper. Also it needs to be made clear that this is not Frama-C per se, but a separate tool (that happens to be based on Frama-C).

lembergerth:
I know that other people got Frama-C to execute on a remote system, so my failure shouldn't mean that we have to remove the original tool-info module.

But I very much like this new approach, thanks for that effort.

MartinSpiessl:
We decided to make a separate tool and tool-info module named `frama-c-sv`. I just updated the PR description and force-pushed a new version where instead of overwriting the tool-info module `frama-c.py`, we create a separate one named `frama-c-sv.py`.

Thanks everyone for the input!

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,43 @@
...
+    REQUIRED_PATHS = ["."]
+    _TOOL_NAME = "frama-c-sv"
+    _SCRIPT_NAME = _TOOL_NAME + ".py"
+
+    def executable(self, _):
+        return util.find_executable(_SCRIPT_NAME)
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
```
PhilippWendler, line 27: This seems to contradict the `executable` method: That method does not search in a subdirectory, but `program_files` still marks everything in the parent directory to be required (which is typical if the executable is in a subdirectory like `bin`, but then `executable` should search there). If the executable is in the tool's main directory, just deleted this method, the default implementation would then work well.

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return _TOOL_NAME
```
PhilippWendler, line 34: No nicer human-readable name available?

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [_SCRIPT_NAME, "--program"] + list(task.input_files_or_identifier)
```
PhilippWendler, line 37: You need to use `executable` here, this is why it is passed. Otherwise it will break for example if `_SCRIPT_NAME` is not in the current directory.

Does your tool support tasks with no or more than one input file? If the answer is no to a part of this question, please consider using `task.single_input_file` or `task.input_files` to produce nice error message for users.

PhilippWendler (changes requested):

benchexec/tools/frama-c-sv.py
MartinSpiessl: Sure, I will delete it then

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,43 @@
...
+    Tool info for Frama-C with a wrapper for usage in the SVCOMP.
+    URL: https://gitlab.com/sosy-lab/software/frama-c-sv
+    """
+
+    REQUIRED_PATHS = ["."]
+    _TOOL_NAME = "frama-c-sv"
+    _SCRIPT_NAME = _TOOL_NAME + ".py"
+
+    def executable(self, _):
+        return util.find_executable(_SCRIPT_NAME)
```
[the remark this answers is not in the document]
MartinSpiessl: What about using `util.find_executable2`? What is its use? Ff I need to use the tool locator then when would I ever be allowed to use `util.find_executable2`?

benchexec/tools/frama-c-sv.py
MartinSpiessl: I actually prefer it this way, because the variable name will still adequately express what the variable contains even if I change the tool name completely. It is totally human-readable, I didn't name it `_TN`.

benchexec/tools/frama-c-sv.py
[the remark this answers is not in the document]
PhilippWendler: Tool-info modules based on `BaseTool2` should never use any of the `util` methods and always call `tool_locator.find_executable()`.

benchexec/tools/frama-c-sv.py
PhilippWendler: Misunderstanding: I was asking about the *result* of the method, not about its implementation.

benchexec/tools/frama-c-sv.py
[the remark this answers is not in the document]
MartinSpiessl: Another question? Why do I even need to bother with this. This is boiler plate. Why isn't there a default implementation that just works. Maybe some field for the tool name in the constructor that I can overwrite.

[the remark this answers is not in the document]
PhilippWendler: Requiring implementors to write a one-line method (that can be taken almost completely from the docs of this method) is as practically as easy and as little effort as requiring them to set a certain field in the constructor, so the boilerplate is just a few characters.

At the same time, having the methods opens up much more possibilities and flexibility without hiding the actual workings under the hood. For example, many tool-info modules actually require to pass two parameters to `find_executable`. Other tool-info modules do something with the result of the method or use several calls to `find_executable` to support different tool versions.

benchexec/tools/frama-c-sv.py
MartinSpiessl: I had it as `BaseTool` module before and just rewrote it on the shortest possible path to fit the `BaseTool2` interface.

I probably just copied this from `cpachecker.py`.

Sure I will just use `task.input_files` then. Does CPAchecker support having no input file?

PhilippWendler: Yes, it is used for example here: https://gitlab.com/sosy-lab/software/cpachecker/-/blob/trunk/test/test-sets/integration-value-java.xml

benchexec/tools/frama-c-sv.py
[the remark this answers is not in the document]
MartinSpiessl: using the tool-locator now => resolved

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,38 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.util as util
```
PhilippWendler, line 8:
```suggestion
```
Unused import.

benchexec/tools/frama-c-sv.py
```diff
@@ -0,0 +1,38 @@
...
+        return tool_locator.find_executable(Tool._SCRIPT_NAME)
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return Tool._TOOL_NAME
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [executable, "--program"] + list(task.input_files_or_identifier)
```
PhilippWendler, line 32:
```suggestion
        cmd = [executable, "--program"] + list(task.input_files)
```
According to your comment.

PhilippWendler (approved):
Looks good, will merge once these two suggestions are applied and CI went through.

benchexec/tools/frama-c-sv.py
MartinSpiessl: Oh yes, I looked up and studied the method definitions, but I must have forgotten to actually implement the change.

PR #648, opened 2020-10-26
Add basic support for IE and add loading/warning messages

### LINKED ISSUE #637: Improve HTML Tables for Edge or InternetExplorer
### Problem
When opening HTML tables with Edge or IE, the page stays just completely blank/white.
This is especially bad for SV-COMP tables.

### Possible Source of Evil
The problem could be a simple bug in the Javascript.
In IE once an error appeared: `Object doesn't support property or method 'flat'`, but did not look further into it.

### Wanted Solution
Either fix the bug (preferred)  
or at least print a proper warning, for example:   
`Edge and IE are not supported, these tables are optimized for Firefox and Chrome`.

### Even better Solution
The latest version of Edge is based on the Chrome rendering engine (or something like that?).
Maybe the tables could even work in Edge directly.

### Additional step for SV-COMP
The SV-COMP tables of 2019 work fine with Edge and IE (at least they show the table and clickable logfiles). 
The SV-COMP tables of 2020 just appear as a blank page.
Perhaps the SV-COMP organizers could manually update these tables afterwards.
That would greatly improve user experience.

lachnerm:
Currently, the app crashes when using IE due to unsupported newer language features, resulting in a blank page. Adding polyfills fixes this by adding backward compatibility for those newer features. However, there are many other bugs when using IE which are not really worth the effort to fix for a deprecated browser, so I've added an error message that is shown beneath the menu when using IE.

Additionally, a template with a loading message was added that is shown when the page is still being loaded across all browsers. This message is automatically replaced by the content after everything has finished rendering.

If there's a syntax errors in our bundled script files or the browser doesn't recognize some language features (which most of the time is handled like a syntax error), it is possible that the user is still confronted with a blank page. This seems to be the default behavior of some browsers when detecting syntax errors for which I didn't find a way to prevent it. However, syntax errors shouldn't occur in our production code any way and browsers besides IE that don't support recent language features are rarely used, so I think we can ignore those cases.

Fixes #637.


PhilippWendler:
Thanks for investigating this! But I would say that the result is not optimal.

First, it increases the bundle size by about 100 KB just for improving support for a practically dead browser that will still not work correctly? (Probably this comes from the polyfill?) I would say this is not worth it.

We also do not need support for IE 11 in development mode, because who would want to use this?

Adding `src/index.html` makes no sense, because none of this will be used in production tables, the HTML code in them comes from `benchexec/tablegenerator/htmltable.py`. For development mode we already have `public/index.html`.

Anyway, I don't think it is worth adding a loading message because usually the table renders quite fast once everything is loaded. Or is there a way to make the loading message show already while the bulk of the JS and JSON stuff is still being transferred over the wire? Then it would be worth it independently of the current issue (and would need to be done in the correct place as mentioned). But we should not mix this with a warning about incompatible browsers, that would be bad UX! Just imagine a user loading a table via a mobile connection and it taking a few minutes, while they are being told that they should use a different browser although they are using a supported one.

So I would say that if there is a way to show a message only on unsupported browsers, this would be good to add. Ideally this would detect some feature that we require instead of hardcoding IE such that it is shown also in other unsupported browsers. (For example, in #637 `Object.flat` is mentioned in an error message, maybe this can be used?) Then we could even remove IE 11 from the list of browsers for which our production bundles are produced (check `npx browserslist` for what the definition in `package.json` results in), although it seems this does not reduce bundle size unfortunately.

With this proposal we would still support the same set of browsers as today, plus better UX for people on unsupported browsers with no disadvantage for people on supported browsers, right?

lachnerm:
> First, it increases the bundle size by about 100 KB just for improving support for a practically dead browser that will still not work correctly? (Probably this comes from the polyfill?) I would say this is not worth it.
> 
Yes, this is due to the polyfills. I see the point that it's not quite worth it though.

> We also do not need support for IE 11 in development mode, because who would want to use this?
> 
I added this when working on the error message in IE. Since the current solution added some functionality that was specific for IE, I thought it may be reasonable to also enable development mode for the browser, so I left it there. We can of course remove it again, especially if we don't keep the current solution.

> Adding `src/index.html` makes no sense, because none of this will be used in production tables, the HTML code in them comes from `benchexec/tablegenerator/htmltable.py`. For development mode we already have `public/index.html`.
> 
Yes, you're right, I seem to have missed that. I'll remove the file (and potentially move the content to where it's supposed to be).

> Anyway, I don't think it is worth adding a loading message because usually the table renders quite fast once everything is loaded. Or is there a way to make the loading message show already while the bulk of the JS and JSON stuff is still being transferred over the wire? 
> 
I'm not really sure which point in time you are referring to. The whole JS-code is placed in `script` tags at the end of the `body` element of the HTML-file, so the earliest point we can show the loading message would be pretty much immediately when the HTML was loaded, before the JS is started to be processed - did you mean this? If so, we could keep the current way the loading message is being handled (after moving it to the right file of course), as it's already currently shown at this point. 

> But we should not mix this with a warning about incompatible browsers, that would be bad UX! Just imagine a user loading a table via a mobile connection and it taking a few minutes, while they are being told that they should use a different browser although they are using a supported one.
>
This would indeed be a case we definitely don't want to happen. If we keep the loading message, I will remove the part about browser compatibility.
> So I would say that if there is a way to show a message only on unsupported browsers, this would be good to add. Ideally this would detect some feature that we require instead of hardcoding IE such that it is shown also in other unsupported browsers. (For example, in #637 `Object.flat` is mentioned in an error message, maybe this can be used?) Then we could even remove IE 11 from the list of browsers for which our production bundles are produced (check `npx browserslist` for what the definition in `package.json` results in), although it seems this does not reduce bundle size unfortunately.
> 
Okay, I thought of some new way we could approach this problem without the need to support the browser itself.
The problem is that if the browser encounters a syntax error (or unsupported language features), all processing of the scripts will be stopped. So within our `index.js` or anything else that is going to be put into our `bundle.js`, we cannot handle those syntax errors, as the code to handle those wouldn't even be executed itself. We are also receiving syntax errors from our `vendor.js`, which makes things even more complicated. 
Since adding a message in the HTML file that will be removed via JS (in case the JS code could be loaded successfully) is also not an option, we could handle this with some JS that works independently of our bundled files. One way to achieve this would be by adding a `script` tag into the HTML that does our feature detection before the bundled files are loaded into the browser. I haven't found a stable and universal way to check if the browser supports all features we are using and it's also kind of difficult to maintain. However, we could start with prechecking whether any syntax errors could occur and act accordingly. This is kind of hacky though, but it should work and is probably the most stable option we have.

In this new `script` tag, we use either the `eval()` function or `Function()` to try to execute some syntax from newer features such as arrow functions. This will raise an exception too, but it won't stop the execution of the script itself. We put this into a `try-catch block` and catch any occurring syntax errors, which would tell us that the browser doesn't support the features we tested in the `try block` and therefore require within our code. In this case we can add an error message to the HTML in the `catch block`. We would still encounter the syntax errors from our bundled files, but since we don't want to show anything besides the error message in this case any way, those errors don't really matter. The code would look something like this:

``` 
try {
  eval("const x = y => 4");
} catch (err) {
  document.getElementById("root").innerHTML = "Your browser is not supported. Please consider using another browser such as Firefox or Google Chrome.";
}
```

I think this should already suffice to solve our problem. If we want to, we could refine the code with syntax errors out of the way, as other errors don't automatically stop the whole JS code any more. The mentioned error from `Object.flat` for example is only a type error and won't crash the whole code. But arrow functions were introduced in ES6, so if the browser does support them, then there shouldn't be any compatibility problems (at least regarding the JS code) any way. 
What would you think of this solution? I'd say it's far from beautiful, but probably the most practical way without having to take every single possible browser into consideration. Also this is pretty much safe regarding possible changes of the browsers themselves.

PhilippWendler:
> > Anyway, I don't think it is worth adding a loading message because usually the table renders quite fast once everything is loaded. Or is there a way to make the loading message show already while the bulk of the JS and JSON stuff is still being transferred over the wire?
> 
> I'm not really sure which point in time you are referring to. The whole JS-code is placed in `script` tags at the end of the `body` element of the HTML-file, so the earliest point we can show the loading message would be pretty much immediately when the HTML was loaded, before the JS is started to be processed - did you mean this? If so, we could keep the current way the loading message is being handled (after moving it to the right file of course), as it's already currently shown at this point.

If browsers do already show this part while they still retrieve the script tags, then yes, this sounds good. Can be implemented separately from the warning, though.

>  One way to achieve this would be by adding a `script` tag into the HTML that does our feature detection before the bundled files are loaded into the browser. I haven't found a stable and universal way to check if the browser supports all features we are using and it's also kind of difficult to maintain. However, we could start with prechecking whether any syntax errors could occur and act accordingly. This is kind of hacky though, but it should work and is probably the most stable option we have.

I think we can keep the check simple. If we can detect whether there were syntax errors in other `<script>` tags, this would of course be good, because if there is a syntax error there is no reason to not show the message about an unsupported browser. For cases where no syntax error occurs or where we cannot detect it but still want to warn about the browser, we could live with a pretty basic heuristic. For example, we could check for some feature that is required by our code but is unlikely to exist in most older browsers, and does definitely not exist in IE. 

This could still mean that in some obscure browser that feature happens to be implemented but some other feature that we require is missing and people will get a blank page, but it would be extremely rare and unlikely.

> In this new `script` tag, we use either the `eval()` function or `Function()` to try to execute some syntax from newer features such as arrow functions. This will raise an exception too, but it won't stop the execution of the script itself. We put this into a `try-catch block` and catch any occurring syntax errors, which would tell us that the browser doesn't support the features we tested in the `try block` and therefore require within our code. In this case we can add an error message to the HTML in the `catch block`. We would still encounter the syntax errors from our bundled files, but since we don't want to show anything besides the error message in this case any way, those errors don't really matter. The code would look something like this:
> 
> ```
> try {
>   eval("const x = y => 4");
> } catch (err) {
>   document.getElementById("root").innerHTML = "Your browser is not supported. Please consider using another browser such as Firefox or Google Chrome.";
> }
> ```
> 
> I think this should already suffice to solve our problem.

Now I think we actually propose the same thing? Wasn't so sure about this for the previous paragraph.

> If we want to, we could refine the code with syntax errors out of the way, as other errors don't automatically stop the whole JS code any more.

This I don't understand?

> The mentioned error from `Object.flat` for example is only a type error and won't crash the whole code. But arrow functions were introduced in ES6, so if the browser does support them, then there shouldn't be any compatibility problems (at least regarding the JS code) any way.

I would leave the choice of the feature used for detection up to you. Checking for existence of a function like `Object.flat` would mean we don't even need a try block, I guess?

The general rule should be that
- the feature must be used by our production code (not sure whether this holds for arrow functions),
- it must not be supported by IE (because we want to make sure that we trigger the warning there), and
- the more modern the better (because then it becomes more unlikely that the warning won't be shown in other unsupported browsers).

> What would you think of this solution? I'd say it's far from beautiful, but probably the most practical way without having to take every single possible browser into consideration. Also this is pretty much safe regarding possible changes of the browsers themselves.

Exactly.

lachnerm:
> > > Anyway, I don't think it is worth adding a loading message because usually the table renders quite fast once everything is loaded. Or is there a way to make the loading message show already while the bulk of the JS and JSON stuff is still being transferred over the wire?
> > 
> > 
> > I'm not really sure which point in time you are referring to. The whole JS-code is placed in `script` tags at the end of the `body` element of the HTML-file, so the earliest point we can show the loading message would be pretty much immediately when the HTML was loaded, before the JS is started to be processed - did you mean this? If so, we could keep the current way the loading message is being handled (after moving it to the right file of course), as it's already currently shown at this point.
> 
> If browsers do already show this part while they still retrieve the script tags, then yes, this sounds good. Can be implemented separately from the warning, though.
> 
Yes, this should work. The HTML file is processed straight from the top to the bottom, so everything that is put before the scripts will also be shown before those are processed. I've moved the loading message to the corresponding part in the `htmltable.py` (and `public/index.html` too).

> > > In this new `script` tag, we use either the `eval()` function or `Function()` to try to execute some syntax from newer features such as arrow functions. This will raise an exception too, but it won't stop the execution of the script itself. We put this into a `try-catch block` and catch any occurring syntax errors, which would tell us that the browser doesn't support the features we tested in the `try block` and therefore require within our code. In this case we can add an error message to the HTML in the `catch block`. We would still encounter the syntax errors from our bundled files, but since we don't want to show anything besides the error message in this case any way, those errors don't really matter. The code would look something like this:
> > ```
> > try {
> >   eval("const x = y => 4");
> > } catch (err) {
> >   document.getElementById("root").innerHTML = "Your browser is not supported. Please consider using another browser such as Firefox or Google Chrome.";
> > }
> > ```
> > 
> > 
> > I think this should already suffice to solve our problem.
> 
> Now I think we actually propose the same thing? Wasn't so sure about this for the previous paragraph.
> 
Yes, I think we were talking about the same procedure.
> 
> > If we want to, we could refine the code with syntax errors out of the way, as other errors don't automatically stop the whole JS code any more.
> 
> This I don't understand?
> 
Those were some thoughts of mine which haven't really proved practical now during the implementation of the solution, so never mind about that.

> 
> The general rule should be that
> 
> * the feature must be used by our production code (not sure whether this holds for arrow functions),
> * it must not be supported by IE (because we want to make sure that we trigger the warning there), and
> * the more modern the better (because then it becomes more unlikely that the warning won't be shown in other unsupported browsers).
> 
I've decided to use `Object.values` for the feature detection as it's probably the newest feature (ES8) that is used consistently in our source code and also kept in our production code. Generally, arrow functions should already suffice (since I don't think there are actually at least somewhat important browsers out there updated to support ES6, but not ES8), but I agree that the most modern feature should be used, just in case.

I've tested the new solution in development mode and with a production HTML-table and everything behaves as expected for Google Chrome, Firefox, Edge and IE, so the first three render the table without any troubles while the last one shows the "unsupported browser" error message.

PhilippWendler (approved):
Thanks! This looks good and can in principle be merged.

One question: When I inject an artificial error in the test `eval` to see how the error message works, I still get the proper tables instead of the error message in Firefox, because first the error message is shown and then the actual table loads and overwrites the error message. I guess this does not happen on browsers where the table load fails? Could this be a problem in other browsers? E.g., error message is triggered, tables load nevertheless and hide the error message, and only later some other error occurs? But I guess in this case the user would at least get some UI, which might be better than just the error message, and the blank screen is still reliably prevented. Correct?

PhilippWendler:
@kfriedberger As you originally requested this, could you please give feedback on whether this is a satisfying solution for you?

kfriedberger (commented):
I did not review the code, but only the resulting functionality, based on 6b8e272c.

On IE11 and Edge44, there is a message now `Please wait while the page is being loaded`.
There could be a hint on `try a another browser`. This message is still missing.

Edge86 shows the table without any problems.

PhilippWendler:
Weird, @lachnerm said that IE shows the "Your browser is not supported." warning. Could you please give us the exact IE version and maybe have a look in the developer console whether there are any messages?

@lachnerm Could you please investigate?

kfriedberger:
I can give more details.
I used the this [follwing file](http://www.cip.ifi.lmu.de/~friedber/CPAchecker/00895.-r_integration-bam.2020-10-16_13-28-14.results.html) for all tests.

For IE11 (11.572.19041.0 on Windows 10):
- a file from a webserver works and shows `Your browser is not supported. Please consider using another browser such as FF or GC.`
- a file from local disk shows only `Please wait while the page is being loaded`, but it also shows a `systax error in FILENAME (39,9)`, which is right after the check for the unsupported `Object.values({})` and tries to get initialize the variable `msgContainer`. There is also a `syntax error in FILENAME (44,1)`, where a `const data` appears. It seems like IE11 is using a compatibility mode for IE7 that is used for local and intranet sites by default. [Disabiling the compatibility mode](https://en.wikipedia.org/wiki/Compatibility_mode#:~:text=In%20IE11%2C%20a%20user%20can,and%20clicking%20Compatibility%20View%20Settings.) helps and then shows `Your browser is not supported. Please consider using another browser such as FF or GC.`. 

For Edge44 (44 on Windows 10):
- a file from webserver is plain white without any message. There is an type error `Object doesn't support property or method "flat" in FILENAME (55,109747)`
- a file from local disk shows `Please wait while the page is being loaded`, and an (three times the same?) error `access denied FILENAME (47,932763)`. I do not know why.

Firefox and Chrome work fine with both the local and the webser-based table.

PhilippWendler:
Ah, that explains it. Thanks for investigating!

@lachnerm I guess there is a way to handle this for IE compat mode? For Legacy Edge, it possible without too much effort we could do it, but given that Legacy Edge should not be present anymore on systems with the latest Windows updates it does not have high importance.

lachnerm:
> For IE11 (11.572.19041.0 on Windows 10):
> 
> * a file from local disk shows only `Please wait while the page is being loaded`, but it also shows a `systax error in FILENAME (39,9)`, which is right after the check for the unsupported `Object.values({})` and tries to get initialize the variable `msgContainer`. There is also a `syntax error in FILENAME (44,1)`, where a `const data` appears. It seems like IE11 is using a compatibility mode for IE7 that is used for local and intranet sites by default. [Disabiling the compatibility mode](https://en.wikipedia.org/wiki/Compatibility_mode#:~:text=In%20IE11%2C%20a%20user%20can,and%20clicking%20Compatibility%20View%20Settings.) helps and then shows `Your browser is not supported. Please consider using another browser such as FF or GC.`.
> 
Strangely I couldn't reproduce the error on my machine, even though I'm using the exact same version of IE11. For me, with and without the compatibility mode everything works, however when emulating IE7 I noticed some bugs that were really similar to the ones reported by you. This is due to IE7 not even supporting the `const` declaration in JS, so I replaced it with `var`, which should be supported by all browsers. Since the error reported by you also refers to both usages of `const`, I strongly believe that this was the reason in your browser too.
> For Edge44 (44 on Windows 10):
> 
> * a file from webserver is plain white without any message. There is an type error `Object doesn't support property or method "flat" in FILENAME (55,109747)`
>
After investigating this further, I found out that Legacy Edge actually supports ES8, so the feature `Object.values` actually does work there. However, the feature `array.flat` from ES10 is not supported by the Legacy version any more, but we do use that in our production code. This shouldn't occur in any still used browser, but fixing this is pretty cheap by using `array.flat` instead of `Object.values` for feature detection, so I've fixed it. I seem to have missed that there was a newer feature than `Object.values` that is preserved in the production code, so it's better to use the more modern feature `array.flat` anyway.
> * a file from local disk shows `Please wait while the page is being loaded`, and an (three times the same?) error `access denied FILENAME (47,932763)`. I do not know why.
> 
I couldn't reproduce this error. When opening local files from my machine, I received the same error as the one from the web server and this was fixed with the solution for the web server for me too.

I had already installed the new version of Edge on my machine and Microsoft pretty much doesn't want the users to access the Legacy version anymore, so enabling this version locally required some really hacky workarounds. I still can't assure that there weren't some settings from the newer version meddling with the Legacy browser, so I'd appreciate if @kfriedberger could check if the new solutions actually work for him too. The same applies to the IE fixes, since I couldn't really reproduce the error with the same settings you had enabled.

kfriedberger (approved):
Now everything works as expected.
I get a nice message on all of my outdated browsers :+1:

PR #649, opened 2020-10-26
support for BaseTool2 in Ultimate tool-info-modules

danieldietsch:


danieldietsch:
@skanav  this should probably take care of the ultimate-based tools (see also #646)

danieldietsch:
Ah I just realized that I also need to take care of the data model, not just update to ``BaseTool2``

benchexec/tools/ultimate.py
```diff
@@ -68,30 +66,32 @@ def __init__(self):
...
+        try:
+            exe = tool_locator.find_executable("Ultimate.py")
+            for (_, dir_names, file_names) in os.walk(os.path.dirname(exe)):
+                if "Ultimate" in file_names and "plugins" in dir_names:
                     return exe
                 break
+        except benchexec.tools.template.ToolNotFoundException:
+            pass
 
         # possibly another Ultimate.py was found or not found at all, check in the current dir
```
PhilippWendler, line 80: Note that tool-info modules based on `BaseTool2` should not look for the tool in the current directory automatically anymore. `ToolLocator` will search in the current directory if and only if this is desired according to the BenchExec configuration.

And AFAIS this should mean that the whole method can and should be simplified to a simple call to `tool_locator.find_executable` as in most other tool-info modules.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +204,116 @@ def _requires_ultimate_data(self, executable):
         # all versions before 0.1.24 do not require ultimatedata
         return not (int(major) == 0 and int(minor) < 2 and int(patch) < 24)
 
-    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits=None):
-        if rlimits is None:
-            rlimits = {}
+    def cmdline(self, executable, options, task, resource_limits):
+        self._uses_propertyfile = task.property_file is not None
```
PhilippWendler, line 208: I did not notice this before, but storing information about a task in `self` is broken, because there is no guarantee that `cmdline` and `determine_result` are called in any particular order (cf. docs of `BaseTool2`). So please remove this field. With the new API this should finally be possible because `determine_result` has access to the run command line.

benchexec/tools/ultimate.py
```diff
@@ -445,14 +442,14 @@ def _determine_result_with_propertyfile(
...
-                return line[start_position:].strip()
-        return None
+    def get_value_from_output(self, output, identifier):
+        regex = re.compile(identifier)
+        for line in output:
+            match = regex.search(line)
+            if match and len(match.groups()) > 0:
+                return match.group(1)
+        logging.debug("Did not find a match with regex %s", identifier)
+        return "N/A"
```
PhilippWendler, line 452: `None` is in general a better return value. Using a string will lead to table columns with mixed numbers and strings, so there won't be any statistics etc. With `None` the respective cells will be correctly interpreted as "missing value".

benchexec/tools/ultimate.py
```diff
@@ -68,30 +66,32 @@ def __init__(self):
         self.java = None
 
     @functools.lru_cache()
```
PhilippWendler, line 68: No reason to cache this, actually.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +204,116 @@ def _requires_ultimate_data(self, executable):
...
-        self._uses_propertyfile = propertyfile is not None
-        if _OPTION_NO_WRAPPER in options:
+        if _OPTION_NO_WRAPPER in combined_options:
             # do not use old wrapper script even if property file is given
             self._uses_propertyfile = False
-            propertyfile = None
-            options.remove(_OPTION_NO_WRAPPER)
+            combined_options.remove(_OPTION_NO_WRAPPER)
 
         if self._is_svcomp17_version(executable):
```
PhilippWendler, line 216: Note that the new API explicitly guarantees that information about the tool version may be stored in a field of `self` and used by methods like `cmdline`, so if you wish, you could simplify this.

benchexec/tools/ultimate.py
```diff
@@ -357,7 +354,7 @@ def _determine_result_without_propertyfile(
         ltl_true_string = "Buchi Automizer proved that the LTL property"
         overflow_false_string = "overflow possible"
 
-        for line in output:
+        for line in run.output:
             if line.find(unsupported_syntax_errorstring) != -1:
```
PhilippWendler, line 358: Just noticed this: Looking for strings in Python is most easily expressed as `if needle in haystack:`

benchexec/tools/ultimate.py
```diff
@@ -13,13 +13,11 @@
 import re
 import shutil
 import subprocess
-import sys
 
 import benchexec.result as result
 import benchexec.tools.template
 import benchexec.util as util
 from benchexec import BenchExecException
```
PhilippWendler, line 20: For the cases where this exception is currently used, `ToolNotFoundException` seems a good replacement.

benchexec/tools/ultimate.py
```diff
@@ -64,34 +62,18 @@ class UltimateTool(benchexec.tools.template.BaseTool):
...
-        for (_dirpath, dirnames, filenames) in os.walk(current):
-            if (
-                "Ultimate" in filenames
-                and "Ultimate.py" in filenames
-                and "plugins" in dirnames
-            ):
-                return "./Ultimate.py"
+    def executable(self, tool_locator):
+        exe = tool_locator.find_executable("Ultimate.py")
+        for (_, dir_names, file_names) in os.walk(os.path.dirname(exe)):
```
PhilippWendler, line 69: I am wondering whether this is as intended. `os.walk` walks recursively through directories, so the check will succeed if there are `Ultimate` and `plugins` directory anywhere below the executable's directory. The error message seems to indicate that all three are expected to be in the same directory instead?

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
-            rlimits = {}
-
-        self._uses_propertyfile = propertyfile is not None
-        if _OPTION_NO_WRAPPER in options:
-            # do not use old wrapper script even if property file is given
-            self._uses_propertyfile = False
-            propertyfile = None
-            options.remove(_OPTION_NO_WRAPPER)
+    def cmdline(self, executable, options, task, resource_limits):
+        combined_options = options + [*task.options] if task.options else []
```
PhilippWendler, line 190: `task.options` is not a list of command-line parameters, but an arbitrary data structure (as defined in the task-definition file in the `options` key). So this needs to be converted to command-line parameters first. Cf. #646 and the `sv_benchmarks_util` that is added there for code that is useful for most cases. Maybe we should delay this PR until #646 is merged and then rebase on top of it?

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+            msg = (
+                "Unsupported argument combination: "
+                "If you specify {}, you also need to give a toolchain (with '-tc' or '--tolchain')".format(
+                    _OPTION_NO_WRAPPER
+                )
+            )
+            raise UnsupportedFeatureException(msg)
 
-            cmdline.append("--full-output")
+        if task.property_file is not None:
```
PhilippWendler, line 212: Simplification:
```suggestion
        if task.property_file:
```

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
-                cmdline += ["--file"] + tasks
-            cmdline += options
-            self.__assert_cmdline(
-                cmdline,
-                "cmdline contains empty or None argument when using default SVCOMP mode: ",
+        # there is no way to run ultimate; not enough parameters
+        msg = (
+            "Unsupported argument combination: You either need a property file or a toolchain option (-tc). "
+            "options={} property_file={} resource_limits={}".format(
+                options, task.property_file, resource_limits
```
PhilippWendler, line 219: `task.property_file` is guaranteed to be `None` here, right?

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+        mem_bytes = resource_limits.memory
+        cmdline = [self.java]
+
+        # -ea has to be given directly to java
+        if "-ea" in options:
+            options = [e for e in options if e != "-ea"]
+            cmdline += ["-ea"]
+
+        if mem_bytes:
+            cmdline += ["-Xmx" + str(mem_bytes)]
```
PhilippWendler, line 234: I noticed this code now and while the decision of how to do this is up to you, I wanted to mention the following (from my experience as a CPAchecker developer, not as BenchExec maintainer):

The memory limit from BenchExec (which is in `mem_bytes`) is a hard limit, reaching it will let the process be killed. Using it as the max heap size of a JVM will mean that the JVM might try to use that much memory for heap plus all the other memory (native allocations, stack, etc.), so it might well exceed the limit. Due to the GC, the JVM could even try to allocate that much memory even if it could cope with less memory (at the cost of more GC runs). So in CPAchecker, for example, we always set the max heap to some smaller value. How much smaller unfortunately depends on a lot of things, including how much memory is needed by non-JVM allocations (e.g., from a native SMT solver).

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+            if "-ultimatedata" not in options and "-data" not in options:
+                if self.api == 2:
+                    cmdline += [
+                        "-data",
+                        "@noDefault",
+                        "-ultimatedata",
+                        os.path.join(os.path.dirname(executable), "data"),
+                    ]
+                if self.api == 1:
+                    raise ValueError(
```
PhilippWendler, line 248: Seems like `UnsupportedFeatureException` would be a good fit here and in similar locations?

Of course you could in principle also remove these checks and just let Ultimate print an error message when such a command line is attempted to be executed. It would be a little less convenient for the user, but reduce the maintenance effort of the tool-info module for you. So I leave this up to you.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+            if "-data" not in options:
+                if self.api == 2 or self.api == 1:
+                    cmdline += [
+                        "-data",
+                        os.path.join(os.path.dirname(executable), "data"),
+                    ]
+
+        cmdline += options
+
+        if task.input_files:
```
PhilippWendler, line 272: `task.input_files` will always be non-empty when it is returned, otherwise an exception is thrown. If you want to support both empty and non-empty list of input files, you would need to use
```suggestion
        if task.input_files_or_empty:
```

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+            cmdline += ["-i"] + [*task.input_files]
```
PhilippWendler, line 273:
```suggestion
            cmdline += ["-i", *task.input_files]
```

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+        self.__assert_cmdline(cmdline, "No_Wrapper")
+        return cmdline
+
+    def _cmdline_default(self, executable, options, task):
+        # use the old wrapper script if a property file is given
+        cmdline = [executable, "--spec", task.property_file]
+        if task.input_files:
```
PhilippWendler, line 280: Same as above.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +186,120 @@ def _requires_ultimate_data(self, executable):
...
+        self.__assert_cmdline(cmdline, "Default")
+        return cmdline
+
+    def _cmdline_svcomp17(self, executable, options, task):
+        cmdline = [executable, task.property_file]
+        cmdline += [
+            option for option in options if option not in _SVCOMP17_FORBIDDEN_FLAGS
+        ]
+        cmdline.append("--full-output")
+        cmdline += [*task.input_files]
```
PhilippWendler, line 292: This is enough:
```suggestion
        cmdline += task.input_files
```

benchexec/tools/ultimate.py
```diff
@@ -321,18 +310,12 @@ def program_files(self, executable):
...
-    def determine_result(self, returncode, returnsignal, output, is_timeout):
-        if self._uses_propertyfile:
-            return self._determine_result_with_propertyfile(
-                returncode, returnsignal, output, is_timeout
-            )
-        return self._determine_result_without_propertyfile(
-            returncode, returnsignal, output, is_timeout
-        )
+    def determine_result(self, run):
+        if any([arg for arg in run.commandline if "--spec" in arg or ".prp" in arg]):
```
PhilippWendler, line 314:
```suggestion
        if any("--spec" in arg or ".prp" in arg for arg in run.commandline):
```

But do you actually want to the check whether there is one argument that contains `--spec`, or whether there is one argument that is equal to `--spec`?
If the latter:
```suggestion
        if any(arg == "--spec" or ".prp" in arg for arg in run.commandline):
```

PhilippWendler (changes requested):
Looks mostly good, but the `task.options` need to be handled differently and I have a few other non-critical questions and suggestions.

benchexec/tools/ultimate.py
danieldietsch: But there is an unconditional ``break`` after the ``if``

benchexec/tools/ultimate.py
danieldietsch: That sounds good! Lets do that.

benchexec/tools/ultimate.py
danieldietsch: Yes, we had the same policy with Ultimate for a long time. Nevertheless, it seems that allowing the VM in prinicpal to go up to this bound reduces OOMs, at least on SVCOMP instances.

benchexec/tools/ultimate.py
danieldietsch: Yes, but I cannot change older Ultimate versions. So to be compatible with them we have a complex tool info module ;)

benchexec/tools/ultimate.py
PhilippWendler: Ok, right. Hard to spot :-)

How about
```
directory = os.path.dirname(exe)
if os.path.isfile(os.path.join(directory, "Ultimate")) and os.path.isdir(os.path.join(directory, "plugins")):
```
instead of the loop?
I think this would be better understandable.

But I am leaving this up to you.

danieldietsch: Ah, I leave it as is; it might be that we need to check for even more files in the near future.

benchexec/tools/ultimate.py
PhilippWendler: #646 is merged now.

danieldietsch:
@PhilippWendler I get the following exception on my Python 3.7 when I run ``python3 -m benchexec.test_tool_info ultimateautomizer --no-container --tool-directory /storage/repos/ultimate/releaseScripts/default/UAutomizer-linux``

My python is 3.7.8. 

```
WARNING: Tool-info module does not support SV-Benchmarks task: “expected str, bytes or os.PathLike object, not list”
  File "/storage/repos/benchexec/benchexec/test_tool_info.py", line 108, in log_if_unsupported
    yield  # call code block to be executed
  File "/storage/repos/benchexec/benchexec/test_tool_info.py", line 272, in print_standard_task_cmdlines
    CURRENT_BASETOOL.ResourceLimits(cputime=900, cputime_hard=1000),
  File "/storage/repos/benchexec/benchexec/model.py", line 219, in cmdline_for_run
    args = [os.path.expandvars(arg) for arg in args]
  File "/storage/repos/benchexec/benchexec/model.py", line 219, in <listcomp>
    args = [os.path.expandvars(arg) for arg in args]
  File "/usr/lib/python-exec/python3.7/../../../lib/python3.7/posixpath.py", line 288, in expandvars
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not list
```

benchexec/tools/ultimate.py
```diff
@@ -204,113 +188,125 @@ def _requires_ultimate_data(self, executable):
...
-            propertyfile = None
-            options.remove(_OPTION_NO_WRAPPER)
+    def cmdline(self, executable, options, task, resource_limits):
+        data_model_param = get_data_model_from_task(
+            task,
+            {ILP32: ["--architecture", "32bit"], LP64: ["--architecture", "64bit"]},
+        )
+        combined_options = options
+        if data_model_param and not self._is_sublist(data_model_param, options):
+            combined_options += [data_model_param]
```
PhilippWendler, line 198: Either change this line to `combined_options += ["--architecture", data_model_param]"` and make the values of the above dict to simple strings, or change this line to `combined_options += data_model_param` (though in the later case, `data_model_param` should really be named `data_model_params`.

This is the cause of the crash due to a nested list being contained in the result value.

benchexec/tools/predatorhp.py
```diff
@@ -31,8 +31,7 @@ def cmdline(self, executable, options, task, rlimits):
         spec = ["--propertyfile", task.property_file] if task.property_file else []
 
         data_model_param = get_data_model_from_task(
-            task,
-            {ILP32: "--compiler-options=-m32", LP64: "--compiler-options=-m64"},
+            task, {ILP32: "--compiler-options=-m32", LP64: "--compiler-options=-m64"},
```
PhilippWendler, line 34: Please don't reformat other files.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +188,125 @@ def _requires_ultimate_data(self, executable):
...
-            self._uses_propertyfile = False
-            propertyfile = None
-            options.remove(_OPTION_NO_WRAPPER)
+    def cmdline(self, executable, options, task, resource_limits):
+        data_model_param = get_data_model_from_task(
+            task,
+            {ILP32: ["--architecture", "32bit"], LP64: ["--architecture", "64bit"]},
+        )
+        combined_options = options
+        if data_model_param and not self._is_sublist(data_model_param, options):
```
PhilippWendler, line 197: This check would mean if a user manually wants to override the architecture of a task by adding `--architecture something` to the command line, the tool-info module will still add `--architecture` as well, leading to duplicate parameters.

I would suggest to simplify this and simply check wether `--architecture` exists in `options`. This is also how the other tool-info modules are currently implemented.

benchexec/tools/ultimate.py
```diff
@@ -321,18 +317,12 @@ def program_files(self, executable):
...
-    def determine_result(self, returncode, returnsignal, output, is_timeout):
-        if self._uses_propertyfile:
-            return self._determine_result_with_propertyfile(
-                returncode, returnsignal, output, is_timeout
-            )
-        return self._determine_result_without_propertyfile(
-            returncode, returnsignal, output, is_timeout
-        )
+    def determine_result(self, run):
+        if any([arg for arg in run.commandline if "--spec" == arg or ".prp" in arg]):
```
PhilippWendler, line 321:
```suggestion
        if any(arg for arg in run.commandline if "--spec" == arg or ".prp" in arg):
```
The tool `flake8` that we use for checking best practices of Python code suggests this, and we use it in CI so we want to apply all of its suggestions when they make sense.

benchexec/tools/ultimate.py
```diff
@@ -204,113 +188,125 @@ def _requires_ultimate_data(self, executable):
...
+        cmdline.append("--full-output")
+        cmdline += task.input_files
+        self.__assert_cmdline(cmdline, "SVCOMP17")
+        return cmdline
+
+    @staticmethod
+    def __assert_cmdline(cmdline, mode):
+        assert all(
+            cmdline
+        ), "cmdline contains empty or None argument when using %s mode: %s".format(
```
PhilippWendler, line 307: `.format()` does not support `%s`, so this needs to changed:
```suggestion
        ), "cmdline contains empty or None argument when using {} mode: {}".format(
```

PhilippWendler (changes requested):

benchexec/tools/predatorhp.py
danieldietsch: Oh, sorry, was not my intention ...

benchexec/tools/ultimate.py
danieldietsch: Do you have a flake8 config somewhere so that I can integrate it into PyCharm? I ran flake8 3.8.4, and it did not complain about that.

PhilippWendler: Sure, our config is checked in already (`.flake8`). It just does not run on Travis CI due to the inflexibility of Travis' configuration, and GitLab CI unfortunately does not work on GitHub PRs yet. This is why I mention it manually.

If you want to check it for yourself, this is of course nice, and you are probably just missing the package `flake8-awesome` with additional plugins. In our `.gitlab-ci.yml` for job `flake8` it is visible that there is nothing necessary besides installing that package and running `flake8`. (Of course there is no expectation for PR authors to look there, I am just mentioning this as a possible reference.)

benchexec/tools/ultimate.py
PhilippWendler: A warning like you added is of course even nicer for the user!

benchexec/tools/ultimate.py
```diff
@@ -204,113 +188,136 @@ def _requires_ultimate_data(self, executable):
...
-
-        self._uses_propertyfile = propertyfile is not None
-        if _OPTION_NO_WRAPPER in options:
-            # do not use old wrapper script even if property file is given
-            self._uses_propertyfile = False
-            propertyfile = None
-            options.remove(_OPTION_NO_WRAPPER)
+    def cmdline(self, executable, options, task, resource_limits):
+        data_model_param = get_data_model_from_task(
+            task, {ILP32: "32bit", LP64: "64bit"},
```
PhilippWendler, line 193:
```suggestion
            task, {ILP32: "32bit", LP64: "64bit"}
```

The version of black used by us would otherwise reformat this. I would actually have committed this myself but could not push to the PR branch, so please apply this suggestion.

PhilippWendler (approved):
Thank you very much for this improvement!

There is just a small technicality that needs to be done before the merge.

PR #650, opened 2020-10-28
Refactor metaval tool-info module:extend BaseTool2

MartinSpiessl:
This updates the metaval tool-info module to inherit from the new `BaseTool2` base class.

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -96,39 +96,36 @@ class Tool(benchexec.tools.template.BaseTool2):
                     "benchexec.tools." + verifierName, fromlist=["Tool"]
                 ).Tool()
 
-        if verifierName in self.wrappedTools:
-            tool = self.wrappedTools[verifierName]
-            assert isinstance(
-                tool, BaseTool2
-            ), "we expect that all wrapped tools extend BaseTool2"
-            wrapped_executable = tool.executable(
-                BaseTool2.ToolLocator(
-                    tool_directory=self.TOOL_TO_PATH_MAP[verifierName]
-                )
-            )
-            wrappedOptions = tool.cmdline(
-                wrapped_executable,
-                options,
-                [os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
-                rlimits,
-            )
-
-            return (
-                [
-                    executable,
-                    "--verifier",
-                    self.TOOL_TO_PATH_MAP[verifierName],
-                    "--witness",
-                    witnessName,
-                ]
-                + additionalPathArgument
-                + witnessTypeArgument
-                + list(task.single_input_file)
-                + ["--"]
-                + wrappedOptions
-            )
-        else:
+        if not verifierName in self.wrappedTools:
             sys.exit("ERROR: Could not find wrapped tool")  # noqa: R503 always raises
+        tool = self.wrappedTools[verifierName]
+        assert isinstance(
+            tool, BaseTool2
+        ), "we expect that all wrapped tools extend BaseTool2"
+        wrapped_executable = tool.executable(
+            BaseTool2.ToolLocator(tool_directory=self.TOOL_TO_PATH_MAP[verifierName])
+        )
+        wrappedOptions = tool.cmdline(
+            wrapped_executable,
+            options,
+            [os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
+            rlimits,
+        )
+
+        return (
+            [
+                executable,
+                "--verifier",
+                self.TOOL_TO_PATH_MAP[verifierName],
+                "--witness",
+                witnessName,
+            ]
+            + additionalPathArgument
+            + witnessTypeArgument
+            + list(task.single_input_file)
+            + ["--"]
+            + wrappedOptions
+        )
 
     def version(self, executable):
         stdout = self._version_from_tool(executable, "--version")
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -29,7 +29,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     """
 
     TOOL_TO_PATH_MAP = {
-        "cpachecker-metaval": "CPAchecker-backend",
+        "cpachecker-metaval": "CPAchecker-frontend",
         "cpachecker": "CPAchecker",
         "esbmc": "esbmc",
         "symbiotic": "symbiotic",
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -9,10 +9,7 @@ import argparse
 import benchexec.util as util
 import benchexec.tools.template
 from benchexec.tools.template import BaseTool2
-import contextlib
-import os
 import re
-import sys
 import threading
 
 
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -104,7 +104,7 @@ class Tool(benchexec.tools.template.BaseTool2):
             BaseTool2.ToolLocator(tool_directory=self.TOOL_TO_PATH_MAP[verifierName])
         )
         wrappedtask = BaseTool2.Task(
-            input_files=[os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
+            input_files=["../output/ARG.c"],
             identifier=task.identifier,
             property_file=task.property_file,
             options=task.options,
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -8,9 +8,11 @@
 import argparse
 import benchexec.util as util
 import benchexec.tools.template
-from benchexec.tools.template import BaseTool2
+import os
 import re
 import threading
+
+from benchexec.tools.template import BaseTool2
 
 
 class Tool(benchexec.tools.template.BaseTool2):
@@ -46,10 +48,10 @@ class Tool(benchexec.tools.template.BaseTool2):
     def name(self):
         return "metaval"
 
-    def determine_result(self, returncode, returnsignal, output, isTimeout):
+    def determine_result(self, run):
         verifierDir = None
         regex = re.compile("verifier used in MetaVal is (.*)")
-        for line in output[:20]:
+        for line in run.output[:20]:
             match = regex.match(line)
             if match is not None:
                 verifierDir = match.group(1)
@@ -66,7 +68,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         assert isinstance(
             tool, BaseTool2
         ), "we expect that all wrapped tools extend BaseTool2"
-        return tool.determine_result(returncode, returnsignal, output, isTimeout)
+        return tool.determine_result(run)
 
     def cmdline(self, executable, options, task, rlimits):
         parser = argparse.ArgumentParser(add_help=False, usage=argparse.SUPPRESS)
@@ -101,7 +103,7 @@ class Tool(benchexec.tools.template.BaseTool2):
             BaseTool2.ToolLocator(tool_directory=self.TOOL_TO_PATH_MAP[verifierName])
         )
         wrappedtask = BaseTool2.Task(
-            input_files=["../output/ARG.c"],
+            input_files=["output/ARG.c"],
             identifier=task.identifier,
             property_file=task.property_file,
             options=task.options,
@@ -123,7 +125,7 @@ class Tool(benchexec.tools.template.BaseTool2):
             ]
             + additionalPathArgument
             + witnessTypeArgument
-            + list(task.single_input_file)
+            + [task.single_input_file]
             + ["--"]
             + wrappedOptions
         )
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -124,6 +124,7 @@ class Tool(benchexec.tools.template.BaseTool2):
             ]
             + additionalPathArgument
             + witnessTypeArgument
+            + ["--property", task.property_file]
             + [task.single_input_file]
             + ["--"]
             + wrappedOptions
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -103,10 +103,16 @@ class Tool(benchexec.tools.template.BaseTool2):
         wrapped_executable = tool.executable(
             BaseTool2.ToolLocator(tool_directory=self.TOOL_TO_PATH_MAP[verifierName])
         )
+        wrappedtask = BaseTool2.Task(
+            input_files=[os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
+            identifier=task.identifier,
+            property_file=task.property_file,
+            options=task.options,
+        )
         wrappedOptions = tool.cmdline(
             wrapped_executable,
             options,
-            [os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
+            wrappedtask,
             rlimits,
         )
 
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -6,7 +6,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
-import benchexec.util as util
 import benchexec.tools.template
 import os
 import re
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -8,6 +8,7 @@
 import argparse
 import benchexec.util as util
 import benchexec.tools.template
+from benchexec.tools.template import BaseTool, BaseTool2
 import contextlib
 import os
 import re
@@ -15,7 +16,7 @@ import sys
 import threading
 
 
-class Tool(benchexec.tools.template.BaseTool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
     This is the tool info module for MetaVal.
 
@@ -42,8 +43,8 @@ class Tool(benchexec.tools.template.BaseTool):
         self.lock = threading.Lock()
         self.wrappedTools = {}
 
-    def executable(self):
-        return util.find_executable("metaval.sh")
+    def executable(self, toolLocator):
+        return toolLocator.find_executable("metaval.sh")
 
     def name(self):
         return "metaval"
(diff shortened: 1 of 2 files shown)


benchexec/tools/metaval.py
```diff
@@ -108,14 +116,20 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
...
         if verifierName in self.wrappedTools:
-            with self._in_tool_directory(verifierName) as oldcwd:
-                wrappedOptions = self.wrappedTools[verifierName].cmdline(
-                    self.wrappedTools[verifierName].executable(),
-                    options,
-                    [os.path.relpath(os.path.join(oldcwd, "output/ARG.c"))],
-                    os.path.relpath(os.path.join(oldcwd, propertyfile)),
-                    rlimits,
+            tool = self.wrappedTools[verifierName]
+            assert isinstance(tool, BaseTool2): "we expect that all wrapped tools extend BaseTool2"
```
PhilippWendler, line 120: Syntax error:
```suggestion
            assert isinstance(tool, BaseTool2), "we expect that all wrapped tools extend BaseTool2"
```

benchexec/tools/metaval.py
```diff
@@ -77,12 +78,19 @@ def determine_result(self, returncode, returnsignal, output, isTimeout):
...
-        with self._in_tool_directory(verifierName):
-            return self.wrappedTools[verifierName].determine_result(
-                returncode, returnsignal, output, isTimeout
-            )
 
-    def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
+        tool = self.wrappedTools[verifierName]
+        if isinstance(tool, BaseTool2):
+            return tool.determine_result(returncode, returnsignal, output, isTimeout)
+        elif isinstance(tool, BaseTool):
```
PhilippWendler, line 85: Should this be removed as well? I would assume that the import of `BaseTool` as well as `_in_tool_directory` can be removed completely now.

PhilippWendler (changes requested):

benchexec/tools/metaval.py
MartinSpiessl: Yep, fixed, thank!

benchexec/tools/metaval.py
MartinSpiessl: Sure, done!

benchexec/tools/metaval.py
```diff
@@ -108,14 +111,22 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
                 ).Tool()
 
         if verifierName in self.wrappedTools:
```
PhilippWendler, line 113: Just noticed now (but it actually had the same behavior before):
There is no way this condition can evaluate to `False`, except if a concurrent thread would delete something from `self.wrappedTools`, but of course this won't happen. So we could remove this condition and have less indentation.

benchexec/tools/metaval.py
```diff
@@ -108,14 +111,22 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
...
+            ), "we expect that all wrapped tools extend BaseTool2"
+            wrapped_executable = self.wrappedTools[verifierName].executable(
+                BaseTool2.ToolLocator(
+                    tool_directory=self.TOOL_TO_PATH_MAP[verifierName]
                 )
+            )
+            wrappedOptions = self.wrappedTools[verifierName].cmdline(
+                wrapped_executable,
+                options,
+                [os.path.relpath(os.path.join(os.getcwd(), "output/ARG.c"))],
```
PhilippWendler, line 126: This actually always produces `["output/ARG.c"]` because you first join it with the current directory and then make it relative to the current directory again. So this can be simplified.

But more importantly, new tool-info modules require a `Task` instance here.

benchexec/tools/metaval.py
```diff
@@ -126,7 +137,7 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
                 ]
                 + additionalPathArgument
                 + witnessTypeArgument
-                + tasks
+                + list(task.input_files_or_empty)
```
PhilippWendler, line 140: Probably `task.input_files` or even `task.single_input_file`? Does MetaVal handle more than one input file?

benchexec/tools/metaval.py
```diff
@@ -108,14 +111,22 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
...
 
         if verifierName in self.wrappedTools:
-            with self._in_tool_directory(verifierName) as oldcwd:
-                wrappedOptions = self.wrappedTools[verifierName].cmdline(
-                    self.wrappedTools[verifierName].executable(),
-                    options,
-                    [os.path.relpath(os.path.join(oldcwd, "output/ARG.c"))],
-                    os.path.relpath(os.path.join(oldcwd, propertyfile)),
-                    rlimits,
+            tool = self.wrappedTools[verifierName]
```
PhilippWendler, line 114: This variable is used only once, wouldn't it make sense to also use it in the remaining places where the dict is accessed in the following code?

benchexec/tools/metaval.py
```diff
@@ -108,14 +111,22 @@ def cmdline(self, executable, options, tasks, propertyfile=None, rlimits={}):
...
+            assert isinstance(
+                tool, BaseTool2
+            ), "we expect that all wrapped tools extend BaseTool2"
+            wrapped_executable = self.wrappedTools[verifierName].executable(
```
PhilippWendler, line 118: I wonder whether it makes sense to also call `tool.version()` afterwards. The BenchExec docs do not strictly guarantee that this method is called, but they do mention that the general methods (which includes `version()`) are called before the run-specific methods, and it specifically allows tool-info modules to store the result from e.g. `version()` in a field.¹

The reason why this is relevant is that some tool-info modules store information about the tool version to support different versions. I think most do this in the call to `executable()`, but some might do it in `version()`. So to make sure, we could call the latter method here.

And to make it efficient (a call to `version()` typically starts the verifier) we could move both calls into the block that loads the wrapped tool-info module, so that they are called only once.

¹ The commit on which this PR is based is so old that it does not include this part of the docs, in case you wonder where it is. Merging current master into this PR might make sense.

PhilippWendler (changes requested):
I think `_in_tool_directory` can also be removed?

And note that `determine_result` also needs to be updated.

benchexec/tools/metaval.py
MartinSpiessl: I would rather get this PR done and not misuse it for forcing me to do refactorings. If there are concerns about the code quality please just open an issue.

>  except if a concurrent thread would delete something

exactly this, so let's be defensive and not change this.

benchexec/tools/metaval.py
MartinSpiessl: This really is overengineering on the benchexec side. Why not throw everything we have on the tools and let them deal with it?

But OK I will change it to task.single_input_file

benchexec/tools/metaval.py
MartinSpiessl: Sure, we can do this. This is actually more feedback than I want. It doesn't have to be super perfect and without spaces for improvement. It is my responsibility anyway to maintain this tool-info module, so please let it up to me to decide when it makes sense improving the code structure.

benchexec/tools/metaval.py
PhilippWendler: There is no existing code that removes stuff from `self.wrappedTools`, so no need to be defensive here. Other threads randomingly changing stuff from your fields would break a lot anyway.

benchexec/tools/metaval.py
PhilippWendler: Well, it gives nicer error messages to the user. But nobody is forced to use this, so I don't see why this is over-engineering? I just don't see a reason to use the longer `task.input_files_or_empty` if a shorter method will be easier to understand and provide better user experience?

benchexec/tools/metaval.py
MartinSpiessl: I am totally lost here and cannot tell what you want me to do.

> we could move both calls into the block that loads the wrapped tool-info module, so that they are called only once.

How will this lead to the tools only being executed once? Which second call apart from the one cause by `version()` are you referring to? What do you mean with `loads the wrapped tool-info module`? Do you mean the import and instantiation of the wrapped verifiers `Tool` class?

> The BenchExec docs do not strictly guarantee that this method is called, but they do mention that the general methods (which includes version()) are called before the run-specific methods, and it specifically allows tool-info modules to store the result from e.g. version() in a field

If there is no strict guarantee that these methods are called, then this should be formalized. Tools really need to work even if their `version()` method is not called. Everything else would be bad software architecture.

benchexec/tools/metaval.py
PhilippWendler: Sorry, but I will not stop giving helpful feedback when I notice something.

> It is my responsibility anyway to maintain this tool-info module, so please let it up to me to decide when it makes sense improving the code structure.

This is not true. Tool-info modules are not the sole responsibility of the original submitter. BenchExec developers always have and will continue to also work on these modules, for example to improve code quality (as has happened for `metaval.py` in the past) or even to add features (example: #646). There might also be other developers/maintainers of a tool (and thus of the tool-info module) in the future. So we do require a certain level of code quality.

Also, you are using strawman arguments here, please do not do this:

> It doesn't have to be super perfect and without spaces for improvement.

Nobody has required or expected the code to be perfect.

> not misuse it for forcing me to do refactorings

(cited here from other comment)

Nobody is forcing you to do this. I carefully worded all my suggestions for code improvements as suggestions.

benchexec/tools/metaval.py
MartinSpiessl: Hmm, this worked alright at the time we drafted the code (I think it was https://github.com/sosy-lab/benchexec/pull/602), what changed inbetween?

benchexec/tools/metaval.py
PhilippWendler: > > we could move both calls into the block that loads the wrapped tool-info module, so that they are called only once.
> 
> How will this lead to the tools only being executed once?

The code block that loads the wrapped tool-info module (inside `if verifierName not in self.wrappedTools:`) is executed only once. If `tool.version()` is called only there, it will also be called only once.

> Which second call apart from the one cause by `version()` are you referring to?

With "both" I meant the calls to `executable()` and `version()` of the wrapped module.

> What do you mean with `loads the wrapped tool-info module`? Do you mean the import and instantiation of the wrapped verifiers `Tool` class?

Exactly.

> > The BenchExec docs do not strictly guarantee that this method is called, but they do mention that the general methods (which includes version()) are called before the run-specific methods, and it specifically allows tool-info modules to store the result from e.g. version() in a field
> 
> If there is no strict guarantee that these methods are called, then this should be formalized. Tools really need to work even if their `version()` method is not called. Everything else would be bad software architecture.

Note that this is not about tools not working if their version is not retrieved, but about tool-info modules not working without `version()` being called. And because many tool-info modules have methods whose implementation depends on the result of `version()` but this method is somewhat costly, we consider it ok to cache its result as a side effect in a field (maybe not the best architecture, but not bad, and we do not want to require tool-info modules to be perfect).

If we are going to formalize this, I would tend to say that we will guarantee that `executable()` and `version()` will be called, because this would make writing tool-info modules easier with little to no cost for BenchExec itself, and there are many tool-info modules but only very few callers. Furthermore, this way we do not introduce a new requirement for tool-info modules right after the new API was introduced.

Of course, MetaVal's tool-info module needs to support only a limited set of other tool-info modules and for these it might be fine if `version()` is never called (it is fine for CPAchecker's module, for example). So you can also check whether this is the case for all tool-info modules that you want to support and keep the current code (though a comment about this would probably be valuable for the future).

benchexec/tools/metaval.py
PhilippWendler: This was done as part of #592, i.e., the introduction of `BaseTool2`. The changes to MetaVal's tool-info module were based on an intermediate draft version of that API.

If you are not yet familiar with it, I recommend having a look at our [migration guide](https://github.com/sosy-lab/benchexec/blob/master/doc/tool-integration.md#migrating-tool-info-modules-to-new-api) that covers the full set of differences between `BaseTool` and `BaseTool2`.

MartinSpiessl: That migration guide can be improved, it doesn't even list the method signatures. It will be quicker reading the actual code.

benchexec/tools/metaval.py
MartinSpiessl: > Nobody is forcing you to do this

Well it is my job, and my boss wants this to be done. Otherwise I would totally ditch it.  
I am just under a lot of stress currently because of the SV-COMP related activities. I would really welcome if you respected my wishes here and concentrated on the most important things. I also need to fix the tool. This tool-info business probably costs me more time than working on the actual tool.

But OK, it is your decision and I cannot force you not to give feedback. So I will just try my best to address everything to get it over with. I am just generally frustrated with the amount of time that is wasted with this totally trivial manner.

For next year (if there is a next year with MetaVal) I will move all the MetaVal-specific code into MetaVal itself and do the wrapped verifier handling there, so the tool-info module will become minimal and I won't have to deal with PRs like this one anymore. You need to be aware that the experience with working on tool-info modules is far from - and never has been - pleasant for me.

> Also, you are using strawman arguments here,

Of course I am using exaggerations like any normal being,you have to read between the lines and don't take phrases like that literally! I wanted to signal you that my priorities currently are different to yours. I would rather get the PR merged in a state where the tool-info module is working/functional, then work on the tool and later come back to work on an issue where the quality improvements that could be done are listed.

It is not like I am gone after this PR is merged, as it might be with random one-time contributors. But I already rage-closed this PR (and reopened it after I calmed down) because it is really pulling me down.

benchexec/tools/metaval.py
MartinSpiessl: I have no nerves to discuss this, so I will just remove it

benchexec/tools/metaval.py
MartinSpiessl: > Well, it gives nicer error messages to the user

But it adds too much complexity just for an error message. The interface of `Task` would be pretty clear, but it is overly complicated with all those `@property` methods.

The whole benchexec tool-info module thing is over-engineered. A uniform wrapper around each tool with a clear command line interface would be way better if you ask me (I know you will probably argue that the tool-info modules are exactly that, but I disagree). There are a lot of YAGNIs and WETs involved in this tool-info module business. It is way too bloaty for what it is actually doing. It also means we have to develop both in the tool repository and at the tool-info module level. Functionality can be shifted between these two. tool-info modules need to disappear if you ask me.

If you want to continue this discussion, please let's not do it here. I implemented the change you asked for so this thread should be resolved.

benchexec/tools/metaval.py
MartinSpiessl: > Of course, MetaVal's tool-info module needs to support only a limited set of other tool-info modules and for these it might be fine if version() is never called (it is fine for CPAchecker's module, for example). So you can also check whether this is the case for all tool-info modules that you want to support and keep the current code (though a comment about this would probably be valuable for the future).

I didn't have problems so far, and would deal with them once and if they occur. We can also have a look into this after SV-COMP, if you want to do this just add an issue, assign me and set the due date to beginning of December.

benchexec/tools/metaval.py
MartinSpiessl: The original change that was requested in this thread has been implemented, so I will resolve the thread.

benchexec/tools/metaval.py
MartinSpiessl: > This actually always produces ["output/ARG.c"] because you first join it with the current directory and then make it relative to the current directory again. So this can be simplified.

Yeah this is probably broken anyway because it actually has to be ["../output/ARG.c"], because the output folder will actually be created outside of the wrapped tool's directory. I guess this is a result of the refactoring where we replace the context manager. At that time the code probably used the cached working directory of metaval instead of `os.getcwd()`. I will need to do some tool testing anyway to see whether this still works.

MartinSpiessl: OK, there are still a lot of tool info modules that are not yet refactored to implement`BaseTool2`, so the missing checks for `BaseTool` or `BaseTool2` that you wanted me to remove actually are now hindering me. I basically have to wait for https://github.com/sosy-lab/benchexec/pull/646 to be merged.

MartinSpiessl: https://github.com/sosy-lab/benchexec/pull/650/commits/4f1313e86e09889897cb099833e3783265b23b3e should fix this by passing a proper `Task` object, but I have no way of knowing until #646 is merged, I manually merge that PR into this PR (with high chance that this will cause more work than it will save) or readd the `BaseTool` checks

PhilippWendler: Do you mean because you need a working `BaseTool2` tool-info module to test your PR? You can do this with CPAchecker's module, it has been upgraded already.

PhilippWendler: From my point of view, this and the (now 3) unused imports are the only remaining things to do. So please ping me after you finished and tested this, then I will merge.

PhilippWendler: Note that now #646 was merged, so not only CPAchecker's tool-info module but also others are now available for testing.

MartinSpiessl: Open points should be addressed via:
https://github.com/sosy-lab/benchexec/pull/650/commits/444a46ce90c679e9f3213d0360a03461afd0add1
https://github.com/sosy-lab/benchexec/pull/650/commits/e9ae4f3dfe583108ecc62286513277e343b1091a
https://github.com/sosy-lab/benchexec/pull/650/commits/8fee3ad6a5fba4ac314ff2bda016b17174a5c993

I also found some bugs via testing and fixed those as well:
https://github.com/sosy-lab/benchexec/pull/650/commits/0bc36059a125f88f91e5c9c9d2e492bc2d693282

Accidentally overlooked/reintroduced some unused imports, those should be fine with:
https://github.com/sosy-lab/benchexec/pull/650/commits/b2adf40319467c20d072c0966b5e2ecce57d8064

PhilippWendler (approved):
I have to fix another unused import, but now it seems fine.

PhilippWendler:
For the future: force-pushes make it harder to review on GitHub, essentially I have to reread all the code every time. For keeping the branch up-to-date, a regular merge would mean less effort for me. Furthermore, now you even copied a commit from another committer onto your branch, probably due to a wrong rebase, such that now we have that commit twice in the history.

Force pushes can be a nice tool for cleaning up the history of a branch to remove small fixups etc. (but they were not used in this regard here).

MartinSpiessl:
> https://gitlab.com/sosy-lab/software/coveriteam-remote-server/-/merge_requests/6

What makes you thing this could happen by a rebase? I cherry-picked that one because I needed it for testing, and forgot to remove it again when I pushed the last change.

> a regular merge would mean less effort for me

Sure, will do next time.

PhilippWendler:
> What makes you thing this could happen by a rebase?

Because I have seen such things happen due to rebases. I would happen if rebasing something else onto a branch instead of the branch on something else.

PR #654, opened 2020-11-10
Fix various issues with number format in table-generator

PhilippWendler:
The fixes include wrong rounding, truncating small values to 0, rounding values to a wrong number of significant digits, not rounding numbers when they should have been, and incorrect alignment of values in the summary table of HTML tables. Both CSV and HTML tables are affected, but in different ways. However, luckily the produced errors are small. The main issue produced rounding errors in the least significant digit, i.e., with two significant digits the rounding error was at most 1%, with more significant digits it was smaller. Values that would be incorrectly truncated to 0 were smaller than 1e-6. Other solved problems issues produced rounding errors that were even smaller, typically in the size of typical floating-point rounding issues.

The main strategy for fixing this is to switch table-generator to use only the `decimal.Decimal` type for representing numeric values and no longer use `float`s. This allows us to afterwards fix the actual rounding problem as well as consistently using a new way of printing numbers that avoids some problems of e.g., `str()`.

Because the diff of this PR will be too large to review it, I have arranged the commits such that the PR can be split into several parts, each part for one group of related changes. I will describe each part in the following in more detail and make sure to push each part separately, such that CI runs for each part and not only once at the end of the PR. (Unfortunately, Travis has problems and GitHub thus shows CI runs as failures, but on [Gitlab](https://gitlab.com/sosy-lab/software/benchexec/-/pipelines?page=1&scope=all&ref=fix-tablegenerator-number-format) one can see that CI succeeds at the crucial points.) Where necessary, I have also separated large test changes from actual code changes. The recommended way to review this PR is to look at the diff of each commit on its own.

In addition to the following changes, there are a small number of refactorings that were necessary or helpful in between. However, I have kept these to a minimum and will do more refactorings and apply several simplifications that are now possible afterwards.

### Part 1: Switch to `Decimal` (7154717, fa1c78e, 6758626, b8fddc2)
This switches tablegenerator to use `Decimal` exclusively. After that we can specify the rounding mode that we want to use instead of relying on the rounding mode that Python uses for `round()` on `float`s without possibility of overriding it. That fixes the biggest rounding problem including the case found in https://github.com/sosy-lab/benchexec/pull/616#discussion_r510832784. Because arithmetic with `Decimal`s does not have the imprecision of `float`, also other rounding problems go away and unfortunately these could not be clearly separated. During this part the integration tests of table-generator are known to fail, but at the end all tests pass again. Cf. commit messages for more details.

### Part 2: Print `Decimal`s correctly (8e53a37, 445edd9, 3876d57, 017c28a, e5c5867, 0356c62, 966d447)
Because `str()` always falls back to scientific notation for small values, and formatting a number with code like `"{:.3f}".format(...)` requires specifying a number of decimal digits (which we do not always know correctly before), the existing code does not correctly output values in decimal notation, and I could not find a method for `Decimal`s in Python that does what we need. So I wrote a method for printing `Decimal`s in decimal notation from scratch. The remaining commits switch more and more cases where `str()` was called to this new utility, each time fixing a different kind of wrong output. In one case this led to so many changes in the integration tests of table-generator that I again splitted the test changes from the actual code changes, and for two commits the integration tests of table-generator are known to fail. At all other times all tests are adjusted at the same time with the code.

### Part 3: Fix tooltip handling (4822427, 5a53125)
This brings the formatting of tooltip values for average and standard deviation of columns with type count (only integral values) back to what it was for the old kind of HTML tables. Cf. commit message of 5a53125.

### Part 4: Fix alignment in summary tables (34ad7b7)
This removes unnecessary alignment whitespace from summary tables. Cf. commit message of 34ad7b7.


lembergerth:
## Review Part 1 (Decimal)

Looks good. There's no more use of any floats in the table-generator module, so I'm confident that all values that are created *inside the module* use Decimal.
I can't reason about values passed from the outside, though.

Typing information and `mypy` could be used to specify `Decimal` or similar as a return type for methods handling decimal values. This would make sure that table_generator really returns Decimal at all points, independent of the type of the input parameters.

benchexec/tablegenerator/util.py
```diff
@@ -139,6 +139,49 @@ def to_decimal(s):
         return None
 
 
+def print_decimal(d):
```
lembergerth, line 142: **Suggestion (non-blocking)**: Don't use a util method `print_decimal`,
but implement an own type for decimals with the corresponding `__str__` method.

The current code seems to work really well, but you still have to remember to call `util.print_decimal(value)` instead of using the common `str(value)`. This is error-prone (and less beautiful).

You could avoid this by implementing a subtype of `Decimal` with the content of `print_decimal` as `__str__` method.

lembergerth:
## Review Part 2 (Print Decimals correctly)

Very nice! Code is way easier to read now, a clear improvement.
Only one suggestion, didn't see any issues.

lembergerth:
I couldn't find commits for Part 3 and 4. Are these not pushed yet?

PhilippWendler:
Thanks for already starting to review! I was pushing the commits step by step in order to get CI run on each crucial point.

PhilippWendler:
> Looks good. There's no more use of any floats in the table-generator module, so I'm confident that all values that are created _inside the module_ use Decimal.
> I can't reason about values passed from the outside, though.

Luckily, table-generator never gets values from the outside except for what it reads from XML files, so we don't have to worry about that.

> Typing information and `mypy` could be used to specify `Decimal` or similar as a return type for methods handling decimal values. This would make sure that table_generator really returns Decimal at all points, independent of the type of the input parameters.

Right. I actually do have a plan on using `pytype`, but to do so I first need to clean up a few cases where type inference does not work, and also fix a few minor bugs found by it. `pytype` so far has worked wonderfully without type annotations, but I'll consider adding a few of them later on for important code parts. For the particular case of this PR, there is a manual `TypeError` thrown if something else than a `Decimal` appears, so we are safe for now.

benchexec/tablegenerator/util.py
PhilippWendler: Good idea, sounds interesting. It would certainly be safer. Not sure yet whether it is worth because we use all the arithmetic methods on `Decimal`s so we would need to add a lot of forwardings methods to our type. But I will keep this in mind for later maybe.

For this PR I checked all occurrences of `str()` in `columns.py`.

PhilippWendler:
Now all four parts are pushed. I do have further refactoring commits, but we can also do these separately (I am pretty sure they do not change any output, and our tests confirm this).

benchexec/tablegenerator/util.py
lembergerth: Would this not be good enough?

```python
import decimal
class PointNotationDecimal(decimal.Decimal): 
   ...:     def __str__(self): 
   ...:         return util.print_decimal(self)  # better: move code here
   ...:
Decimal = PointNotationDecimal  # for a smaller changeset, if desired                           
```

> For this PR I checked all occurrences of str() in columns.py

I also went through it and didn't find any issues.

lembergerth: Sorry, now I get it!
`PointNotationDecimal(1) + PointNotationDecimal(1)` would still produce a `Decimal`, of course.

lembergerth:
## Review Part 3 and Part 4 (adjustments to tooltips and summary footer)

Looks good to me. Small adjustments that produce cleaner table-cell values.

lembergerth (approved):

benchexec/tablegenerator/test_util.py
```diff
@@ -51,3 +52,49 @@ def test_split_string_at_suffix(self):
...
+            self.assertEqual(expected, util.print_decimal(Decimal(value)))
+
+    def test_print_decimal_int(self):
+        # These values should be printed like Decimal prints them after quantizing
+        # to remove the exponent.
+        test_values = ["0e0", "-0e0", "0e20", "1e0", "1e20", "0e10"]
+        for value in test_values:
+            value = Decimal(value)
+            expected = str(value.quantize(1))
+            assert "e" not in expected
```
kfriedberger, line 91: [non-blocking minor] inconsistent usage of assertion, no message available

benchexec/tablegenerator/test_util.py
```diff
@@ -51,3 +52,49 @@ def test_split_string_at_suffix(self):
...
+            self.assertEqual(expected, util.print_decimal(value))
+
+    def test_print_decimal_float(self):
+        # These values should be printed like str prints floats.
+        test_values = ["1e-4", "123e-4", "1234e-4", "1234e-5", "1234e-6"]
+        for value in test_values:
+            expected = str(float(value))
+            assert "e" not in expected, expected
```
kfriedberger, line 99: [non-blocking minor] inconsistent usage of assertion, message available, see above.

kfriedberger (approved):
I found a small inconsistency, that is minor and does not block the merge.
Overall, the commits seem to be valid.

The change for the user is minimal.
It's interesting that nobody complained about rounding issues before. :-)

benchexec/tablegenerator/test_util.py
PhilippWendler: Thanks for looking at this!

Yes, assertion messages are in general inconsistent in BenchExec, which is ok for me because we don't expect them to be shown to users. And if users do happen to see them, they need to report a bug and we need to debug this anyway, so it is just a minor hurdle then.

PR #660, opened 2020-11-16
Add parameter to htmltable generation for a default (fallback) URL state

### LINKED ISSUE #658: Allow specifying initial state for hash routing
It would be nice if table-generator would be able to insert a string into the generated HTML file that would be used as source of initial state for the table if it is opened without a hash part in the URL. Ideally, this would
- work no matter which parameters are set in the URL and need no changes when more parameters are added,
- be ignored if table is opened with empty hash part (`...table.html#` or `...table.html#/` or a similar syntax) such that there is still a way to open it with default state,
- apply the settings before the UI is rendered for the first time in order to not confuse the user with a short flash of different content, and
- modify the hash part of the URL when being applied, such that it is visible in the URL bar afterwards.

More concrete example: After running table-generator with a to-be-added parameter like `table-generator ... --initial-table-state '/table'` and opening the produced HTML table, our code should change the content of the URL bar to `...#/table` and produce an initial view of the table that looks as if this URL would have been specified by the user.

How and where the string is added to the generated table source does not matter for table-generator, we can decide this based on what is the easiest way for the JS code. I would then take care of the integration in table-generator.

lachnerm:
Adds a parameter called `initial-table-state` to the generation of tables that sets the default URL state (i.e. the selected tab and any URL parameters) for the file.
The value of the new parameter needs to exactly match the way our JS scripts create the URL parameters - **there is no error checking or fallback for this string**. I don't think this is necessary since everybody who uses this parameter should know what he's doing - and in case the string actually contains an error, one probably wants to generate the HTML file anew anyway.

The default state is only applied during the first loading of the file and in case there is no selected tab and no URL parameters defined. It is therefore still possible to send links containing URL parameters to files without the default state overriding anything. Simply appending a `#` to the end of the URL will disable any default state behavior and open the `Summary` tab like before.

Fixes #658.


lachnerm:
In case the changes to the files are fine like this, I will upload the fixes to the test files. If there's anything I'm supposed to change, I will upload the fixes afterwards.

PhilippWendler:
I think this is good. I still have to test it and I will think about a better name for the command-line parameter, but in principle this is what we want.

PhilippWendler:
Hm, it does not work for me currently. I tried with `--initial-table-state 'table'` (and `#/table` and `/table` as well) and always the `#/` URL was opened.

For the syntax of the string, I would suggest that we either define the JSON to contain strings like `/table` or strings like `#/table`. table-generator should always accept both forms (with and without `#`) and convert it into the form that is required for the JSON. As error check I think we can check in table-generator that the string starts with `/` (after the optional `#`) because this is what all our hash parts start with, right?

lachnerm:
> Hm, it does not work for me currently. I tried with `--initial-table-state 'table'` (and `#/table` and `/table` as well) and always the `#/` URL was opened.
> 
There was a small bug left when opening a local file for the first time, which I've probably missed since I've always been refreshing the file on testing instead of opening it anew. Well, sorry about that - it should be working now.

> For the syntax of the string, I would suggest that we either define the JSON to contain strings like `/table` or strings like `#/table`. table-generator should always accept both forms (with and without `#`) and convert it into the form that is required for the JSON. 
>
It is now possible to add an additional `/` or `#/` in front of the parameter with the same behavior. So `table`, `/table` and `#/table` should work now and all do the same.
>As error check I think we can check in table-generator that the string starts with `/` (after the optional `#`) because this is what all our hash parts start with, right?
>
Yes this would be possible, I'd just need to remove the possibility to allow the parameter to start without an initial `#/` or `/`. However, I don't think this would be of much benefit compared to also allowing the parameter to start without the other characters.

PhilippWendler:
Thanks. I guess we could even be more strict.

However, in the current form if I open `table.html#/` I still get what is set with `--initial-table-state`, only `table.html#` properly resets it. I am also wondering whether we can simplify both the condition `shouldSetInitial` as well as the creation of `hrefString` in `setConstantHashSearch`. Shouldn't it be possible to only use the initial table state if there is no `#` in the URL? Such that both `table.html#` as well as `table.html#/` ignore the initial state and `hrefString` can be identical to `document.location.href`? Or do I overlook a use case here?

lachnerm:
> However, in the current form if I open `table.html#/` I still get what is set with `--initial-table-state`, only `table.html#` properly resets it. 
>
If you want, I can add the option that the parameter won't be applied if the URL ends with `#/` too. The initial parameter will in this case only be applied on the very first opening of the page, since afterwards there will either be URL parameters blocking it or the automatically added `/#/` from (I guess) the `Router` component. So refreshing without opening a tab and without any URL parameters still won't apply the initial parameter any more. This may of course be a desired behaviour too though. 

>I am also wondering whether we can simplify both the condition `shouldSetInitial` as well as the creation of `hrefString` in `setConstantHashSearch`. Shouldn't it be possible to only use the initial table state if there is no `#` in the URL? Such that both `table.html#` as well as `table.html#/` ignore the initial state and `hrefString` can be identical to `document.location.href`? Or do I overlook a use case here?
>
The current workarounds and multiple conditions are necessary to make the initial parameter work on refreshing. Assuming that we apply the above-mentioned changes, your suggestions for simplification should be possible, yes.

PhilippWendler:
Regarding refreshing: I would say that refreshing should not change the visible state of the table, and as far as I understand this would be the case with my proposal, but not with the current code of the PR, right? (When the table is first loaded, the initial table state gets appended to the URL so any further refreshes should just be based on what is in the URL at that point. Currently, when refreshing the table while looking at the summary tab with no further state set, the initial table state would be reapplied on refresh.)

lachnerm:
> Regarding refreshing: I would say that refreshing should not change the visible state of the table, and as far as I understand this would be the case with my proposal, but not with the current code of the PR, right? (When the table is first loaded, the initial table state gets appended to the URL so any further refreshes should just be based on what is in the URL at that point. Currently, when refreshing the table while looking at the summary tab with no further state set, the initial table state would be reapplied on refresh.)

Yes, exactly as you said. So I'll disable the reapplication of the initial state when refreshing.

PhilippWendler (approved):
Thanks, this looks good so far. I just polished handling and messages a little bit.

Is it possible to write a meaningful test for this?

lachnerm:
Hm, I think writing tests for the selected menu item part of the parameter should be possible. So we insert e.g. `table` as initial state and test whether the `table` tab is open when rendering the app (and of course for the other menu items too).
Everything beyond that becomes pretty specific for the corresponding part of the application and the used test data. It should theoretically be possible to insert an initial state like 
```
scatter?toolX=1&columnX=3&toolY=0&columnY=0&scaling=Linear&results=All&line=1000
``` 
and test whether all the parameters are correctly applied, however this would be more like a test for the `ScatterPlot` itself instead of the initial parameter, so I don't think this would make much sense. Also, the initial parameter doesn't contain much logic itself - it actually only sets the URL and lets the app handle the rest. I still think testing whether the menu item part of the parameter is applied would make sense, and if this part works, we can assume that the rest will be applied too - whether or not the rest does what it's supposed to be, isn't part of the initial parameters job.


So basically I'd suggest only testing for whether string like `scatter`, `scatter?toolX=0&columnX=3&results=All` and `scatter?notExistingParam=notExisting` or whatever actually lead to the `ScatterPlot` and ignore the rest of the parameter. A test with a parameter referring to a non-existent tab is possible too. What do you say?

Oh, and we can of course add a test for the deactivation of the initial parameter by forcefully setting the `document.location.href` attribute before the rendering and checking the URL afterwards.

PhilippWendler:
Sounds good!

lachnerm:
I had to make some small changes regarding how the parameters that define whether the plots should be rendered flexible or with a fixed width and height are propagated through the app. This doesn't really change any functionalities though, it was just necessary for the tests to work.

PhilippWendler (approved):

PR #666, opened 2020-11-21
update TI - predatorHP - removed use of stderr in get version

skanav:
update tool info module of predatorHP

### DIFF
--- a/benchexec/tools/predatorhp.py
+++ b/benchexec/tools/predatorhp.py
@@ -25,7 +25,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "PredatorHP"
 
     def version(self, executable):
-        return self._version_from_tool(executable, use_stderr=True)
+        return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, task, rlimits):
         spec = ["--propertyfile", task.property_file] if task.property_file else []


PhilippWendler:
I guess this would break backwards compatibility with older tool versions?

@versokova Is there a way to get the version that is compatible with all tool versions?

If not, I guess we could try a fallback solution.

skanav:
I am surprised that it worked before. version is printed on `stdout` and `stderr` is empty. That is causing the version number to be `empty`.

versokova:
@skanav I am pretty sure that the version is printed on `stderr`. See [code](https://github.com/versokova/predatorhp/blob/master/predatorHP.py#L108).

skanav:
> @skanav I am pretty sure that the version is printed on `stderr`. See [code](https://github.com/versokova/predatorhp/blob/master/predatorHP.py#L108).

yes, thanks for the clarification. It seems that it prints on `stdout` in python3 and `stderr` in python2.

skanav:
Closing the PR as the CI in verifier archives passed after adding packages `python2`and `python-is-python2` to `Dockerfile.user`.

For reference:
https://gitlab.com/sosy-lab/sv-comp/archives-2021/-/merge_requests/60
https://gitlab.com/sosy-lab/sv-comp/archives-2021/-/merge_requests/58

PR #669, opened 2020-11-23
Update tool-info for witnesslint.

SvenUmbricht:
This incorporates the changes requested by @kfriedberger.

### DIFF
--- a/benchexec/tools/witnesslint.py
+++ b/benchexec/tools/witnesslint.py
@@ -7,6 +7,7 @@
 
 import benchexec.result as result
 import benchexec.tools.template
+import re
 
 
 class Tool(benchexec.tools.template.BaseTool2):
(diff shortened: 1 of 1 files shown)


benchexec/tools/witnesslint.py
```diff
@@ -25,10 +26,17 @@ def version(self, executable):
         version_string = self._version_from_tool(executable)
         return version_string.partition("version")[2].strip().split(" ")[0]
 
+    def _code_matches_tool_output(self, tool_output, exit_code):
+        pattern = re.compile(rf"Exit code: {exit_code}")
```
PhilippWendler, line 30: f-strings are not supported by all Python versions that BenchExec supports, so we would prefer this:
```suggestion
        pattern = re.compile(r"Exit code: {}".format(exit_code))
```

Orthogonally: Why use regexps here? `return "Exit code: {}".format(exit_code) in tool_output[-1]` would work just as well? (And if you expect the last line to be exactly this string, just use `==` instead of `in`.

benchexec/tools/witnesslint.py
```diff
@@ -25,10 +26,17 @@ def version(self, executable):
...
+        if pattern.match(tool_output[-1]):
+            return True
+        return False
+
     def determine_result(self, run):
         if run.exit_code.value == 0:
-            return result.RESULT_TRUE_PROP
+            if self._code_matches_tool_output(run.output, 0):
```
PhilippWendler, line 37: It is up to you to decide, but is it worth double checking here? (I find it redundant and surprising for a tool to output its own exit code.)

In general, I would expect that a tool either outputs its result as some kind of meaningful message (not `Exit code: 0`), and/or sets its exit code appropriately. The tool-info module could then check either, or both, of course.

benchexec/tools/witnesslint.py
SvenUmbricht: You're right, `==` is enough, I was just overthinking it.

benchexec/tools/witnesslint.py
SvenUmbricht: I am not 100% sure that this is necessary here, but in the next line it is, to prevent misinterpretations of the exit code.
For example, in this table (https://sv-comp.sosy-lab.org/2021/results/results-verified/cpa-seq.results.SV-COMP21.All.table.html#/table - last column) we get a wrong result: an exception occurs and we get 1 as exit code - but not because this was set by the linter, so it should not be reported as `false`.

The additional check is to make sure that the perceived exit code was indeed set by witnesslint.

PR #675, opened 2020-11-24
Automatic timeout from rlimits

gernst:
Leave some time at the end to write test suite

### DIFF
--- a/benchexec/tools/cmaesfuzz.py
+++ b/benchexec/tools/cmaesfuzz.py
@@ -32,7 +32,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         # add a time limit if not given
         # that is hopefully sufficient to write all tests
         if "-t" not in options and rlimits.cputime:
-            # at least three seconds + 1% of overall time
+            # at least five seconds + 1% of overall time
             timeout = int(rlimits.cputime * 0.99 - 5)
             # but don't add negative timeout
             if timeout > 0:
--- a/benchexec/tools/cmaesfuzz.py
+++ b/benchexec/tools/cmaesfuzz.py
@@ -33,7 +33,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         # that is hopefully sufficient to write all tests
         if "-t" not in options:
             # at least three seconds + 1% of overall time
-            timeout = int(rlimits.cputime*0.99 - 5)
+            timeout = int(rlimits.cputime * 0.99 - 5)
             # but don't add negative timeout
             if timeout > 0:
                 options = options + ["-t", timeout]
--- a/benchexec/tools/cmaesfuzz.py
+++ b/benchexec/tools/cmaesfuzz.py
@@ -32,8 +32,8 @@ class Tool(benchexec.tools.template.BaseTool2):
         # add a time limit if not given
         # that is hopefully sufficient to write all tests
         if "-t" not in options and rlimits.cputime:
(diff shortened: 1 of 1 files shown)


benchexec/tools/cmaesfuzz.py
```diff
@@ -29,6 +29,16 @@ def executable(self, tool_locator):
         return tool_locator.find_executable("fuzzer")
 
     def cmdline(self, executable, options, task, rlimits):
+        # add a time limit if not given
+        # that is hopefully sufficient to write all tests
+        if "-t" not in options:
```
PhilippWendler, line 34:
```suggestion
        if "-t" not in options and rlimits.cputime:
```
A time limit does not always exist.

benchexec/tools/cmaesfuzz.py
```diff
@@ -29,6 +29,16 @@ def executable(self, tool_locator):
...
+            # at least three seconds + 1% of overall time
+            timeout = int(rlimits.cputime*0.99 - 5)
```
PhilippWendler, line 36: Mismatch with comment?

benchexec/tools/cmaesfuzz.py
```diff
@@ -29,6 +29,16 @@ def executable(self, tool_locator):
...
+            # but don't add negative timeout
+            if timeout > 0:
+                options = options + ["-t", timeout]
```
PhilippWendler, line 39:
```suggestion
                options = options + ["-t", str(timeout)]
```

benchexec/tools/cmaesfuzz.py
```diff
@@ -29,6 +29,16 @@ def executable(self, tool_locator):
...
+            else:
+                options = options + ["-t", rlimits.cputime]
```
PhilippWendler, line 41:
```suggestion
                options = options + ["-t", str(rlimits.cputime)]
```

PhilippWendler (changes requested):
Please test my suggestions. Didn't this crash in your tests before the `str()` calls are added? Would be weird if it didn't...

benchexec/tools/cmaesfuzz.py
PhilippWendler: Maybe you would also like to implement this offset within your tool? I.e., just passing the actual time limit to the tool, and the tool itself then deduces when it should stop?

Might be easier to update that way and more convenient for users who use the tool outside of BenchExec. The tool could also make a better estimate, i.e., depending on the size of the test suite... Or update the calculation if later versions are faster/slower.

gernst: Maybe, but I'd rather not (the tool's time limit should be interpreted for finding test-cases, not writing them out)

PhilippWendler: As you prefer.

But please make the comment consistent with the code and confirm that it works in tests now.

dbeyer:
@gernst and @PhilippWendler There are unresolved conversations. Anything to add?

PhilippWendler:
Yes, waiting for https://github.com/sosy-lab/benchexec/pull/675#discussion_r529677243 to be addressed.

gernst:
Sure. I'll get to it soon

gernst:
Ok this is fixed and tested now, hence ready to merge, thanks for the feedback.

PhilippWendler (approved):

PR #676, opened 2020-11-24
Updated required paths for SMACK

zvonimir:
This should hopefully fix the problem with the first preliminary run.
I forgot to update our paths here.

### DIFF
--- a/benchexec/tools/smack.py
+++ b/benchexec/tools/smack.py
@@ -14,8 +14,18 @@ import re
 
 class Tool(benchexec.tools.template.BaseTool):
 
-    REQUIRED_PATHS = ["bin", "lib", "share", "smack-deps", "smack.sh",
-                      "boogie", "corral", "llvm", "lockpwn", "smack"]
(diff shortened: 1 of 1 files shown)


PhilippWendler (commented):
Would it be possible to have the union of dependencies for old and new versions here? Then it would still be possible to benchmark older versions.

Patterns that don't match anything are not a problem, they will be ignored.

You could test the result with the usual `benchexec.test_tool_info` command.

zvonimir:
Good point. Done!

PhilippWendler:
Thanks!

Our style guide requires this to look like this, though:
```
    REQUIRED_PATHS = [
        "bin",
        "lib",
        "share",
        "smack-deps",
        "smack.sh",
        "boogie",
        "corral",
        "llvm",
        "lockpwn",
        "smack",
    ]
```
I would reformat it myself, but I don't have the permission to push to your PR. Can you please either run our code formatter `black .` or update the formatting manually?

PhilippWendler (approved):

PR #677, opened 2020-11-25
Update fusebmc tool

kaled-alshmrany:
We have changed the REQUIRED_PATHS - adding the fusebmc_inustrment

### DIFF
--- a/benchexec/tools/fusebmc.py
+++ b/benchexec/tools/fusebmc.py
@@ -7,18 +7,21 @@
 
 import benchexec.tools.esbmc as esbmc
 
+
 class Tool(esbmc.Tool):
     """
     This class serves as tool adaptor for FuSeBMC (https://github.com/kaled-alshmrany/FuSeBMC)
     """
+
     REQUIRED_PATHS_TESTCOMP20 = ["esbmc", "esbmc-wrapper.py", "my_instrument"]
     REQUIRED_PATHS_TESTCOMP21 = [
         "esbmc",
         "fusebmc.py",
         "FuSeBMC_inustrment/FuSeBMC_inustrment",
         "fusebmc_output",
-        "map2check-fuzzer"
+        "map2check-fuzzer",
     ]
+
     def name(self):
         return "FuSeBMC"
 
@@ -38,4 +41,4 @@ class Tool(esbmc.Tool):
             paths = self.REQUIRED_PATHS_TESTCOMP20
         elif self._version > 20:
             paths = self.REQUIRED_PATHS_TESTCOMP21
-        return paths
+        return self._program_files_from_executable(executable, paths)
--- a/benchexec/tools/fusebmc.py
+++ b/benchexec/tools/fusebmc.py
@@ -19,4 +19,7 @@ class Tool(esbmc.Tool):
         return "FuSeBMC"
     
     def executable(self, tool_locator):
-        return tool_locator.find_executable("fusebmc.py")
+        try:
+            return tool_locator.find_executable("fusebmc.py")
+        except ToolNotFoundException:
+            return super().executable(tool_locator)
--- a/benchexec/tools/fusebmc.py
+++ b/benchexec/tools/fusebmc.py
@@ -7,19 +7,35 @@
 
 import benchexec.tools.esbmc as esbmc
 
-
 class Tool(esbmc.Tool):
     """
     This class serves as tool adaptor for FuSeBMC (https://github.com/kaled-alshmrany/FuSeBMC)
     """
-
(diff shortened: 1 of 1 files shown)


PhilippWendler (commented):
Currently, this would break benchmarking of older FuSeBMC versions, correct?

Could you make this backwards compatible? For `REQUIRED_PATHS`, it is easy, just use the union of all required paths. For `executable`, a fallback solution could be used.

Are all other methods still working if there is a new main executable of the tool? Or should FuSeBMC now be decoupled from ESBMC's tool-info module?

rafaelsamenezes:
> Currently, this would break benchmarking of older FuSeBMC versions, correct? Could you make this backwards compatible?

You mean for the competition? Yes it would 

> For REQUIRED_PATHS, it is easy, just use the union of all required paths.

Ok

> For executable, a fallback solution could be used.

Checking for the file existence should be enough, right? 

```python3
    if(os.path.isfile("fusebmc.py)):
        return tool_locator.find_executable("fusebmc.py")
    return super().executable(tool_locator)
```

Would this be ok?

PhilippWendler:
> > For executable, a fallback solution could be used.
> 
> Checking for the file existence should be enough, right?
> 
> ```python
> def executable(self, tool_locator):
>     if(os.path.isfile("fusebmc.py)):
>         return tool_locator.find_executable("fusebmc.py")
>     return super().executable(tool_locator)
> ```
> 
> Would this be ok?

`tool_locator.find_executable` actually does not necessarily look in the current directory, and we want to avoid hardcoding any dependency on the current directory here (which your `isfile` call would do). So just call `tool_locator.find_executable`, catch the `ToolNotFoundException` and do the fallback in case of exception.

rafaelsamenezes:
> For REQUIRED_PATHS, it is easy, just use the union of all required paths.

@PhilippWendler don't you mean intersection?

BTW, we are adding a new folder because we need the existence of that folder when running benchexec. When we run on the competition environment we are having an issue with the folder not existing.

PhilippWendler:
> > For REQUIRED_PATHS, it is easy, just use the union of all required paths.
> 
> @PhilippWendler don't you mean intersection?

No, you need to list all the paths that are required as dependency for any of the tool versions.

Is `inustrment` a typo?

Please also add an import for the exception and remove the superfluous whitespace from line 20.

rafaelsamenezes:
> No, you need to list all the paths that are required as dependency for any of the tool versions.

I still don't understand, the older versions didn't contain `fusebmc.py`, if I add it to the required paths it will break those. Also `esbmc-wrapper.py` will not exist anymore if I add it to the required paths it will cause issues for newer releases

kaled-alshmrany:
> Is inustrment a typo?

No, it is not, you can keep it

PhilippWendler:
>  Also `esbmc-wrapper.py` will not exist anymore if I add it to the required paths it will cause issues for newer releases

This wouldn't cause problems, patterns that don't match anything in `REQUIRED_PATHS` will simply be ignored. But the solution with the version check is also ok, of course, and maybe clearer. However, `program_files` cannot simply return the paths from `REQUIRED_PATHS`, it must handle the case where the tool is not in the current directory. The easiest way to do this is to call `_program_files_from_executable` like the default implementation of `program_files` does.

And could you please adjust the format as the CI [says](https://travis-ci.org/github/sosy-lab/benchexec/jobs/745931068#L955-L981)? (Or simply use the Python formatter `black`.)

rafaelsamenezes:
> This wouldn't cause problems, patterns that don't match anything in `REQUIRED_PATHS` will simply be ignored. But the solution with the version check is also ok, of course, and maybe clearer. However, `program_files` cannot simply return the paths from `REQUIRED_PATHS`, it must handle the case where the tool is not in the current directory. The easiest way to do this is to call `_program_files_from_executable` like the default implementation of `program_files` does.

You mean this? 

```python3
    def program_files(self, executable):
        if self._version == 20:
            paths = self.REQUIRED_PATHS_TESTCOMP20
        elif self._version > 20:
            paths = self.REQUIRED_PATHS_TESTCOMP21
        return self._program_files_from_executable(executable, paths)
```
Or should I do the same as the `program_files` from `Base2`?

So instead of `return self._program_files_from_executable(executable, paths)` we'd use `return [executable] + self._program_files_from_executable(executable, paths)`

PhilippWendler:
Yes, this is what I meant. Whether you need to return `[executable]` in addition to the result of `_program_files_from_executable` simply depends on whether `REQUIRED_PATHS` includes the main executable or not.

PhilippWendler (approved):

PR #680, opened 2020-11-27
Update data model smack

skanav:
updated tool info module for SMACK to BaseTool2

The changes in the function `determine_result` need more thorough review.

### DIFF
--- a/benchexec/tools/smack.py
+++ b/benchexec/tools/smack.py
@@ -58,7 +58,10 @@ class Tool(benchexec.tools.template.BaseTool2):
         modifications to make just before calling SMACK.
         """
         data_model_param = get_data_model_from_task(task, {ILP32: "-m32", LP64: "-m64"})
-        if data_model_param and "--clang-options" not in options:
+        print(options)
+        if data_model_param and not any(
+            option.startswith("--clang-options=") for option in options
+        ):
             options += ["--clang-options=" + data_model_param]
 
         if task.property_file:
--- a/benchexec/tools/smack.py
+++ b/benchexec/tools/smack.py
@@ -6,13 +6,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import benchexec.result as result
-import benchexec.util as util
 import benchexec.tools.template
+from benchexec.tools.template import UnsupportedFeatureException
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 
 import re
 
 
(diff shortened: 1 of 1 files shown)


benchexec/tools/smack.py
```diff
@@ -53,37 +52,43 @@ def name(self):
...
         """
         Allows us to define special actions to be taken or command line argument
         modifications to make just before calling SMACK.
         """
-        assert len(tasks) == 1
-        assert propertyfile is not None
-        prop = ["--svcomp-property", propertyfile]
-        return [executable] + options + prop + tasks
+        data_model_param = get_data_model_from_task(task, {ILP32: "-m32", LP64: "-m64"})
+        if data_model_param and "--data-model" not in options:
```
PhilippWendler, line 61: `--data-model` is wrong here, isn't it? But `"--clang-options" not in options` would also not make sense, because it probably never appears on its own.

@zvonimir Is `--clang-options=...` allowed to appear more than once? If not, we could use something like `any(option.startswith("--clang-options=") for option in options)`. If it is allowed to appear more than once, we might simply always append the option for the data model.

skanav: Sorry, it was a copy paste mistake. fixed now.

PhilippWendler: As I wrote in my comment: `"--clang-options" not in options` does not make sense because it will never be true.

skanav: updated

zvonimir (approved):
I think this looks ok. We'll have to test it though.

benchexec/tools/smack.py
zvonimir: For SVCOMP, `--clang-options` will be set only once, which is exactly here to determine 32 vs 64 bits. So this should be ok I think.

PhilippWendler:
There was an error in the result-handling logic, I fixed this.

Note that the result-handling logic is still different from the previous version, though. The current version checks whether for example `invalid pointer dereference` occurs in any line of the output, whereas the previous version made a much stricter check and required that this string occurs after `SMACK found an error:`. So the new version could lead to wrong results.

I merge this now such that you can get pre-run results, but please @zvonimir confirm whether this is ok or needs to be changed.

zvonimir:
Thanks! Yeah, I noticed the change in logic. Currently I can't think of a reason for why the new logic would not work. But maybe I am missing some corner case. Let me consult with my students as well.

PR #684, opened 2020-11-29
Scoring Schema for Results with Expected Verdict TRUE

dbeyer:
Until last year, 1 point was assigned to unconfirmed witnesses for correct results true.
A few years back, the community decided that from SV-COMP 2020 onwards, this point should not be assigned anymore. Points should be assigned only for results with confirmed witnesses. Results without witnesses are considered value-less.

### DIFF
--- a/benchexec/result.py
+++ b/benchexec/result.py
@@ -75,7 +75,7 @@ RESULT_CLASS_OTHER = "other"
 # Score values taken from http://sv-comp.sosy-lab.org/
 # (use values 0 to disable scores completely for a given property).
 _SCORE_CORRECT_TRUE = 2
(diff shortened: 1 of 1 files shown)


PhilippWendler:
For BenchExec, the background considerations in the SV-COMP community are not so relevant as source of such changes, we simply implement what is stated in the [SV-COMP rules](https://sv-comp.sosy-lab.org/2021/rules.php#scores). But those don't match yet, will be they be updated?

dbeyer:
Yes, I am working on correcting the web page.  In the community meeting, somebody reminded my of this, and the 2019 report and the slides of the presentation on 2020-04-06 are very clear on this. Last year, an exception was approved by the jury.

dbeyer:
Web pages are corrected now.

PR #691, opened 2020-12-02
Improve and update MetaVal tool-info module

MartinSpiessl:
- This makes the new python-implementation of MetaVal work
- It is backwards compatible, since we still call it via metaval.sh
- This should improve the ability to execute MetaVal from outside its
  tool directory
- We also fail early now if the user doesn't specify a property

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -8,9 +8,11 @@
 import argparse
 import benchexec.tools.template
 import re
+import pathlib
 import threading
 
 from benchexec.tools.template import BaseTool2
+from benchexec.tools.template import UnsupportedFeatureException
 
 
 class Tool(benchexec.tools.template.BaseTool2):
@@ -34,7 +36,11 @@ class Tool(benchexec.tools.template.BaseTool2):
         "ultimateautomizer": "UAutomizer-linux",
     }
     PATH_TO_TOOL_MAP = {v: k for k, v in TOOL_TO_PATH_MAP.items()}
-    REQUIRED_PATHS = list(TOOL_TO_PATH_MAP.values()) + ["VERSION.txt"]
+    REQUIRED_PATHS = list(TOOL_TO_PATH_MAP.values()) + [
+        "VERSION.txt",
+        "metaval.py",
+        "metaval.sh",
+    ]
 
     def __init__(self):
         self.lock = threading.Lock()
@@ -44,13 +50,13 @@ class Tool(benchexec.tools.template.BaseTool2):
         return toolLocator.find_executable("metaval.sh")
 
     def name(self):
-        return "metaval"
+        return "MetaVal"
 
     def determine_result(self, run):
         verifierDir = None
         regex = re.compile("verifier used in MetaVal is (.*)")
         for line in run.output[:20]:
-            match = regex.match(line)
+            match = regex.search(line)
             if match is not None:
                 verifierDir = match.group(1)
                 break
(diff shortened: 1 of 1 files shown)


MartinSpiessl:
@PhilippWendler could you have a look?

1. Is usage of `pathlib` OK or is there a better replacement in BenchExec?
There is `_program_files_from_executable` but I am not sure whether that one is allowed to be used and it also works on lists and filename patterns, neither is needed here.

2. Is changing the capitalization of the reported tool name a problem? Changing this is not a hard requirement, but it would be nicer.

3. MetaVal should now be executable from outside its tool directory (which I tested and verified). There are still some problems when this is done via a cloud-based executor though, but these problems are probably bugs in these executors and not in the tool-info module of MetaVal


I still need to test this against the current tool implementation and will remove the draft status afterwards.

PhilippWendler:
> 1. Is usage of `pathlib` OK or is there a better replacement in BenchExec?

We don't have anything better. However, BenchExec itself currently never uses `pathlib` and always uses strings for paths. The tool-info API is also documented to use strings. So if you now pass a `Path` to another tool-info module and maybe also end up returning `Path` instances to BenchExec, things might break? I am not sure, most path-handling methods of the Python API accept both `Path` and strings, but the `Path` instances might end up somewhere where only strings are allowed. It is ok for me if you want to do that, but if you want to avoid that risk, you could also convert the `Path` to a string before passing it on, or just use the `os.path` methods.

> There is `_program_files_from_executable` but I am not sure whether that one is allowed to be used and it also works on lists and filename patterns, neither is needed here.

It is certainly allowed to be used. But of course the need for unwrapping the resulting list makes it unwieldy.

> 2. Is changing the capitalization of the reported tool name a problem? Changing this is not a hard requirement, but it would be nicer.

The new result is actually the preferred form for this method. And changing is never a problem, this is only intended for humans.

> I still need to test this against the current tool implementation and will remove the draft status afterwards.

For me it looks good. Just tell me whether you want to change the path handling or keep it as it is.

MartinSpiessl:
> We don't have anything better. However, BenchExec itself currently never uses pathlib and always uses strings for paths. The tool-info API is also documented to use strings. So if you now pass a Path to another tool-info module and maybe also end up returning Path instances to BenchExec, things might break? I am not sure, most path-handling methods of the Python API accept both Path and strings, but the Path instances might end up somewhere where only strings are allowed. It is ok for me if you want to do that, but if you want to avoid that risk, you could also convert the Path to a string before passing it on, or just use the os.path methods.

These were exactly my considerations as well. I don't like the idea of having bugs because someone expects a string object and I suddenly pass `Path` objects arround. I will rewrite it to use `os.path` then, something like:
```
def _resource(self, executable, relpath):
        return os.path.join(os.path.dirname(executable), relpath)
```
 Generally I have to say I like `pathlib` much more than `os.path`, it just feels more high-level not having to manipulate strings all the time. Also we can call methods on those objects, which makes the code more concise.

PhilippWendler:
> Generally I have to say I like `pathlib` much more than `os.path`, it just feels more high-level not having to manipulate strings all the time. Also we can call methods on those objects, which makes the code more concise.

Definitely! But migrating a project like BenchExec is not so easy probably. I do have a medium-term plan of making sure that everything returned from tool-info modules is accepted whether it is a string or a `Path`, though, and then we can also give `Path` instances to tool-info modules in a later API version (maybe even still on `BaseTool2` via some opt-in mechanism).

PhilippWendler (approved):

MartinSpiessl:
> Definitely! But migrating a project like BenchExec is not so easy probably.

I totally agree on that! It is already quite good that it is handled consistently.

> I do have a medium-term plan of making sure that everything returned from tool-info modules is accepted whether it is a string or a Path, though, and then we can also give Path instances to tool-info modules in a later API version (maybe even still on BaseTool2 via some opt-in mechanism).

Good to know!

PR #695, opened 2020-12-04
Upgrade almost all JS libraries to their newest version

PhilippWendler:
The largest updates are React (version 17) and create-react-scripts (version 4). So far everything seems to work.

@lachnerm Could you please double check? And could you please check the release notes of those libraries where we use a new major version if there is something important mentioned?


lachnerm (approved):
Looks good! I didn't find any bugs either, everything seems to be working like it's supposed to be.

The only noteworthy change may be the [support for Fast Refresh by react-scripts](https://github.com/facebook/create-react-app/pull/8582). This enables the preservation of react state of functional components and hooks on reloads. Since we mostly use classes in our project, we probably won't notice that much of it, but especially if we switch to using hooks sometime in the future, this might be quite handy. There are no changes in functionality and it's only relevant when developing though.
Apart from this, I couldn't really find anything interesting. Even `React v17` itself [doesn't introduce any developer-facing features](https://reactjs.org/blog/2020/10/20/react-v17.html), but many under-the-hood changes. It's main focus lies on enabling gradual React upgrades, which may be useful in some cases, but doesn't really provide any immediate merits for us I think.

PR #696, opened 2020-12-04
[draft] contrib/mergeBenchmarkSets.py: take correct + invalid witness file as correct-unconfirmed

mchalupa:
This PR is meant to be mainly a remainder for the next SV-COMP, do not merge until the discussion is resolved.

Take correct answer + invalid witness false as correct-unconfirmed instead of error. This PR is related to the discussion in
https://gitlab.com/sosy-lab/sv-comp/archives-2021/-/issues/23#note_460576708 .

### DIFF
--- a/contrib/mergeBenchmarkSets.py
+++ b/contrib/mergeBenchmarkSets.py
@@ -69,12 +69,7 @@ def getWitnessResult(witness, verification_result):
     # then leave status and category as is.
     if status_from_validation == status_from_verification:
         return status_from_verification, category_from_verification
-    # An invalid witness counts as error of the verifier.
-    if status_from_validation == "ERROR (invalid witness syntax)":
-        return (
-            "witness invalid (" + status_from_verification + ")",
-            result.CATEGORY_ERROR,
-        )
+
     # Other unconfirmed witnesses count as CATEGORY_CORRECT_UNCONFIRMED.
     if category_from_verification == result.CATEGORY_CORRECT:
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Is this still relevant? Should the conflicts be resolved and the PR marked as ready for review?

PhilippWendler:
@dbeyer ping What is your opinion on this?

PhilippWendler:
@dbeyer ping, this seems to be the last chance to get it done for this year's competitions? If you want it merged, would be good to have it in the 3.10 tag of BenchExec.

dbeyer:
After a long discussion, we found out that the distinction between `unconfirmed` and `error` was a wanted design choice.
https://gitlab.com/sosy-lab/sv-comp/archives-2021/-/issues/23#note_460680428

"
- `error`: If a witness is invalid syntactically, then this is clearly an error of the tool, similar to time out or out of memory.
- `unconfirmed`: If a witness is valid and also correctly points out an error path, then it could still be unconfirmed, due to a weakness in the validators. This is addresses by allowing many validators, and taking the any rule (one validator confirms => result is considered confirmed).
"
A syntax error in the witness is a definitive problem of the tool, and there is no hope that a better result validator can confirm it.

An unconfirmed witness is a witness that is syntactically correct, but no validator could reconstruct the proof based on information from that witness. The conclusion that either the error path or invariants in the witness are wrong, or the validators did not understand the reasoning in the witness.

dbeyer:
In sum, this should not be changed.

Issue #698, opened 2020-12-08
required paths in tool info module


skanav:
[REQUIRED_PATHS](https://github.com/sosy-lab/benchexec/blob/33029b5e38f18a5d13129d4969f8e30c219a2e18/benchexec/tools/template.py#L78)  is used in a tool info module to specify the paths that are necessary to run the tool. 
This variable is then used to compute the list of files to be transported to the cloud.

The default value of `REQUIRED_PATHS` is an `empty list`. This means that `benchexec` will only the `executable` is transferred if this variable is not explicitly defined in the tool info module of the tool.

I would propose to change the default behavior of `benchexec` to transfer the complete tool directory. This would also simplify tool info modules.

Now, `benchexec` does not know about the tool directory.  So we could do one of the following:
1. use `cwd` or `tool_directory` if explicitly specified
2. use the computed  `base_dir` of the `executable` in the function [`_program_files_from_executable`](https://github.com/sosy-lab/benchexec/blob/33029b5e38f18a5d13129d4969f8e30c219a2e18/benchexec/tools/template.py#L234)

PhilippWendler:
What is the `base_dir` of an executable?

skanav:
> What is the `base_dir` of an executable?

updated the comment

MartinSpiessl:
> I would propose to change the default behavior of benchexec to transfer the complete tool directory. This would also simplify tool info modules.

This can backfire in case someone has a symlink to sv-benchmarks in the tool directory. Been there, done that, cannot recommend! Our cloud system fortunately complained and refused to accept such a large number of files (136k currently in sv-`benchmarks/c`).

> 1. use cwd or tool_directory if explicitly specified
> 2. use the computed base_dir of the executable in the function _program_files_from_executable

Same problem. Greedily declaring files as required might also be a security issue.

I personally don't want all my private test programs to end up in some cloud service just because a tool was lazy in defining what its required files are.

What about instead of changing the default behavior we add a flag that can be set by the user (at their own risk) and that will do what you want?

PhilippWendler:
The `base_dir` you are referring to is simply the directory of the executable, unless the authors of the tool-info module specifically override `program_files`. However, this just overcomplicates things, instead of overriding `program_files`, it is much easier to simply set `REQUIRED_PATHS` accordingly. Your proposal 2 should be 'use `REQUIRED_PATHS = ["."]` as default'.

In general, I see the following problems:
[...]

PR #699, opened 2020-12-11
Fix Symbiotic's tool-info module: Deploy all components of the archive

dbeyer:
This PR fixes issue #698 for Symbiotic in order to successfully participate in SV-COMP 2021.

### DIFF
--- a/benchexec/tools/symbiotic4.py
+++ b/benchexec/tools/symbiotic4.py
@@ -15,15 +15,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     Symbiotic tool info object
     """
 
-    REQUIRED_PATHS = [
-        "bin",
-        "include",
-        "share",
-        "instrumentations",
-        "lib",
-        "lib32",
-        "symbiotic",
-    ]
+    REQUIRED_PATHS = ["."]
 
     def executable(self, tool_locator):
         """
--- a/benchexec/tools/symbiotic.py
+++ b/benchexec/tools/symbiotic.py
@@ -17,12 +17,6 @@ class Tool(OldSymbiotic):
     Symbiotic tool info object
     """
 
-    REQUIRED_PATHS_4_0_1 = ["bin", "include", "lib", "lib32", "llvm-3.8.1"]
-
-    REQUIRED_PATHS_5_0_0 = ["bin", "include", "lib", "lib32", "llvm-3.9.1"]
-
-    REQUIRED_PATHS_NEW = ["bin", "include", "properties", "lib", "llvm-*"]
-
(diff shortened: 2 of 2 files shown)


dbeyer:
@PhilippWendler Unfortunately, I need a new tag for this version and re-run Symbiotic in SV-COMP 2021 with the new version. Perhaps 3.5.1 as a tag for the branch?

benchexec/tools/symbiotic4.py
```diff
@@ -15,15 +15,7 @@ class Tool(benchexec.tools.template.BaseTool2):
...
-    REQUIRED_PATHS = [
-        "bin",
-        "include",
-        "share",
-        "instrumentations",
-        "lib",
-        "lib32",
-        "symbiotic",
-    ]
+    REQUIRED_PATHS = ["."]
```
mchalupa, line 18: By the way, this could be a good default value of this attribute.

mchalupa (approved):
The change looks good, I tested with `benchexec.test_tool_info` and everything seems alright.

benchexec/tools/symbiotic4.py
mchalupa: Oh, that is exactly what is in the issue #698 :)

PhilippWendler:
Ok, I will merge this and create release 3.6.

In the future, please do not start branches at exactly the tag revision, because then we have more than one commit in the repository where `benchexec --version` produces a number like `3.5`. I fixed this now by rebasing the branch on top of 473c0eaf904a4fecc16d36841dfd04daa2d2bb03.

dbeyer:
> Ok, I will merge this and create release 3.6.

Thank you!

> rebasing the branch on top of [473c0ea](https://github.com/sosy-lab/benchexec/commit/473c0eaf904a4fecc16d36841dfd04daa2d2bb03).

But what if I do not want all those other commits? On the branch I was sure that I changed only the Symbiotic file and nothing else. Now I have all those other changes. I think they are not touching the code that I had executed for the competition, but what if there is a strange interference and then the results are not reproducible?  Shouldn't a bug-fix release be an end of a (not merged) branch, while the commits leading there are merged into the master later on (by cherry-picking?), and I can still take the branch with as few commits as possible?

PhilippWendler:
Misunderstanding: I did not rebase the branch on top of yesterday's master, I rebased it on top of the linked commit, which is exactly 1 commit after the tag 3.5 and which contains only a changed version number. So the bug fix release contains exactly what it should contain.

dbeyer:
Indeed!  Very nice.  Thanks a lot!

PR #700, opened 2020-12-14
Module for delegating benchmarking to VerifierCloud

skanav:
This adds a script to `contrib/` that allows delegating the actual benchmarking to a cloud service, just like `contrib/aws-benchmark.py` does. The cloud service in this case is the [VerifierCloud](https://gitlab.com/sosy-lab/software/verifiercloud/) (internal repo) that is used for example by CPAchecker developers, SV-COMP, and Test-Comp. So it is easier to maintain this code in the BenchExec repo.

The source of this code is in the [CPAchecker repo](https://gitlab.com/sosy-lab/software/cpachecker/-/tree/74cb6d1c1e9d9449f1215efd70f5f4af4ec3f269/scripts) (`benchmark.py` and `benchmark` directory but without the code for the WebClient mode of the VerifierCloud).

---
### Original Description

As discussed, a PR adding the `benchmark.py` to the benchexec repo.

There are two deviations from what we discussed:
1) I am still using `ant`instead of directly using `ìvy`.
2) I have split the `benchmark.py` 

The reason for the deviations: I think it makes it easier to maintain the code duplication as these files would also be present in the `CPAchecker` directory.


contrib/vcloud/util.py
```diff
@@ -0,0 +1,90 @@
...
+            or key.startswith("cputime-cpu")
+        ):
+            result_values[key] = value
+        elif key not in ["command", "timeLimit", "coreLimit", "memoryLimit"]:
+            result_values["vcloud-" + key] = value
+
+    return result_values
+
+
+def parse_frequency_value(s):
```
PhilippWendler, line 60: This is duplicated from `benchexec/utils.py`. Probably it is good to check what else is duplicated in the `util` module.

contrib/vcloud/util.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import collections
+import os
+import sys
+
+import benchexec.util
```
PhilippWendler, line 12: This highlights that we have now another module named `util`, which can be confusing (unfortunately, there are already two such modules in BenchExec itself). So if this module is actually needed, it should be renamed.

contrib/vcloud/build.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PhilippWendler, line 1: This is quite a lot of stuff just to basically make a single call to Ivy, which should be doable with one command, right? Furthermore, it adds the dependency to `ant` to something that would otherwise require only Python and Java. Can't we get rid of this?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,449 @@
...
+
+DEFAULT_CLOUD_TIMELIMIT = 300  # s
+
+DEFAULT_CLOUD_MEMORY_REQUIREMENT = 7000000000  # 7 GB
+DEFAULT_CLOUD_CPUCORE_REQUIREMENT = 2  # one core with hyperthreading
+DEFAULT_CLOUD_CPUMODEL_REQUIREMENT = ""  # empty string matches every model
+
+STOPPED_BY_INTERRUPT = False
+
+# Assuming that benchmark.py is executed from the contrib folder.
```
PhilippWendler, line 29: We should not make such assumptions on the current working directory.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,449 @@
...
+_ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__)))
+
+# Check if benchmark.py is executed from the contrib folder,
+# otherwise it is being used from the scripts folder in CPAchecker.
+if not os.path.isfile(os.path.join(_ROOT_DIR, "build.xml")):
+    _ROOT_DIR = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), os.pardir, os.pardir)
+    )
+
+# Keeping the variable name as before but adding a noqa
```
PhilippWendler, line 39: I don't understand this comment?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,449 @@
...
+DEFAULT_CLOUD_CPUCORE_REQUIREMENT = 2  # one core with hyperthreading
+DEFAULT_CLOUD_CPUMODEL_REQUIREMENT = ""  # empty string matches every model
+
+STOPPED_BY_INTERRUPT = False
+
+# Assuming that benchmark.py is executed from the contrib folder.
+_ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__)))
+
+# Check if benchmark.py is executed from the contrib folder,
+# otherwise it is being used from the scripts folder in CPAchecker.
```
PhilippWendler, line 33: Isn't the comment referring to CPAchecker outdated?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,449 @@
...
+                    "revision": benchmark.tool_version,
+                    "benchexec-module": benchmark.tool_module,
+                },
+                "benchmark": benchmark.name,
+                "timestamp": benchmark.instance,
+                "generator": "benchmark.vcloud.py",
+            }
+        )
+
+        if not os.getenv("TAKE_LOCAL_VCLOUD_JAR"):
```
PhilippWendler, line 82: Having an environment variable like this is quite an obscure UI. Is it really necessary or can we detect somehow whether to use Ivy? If not, I would suggest a parameter like `--vcloudJar` that users could use for passing the path to the JAR, and if it is not set, the script downloads it using Ivy.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,449 @@
...
+                    output_handler.output_after_run(run)
+            else:
+                logging.warning("No results exist for file %s.", run.identifier)
+                output_handler.set_error("missing results", runSet)
+                executedAllRuns = False
+
+            if os.path.exists(run.log_file + ".stdError"):
+                runsProducedErrorOutput = True
+
+            # The directory structure differs between direct and webclient mode when using VCloud.
```
PhilippWendler, line 361: This comment also seems outdated, there should not need to be a reference to the WebClient in this script. Instead, it should make the result look like the result from plain BenchExec.

PhilippWendler (changes requested):
Somehow, this does not print the usual `In order to get HTML and CSV tables, run...` message at the end?

Also, the output files are produced in `test/results/`, which does not make sense for a script outside of the CPAchecker repo. They should be in `results/`.

And the directories for the JAR should be in `.gitignore`.

contrib/benchmark.py
```diff
@@ -0,0 +1,50 @@
...
+] + benchexec.tools.__path__
+
+
+class Benchmark(BenchmarkBase):
+    """
+    Benchmark class that defines the load_executor function.
+    """
+
+    def load_executor(self):
+        if self.config.cloud:
```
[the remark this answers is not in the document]
skanav: I thought about it, but then kept it. 

I you want it to be removed, I can remove it. I have no strong opinions on it.

contrib/vcloud/util.py
skanav: This does not raises an exception. I wonder if that was the reason to maintain two copies.

contrib/vcloud/util.py
skanav: renamed.

contrib/vcloud/benchmarkclient_executor.py
skanav: I think I have mentioned it in the PR comment that I have tried to keep stuff as it is because it makes the maintenance of  code duplication simple.

But if you prefer to keep two slightly different files for the `benchmarkclient_executor` then so be it.

PhilippWendler: No, of course as many files as possible should be identical.

But still, there is no reason to adjust something to the way the webclient mode does something. It should be adjusted to the way BenchExec does things.

contrib/vcloud/util.py
PhilippWendler: Hm, both methods contain `raise ValueError`? If there are differences in behavior, we should still deduplicate the common parts.

skanav: You have more knowledge on this than me. I don't know why was the code duplicated originally.
If you want I can remove the call and use the function from the `benchexec.util`

PhilippWendler: I just noticed that the copy in `vcloudutil.py` uses `KHz` instead of `kHz` (which is a wrong unit), and I am not sure whether the unit handling is otherwise identical. So we would need to check which format and unit the VerifierCloud actually produces and then we can decide whether the copy in BenchExec correctly parses this.

contrib/vcloud/benchmarkclient_executor.py
skanav: I have removed it for the moment.
I would prefer to go ahead with this PR first and then add this feature when the need arise.

PhilippWendler: Please do not check in JAR files. To not needlessly increase the history size I would prefer if the last commit could be edited to not contain the JAR file and then be force-pushed.

PhilippWendler: Futhermore, the code that you said is removed is still there?

We should also think about how to make this work in way that allows easy sharing of this file with the CPAchecker repository. I would suggest to move the code that is responsible for retrieving the JAR file to `vcloud-benchmark.py` and just expect a path to the JAR to be passed to `benchmarkclient_executor.py` somehow. Then `vcloud-benchmark.py` could retrieve the JAR via Ivy or use a user-supplied one, and CPAchecker's `benchmark.py` could use the existing `ant`-based setup.

contrib/vcloud/benchmarkclient_executor.py
skanav: updated the comment

contrib/vcloud-benchmark.py
```diff
@@ -0,0 +1,43 @@
...
+sys.dont_write_bytecode = True  # prevent creation of .pyc files
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from vcloud.benchmarkbase import BenchmarkBase  # noqa E402
+from benchexec import __version__  # noqa E402
+import benchexec.benchexec  # noqa E402
+import benchexec.tools  # noqa E402
+
+
+class Benchmark(BenchmarkBase):
```
PhilippWendler, line 23: I guess we should use more descriptive names for both classes. There exists already a class `Benchmark` in BenchExec, for example.

How about something with `Vcloud` in the name? Maybe we can also unify both classes given that one is really short (if keeping them separate is not required for making the integration with CPAchecker's scripts easier).

skanav: I have tried to keep it such that one needs to only copy the folder `vcloud` to the `CPAchecker` repo.
This file is not required there.

On the other hand, we can also have only one file that would pass different parameters to the main.
I have no strong opinions.

PhilippWendler: > I have tried to keep it such that one needs to only copy the folder `vcloud` to the `CPAchecker` repo.

Yes, this is something that we should aim for and other simplifications should be done only if they do not interfere with this goal.

contrib/vcloud/benchmarkclient_executor.py
skanav: > Please do not check in JAR files. To not needlessly increase the history size I would prefer if the last commit could be edited to not contain the JAR file and then be force-pushed.

Which jar files? We would need `ivy.jar`.

skanav: > Futhermore, the code that you said is removed is still there?
It is there now. I had forgotten to commit.

> We should also think about how to make this work in way that allows easy sharing of this file with the CPAchecker repository. I would suggest to move the code that is responsible for retrieving the JAR file to `vcloud-benchmark.py` and just expect a path to the JAR to be passed to `benchmarkclient_executor.py` somehow. Then `vcloud-benchmark.py` could retrieve the JAR via Ivy or use a user-supplied one, and CPAchecker's `benchmark.py` could use the existing `ant`-based setup.

Why not just use the same thing for both, i.e., remove the `resolve-benchmark-dependencies` from CPAchecker's `build.xml`?

contrib/vcloud-benchmark.py
skanav: > How about something with `Vcloud` in the name? Maybe we can also unify both classes given that one is really short (if keeping them separate is not required for making the integration with CPAchecker's scripts easier).

done

contrib/vcloud/vcloudbenchmarkbase.py
```diff
@@ -0,0 +1,128 @@
...
+import benchexec.util
+
+
+class VcloudBenchmarkBase(benchexec.benchexec.BenchExec):
+    """
+    An extension of BenchExec that supports executing the benchmarks in the
+    VerifierCloud (internal project at https://gitlab.com/sosy-lab/software/verifiercloud/).
+    """
+
+    DEFAULT_OUTPUT_PATH = "results/"
```
PhilippWendler, line 21: I still don't understand why this is necessary?

Did you test this and it does not work without?

`benchexec.benchexec.BenchExec` already declares this.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: > > Please do not check in JAR files. To not needlessly increase the history size I would prefer if the last commit could be edited to not contain the JAR file and then be force-pushed.
> 
> Which jar files? We would need `ivy.jar`.

We should just download it like the `ant` file did (whether we use `ant` or not does not change anything regarding the arguments of having it checked in or not).

[`urllib.request.urlretrieve`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve) should be usable here, it is the shortest way I know to download a file in Python.

PhilippWendler: > Why not just use the same thing for both, i.e., remove the `resolve-benchmark-dependencies` from CPAchecker's `build.xml`?

Hmm, interesting idea. It would have the advantage of more clearly separating the benchmarking dependencies from CPAchecker's actual dependencies (by removing the former from `ivy.xml`). On the other hand there would be two `ivy.jar` lying around, and JAR files would probably be lying somewhere below `scripts` instead of `lib`. But we can try out and see how it looks?

contrib/vcloud/vcloudbenchmarkbase.py
skanav: sorry, this was my mistake. fixed now

contrib/vcloud/benchmarkclient_executor.py
skanav: >urllib.request.urlretrieve should be usable here, it is the shortest way I know to download a file in Python.

from here: https://www.sosy-lab.org/ivy/org.apache.ivy/ivy/ivy-2.5.0.jar ?

PhilippWendler: Yes.

Thinking about it, it would be good to use the name `ivy-2.5.0.jar` for the local file. Then we can skip retrieving the file if it already exists, and still will be able to update the used Ivy version by changing the file name. If we use `ivy.jar` we would either have to download every time or there would be update problems.

skanav: done

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def get_system_info():
+    return None
+
+
+def download_required_jars():
+    # download ivy if needed
+    ivy_path = os.path.join(_ROOT_DIR, "lib", IVY_JAR_NAME)
+    ivy_download_url = "https://www.sosy-lab.org/ivy/org.apache.ivy/ivy/" + IVY_JAR_NAME
```
PhilippWendler, line 57: nitpick: make a module-level constant out of this?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,462 @@
...
+
+    # prepare command
+    cmd = ["java", "-Divy.default.resolver=Sosy-Lab", "-jar", "lib/" + IVY_JAR_NAME]
+    cmd += ["-settings", "lib/ivysettings.xml"]
+    cmd += ["-dependency", "org.sosy_lab", "vcloud", "0.+"]
+    cmd += ["-confs", "runtime", "-mode", "dynamic", "-refresh"]
+    cmd += ["-retrieve", "lib/vcloud-jars/[artifact](-[classifier]).[ext]"]
+
+    # install cloud and dependencies
+    ant = subprocess.Popen(
```
PhilippWendler, line 70: nitpick: not really related to ant anymore :-)

We could also simply use `subprocess.run()` instead of `communicate()` and `wait()`.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -0,0 +1,462 @@
...
+def download_required_jars():
+    # download ivy if needed
+    ivy_path = os.path.join(_ROOT_DIR, "lib", IVY_JAR_NAME)
+    ivy_download_url = "https://www.sosy-lab.org/ivy/org.apache.ivy/ivy/" + IVY_JAR_NAME
+    if not os.path.isfile(ivy_path):
+        # let the process exit if an exception occurs.
+        urllib.request.urlretrieve(ivy_download_url, ivy_path)  # noqa S310
+
+    # prepare command
+    cmd = ["java", "-Divy.default.resolver=Sosy-Lab", "-jar", "lib/" + IVY_JAR_NAME]
```
PhilippWendler, line 63: Is the default resolver necessary both here and in `ivysettings.xml`? I guess one can go away (preferably this one)?

contrib/vcloud/lib/ivysettings.xml
```diff
@@ -0,0 +1,29 @@
...
+https://gitlab.com/sosy-lab/software/java-project-template
+
+SPDX-FileCopyrightText: 2018-2020 Dirk Beyer <https://www.sosy-lab.org>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<!DOCTYPE ivysettings> <!-- generic AcceptAllDTD to make the Eclipse Warning go away -->
+<ivysettings>
+    <!-- DO NOT EDIT LOCALLY!
```
PhilippWendler, line 15: This is not longer true, so I would suggest to do more cleanup in this file:
- update license header
- remove this comment
- remove unused properties

contrib/vcloud/vcloudbenchmarkbase.py
```diff
@@ -0,0 +1,126 @@
...
+            "--vcloudCPUModel",
+            dest="cpu_model",
+            type=str,
+            default=None,
+            metavar="CPU_MODEL",
+            help="Only execute runs in the VerifierCloud on CPU models that contain the given string.",
+        )
+
+        vcloud_args.add_argument(
+            "--vcloudUser",
```
PhilippWendler, line 55: This parameter is not used here, and the same is true for several others, please remove them.

PhilippWendler (commented):
Please also remember the comments from my other review that were not yet addressed.

And please cleanup the git history such that `ivy.jar` is gone (I would suggest to squash 10f08ddf75abe2bde3800dd8d544a5324125ffed, 197ba15f6f3aad9364b5a30e27db039723a35e38, and d0a965773e9cf704758af0185d263da120b9f440 into 5ad9c941052b64ff6869b2a6f6ca54cfd3a27042, that should apply cleanly, and no other changes.)

contrib/vcloud/vcloudbenchmarkbase.py
skanav: I would need some help here. I actually don't know which other parameters are not used here.

contrib/vcloud/util.py
skanav: I am also not sure. I also think that I am not the right person to address this.

contrib/vcloud/vcloudbenchmarkbase.py
PhilippWendler: They are explicitly documented as such.

skanav: done

contrib/vcloud/vcloudbenchmarkbase.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    An extension of BenchExec that supports executing the benchmarks in the
+    VerifierCloud (internal project at https://gitlab.com/sosy-lab/software/verifiercloud/).
+    """
+
+    def create_argument_parser(self):
+        parser = super(VcloudBenchmarkBase, self).create_argument_parser()
+        vcloud_args = parser.add_argument_group("Options for using VerifierCloud")
+        vcloud_args.add_argument(
+            "--vcloud",
```
PhilippWendler, line 25: Wasn't this parameter removed?

skanav: that's what I remember too. Anyways, removed it.

contrib/vcloud/util.py
skanav: @PhilippWendler actually you might be better positioned to answer this.
This function was introduced by you in [this commit](https://gitlab.com/sosy-lab/software/cpachecker/-/commit/508661e56ec6f385dd64a339abec229e4a9891b1)
and in benchexec utils it was introduced by @TBunk in 494e6009b006c9ccaaf49cef5c364e10692ae8c3

As I said before, I am happy to call whatever of the two is better. But I don't think that I should be making this decision.

PhilippWendler: Well, I also have to check what the VerifierCloud actually outputs and compare that against the two implementations of this method, it is not easier for me.

But I did so know, found a bug in the method (6750c9663febde492761d0c4111dfefb3bc0c503), and found out why the standard BenchExec method for this cannot be used (f71a4ca03b8c7c63816bab57a71f9d28a7139e2d).

PhilippWendler (approved):
So, there were several more cleanups necessary, but I think now this is mergeable from BenchExec's point of view.

What is the further plan regarding the integration with the code in the CPAchecker repo? Did you already test that the modules that were changed for this PR still work as expected when copied over to the CPAchecker repo? And does the integration of this code together with CPAchecker's `benchmark.py` work or will some changes be necessary for the code here as well?

skanav:
> What is the further plan regarding the integration with the code in the CPAchecker repo?

No idea. I limited my concern to this PR. 

>  Did you already test that the modules that were changed for this PR still work as expected when copied over to the CPAchecker repo? 

No

>And does the integration of this code together with CPAchecker's `benchmark.py` work or will some changes be necessary for the code here as well?

I expect that it would require some changes.

PhilippWendler:
So what is the suggested next step?

skanav:
One of us should port the changes to the CPAchecker repo. I can do it if you prefer so , but I wouldn't mind you doing it either.

Also, I don't think I am the right person to test these changes in the CPAchecker repo. Till now I have tested the changes in this PR wrt CoVeriTeam service.

PhilippWendler:
Ok, feel free to do it as I will be on holiday for the next weeks.

PR #701, opened 2020-12-14
removed required path from various tools

skanav:
PR to remove the variable `REQUIRED_PATH` from various tool info modules.

### DIFF
--- a/benchexec/tools/brick.py
+++ b/benchexec/tools/brick.py
@@ -16,8 +16,6 @@ class Tool(benchexec.tools.template.BaseTool2):
     https://github.com/brick-tool-dev/brick-tool
     """
 
-    REQUIRED_PATHS = ["bin", "lib"]
-
     def executable(self, tool_locator):
         return tool_locator.find_executable("brick", subdir="bin")
 
--- a/benchexec/tools/cbmc.py
+++ b/benchexec/tools/cbmc.py
@@ -21,8 +21,6 @@ class Tool(benchexec.tools.template.BaseTool2):
     SV-COMP conditions is assumed.
     """
 
-    REQUIRED_PATHS = ["cbmc", "cbmc-binary", "goto-cc"]
-
     def executable(self, tool_locator):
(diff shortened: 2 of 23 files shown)


PhilippWendler:
The response in #698 was quite negative (in favor of keeping `REQUIRED_PATHS`), so why remove this now?

dbeyer:
But the PR does not touch tool-info modules of users that use the `REQUIRED_PATHS` (cf. #698).

The `REQUIRED_PATHS` that are touched in this PR are not tested and most likely contain omissions, as we witnessed incidentally after many weeks (years?) for Symbiotic in #699.

PhilippWendler:
@lembergerth @MartinSpiessl @gernst Could you please check whether this PR removes `REQUIRED_PATHS` from any tool-info module where you find it useful for you? Please tell us, then we would keep these and remove only the remaining.

lembergerth (approved):
Looks good to me. I sporadically use cbmc and klee in the VCloud, but I can put the result files in a parent directory.

gernst:
Looks good to me, too, my tools are unaffected. Do I understand correctly: the default is `REQUIRED_PATHS=["."]`?

PhilippWendler:
No, the default is just the `executable`. So at least it would not match lots of useless stuff.

lembergerth:
@gernst The workaround would be to use `benchmark.py` with something like the following: `--cloudAdditionalFiles lib/ bin/ src/`

gernst:
Uh, what's the use case in which the default actually works? Almost all tools do have more than one required file. Just curious - feel free to ignore me.

PhilippWendler:
Yes, but there is no better default. As discussed in #698, `.` is not a good default and would still fail for many tools (those with the executable in a subdirectory, for example).

gernst (approved):

MartinSpiessl (approved):
Looks good to me.
None of "my" tools I use is affected.

PR #703, opened 2020-12-16
Determine result for coveriteam-verifier-validator more precisely

lembergerth:
If multiple verdicts are printed (happens for intermediate steps with option --debug)
take the last verdict printed.
If there's an exception, return EXCEPTION-status instead. And if there's no verdict,
do not assume error, but assume UNKNOWN.

Also make parsing of the verdict-result more robust:
Don't parse the string printed by CoVeriTeam into a dict, but use a regular expression to get the verdict.
Parsing of the string into a dict does not work if any output artifact of CoVeriTeam is None, which results in CoVeriTeam printing something like the following:
`{'verdict': 'true', 'witness': }`. This can't be parsed into a dict then.

### DIFF
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -52,7 +52,7 @@ class Tool(coveriteam.Tool):
         If more than one dict is printed, the first matching one.
         """
         verdict = None
-        verdict_regex = re.compile(r"'verdict': '([a-zA-Z\(\)\ ]*)'")
+        verdict_regex = re.compile(r"'verdict': '([a-zA-Z\(\)\ \-]*)'")
         for line in reversed(run.output):
             line = line.strip()
             verdict_match = verdict_regex.search(line)
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -59,7 +59,7 @@ class Tool(coveriteam.Tool):
             if verdict_match and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
                 verdict = verdict_match.group(1).lower()
-            if 'Traceback (most recent call last)' in line:
+            if "Traceback (most recent call last)" in line:
                 verdict = "EXCEPTION"
         if verdict is None:
             return result.RESULT_UNKNOWN
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -52,15 +52,13 @@ class Tool(coveriteam.Tool):
         If more than one dict is printed, the first matching one.
         """
         verdict = None
+        verdict_regex = re.compile(r"'verdict': '([a-zA-Z\(\)\ ]*)'")
         for line in reversed(run.output):
             line = line.strip()
-            if "'verdict':" in line and verdict is None:
+            verdict_match = verdict_regex.search(line)
+            if verdict_match and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
-                try:
-                    verdict = re.search(r"'verdict': '([a-zA-Z\(\)\ ]*)'", line).group(1).lower()
-                except AttributeError:
-                    # re.search didn't find anything, so this is not a valid verdict line
-                    pass
+                verdict = verdict_match.group(1).lower()
             if 'Traceback (most recent call last)' in line:
                 verdict = "EXCEPTION"
         if verdict is None:
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -10,6 +10,7 @@ import benchexec.result as result
 from benchexec.tools.template import UnsupportedFeatureException
 from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import ast
+import re
 
 
 class Tool(coveriteam.Tool):
@@ -50,14 +51,17 @@ class Tool(coveriteam.Tool):
         will print out the produced aftifacts.
         If more than one dict is printed, the first matching one.
         """
-        for line in run.output:
+        verdict = None
+        for line in reversed(run.output):
             line = line.strip()
-            if "verdict" in line:
+            if "'verdict':" in line and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
                 try:
-                    d = ast.literal_eval(line)
-                    if isinstance(d, dict):
-                        return d.get("verdict", result.RESULT_ERROR)
-                except SyntaxError:
+                    verdict = re.search(r"'verdict': '([a-zA-Z\(\)\ ]*)'", line).group(1).lower()
+                except:
                     pass
-        return result.RESULT_ERROR
+            if 'Traceback (most recent call last)' in line:
+                verdict = "EXCEPTION"
+        if verdict is None:
+            return result.RESULT_UNKNOWN
+        return verdict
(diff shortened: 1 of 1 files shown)


benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,17 @@ def determine_result(self, run):
...
             line = line.strip()
-            if "verdict" in line:
+            if "'verdict':" in line and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
                 try:
-                    d = ast.literal_eval(line)
-                    if isinstance(d, dict):
-                        return d.get("verdict", result.RESULT_ERROR)
-                except SyntaxError:
+                    verdict = re.search(r"'verdict': '([a-zA-Z\(\)\ ]*)'", line).group(1).lower()
```
PhilippWendler, line 60: I guess it would be slightly less error-prone to avoid the duplication of `'verdict'` and just search all lines with the regexp, taking the group content it there is a match?

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,17 @@ def determine_result(self, run):
...
+                except:
```
PhilippWendler, line 61: What kind of exceptions are expected here?

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,18 @@ def determine_result(self, run):
...
-                    if isinstance(d, dict):
-                        return d.get("verdict", result.RESULT_ERROR)
-                except SyntaxError:
+                    verdict = re.search(r"'verdict': '([a-zA-Z\(\)\ ]*)'", line).group(1).lower()
+                except AttributeError:
+                    # re.search didn't find anything, so this is not a valid verdict line
                     pass
-        return result.RESULT_ERROR
+            if 'Traceback (most recent call last)' in line:
+                verdict = "EXCEPTION"
```
skanav, line 65: I guess you want to keep it separate than the verdict `error`?

benchexec/tools/coveriteam-verifier-validator.py
lembergerth: Oh man, sorry, that was very error-prone.
I added the expected `AttributeError` from the `.group(1)` call in 0eb8a6a for documentation purposes, but it is gone in the latest version.

benchexec/tools/coveriteam-verifier-validator.py
lembergerth: Good idea, thanks!
I had performance issues in mind, but the regex is probably not that much slower than the 'in' check.

Does the current code match your proposal?

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,16 @@ def determine_result(self, run):
...
+            if verdict_match and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
-                try:
-                    d = ast.literal_eval(line)
-                    if isinstance(d, dict):
-                        return d.get("verdict", result.RESULT_ERROR)
-                except SyntaxError:
-                    pass
-        return result.RESULT_ERROR
+                verdict = verdict_match.group(1).lower()
```
skanav, line 61: would returning the verdict at this point be better? or do you see a case where an exception is logged but still a verdict is produced later, and we are interested in the exception instead of the verdict?

lembergerth: Good point, that's exactly the case!
If `--debug` is used, the debug log may still print the output artifacts, which includes an (invalid) verdict.
Since we go through the list from bottom to top, this may be falsely identified as the final verdict if we don't continue.

But you got me thinking, maybe we should ignore lines that start with DEBUG, even if they match the regular expression for a verdict?
I'll try to find an output log that matches my above description.

benchexec/tools/coveriteam-verifier-validator.py
PhilippWendler: Yes, thanks. The performance of a regexp is certainly acceptable, and my understanding of a decent regexp engine is that cases like this where there is a fixed prefix (`'verdict': '`) of a match are just as fast as a simple search for that string (though even if it weren't, I would consider this unnecessary micro-optimization).

But if you wish: an obvious performance improvement would be to move the `verdict_regex.search(line)` call into an `if verdict is None:` condition. (In modern Python we could use the walrus operator, but here we need a nested if.)

lembergerth: Yeah, I thought about it. But I would prefer to keep the nesting-level flat and don't worry about performance, if you agree.
I'm looking forward to being able to use the walrus operator, though.

benchexec/tools/coveriteam-verifier-validator.py
lembergerth: Sorry for the long wait. I couldn't create a log that creates an exception and prints a verdict afterwards, but I guess the current behavior would still be more robust in case such a case ever occurs. Are you ok with that?

*Just for completeness:*

We also can't ignore DEBUG lines, unfortunately, because CoVeriTeam logs messages with newlines.

```
2021-01-04 14:40:45,709 DEBUG Actor cpachecker-predicate-no-cegar acted on artifacts:
 {'program': '../../sv-benchmarks/c/seq-mthreaded/rekcba_ctm.3-2.c', 'spec': '../../sv-benchmarks/c/properties/unreach-call.prp', 'predicates': ''} 
and produced artifacts:
 {'verdict': 'false(unreach-call)', 'witness': '', 'predicates': 'cvt-output/4cfccc45-4f98-4da1-8ace-481f9f79fc87/cpachecker-predicate-no-cegar/c19d57ab-f16f-4416-a6e5-9ff67e55e128/output/predmap.txt'}
```

So that idea didn't help.

skanav (approved):

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,16 @@ def determine_result(self, run):
         will print out the produced aftifacts.
         If more than one dict is printed, the first matching one.
         """
-        for line in run.output:
+        verdict = None
+        verdict_regex = re.compile(r"'verdict': '([a-zA-Z\(\)\ ]*)'")
```
skanav, line 55:
```suggestion
        verdict_regex = re.compile(r"'verdict': '([a-zA-Z\(\)\ \-]*)'")
```

skanav: otherwise it is not parsing `false(unreach-call)` correctly.

lembergerth:
Thanks for the reviews, @PhilippWendler and @skanav!

@skanav You self-requested your review after approving. Was that on purpose, or can the current state be merged?

skanav:
> @skanav You self-requested your review after approving. Was that on purpose, or can the current state be merged?

It was on purpose. I am using this TI and found a couple of minor points (e.g., the one with `-`).

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -50,14 +51,16 @@ def determine_result(self, run):
...
+            if verdict_match and verdict is None:
                 # CoVeriTeam outputs benchexec result categories as verdicts.
-                try:
-                    d = ast.literal_eval(line)
-                    if isinstance(d, dict):
-                        return d.get("verdict", result.RESULT_ERROR)
-                except SyntaxError:
-                    pass
-        return result.RESULT_ERROR
+                verdict = verdict_match.group(1).lower()
```
skanav, line 61: Why do we call `lower()` here? Does it bring us something?
What I have noticed is that if I run `CPAchecker` using `CoVeriTeam` and an exception occurs, then the statuses differ.
e.g.,  `CPAchecker`  would return the status `EXCEPTION` whereas `CoVeriTeam` would return `exception`

skanav:
```suggestion
                verdict = verdict_match.group(1)
```

PR #706, opened 2020-12-21
Added a tool info module for IGML

jfh1911:
The info-module will firstly be used to test and evaluate the
approach, the tool will be available soon.

Hence, the file is unfortunately not finalized and there might be slight changes to this info-module.


jfh1911:
The tool info module is now finished and can be merged.
I guess that it makes sense to squash all commits before merging.
Unfortunately, I dont know how to quash the commits (cannot find the option).

@PhilippWendler, can you tell me how to do this (or just quash everything, if this is easier)?

benchexec/tools/igml.py
```diff
@@ -0,0 +1,136 @@
...
+
+    REQUIRED_PATHS = ["lib/*.jar", "scripts", "igml.jar"]
+
+    def executable(self, tool_locator):
+        executable = tool_locator.find_executable("igml.sh", subdir="scripts")
+        base_dir = os.path.join(os.path.dirname(executable), os.path.pardir)
+        jar_file = os.path.join(base_dir, "igml.jar")
+        bin_dir = os.path.join(base_dir, "target")
+        src_dir = os.path.join(base_dir, "src")
+        return executable
```
PhilippWendler, line 33:
```suggestion
        return tool_locator.find_executable("igml.sh", subdir="scripts")
```
These variables are never used.

benchexec/tools/igml.py
```diff
@@ -0,0 +1,136 @@
...
+        if task.property_file:
+            options += ["-p", task.property_file]
+
+        # try:
+        #     options += ["-f", task.single_input_file]
+        # except benchexec.tools.template.UnsupportedFeatureException:
+        #     raise benchexec.tools.template.UnsupportedFeatureException(
+        #         "Unsupported task with multiple input files for task '{}'".format(task)
+        # )
+
```
PhilippWendler, line 60:
```suggestion

```

benchexec/tools/igml.py
```diff
@@ -0,0 +1,136 @@
...
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import os
+import re
+import sys
+
+from numpy.matlib import identity
```
PhilippWendler, line 13:
```suggestion
```
unused imports

benchexec/tools/igml.py
```diff
@@ -0,0 +1,136 @@
...
+        reason = ""
+        for line in run.output:
+            if "de.upb.cs.sms.igml.errors.LearningFailedException" in line:
+                reason = "(Learner failed)"
+            if "de.upb.cs.sms.igml.errors.DataGenerationFaildException" in line:
+                reason = "(Data generation failed)"
+            if "Error: Invalid configuration (Cannot handle programs with arrays!)" in line:
+                return result.RESULT_UNKNOWN + "(Arrays)"
+            if "org.sosy_lab.common.configuration.InvalidConfigurationException: " \
+               "The generation of the strongest post conditions failed!" in line:
```
PhilippWendler, line 96:
```suggestion
            if (
                "Error: Invalid configuration (Cannot handle programs with arrays!)"
                in line
            ):
                return result.RESULT_UNKNOWN + "(Arrays)"
            if (
                "org.sosy_lab.common.configuration.InvalidConfigurationException: "
                "The generation of the strongest post conditions failed!" in line
            ):
```
Formatting according to what our formatter black would do.

benchexec/tools/igml.py
```diff
@@ -0,0 +1,136 @@
...
+                    raise benchexec.tools.template.UnsupportedFeatureException(
+                        "Unsupported data_model '{}' defined for task '{}'".format(
+                            data_model, task
+                        )
+                    )
+        return options
+
+    def cmdline(self, executable, options, task, rlimits):
+        additional_options = self._get_additional_options(options, task, rlimits)
+        return [executable] + options + additional_options
```
PhilippWendler, line 80: This completely misses the actual input file(s) of the given task, is this intended?

PhilippWendler (changes requested):
Thanks!

I have a few minor nitpicks and one question.

Regarding squashing, I can easily do so when performing the merge. If you prefer to do it yourself, you would need to do it locally and force-push.

benchexec/tools/igml.py
jfh1911: Yes, as the yml file is given to the tool, where the task is extracted from

benchexec/tools/igml.py
jfh1911: Thanks for the hint :-)

benchexec/tools/igml.py
PhilippWendler: Where would the yml file be given to the tool? I don't see this either in the code. And in fact, BenchExec intentionally hides the name and full contents of the yml file from the tool-info module and the tool, because the yml file may contain the expected verdict.

Are you confusing this with the property file (which could in principle contain input files, but does not in environments like SV-COMP and Test-Comp, which given that you import `sv_benchmarks_util` seems likely to be your target use case)?

jfh1911: I hand the filename of the yml file to the tool via the benchmark-definition (` <option name="-f">${taskdef_path_abs}/${taskdef_name}</option>`).

The tool contains two parts. A invariant generator and a validator.
Internally, the validator is given the .yml file, extracts the expected verdict and hand the task from the yml file to the invariant generator.
Thus, I want to give the expected verdict to the tool by purpose.

This may not be the best way to handle this and may abuse Benchexec.
(But as the tool is not designed to participate in the SV-COMP, this is in my point of view ok.

(A cleaner way would be to split the invariant generator and the verifier into two separate tools, but i am currently not able to do the refactoring needed. )

PhilippWendler: I understand.

PhilippWendler:
All discussions were marked as resolved, but there are no changes. Did you maybe forget to push?

PhilippWendler:
Are you still interested in getting this merged?

PhilippWendler:
It seems not.

PR #707, opened 2020-12-22
Add tool info module for Vampire

JakobR:
This PR adds a tool info module for the first-order theorem prover [Vampire](https://github.com/vprover/vampire). I have done a few runs and it seems to work well.

However, I'm not sure about the method `environment`. It seems to be working and correct (according to the doc-comment in `BaseTool2`), but `test_tool_info` gives the following warning:
```
WARNING: Tool module returned invalid entries for environment, these will be ignored: “{'keepEnv': {'TPTP': 1}}”
```

### DIFF
--- a/benchexec/tools/vampire.py
+++ b/benchexec/tools/vampire.py
@@ -31,27 +31,25 @@ class Tool(benchexec.tools.template.BaseTool2):
         return {"keepEnv": {"TPTP": 1}}
 
     def cmdline(self, executable, options, task, rlimits):
-        assert len(task.input_files) <= 1, "only one input file supported"
-
         if "-t" not in options and "--time_limit" not in options:
-            if rlimits.walltime is None:
+            if rlimits.walltime:
                 # Default timeout of Vampire is 60s,
                 # so we set value of 0 explicitly (means unlimited)
                 options += ["-t", "0"]
             else:
-                options += ["-t", f"{rlimits.walltime}s"]
+                options += ["-t", "{}s".format(rlimits.walltime)]
 
         if "-m" not in options and "--memory_limit" not in options:
-            if rlimits.memory is None:
+            if rlimits.memory:
                 # No memory limit has been set
                 # TODO: should we warn in this case? Vampire is going to use the default of 3000MiB
                 pass
             else:
                 # Vampire's option '--memory_limit/-m' takes the value in MiB
-                memory_mb = int(rlimits.memory / (1024 * 1024))
-                options += ["-m", f"{memory_mb}"]
+                memory_mib = rlimits.memory // (1024 * 1024)
+                options += ["-m", str(memory_mib)]
 
-        return [executable, *options, *task.input_files]
+        return [executable, *options, task.single_input_file]
 
     def determine_result(self, run):
         status = self.get_szs_status(run.output)
@@ -125,4 +123,5 @@ class Tool(benchexec.tools.template.BaseTool2):
 
     def get_value_from_output(self, output, identifier):
         if identifier == "szs-status":
-            return self.get_szs_status(output) or "--"
+            return self.get_szs_status(output)
+        return None
--- a/benchexec/tools/vampire.py
+++ b/benchexec/tools/vampire.py
@@ -1,0 +1,128 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Thanks! I am currently on holidays and would look into this after I return, if it is not urgent.

For the `keepEnv` warning message, I think it is just that the respective handling is missing in `test_tool_info`, but I would confirm this later.

benchexec/tools/vampire.py
```diff
@@ -0,0 +1,128 @@
...
+        line = self._version_from_tool(executable, line_prefix="Vampire")
+        line = line.strip()
+        line = line.split(" ")[0]
+        return line.strip()
+
+    def environment(self, executable):
+        return {"keepEnv": {"TPTP": 1}}
+
+    def cmdline(self, executable, options, task, rlimits):
+        assert len(task.input_files) <= 1, "only one input file supported"
```
PhilippWendler, line 34: Is a run with no input files actually supported?

If not, it would be easier to replace `*task.input_files` at the end of this function with `task.single_input_file` and remove this `assert`. Then there would even be a nice error message for users instead of a crash.

benchexec/tools/vampire.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if "-t" not in options and "--time_limit" not in options:
+            if rlimits.walltime is None:
```
PhilippWendler, line 37: Just FYI: limits are never set to `0` (wouldn't make sense after all) and thus `if rlimits.walltime:` is equivalent (same for other limits).

benchexec/tools/vampire.py
```diff
@@ -0,0 +1,128 @@
...
+                options += ["-t", f"{rlimits.walltime}s"]
+
+        if "-m" not in options and "--memory_limit" not in options:
+            if rlimits.memory is None:
+                # No memory limit has been set
+                # TODO: should we warn in this case? Vampire is going to use the default of 3000MiB
+                pass
+            else:
+                # Vampire's option '--memory_limit/-m' takes the value in MiB
+                memory_mb = int(rlimits.memory / (1024 * 1024))
```
PhilippWendler, line 51:
```suggestion
                memory_mb = rlimits.memory // (1024 * 1024)
```

benchexec/tools/vampire.py
```diff
@@ -0,0 +1,128 @@
...
+                options += ["-m", f"{memory_mb}"]
+
+        return [executable, *options, *task.input_files]
+
+    def determine_result(self, run):
+        status = self.get_szs_status(run.output)
+        if run.exit_code:
```
PhilippWendler, line 58: `run.exit_code´ will always be an object instance, so this condition is always true.

benchexec/tools/vampire.py
```diff
@@ -0,0 +1,128 @@
...
+        prefix = "% Termination reason: "
+        reasons = []
+        for line in output:
+            if line.startswith(prefix):
+                reasons.append(line[len(prefix) :])
+        return reasons
+
+    def get_value_from_output(self, output, identifier):
+        if identifier == "szs-status":
+            return self.get_szs_status(output) or "--"
```
PhilippWendler, line 128: Note that if you just return `None` in case the identifier is not found, BenchExec will handle this gracefully (e.g., empty value in CSV output, long dash rendered in HTML tables).

PhilippWendler (commented):
Thanks!

I mostly have a few minor comments and suggestions.

benchexec/tools/vampire.py
PhilippWendler: One more thing:
An additional `return None` is required by our style guide at the end of this function, it should not just fall through to the end because there is also a return value in some branch.

PhilippWendler:
Regarding `keepEnv`: Your code is correct, and I added the respective feature to `test_tool_info` on the master branch.

benchexec/tools/vampire.py
JakobR: I think this works as it is because `run.exit_code` is an instance of `ProcessExitCode`, which has an implementation of `__bool__`. (And my tests seemed to return the right results.)

JakobR:
Thanks for the comments! I have updated the PR.

benchexec/tools/vampire.py
PhilippWendler: Right, of course you are correct. Seems you know my code better than me ;-)

PhilippWendler (approved):
Thanks for the changes! Good idea to rename `memory_mb` to `memory_mib` as well!

Overall it's a very nice tool-info module that seems well-thought-out. :+1:

PR #708, opened 2020-12-23
Add tests for contrib/mergeBenchmarkSets.py

### LINKED ISSUE #693: Tests for contrib/mergeBenchmarkSets.py

There are two general ideas:

SvenUmbricht:
Closes #693.


.flake8
```diff
@@ -32,7 +32,7 @@ ignore =
     # no warnings about pickle,
     S301,S403,
     # no warnings about XML parsing
-    S318,S320,S405,S408,S410,
+    S314,S318,S320,S405,S408,S410,
```
PhilippWendler, line 35: What warning is `S314`?

contrib/mergeBenchmarkSets.py
```diff
@@ -22,6 +23,27 @@
...
+        help="XML-file containing the verification results.",
+    )
+    parser.add_argument(
+        "witnessXML",
+        nargs="*",
+        metavar="witness-xml",
+        help="Any number of XML-files containing validation results.",
+    )
+    parser.add_argument(
+        "--no-overwrite-status-true",
```
PhilippWendler, line 40: It is less confusing to have variables and fields in source code have "positive" identifiers. I.e., it would be better to have the (internal) option named `overwrite_status_true` (or simply `overwrite_status`?). The option for the user can still be negated where necessary, `argarse` has `store_false`.

contrib/mergeBenchmarkSets.py
```diff
@@ -82,98 +104,85 @@ def getWitnessResult(witness, verification_result):
...
-    for result_tag in resultXML.findall("run"):
-        run = result_tag.get("name")
+def getValidationResult(
+    run, witnessSets, status_from_verification, category_from_verification
+):
+    statusWit, categoryWit = None, None
+    name = run.get("name")
+    for witnessSet in witnessSets:
+        witness = witnessSet.get(name)
+        if witness is None or not len(witness):
```
PhilippWendler, line 114: This should be the same as `if not witness`, shouldn't it?

contrib/test_integration/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    os.path.join(base_dir, "contrib", "mergeBenchmarkSets.py"),
+]
+
+# Set to True to let tests overwrite the expected result with the actual result
+# instead of letting them fail.
+# Use this to update expected files if necessary. Do not commit this flag set to True!
+OVERWRITE_MODE = False
+
+
+class MergeBenchmarkSetsTests(unittest.TestCase):
```
PhilippWendler, line 29: I would prefer to restructure the tests. `contrib/test_integration` sounds like it has tests for all of `contrib`, but if were to add tests for other components of `contrib`, we would want to keep them separate from the tests in this directory. Furthermore, a casual look at the `contrib` directory won't tell developers and users of `mergeBenchmarkSets.py` that there are tests for it. So better name the directory something like `test_mergeBenchmarkSets` and then structure the tests inside this according to the usual test structures with modules and classes.

PhilippWendler (changes requested):
It seems this is not picked up in the usual test runs, e.g., in CI. Can you please add this, such that these tests always get executed?

contrib/mergeBenchmarkSets.py
MartinSpiessl: Not quite, `not witness` would also be `True` in case the witness were an empty list, `witness is None` guarantees that it is actually the `None` object. Depending on the python implementation there might be different `None` objects though :see_no_evil:.

But I agree that `not witness` is probably less errorprone here.

.flake8
MartinSpiessl: parsing untrusted XML. In case we use `etree` this would only mean DoS attacks like billion laughs/ quadratic blowup(cf. https://laconicwolf.com/2018/08/13/writing-and-fixing-vulnerable-xml-parsing-code-in-python/ ), no ACE.

I don't see the parser call though, is probably in code not touched by this PR? In general I would not white list this warning per se in `.flake8` and instead on a per call basis at the location where the XML is parsed. Otherwise we might miss cases in which the warning would actually be justified.

contrib/mergeBenchmarkSets.py
PhilippWendler: The code here is `if witness is None or not len(witness):` and not only `if witness is None:`. What's the difference to `if not witness:`?

MartinSpiessl: `if not witness` is less likely to break, e.g. if you pass an integer.

I am not arguing against that `if not witness` :wink:.

contrib/mergeBenchmarkSets.py
SvenUmbricht: You're right, this option's name certainly is confusing.
Internally it is not used though, but stored as `isOverwrite` (which admittedly is a pretty bad name itself), and I didn't want to rename the option in case some scripts rely on it being called that.

contrib/mergeBenchmarkSets.py
SvenUmbricht: `witness` should actually a be an `xml.etree.ElementTree.Element`.
`len(witness)` returns the number of children but I wasn't sure if that would be considered when using `if not witness`.
I have confirmed now that it is indeed safe to use `if not witness` here and changed it accordingly.

contrib/test_integration/__init__.py
SvenUmbricht: > and then structure the tests inside this according to the usual test structures with modules and classes.

I renamed the directory but am not sure what you mean by the above. Those tests are already structured like the other integration tests, no?

.flake8
SvenUmbricht: More specifically, S314 warns about parsing untrusted XML via `xml.etree.ElementTree`.
This warning appears for the `fromstring` calls in `test_mergeBenchmarkSets.py`.
I mainly added this to the `.flake8` file seeing that S318 and S320 are already there as well which are basically the same for `xml.dom.minidom` and `lxml.etree` respectively, but putting `#noqa` to each call site is no problem -> done.

contrib/mergeBenchmarkSets.py
PhilippWendler: But this is a new option, so how can scripts rely on it?

Anyway, I didn't recommend renaming the option (what would a better name be?), but for renaming the internal field in which it is stored.

Note that `isOverwrite` would better be called `doOverwrite` but even better something like `overwriteStatus`.

contrib/__init__.py
```diff
@@ -0,0 +1,6 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: I don't think it makes sense to declare `contrib` to be a Python package. The scripts in this directory are not meant to be imported, and the existing subpackages expect to be top-level packages. I guess this breaks other stuff.

contrib/test_integration/__init__.py
PhilippWendler: I didn't specifically mean the other integration tests, just that the tests should have a good structure. Right now for example we have `test_mergeBenchmarkSets.TestXMLToString, test_mergeBenchmarkSets.TestWitnesses`, and `test_mergeBenchmarkSets.TestMerge` (all as classes in the same module) and `test_mergeBenchmarkSets_integration.MergeBenchmarkSetsIntegrationTests` (as a class in a package). Wouldn't it make more sense to have a package `test_mergeBenchmarkSets` and then structure all the tests appropriately within it using classes and modules as required?

SvenUmbricht: I see your point, but I deliberately separated unit- and integration tests like this to be consistent with other parts of benchexec.
I suppose having several classes in `test_mergeBenchmarkSets` is kind of arbitrary and makes it seem like there is not much of a connection between the tested functions, even though that isn't the case. My proposal would be to have only one class in `test_mergeBenchmarkSets` but otherwise leave the structure as-is.

contrib/__init__.py
SvenUmbricht: That's true. In that case I don't know how to make `nose.collector` find the tests for mergeBenchmarkSets though.
I found this in the docs[(source)](https://nose.readthedocs.io/en/latest/finding_tests.html):

> Directories that don’t look like tests and aren’t packages are not inspected.

Do you know how to change this?

PhilippWendler: Not right now, but the docs says that this is only the default and that "that plugins and command line options can change any of those rules." I suspect there is a way to additionally add `contrib` to the list of directories where it should search for tests.

contrib/test_integration/__init__.py
PhilippWendler: What is the argument against moving the integration tests into the package `test_mergeBenchmarkSets`, e.g., to a module like `test_mergeBenchmarkSets.integration`?

contrib/mergeBenchmarkSets.py
SvenUmbricht: The option already existed before, I just added usage of argparse.
I renamed `isOverwrite` to `overwrite_status`.

contrib/test_integration/__init__.py
SvenUmbricht: Mainly consistency. It makes sense though, so I changed the test structure.

contrib/__init__.py
SvenUmbricht: Found it: The option is `--include contrib`. Added to setup.cfg.

contrib/mergeBenchmarkSets.py
PhilippWendler: Ok, I understand. Then let's no mix this refactoring with interface changes.

contrib/test_mergeBenchmarkSets/test_integration/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        expected = os.path.join(
+            here,
+            "expected",
+            os.path.basename(resultfile) + ".merged" + merge_suffix + ".xml.bz2",
+        )
+        if OVERWRITE_MODE:
+            os.replace(result, expected)
+        else:
+            self.assert_content_equals(result, expected)
+            os.remove(result)  # TODO: More reliable cleanup
```
PhilippWendler, line 59: How about using `tempfile.TemporaryDirectory` as a context manager (using `with`)? We actually prefer tests to not write somewhere inside the user's directories anyway.

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
```diff
@@ -0,0 +1,538 @@
...
+        new_results = mergeBenchmarkSets.xml_to_string(
+            results_xml, qualified_name, public_id, system_id
+        )
+        new_witness_1 = mergeBenchmarkSets.xml_to_string(
+            witness_xml_1, qualified_name, public_id, system_id
+        )
+        new_witness_2 = mergeBenchmarkSets.xml_to_string(
+            witness_xml_2, qualified_name, public_id, system_id
+        )
+        self.assertTrue(element_trees_equal(results_xml, ET.fromstring(new_results)))
```
PhilippWendler, line 308: Right now `flake8` complains about these calls. Using them in tests is no problem, so just silence `S314`. If you prefer, you can also do this for the whole file.

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
```diff
@@ -0,0 +1,538 @@
...
+import sys
+import unittest
+import xml.etree.ElementTree as ET  # noqa: What's wrong with ET?
+
+from contrib import mergeBenchmarkSets
+from benchexec import result
+
+sys.dont_write_bytecode = True  # prevent creation of .pyc files
+
+mock_results = """<?xml version="1.0"?>
```
PhilippWendler, line 18: Wouldn't it make the tests easier to read if these were separate files?

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
```diff
@@ -0,0 +1,538 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: I guess this file could be `__init__.py`? Having `test_mergeBenchmarkSets.test_mergeBenchmarkSets.TestMergeBenchmarkSets.test_*` feels quite redundant.

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
SvenUmbricht: Accidentaly committed this part, still looking into why one of the tests fails in CI (it doesn't on my machine). But yeah, I'll silence it.

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
SvenUmbricht: I suppose it would -> changed.

contrib/test_mergeBenchmarkSets/test_mergeBenchmarkSets.py
SvenUmbricht: Renamed it.

PhilippWendler:
One more thing: We needed proper license and copyright declarations for the added files. Please add them either as a header or a separate `.license` file such that `reuse lint` is happy. CI doesn't show this because this is a PR from a fork.

@MartinSpiessl Could you also please review this, for example whether it covers everything that it should?

contrib/test_mergeBenchmarkSets/test_integration/__init__.py
SvenUmbricht: Had to add an option to `mergeBenchmarkSets` to be able to put the result files in the temp dir, but should be fine.

MartinSpiessl:
Still on my todo list, I will see to get the review done today

contrib/mergeBenchmarkSets.py
```diff
@@ -41,6 +41,13 @@ def parse_args(argv):
         action="store_true",
         help="Do not overwrite true results with results from validation.",
     )
+    parser.add_argument(
+        "-o",
+        "--outputpath",
+        default=".",
+        metavar="OUT_PATH",
+        help="Folder wherein the generated output files will be placed."
```
MartinSpiessl, line 49: black formatter expects a "," at the end of line 49 :see_no_evil:

contrib/mergeBenchmarkSets.py
```diff
@@ -82,102 +111,97 @@ def getWitnessResult(witness, verification_result):
...
-        argv = sys.argv
+def get_validation_result(
+    run, witness_sets, status_from_verification, category_from_verification
+):
+    status_wit, category_wit = None, None
+    name = run.get("name")
+    for witnessSet in witness_sets:
+        witness = witnessSet.get(name)
+        if not witness:
+            continue
```
MartinSpiessl, line 122: This case is not currently covered by any of the unit tests.

contrib/mergeBenchmarkSets.py
```diff
@@ -82,102 +111,97 @@ def getWitnessResult(witness, verification_result):
...
+                coverage_value = "0.00"
+            status_wit, category_wit = (
+                status_from_verification,
+                result.CATEGORY_CORRECT,
+            )
+            category_from_verification = result.CATEGORY_CORRECT
+            try:
+                coverage_percentage = Decimal(coverage_value) / 100
+            except ValueError:
+                continue
```
MartinSpiessl, line 153: This case is not currently covered by any of the unit tests.

contrib/mergeBenchmarkSets.py
```diff
@@ -82,102 +111,97 @@ def getWitnessResult(witness, verification_result):
...
+            status_from_verification = run.find('column[@title="status"]').get("value")
+            category_from_verification = run.find('column[@title="category"]').get(
                 "value"
             )
-            category_from_verification = result_tag.find(
-                'column[@title="category"]'
-            ).get("value")
         except AttributeError:
             status_from_verification = "not found"
             category_from_verification = "not found"
```
MartinSpiessl, line 189: This case is not currently covered by any of the unit tests.

contrib/mergeBenchmarkSets.py
```diff
@@ -186,18 +210,46 @@ def main(argv=None):
             and categoryWit is not None
         ):
             try:
-                result_tag.find('column[@title="status"]').set("value", statusWit)
-                result_tag.find('column[@title="category"]').set("value", categoryWit)
+                run.find('column[@title="status"]').set("value", statusWit)
+                run.find('column[@title="category"]').set("value", categoryWit)
             except AttributeError:
                 pass
```
MartinSpiessl, line 216: This case is not currently covered by any of the unit tests.

contrib/test_mergeBenchmarkSets/test_integration/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+import unittest
+
+sys.dont_write_bytecode = True  # prevent creation of .pyc files
+
+here = os.path.relpath(os.path.dirname(__file__))
+base_dir = os.path.join(here, "..", "..", "..")
+merge_benchmark_sets = [
+    sys.executable,
+    os.path.join(base_dir, "contrib", "mergeBenchmarkSets.py"),
+]
```
MartinSpiessl, line 22: Couldn't we just run the main function from `mergeBenchmarkSets` instead of spawning subprocesses?
For handling stdout one could e.g. use contextlib. redirect_stdout.

MartinSpiessl (changes requested):
There are 4 cases that are currently not checked by any of the unit cases, I marked them with a comment.   
It is probably best to take care of these cases in a different issue/PR, since imho these cases should at least be logged to stderr, if not lead to a hard fail.

For the integration tests I would suggest to do those by calling the main function instead of creating subprocesses, which would simplify the code and make it easier for python coverage tools to determine what is covered by the integration tests.

MartinSpiessl (approved):

contrib/test_mergeBenchmarkSets/test_integration/__init__.py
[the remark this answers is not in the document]
SvenUmbricht: Removed all usages of subprocess. Handling stdout is not even necessary, `mergeBenchmarkSets.main` just prints the path to the generated result file.

PhilippWendler (approved):

PhilippWendler:
So, what is the plan? Finish the remaining tests first or merge this now?

SvenUmbricht:
If it's just the tests I'd write them before merging, but if we want to change the behavior of `mergeBenchmarkSets` for these cases it's probably better to merge first and do that later.

SvenUmbricht:
@PhilippWendler Could you merge this? I took another look and we do need to change `mergeBenchmarkSets`. I would prefer to do this in another PR, since so far I avoided doing any behavioral changes in this one.

contrib/mergeBenchmarkSets.py
SvenUmbricht: This is covered by `test_getValidationResult_multiple_witnesses`. It also doesn't need changing because it makes perfect sense to have some witnesses not in every witness set, e.g., if a tool only validated correctness witnesses.

PR #709, opened 2020-12-31
Fix state for highlighting the name of the current view

### LINKED ISSUE #697: HTML tables have inconsistent state for highlighting the name of the current view
I have noticed the following behaviour that results in an unwanted highlighting of view names in the top row.
I have tested this in Chromium 87.0.4280.66 and Firefox 83.0.

- open a table like [this one](https://buildbot.sosy-lab.org/cpachecker/results/ldv-klever/00269.ldv-klever.table.html)
- click on the table-tab (button "table" on the top left), the table is shown and the word "table" is highlighted with white background
- **alternative 1**: press the browser's back button.
- **alternative 2**: edit url in the browser, delete the suffix "table", keep the suffix "#/" and press ENTER.
- the summary view appears again, but there is the bug: the word "table" remains with white background, "summary" is not highlighted.

lachnerm:
Fixes #697.


PhilippWendler (approved):
Thanks!

PR #710, opened 2021-01-01
Add column sorting settings to hash routing

### LINKED ISSUE #685: Add column sorting to hash routing of HTML tables
Almost all state of the HTML table is handled by adding to the hash part of the URL such that one can use the browsers back button and create links to specific views. One part that is not handled this way is column sorting. For consistency and usability it would be nice to add this.

lachnerm:
Fixes #685.


PhilippWendler:
Thanks!

This currently does not seem to support multi-column sorting (which can be enabled when pressing <kbd>Shift</kbd> while clicking on columns headers), correct? Could you please add this?

I also noticed that it seems to crash if I first click on a column header and then shift-click on the same column header.

lachnerm:
Oh, I  didn't even know about this feature. It should support the multi-column sorting know and shouldn't crash any more.

PhilippWendler:
Looks good! Could you please have a look at the conflict because I merged #713?

lachnerm:
Should be solved now!

PhilippWendler (approved):
Thanks!

Issue #569, opened 2020-04-20
Upgrade to react-table v7


PhilippWendler:
Version 7 of [react-table](https://react-table.js.org/) is a major rewrite and upgrading is [not trivial](https://github.com/tannerlinsley/react-table#i-want-to-migrate-from-v6-to-v7-how-do-i-do-that) because in the new version, react-table itself [does not render anything anymore](https://react-table.js.org/concepts/#react-table-is-a-headless-ui-library), we need to do this ourselves. But because v6 won't get support any longer, we need to migrate. Maybe we can use some of the examples or other projects as base for rendering our table.

PR #711, opened 2021-01-02
Add pagination to hash routing

### LINKED ISSUE #657: Add pagination to hash routing of HTML tables
Almost all state of the HTML table is handled by adding to the hash part of the URL such that one can use the browsers back button and create links to specific views. One part that is not handled this way is the pagination (both the page size and the currently shown page). For consistency and usability it would be nice to add this (e.g., pressing back on the second page of the table will jump to the previously opened tab instead of the previous page of the table).

lachnerm:
Fixes #657.

There is a small bug in `react-table v6` that shows the wrong page for the `Jump to page` value when opening the table with a page that isn't the first one. This bug has already been reported [here](https://github.com/tannerlinsley/react-table/issues/1789), but there doesn't seem to be a fix for this in v6. It seems to be working for v7 though.

I've now disabled the possibility to jump to a certain page by adding the `showPageJump={false}` property to the table which results in the correct page number being shown. We can also reactivate it, but this may sometimes show the wrong page number.

In case we leave it disabled for now, we should reenable it after upgrading to v7 (#569).


PhilippWendler:
Hm, we definitely need to keep the possibility to jump to arbitrary pages without clicking through each individual page with `Next`. Otherwise even simply going to the last page becomes cumbersome with large tables.

On the other hand, if we show a wrong page number this is also not nice, *especially* in the situation where users opened a table that immediately shows a different page: there would be no way for them to see that they are not on page 1.

So if there is no workaround, it might be best to not merge this until we have v7 of react-table?

We could add the setting of how many rows are shown to the hash routing, though?

lachnerm:
> So if there is no workaround, it might be best to not merge this until we have v7 of react-table?
> 
This would probably be the best way for now I think.
> We could add the setting of how many rows are shown to the hash routing, though?
> 
Yes, we could add this. I will open a PR with only this feature to merge.

PR #712, opened 2021-01-02
Fix bug that makes the slider of filters jump between values

### LINKED ISSUE #672: Fix bug that makes the slider of filters jump between values
I encountered a bug when using the slider of the filter on the data from the development server. It's likely that this also occurs in other tables and isn't directly related to the development data, however I can't confirm this since I haven't investigated that far.

When applying a filter for the column `cputime`, if I move the handle of the upper bound of the slider between the values `2090` and `2100`, the handle automatically jumps to the lowest value `2060` at some threshold inbetween. After reaching some other threshold, it jumps to `2100` as expected. 
Something similar happens if I move the handle between `2190` and `2200`, where it jumps to the value `2110` inbetween. I also find the same behavior when moving the lower handle around.

This bug occurs in both Google Chrome and Firefox.

lachnerm:
Fixes #672.

The bug was inside the `build` function of the `NumberFormatterBuilder` in the `utils/utils.js`. The case of a carry over appearing for the last digit was not handled here and therefore the wrong number was returned. For example:
When passing the value `2196` and `3` as the number of significant digits (so basically the number of digits from the left that should stay unchanged, while all following will be rounded up), the builder breaks this number into all preceding digits (`21`) and the last two digits (`96`), turns the latter into a decimal (`9.6`) and rounds it (`10`). This value is supposed to be single-digit all the time, which is not the case for initial values with the last two digits equal or greater than `95`. The rounded value is then appended to all preceding digits and 0s are added at the end to fill up all missing digits.
So in our case, the builder appends `10` to the `21` which just happens to be a valid value too (`2110`) and no 0s are appended. This is why in the interval `[2195,2199]` the slider jumps to the value `2110` as stated in the original issue. What it should do however, is add the carried over digit to our `20`, resulting in the value `21`, and filling up the rest with 0s to reach the value `2200`.

I've added the calculation for carry overs and didn't encounter any bugs any more.


DennisSimon:
Hi @lachnerm , thanks for taking a look at this. 
I reviewed the code and found out that this issue is also already fixed in https://github.com/sosy-lab/benchexec/pull/616. 
I verified the issue with the numbers that you used as an example. Could you check out the branch and verify that the issue is resolved on there?

lachnerm:
> Hi @lachnerm , thanks for taking a look at this.
> I reviewed the code and found out that this issue is also already fixed in #616.
> I verified the issue with the numbers that you used as an example. Could you check out the branch and verify that the issue is resolved on there?

Yes you are right, the issue is indeed already fixed there, thanks for figuring out! I didn't think about checking this beforehand. Well, luckily it was only a small fix any way, so not too much work was done twice.
I'll close this PR then and link the issue to #616.

PR #713, opened 2021-01-18
Add number of rows per page to hash routing

lachnerm:
As stated in #711, this will add the settings of the number of columns to be shown in the html table to the url hash state.

I've noticed that the arrow buttons in the "jump-to-page" field don't work. The `previous` and `next` buttons work perfectly however. This seems to be the case in the current master too for me. Can anyone confirm this? If so, I'd look into what's the reason for this.


PhilippWendler:
Right, I also noticed this. Does it make sense to fix this together or separately?

lachnerm:
It actually seems that this is the default behavior from `react-table`, see [here](https://github.com/tannerlinsley/react-table/issues/677) or [here](https://github.com/tannerlinsley/react-table/issues/1177). Seems like there would be no other way around except for implementing our own pagination solution. If I understand the answer from the second issue correctly, this is already implemented in v7 though and in case it isn't, I would still wait until we update to the newer version any way. So I would suggest to just leave it as it is for now.

PR #718, opened 2021-02-02
Add remaining tests for contrib/mergeBenchmarkSets.py

### LINKED ISSUE #693: Tests for contrib/mergeBenchmarkSets.py

SvenUmbricht:
Closes #693 for real.


SvenUmbricht:
@MartinSpiessl Does it make sense to have these `except`s that need tests in the first place? There is in theory a use case where neither status nor category need to be present (when checking coverage, used in `test_merge_no_status_no_category`) but is this case practically relevant?

I added tests for the current behavior, but it might be better to change that behavior to just raise an error or print a warning.

PhilippWendler:
@MartinSpiessl (Not sure if GitHub notifies about review requests for draft PRs.)

PhilippWendler:
@SvenUmbricht I don't understand why, but when I renamed the default branch on this repo, somehow this PR got closed instead of updated like the other PRs. Maybe it is because it lists your fork as deleted now? Please create a new PR.

MartinSpiessl:
> @MartinSpiessl Does it make sense to have these excepts that need tests in the first place? There is in theory a use case where neither status nor category need to be present (when checking coverage, used in test_merge_no_status_no_category) but is this case practically relevant?

For Test-Comp there will usually also be a status and category column, regardless of whether it is needed (category will just have the value "missing" if the status is "done"). Absence of these columns would indicate something is wrong, so at least a warning in those cases would make sense. I think @dbeyer currently checks this via other means (checking whether there are empty entries in the tables separately).

In case status+category are missing for a coverage task we would currently overwrite the category with `CATEGORY_CORRECT` and keep the status from the "verification" as is, which might break things further down the road if someone looks at the category in the merged table. But of course also no witness information should be present for such a "broken" run, in which case we would not overwrite anything at all.

> I added tests for the current behavior, but it might be better to change that behavior to just raise an error or print a warning.

Generally I think a warning would be most appropriate, no need to abort creating a merged table in case those values are missing(which an error would do). Warning is a concern for `mergeBenchmarkSets.py`, not for how it is tested. So I would say the tests are fine. Of course checking whether the status/category is set to `"not found"` specifically does not make too much sense/is a bit too specific, but by doing so we at least check all the edge cases. What we want to check is that the script doesn't fail in case such an input is provided

PR #719, opened 2021-03-29
569 update react table

### LINKED ISSUE #569: Upgrade to react-table v7

lachnerm:
Alright, I'm finally able to open this PR. This PR updates the `react-table` package from v6 to v7 and closes #569.
The changes in the update to v7 are huge, changing a ton of the API and removing all of the frontend rendering, so I basically had to rewrite a lot from scratch. 

I tried to stay as close to the old design of the table, doing only some minor style changes (cell vertical alignment, transition on hovering etc.). If they are unwanted, I can remove those adjustments of course. Likewise, if desired, I can also change a lot more of the style, since we now have full control over the frontend.

I have tested the table a lot and added all functionalities that our old table offered that I know of, but it's still possible that I missed something. It may be that there is something that doesn't work properly or the table behaves strangly on some occasions, so it'd be helpful if someone else could also do some manual tests.

I have added a `Next` and `Last` button to the pagination next to the `Previous` and `Next` buttons. I think this might be helpful for large tables, but I can also remove them again if they are unwanted.

Additionally, the bug mentioned [here](https://github.com/sosy-lab/benchexec/pull/711#issue-547762813) as well as the bug from [here](https://github.com/sosy-lab/benchexec/pull/713#issue-556779029) are now fixed too.

The code I uploaded is still not in it's final version - I will still do some refactoring and add documentations. However, I think it's a good idea to already upload the changes for testings, so I can fix any bugs or desired changes while I'm still working on the code any way. So for now, don't mind the code itself and only the resulting table instead. I will also fix the failing tests soon.


lachnerm:
Oh, and there's one more thing: I have added min widths to the table columns, so there is a threshold to how small they can become when resizing. What do you think of this? I've found some tables to look rather strange without this, because the cells were too small, so I thought we might take this feature into consideration.

lachnerm:
The code would now be ready for review too.

PhilippWendler (commented):
I noticed that there a few other unrelated changes as well. Could you please move them out, too?

Especially for complex changes like this we should aim at keeping the PR as small as reasonably possible to make it easier to review, test, and debug in case something goes wrong in the future (e.g., when the need for bisecting or reverting arises).

lachnerm:
> Thanks for finding this! Neither the bug nor the solution is specific to react-table v7, am I right? Could you then please move this out of this PR such that we can fix the bug right now for our users and don't have to make the refactoring PR more complex than necessary?

Yes, you are right, this is not specific to the react-table update. I've created a separate PR for the fix. Would it be fine to still leave the commit with the fix here too? Or should I remove the changes and take care of it later when merging?

> I noticed that there a few other unrelated changes as well. Could you please move them out, too?

I've gone through all changes again and I think the other changes are related to the update. I've removed some leftover `console.log` statements which of course is unrelated, but doesn't have any influence on any features. Of course, I can move these out if you want to though.
Also I've switched back to the default `data.json` file for the js development server and moved the declaration for `uniqid` to the `devDependencies` instead of `dependencies`.
Except for that, I don't think there is anything to move out of the PR, or did I miss something? 
The test warnings and the need to add `uniqid` to the `devDependencies` actually only emerged with the new update by the way. I'm not entirely sure why, but the changes relating to this were reasonable to do any way, so I didn't investigate any further.

PhilippWendler:
> > Thanks for finding this! Neither the bug nor the solution is specific to react-table v7, am I right? Could you then please move this out of this PR such that we can fix the bug right now for our users and don't have to make the refactoring PR more complex than necessary?
> 
> Yes, you are right, this is not specific to the react-table update. I've created a separate PR for the fix. Would it be fine to still leave the commit with the fix here too? Or should I remove the changes and take care of it later when merging?
> 
> > I noticed that there a few other unrelated changes as well. Could you please move them out, too?
> 
> I've gone through all changes again and I think the other changes are related to the update. I've removed some leftover `console.log` statements which of course is unrelated, but doesn't have any influence on any features. Of course, I can move these out if you want to though.

Would be nice. The more complex a PR is, the more important is it to keep it minimal and well structured. This helps incredibly when the need for bisecting changes should arise later. For example, if there is some back-and-forth like with `data.json` it is difficult to cleanly bisect the main changes.

For this PR, I would also prefer that we rebase it to clean up all the changes that were moved out of it. So please commit everything separately that it is independent (small changes can go directly to master, as usual) and then rebase this PR and make it consist of a well structured and clear sequence of commits for the steps of this refactoring.

You could also have a look at how this interacts with #616, please. Will there be changes necessary to adapt?

lachnerm:
> > > Thanks for finding this! Neither the bug nor the solution is specific to react-table v7, am I right? Could you then please move this out of this PR such that we can fix the bug right now for our users and don't have to make the refactoring PR more complex than necessary?
> > 
> > 
> > Yes, you are right, this is not specific to the react-table update. I've created a separate PR for the fix. Would it be fine to still leave the commit with the fix here too? Or should I remove the changes and take care of it later when merging?
> > > I noticed that there a few other unrelated changes as well. Could you please move them out, too?
> > 
> > 
> > I've gone through all changes again and I think the other changes are related to the update. I've removed some leftover `console.log` statements which of course is unrelated, but doesn't have any influence on any features. Of course, I can move these out if you want to though.
> 
> Would be nice. The more complex a PR is, the more important is it to keep it minimal and well structured. This helps incredibly when the need for bisecting changes should arise later. For example, if there is some back-and-forth like with `data.json` it is difficult to cleanly bisect the main changes.
> 
Alright! I have moved out any changes that are unrelated to the update of the lib. 
> For this PR, I would also prefer that we rebase it to clean up all the changes that were moved out of it. So please commit everything separately that it is independent (small changes can go directly to master, as usual) and then rebase this PR and make it consist of a well structured and clear sequence of commits for the steps of this refactoring.
> 
I've gone through all changes again during rebasing and think that it's actually better to leave everything in a single commit. As you said, a lot of the commits were only due to some back-and-forth or unrelated changes, leaving only changes that are directly related to the update. The rest of the commits were built on top of each other and can easily be combined. So everything that's left includes the update of the table and directly correlated changes, such as fixing the tests etc. If we were to revert anything later on, it would have to be the whole commit anyway to not break anything, so I hope it's fine this way.

> You could also have a look at how this interacts with #616, please. Will there be changes necessary to adapt?

There will be some merge conflicts that need to be taken care of as there are some changes in the `ReactTable` and `Summary` classes in #616. However, this didn't seem to be too complex and should be doable. As far as I could tell, all changes can be integrated into the new table.

PhilippWendler:
> > For this PR, I would also prefer that we rebase it to clean up all the changes that were moved out of it. So please commit everything separately that it is independent (small changes can go directly to master, as usual) and then rebase this PR and make it consist of a well structured and clear sequence of commits for the steps of this refactoring.
> 
> I've gone through all changes again during rebasing and think that it's actually better to leave everything in a single commit. As you said, a lot of the commits were only due to some back-and-forth or unrelated changes, leaving only changes that are directly related to the update. The rest of the commits were built on top of each other and can easily be combined. So everything that's left includes the update of the table and directly correlated changes, such as fixing the tests etc. If we were to revert anything later on, it would have to be the whole commit anyway to not break anything, so I hope it's fine this way.

Note that the issue is not so much with reverting, because one can revert the final merge commit of the PR anyway. Having separate commits is more useful for reading the code, understanding changes, bisecting and debugging.

For example, feature changes like additional navigation buttons and style changes are better in separate commits. If it is effort now to separate them, we can leave it as it is, though.

But could you please explicitly write down which style changes you made intentionally (as mentioned in the PR description)?

Apart from this, I had a look and found that the functionality seems to work fine. Nice!

The style does not need to be identical with the previous style, but I think the following looked nicer before:
- No visual separation between header cells and table content anymore.
- Header cells are now huge for me.
- Word wrapping seems enabled now? This uses lots of vertical space.
- Input fields for pagination are vertically tight.
- Transition animations are slower, I think this distracts.
- No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).

So it would be cool if you could work on polishing theses things (and anything else that you find).

lachnerm:
> > > For this PR, I would also prefer that we rebase it to clean up all the changes that were moved out of it. So please commit everything separately that it is independent (small changes can go directly to master, as usual) and then rebase this PR and make it consist of a well structured and clear sequence of commits for the steps of this refactoring.
> > 
> > 
> > I've gone through all changes again during rebasing and think that it's actually better to leave everything in a single commit. As you said, a lot of the commits were only due to some back-and-forth or unrelated changes, leaving only changes that are directly related to the update. The rest of the commits were built on top of each other and can easily be combined. So everything that's left includes the update of the table and directly correlated changes, such as fixing the tests etc. If we were to revert anything later on, it would have to be the whole commit anyway to not break anything, so I hope it's fine this way.
> 
> Note that the issue is not so much with reverting, because one can revert the final merge commit of the PR anyway. Having separate commits is more useful for reading the code, understanding changes, bisecting and debugging.
> 
> For example, feature changes like additional navigation buttons and style changes are better in separate commits. If it is effort now to separate them, we can leave it as it is, though.
> 
Hm yes, I see. Most of the changes didn't feel like actually _changing_ anything, but instead completely writing the table from scratch, so almost all changes are depending on each other and difficult to split across commits. But the newly added features can definitely be separated. Also it would be possible to split the `Summary` and `ReactTable` tab. Except for that, I could only think of gradually adding all features of the table (filtering, sorting etc.), but I don't think this will be too helpful. So I would restructure the PR into:
1. Updating of `ReactTable`
2. Updating of `Summary`
3. \- n. Additional features

Would that be okay? I would wait until the end of the PR though.
The style is a little bit more difficult, since most of the styling was previously done by the lib itself. Therefore, I didn't have a base where I could change the CSS from, but instead had to write most of it myself. If we wanted to separate the style updates, I would first have to write the necessary CSS to depict the old table, just to remove it again in the next commit with the new styles. 

> But could you please explicitly write down which style changes you made intentionally (as mentioned in the PR description)?
> 
Yes, of course. Do you mean here in the PR or in the commits?

> The style does not need to be identical with the previous style, but I think the following looked nicer before:
> 
> * No visual separation between header cells and table content anymore.
> * Header cells are now huge for me.
> * Word wrapping seems enabled now? This uses lots of vertical space.
> * Input fields for pagination are vertically tight.
> * Transition animations are slower, I think this distracts.
> * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> 
> So it would be cool if you could work on polishing theses things (and anything else that you find).

Alright, I will work on this.

PhilippWendler:
> Hm yes, I see. Most of the changes didn't feel like actually _changing_ anything, but instead completely writing the table from scratch, so almost all changes are depending on each other and difficult to split across commits.

That's clear, and it is not necessary to split everything.

> Also it would be possible to split the `Summary` and `ReactTable` tab.

I guess that would mean we would have a state in between with both library versions, which seems additional effort that is not worth it (and unclear whether it would work at all anyway).

So if possible split only additional features, and if not, then don't split anything.

> The style is a little bit more difficult, since most of the styling was previously done by the lib itself. Therefore, I didn't have a base where I could change the CSS from, but instead had to write most of it myself. If we wanted to separate the style updates, I would first have to write the necessary CSS to depict the old table, just to remove it again in the next commit with the new styles.

Yes, for cases where the first part of the CSS is not too much effort, that would be nice. I.e., don't write 50 lines of CSS that will immediately be deleted, but for other intentional changes like hover transitions, margin adjustments, etc. it would be good to have them separate. The reason is that now it is not clear at all which of the style differences were intentional because you think that makes the style better, which of the style differences are still on the TODO list, and which of the style differences are unintentional and unnoticed.

> > But could you please explicitly write down which style changes you made intentionally (as mentioned in the PR description)?
> 
> Yes, of course. Do you mean here in the PR or in the commits?

Doesn't really matter, PR is good enough.

lachnerm:
> * Header cells are now huge for me.

Do you refer to the padding of the cells? I've reduced it from `10px` to `5px` (previous value), I hope it is better now.
> * Transition animations are slower, I think this distracts.

I've reduced it to half, from .3s to .15s. It was actually instant previously, but I think a small animation looks better. If you want to, I can also remove the transition though.
> * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> 
I don't really understand what you're referring to here, could you explain one more time please?

If that's all for the requested changes, I would start rebasing the PR the way you explained above.

PhilippWendler:
> > * Header cells are now huge for me.
> 
> Do you refer to the padding of the cells? I've reduced it from `10px` to `5px` (previous value), I hope it is better now.

It might be caused by word wrapping actually.

> > * Transition animations are slower, I think this distracts.
> 
> I've reduced it to half, from .3s to .15s. It was actually instant previously, but I think a small animation looks better. If you want to, I can also remove the transition though.

Hm. Is there really an advantage of this short barely visible animation?

> > * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> 
> I don't really understand what you're referring to here, could you explain one more time please?

This hover feature is gone (when hovering over the cell but not over the text):

grafik

> If that's all for the requested changes, I would start rebasing the PR the way you explained above.

Hm, the rest of my requests like separation between header and table cells is not there yet?

lachnerm:
> > > * Header cells are now huge for me.
> > 
> > 
> > Do you refer to the padding of the cells? I've reduced it from `10px` to `5px` (previous value), I hope it is better now.
> 
> It might be caused by word wrapping actually.
> 
Hm okay, I wasn't really able to reproduce aynthing similar. I've now deactivated word wrapping for the entire table, did that fix it?
> > > * Transition animations are slower, I think this distracts.
> > 
> > 
> > I've reduced it to half, from .3s to .15s. It was actually instant previously, but I think a small animation looks better. If you want to, I can also remove the transition though.
> 
> Hm. Is there really an advantage of this short barely visible animation?
> 
Well, there is no real advantage of course, it just makes the effect smoother and is a very common effect that is used for hovering-related style changes. But if we prefer to keep it without without transitions, that's fine too of course, it's still only a stylistic choice after all.
> > > * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> > 
> > 
> > I don't really understand what you're referring to here, could you explain one more time please?
> 
> This hover feature is gone (when hovering over the cell but not over the text):
> 
> grafik
> 
Ah I see, I actually thought this was a bug, but it makes sense if it was to differentiate between the two features. A small suggestion: How about making two hovering effects, where the background color is the same initially and only when hovering over the inner element. the corresponding part gets darker? Something like this:

hover

I don't think it looks really good to keep the background of the inner element white, it creates some sort of too strong contrast (and even made me think it's a bug). However, it's again only a stylistic choice, so we can also keep it as it is if you prefer.
 
> > If that's all for the requested changes, I would start rebasing the PR the way you explained above.
> 
> Hm, the rest of my requests like separation between header and table cells is not there yet?

I actually uploaded these, but maybe I misunderstood and changed the wrong things. So just to make sure:

> No visual separation between header cells and table content anymore.

I removed this horizontal darkgray border:
image
I just noticed that I may have misunderstood what you meant though. Did you maybe refer to _adding_ this lightgray border again?
image

> Word wrapping seems enabled now? This uses lots of vertical space.

This should be fixed I think, I should have disabled the wrapping.

> Input fields for pagination are vertically tight.

By "vertically tight", I thought you were referring to the padding/margin that is added above/below the elements of the pagination, so I reduced it. Did you mean something else?

PhilippWendler:
> > > > * Header cells are now huge for me.
> > > 
> > > 
> > > Do you refer to the padding of the cells? I've reduced it from `10px` to `5px` (previous value), I hope it is better now.
> > 
> > 
> > It might be caused by word wrapping actually.
> 
> Hm okay, I wasn't really able to reproduce aynthing similar. I've now deactivated word wrapping for the entire table, did that fix it?

Yes, but now the behavior on overflow is still different than before:
grafik
grafik


> > > > * Transition animations are slower, I think this distracts.
> > > 
> > > 
> > > I've reduced it to half, from .3s to .15s. It was actually instant previously, but I think a small animation looks better. If you want to, I can also remove the transition though.
> > 
> > 
> > Hm. Is there really an advantage of this short barely visible animation?
> 
> Well, there is no real advantage of course, it just makes the effect smoother and is a very common effect that is used for hovering-related style changes. But if we prefer to keep it without without transitions, that's fine too of course, it's still only a stylistic choice after all.

Let's leave it out or we think globally about animations and then add them consistently where it makes sense (but separately from this PR).

> > > > * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> > > 
> > > 
> > > I don't really understand what you're referring to here, could you explain one more time please?
> > 
> > 
> > This hover feature is gone (when hovering over the cell but not over the text):
> > grafik
> 
> Ah I see, I actually thought this was a bug, but it makes sense if it was to differentiate between the two features. A small suggestion: How about making two hovering effects, where the background color is the same initially and only when hovering over the inner element. the corresponding part gets darker? Something like this:
> 
> hover

Sure, that's a really good idea (but also should be done separately).

> I don't think it looks really good to keep the background of the inner element white, it creates some sort of too strong contrast (and even made me think it's a bug). However, it's again only a stylistic choice, so we can also keep it as it is if you prefer.

It might not look good, but without it, when hovering in the outer area people will assume that it triggers the effect of the text "click here to select columns". They would only notice that this is not the case after hovering over the inner area.

> > Hm, the rest of my requests like separation between header and table cells is not there yet?
> 
> I actually uploaded these, but maybe I misunderstood and changed the wrong things. So just to make sure:
> 
> > No visual separation between header cells and table content anymore.

This was meant to say that in contrast to the previous table, in this PR there is no visual separation anymore. It should be re-added.

> I removed this horizontal darkgray border:

That just makes it worse :-)

> image
> I just noticed that I may have misunderstood what you meant though. Did you maybe refer to _adding_ this lightgray border again?
> image

Not only the border, previously there was much more distinction (background, shadows).

> > Input fields for pagination are vertically tight.
> 
> By "vertically tight", I thought you were referring to the padding/margin that is added above/below the elements of the pagination, so I reduced it. Did you mean something else?

The input fields are squeezed now:
grafik
Before:
grafik

While reducing the height of the pagination bar is nice (more space for the table), it should not be so small as to squeeze the elements like this. In the first state of the PR, the pagination bar had the original height but dropdowns were still squeezed.

lachnerm:
> > > > > * Header cells are now huge for me.
> > > > 
> > > > 
> > > > Do you refer to the padding of the cells? I've reduced it from `10px` to `5px` (previous value), I hope it is better now.
> > > 
> > > 
> > > It might be caused by word wrapping actually.
> > 
> > 
> > Hm okay, I wasn't really able to reproduce aynthing similar. I've now deactivated word wrapping for the entire table, did that fix it?
> 
> Yes, but now the behavior on overflow is still different than before:
> grafik
> grafik
> 

Okay, got it, the text overflow is now handled with ellipsis again.

> > > > > * Transition animations are slower, I think this distracts.
> > > > 
> > > > 
> > > > I've reduced it to half, from .3s to .15s. It was actually instant previously, but I think a small animation looks better. If you want to, I can also remove the transition though.
> > > 
> > > 
> > > Hm. Is there really an advantage of this short barely visible animation?
> > 
> > 
> > Well, there is no real advantage of course, it just makes the effect smoother and is a very common effect that is used for hovering-related style changes. But if we prefer to keep it without without transitions, that's fine too of course, it's still only a stylistic choice after all.
> 
> Let's leave it out or we think globally about animations and then add them consistently where it makes sense (but separately from this PR).
> 
Alright, I'll try to give a suggestion after the PR is done.

> > > > > * No distinction between hovering on "Click here to select columns" or the header cell for sorting the first column (on table tab).
> > > > 
> > > > 
> > > > I don't really understand what you're referring to here, could you explain one more time please?
> > > 
> > > 
> > > This hover feature is gone (when hovering over the cell but not over the text):
> > > grafik
> > 
> > 
> > Ah I see, I actually thought this was a bug, but it makes sense if it was to differentiate between the two features. A small suggestion: How about making two hovering effects, where the background color is the same initially and only when hovering over the inner element. the corresponding part gets darker? Something like this:
> > hover
> 
> Sure, that's a really good idea (but also should be done separately).
> 
> > I don't think it looks really good to keep the background of the inner element white, it creates some sort of too strong contrast (and even made me think it's a bug). However, it's again only a stylistic choice, so we can also keep it as it is if you prefer.
> 
> It might not look good, but without it, when hovering in the outer area people will assume that it triggers the effect of the text "click here to select columns". They would only notice that this is not the case after hovering over the inner area.
> 
Yes, I see the point. I'll keep this out for now and revert to the previous behavior.

> > > Hm, the rest of my requests like separation between header and table cells is not there yet?
> > 
> > 
> > I actually uploaded these, but maybe I misunderstood and changed the wrong things. So just to make sure:
> > > No visual separation between header cells and table content anymore.
> 
> This was meant to say that in contrast to the previous table, in this PR there is no visual separation anymore. It should be re-added.
> 
> > I removed this horizontal darkgray border:
> 
> That just makes it worse :-)
> 
> > image
> > I just noticed that I may have misunderstood what you meant though. Did you maybe refer to _adding_ this lightgray border again?
> > image
> 
> Not only the border, previously there was much more distinction (background, shadows).
> 

I think I found and restored all previous styles now. 

> > > Input fields for pagination are vertically tight.
> > 
> > 
> > By "vertically tight", I thought you were referring to the padding/margin that is added above/below the elements of the pagination, so I reduced it. Did you mean something else?
> 
> The input fields are squeezed now:
> grafik
> Before:
> grafik
> 
> While reducing the height of the pagination bar is nice (more space for the table), it should not be so small as to squeeze the elements like this. In the first state of the PR, the pagination bar had the original height but dropdowns were still squeezed.

I've increased the padding of the input field to the previous value.

I also found another small style bug:
image
The border below the first headergroup does not go all the way to the end of the table. I'll fix this while rebasing the PR.
Is there anything else I should take note of or have missed? Otherwise I'd start the rebase next.

PhilippWendler:
Thanks!

I found a few more minor details:
- Separator between header and table is only in statistics table, not in main table.
- Separator between header and table is missing in first column of statistics table.
- I still think that the header is not as nicely separated from the table as before. Consider the statistics table, where the row separator in the body of the table is darker and more visually apparent than the separator between header and body.
- Border around statistics table is inconsistent with environment table.
- Vertical column separators missing. Note that the thickness and color of all row and column separators was deliberately chosen to give a good relation between more and less important separators (e.g., column separators within run set vs. across run set vs. row separators), so please replicate them carefully.
- In the pagination bar, the input fields now have different height. I would suggest to make all of them the same height as the buttons.

The plan would be to fix the 2-3 recently found bugs first and only afterwards merge this PR.

lachnerm:
Alright, I've rebased the PR such that react-table is now updated to v7, but the tables still have the same functionality and the same stylings as before. Do you find anything that is still missing or different from before? If not, I would add the `First` and `Last` buttons to the table next.

PhilippWendler:
> Alright, I've rebased the PR such that react-table is now updated to v7, but the tables still have the same functionality and the same stylings as before. Do you find anything that is still missing or different from before?

In the new table, the header of the statistics table and the main table look slightly different (shadow missing in the main table). While checking this, I noticed that (in the existing design) the shadow is between the rows with column headers and filters, but scrolling does not affect the filter row. I guess it makes sense to put the shadow below the filter row, such that it indicates which part of the table is fixed while scrolling and which is not? (I wonder why we never noticed this so far.)

Also the gray background of the row with the run-set names is still missing.

> If not, I would add the `First` and `Last` buttons to the table next.

I would say lets do this in a separate PR.

lachnerm:
> > Alright, I've rebased the PR such that react-table is now updated to v7, but the tables still have the same functionality and the same stylings as before. Do you find anything that is still missing or different from before?
> 
> In the new table, the header of the statistics table and the main table look slightly different (shadow missing in the main table). While checking this, I noticed that (in the existing design) the shadow is between the rows with column headers and filters, but scrolling does not affect the filter row. I guess it makes sense to put the shadow below the filter row, such that it indicates which part of the table is fixed while scrolling and which is not? (I wonder why we never noticed this so far.)
>
> Also the gray background of the row with the run-set names is still missing.
> 
Okay, I finally found the problem. Actually, with the standard color display of my monitor, the shadow of the row is not visible. I just noticed this by changing the tone, which makes it visible for me. The same accounts for the background of the row for the runsets. Well, this explains why it was so difficult to agree on a design so far. 

I assume that this is the case because the shadow is created using black as color with a quite small alpha value. I've cross-checked on another device and the shadow is visible there as well, so maybe just my screen is the exception here. Has there been anything similar known before? I could try to see if I can recreate the shadow and background without using the alpha channel, such that it's visible in all cases. We can also leave it as it is though and I recreate the previous design (now that I have a way to see the shadow as well). In both cases, I can shift the shadow below the filter row of course.

This is by the way what the header looks like on my end (with the current master):



> > If not, I would add the `First` and `Last` buttons to the table next.
> 
> I would say lets do this in a separate PR.

Alright.

PhilippWendler:
> Okay, I finally found the problem. Actually, with the standard color display of my monitor, the shadow of the row is not visible. I just noticed this by changing the tone, which makes it visible for me. The same accounts for the background of the row for the runsets. Well, this explains why it was so difficult to agree on a design so far.

Right :-)
But this is still weird: In your screen shot, the background is all white and no shadow for me, too. How can a screen influence what is shown on a screen shot when that screen shot is displayed on a different screen? I would have understood the situation if the gamma settings of your screen or something similar would have made the light gray invisible, but then screen shot would look just normal. It seems like on your device the browser does not even render these effects? Are there any differences in browser version, OS, browser settings etc. across your devices where you see the difference in rendering behavior?

> Has there been anything similar known before?

No.

> I could try to see if I can recreate the shadow and background without using the alpha channel, such that it's visible in all cases.

I think that using the alpha channel is the only real way to render shadows, at least? But for the background feel free to implement whatever works, as long as the outcome is the same. If I remember correctly, the gray background has a deliberate choice of gray that is consistent with the other gray tones like cell borders, so it would be good to stay consistent.

For reference, this is how it should look on current master:
image

lachnerm:
I've restored the shadows and background colors and moved the shadow below the filter row. Is there anything else that is missing?

PhilippWendler:
Now I understand ;-)

Looks fine now! After the conflicts are resolved I would merge this.

PR #721, opened 2021-04-19
add launch config and npm script for debugging tests

DennisSimon:
This PR adds a configuration for vscode to attach its debugger to a running node instance started with `--inspect` or `--inspect-brk`. 
A command for running tests in debugging mode has been added additionally.


PhilippWendler:
I can't comment on the usefulness, but I would have nothing against adding it, if @lachnerm agrees. We need a `.license` file though, and some docs in `README.md`, for example if users of other editors want to use the debugging feature.

lachnerm:
I haven't tried out the vscode debugger this way so far, but I'd generally say it sounds like a good idea. I wouldn't have anything against it.

PhilippWendler:
We'll have a look on debugger integration again once the need arises.

PR #725, opened 2021-04-24
Memoize filter input fields of React Table

### LINKED ISSUE #722: Filtering in tables moves focus
In html tables, the filter for tasknames removes the focus from the textfield.
If I filter for a prefix and then wait 1-2 seconds, the filter is applied.
However, the cursor focus is lost here, before I can type the suffix of a filename.
Thus, I need to click in the textfield again to further extend the filename.

I assume the same problem appears when filtering for other columns, like cputime and memory.

The problem is available BenchExec 3.7 (and also earlier), using Firefox or Chrome.

lachnerm:
Previously, the filter input fields were remounted on every render, as they were simply passed to the table via an arrow function. When memoizing them, they will stay fixed during different rendering phases (except if relevant parameters changed) and unwanted behavior (such as losing the focus of the input field) as well as unnecessary computations can be prevented. Fixes #722.


PhilippWendler:
Thanks! I noticed that while focus now stays in the field, if I type too fast, there are intermediate re-renders with outdated state, such that temporarily my input is lost. Sometimes there are later renders with the correct state, but I think I also got it to completely loose some of my input:

https://user-images.githubusercontent.com/2545335/116520837-894ca700-a8d3-11eb-9f84-506afeef274f.mp4

lachnerm:
> Thanks! I noticed that while focus now stays in the field, if I type too fast, there are intermediate re-renders with outdated state, such that temporarily my input is lost. Sometimes there are later renders with the correct state, but I think I also got it to completely loose some of my input:
> 
>  benchexec-725-2021-04-29_10.11.21.mp4 
> Could you please have a look?

There was a problem with the timer that is set after typing into the input field before the rows are filtered. This timer is supposed to be reset when you keep typing, but it didn't due to the memoizing. This created the updates with outdated state. I've added the timer to the state of the component, which ensures that it's actually reset when re-rendering.

PhilippWendler (approved):
Nice! Thanks!

Issue #727, opened 2021-04-26
benchexec tables cannot be filtered if result contains ":" 


danieldietsch:
Ultimate has multiple results that contain the character `:`. The benchexec tables seem to have issues with filtering for these result types.

E.g., without filter: 
![image](https://user-images.githubusercontent.com/3883391/116114498-b291eb00-a6b9-11eb-95ed-aad588b69c54.png)

Dropdown:
![image](https://user-images.githubusercontent.com/3883391/116114563-c2a9ca80-a6b9-11eb-90a3-4160be7956e3.png)

Selecting "UNKNOWN: OverapproxCex"
![image](https://user-images.githubusercontent.com/3883391/116114615-cfc6b980-a6b9-11eb-9e77-db39fb23cd5d.png)

I would expect that I see all the runs with status "UNKNOWN: OverapproxCex", but instead, I see nothing.

PhilippWendler:
Thanks for reporting this! I can confirm it, it is caused by the filter logic treating any filter value with a `:` in it as a numberic range filter:

https://github.com/sosy-lab/benchexec/blob/ccc5197058bd0faffaa1ef299a611fd54e62d8e0/benchexec/tablegenerator/react-table/src/utils/filters.js#L168

For fixing this we need to pass the column types down to this method, which I think is feasible.

@lachnerm Could you please work on this and add a unit test? Note that I already added such a case to the development data for debugging. Afterwards, please have a look at whether you find similar cases where the column type should be checked but is not. For example, the same method also does `if (name === "status") {`, but we don't want to hard-code column names (they can be changed).

PR #730, opened 2021-05-02
727 fix table for results with colon

lachnerm:
The column type is now passed to the filtering logic, so it doesn't depend on the name of a column any more. Direct fix for #727.
I've enabled the filtering logic using a colon for size ranges (e.g. `100:200`) only for columns of the type `measure`. Is that correct? I could also change it to for example `all types except for status` if that would fit better in practice.

I've also added some unit tests for category and status filtering, including a case where a colon is in the name of a column.


benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -160,6 +167,7 @@ export default class Overview extends React.Component {
     const deserializedFilters = this.filterUrlRetriever() || [];
     if (!deepEqual(this.lastFiltered, deserializedFilters)) {
       // we only want to kick off filtering when filters changed
+      this.addTypeToFilter(deserializedFilters);
```
PhilippWendler, line 170: Won't this break the caching because `deepEqual` in the line above will never return `true` anymore?

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -415,6 +415,15 @@ const TableRender = (props) => {
     return out;
   };
 
+  const addTypeToFilter = (filters) =>
```
PhilippWendler, line 418: Is the code duplication really needed? And why does this method some differently for `id` filters?

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -141,6 +141,13 @@ export default class Overview extends React.Component {
     }
   }
 
+  addTypeToFilter = (filters) =>
+    filters.forEach((filter) => {
+      const [runsetIdx, name, columnIdx] = filter.id.split("_");
```
PhilippWendler, line 146: For reference: #731

benchexec/tablegenerator/react-table/src/components/Overview.js
lachnerm: Yes, you're right. I've moved the function that adds the column type upwards, before the `deepEqual` check is made.

benchexec/tablegenerator/react-table/src/components/ReactTable.js
lachnerm: I've put the function in the `Overview` component and unified the handling for `id` filters, i.e. column types are only added to non-`id` filters.

benchexec/tablegenerator/react-table/src/components/Overview.js
lachnerm: Yes I agree, there are a lot of things that could be refactored. I assume we leave it for now in this PR and handle this separately?

PhilippWendler: Probably better that way, that's why I created the issue.

benchexec/tablegenerator/react-table/src/utils/filters.js
```diff
@@ -165,15 +165,15 @@ const buildMatcher = (filters) => {
       acc[tool] = {};
     }
     let filter;
-    if (value.includes(":")) {
+    if (type === "measure" && value.includes(":")) {
```
PhilippWendler, line 168: I just noticed that this misses `count` columns. In `utils.js` we have `isNumericColumn`, which unfortunately is not directly applicable here, but this could be refactored to reuse the code there?

lachnerm: I've fixed this by using the `isNumericColumn` function and passing the `type` variable as an object. Might still not be the best solution, since we don't have a real column object here, but I think it's still better than just copying the code. What do you think?

PhilippWendler (approved):
Thanks!

PR #732, opened 2021-05-06
Fix score-based Quantile Plot for runsets with tasks without scores

### LINKED ISSUE #728: Weird NaN values in data points in score-based quantile plots
[This plot](https://buildbot.sosy-lab.org/cpachecker/results/predicate-linear/01024.-r_integration-predicate-linear.2021-04-27_23-04-18.results.html#/quantile?scaling=Linear&selection=cputime) has obviously wrong data points where the x value is set to `NaN` according to the tool tip:
grafik
The y values contain valid values that are expected to be present in the plot.

It also occurs for some other plots, but not for our example table. Other kinds of quantile plot do not show this behavior, only the score-based quantile plot.

It occurs in BenchExec 3.7, not clear when this started.

Thanks to @kfriedberger for reporting this.

lachnerm:
Fixes #728.


PhilippWendler (approved):
Thanks!

PR #737, opened 2021-05-18
704 fix empty cell filtering

### LINKED ISSUE #704: Small issues with filters in HTML tables
I encountered at least two slight problems with the current state of filters in HTML tables:
- When specifying a task-name filter via the table tab (not the filter overlay), this is not added to the URL.
- In tables with missing results (either empty cells or cells rendered as `--`), these seem to be always excluded if any filter is used. Even when selecting "Show all" in the filter drop down, such results are still excluded (even reflected by the count in the top right corner) and can be made visible only by using the trash-can icon. Selecting "Empty rows" in the filter always shows nothing.

lachnerm:
Fixes the filtering for empty cells, which were previously filtered out all the time. Additionally, the two design adjustments suggested [here](https://github.com/sosy-lab/benchexec/issues/704#issuecomment-842882107) were implemented.
Fixes #704.


PhilippWendler (approved):
Thanks, looks good!

Could you please add some unit tests for this case?

lachnerm:
Sure, I've added a unit test that checks whether the empty rows are not generally filtered out and one that checks whether they can be particularly filtered for. Is there anything else?

PhilippWendler:
Thanks, together with the empty row being part of the test data for the other unit tests I think this is fine.

PR #739, opened 2021-06-28
Remove object state for MetaVal tool-info module

lembergerth:
Currently, the MetaVal tool-info module is stateful: It only imports required (sub-)tool-info modules when constructing the commandline (`cmdline`), stores it in an object attribute `self.wrappedTools` and then reuses that for `determine_result`.

This means that `determine_result` will not work if `cmdline` was not executed on the same Tool-object before.

This PR changes that:  
Do not keep a state in MetaVal tool-info module,
but just load the required tool-info module of used components
when cmdline and result are determined.

@MartinSpiessl As the maintainer of MetaVal, does this PR make sense to you?

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -44,13 +44,22 @@ class Tool(benchexec.tools.template.BaseTool2):
 
     def __init__(self):
         self.lock = threading.Lock()
-        self.wrappedTools = {}
 
     def executable(self, toolLocator):
         return toolLocator.find_executable("metaval.sh")
 
     def name(self):
         return "MetaVal"
+
+    def _import_tool(self, verifierName):
+        with self.lock:
+            tool = __import__(
+                "benchexec.tools." + verifierName, fromlist=["Tool"]
+            ).Tool()
+        assert isinstance(
+            tool, BaseTool2
+        ), "we expect that all wrapped tools extend BaseTool2"
+        return tool
 
     def determine_result(self, run):
         verifierDir = None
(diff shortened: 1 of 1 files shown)


MartinSpiessl:
Sure, this doesn't reduce the functionality, only extends it. Probably at the cost of some performance (constructing Tool objects for each run?).

Why can't we have both, i.e., keep the caching as such and load the module on a cache miss?

lembergerth:
Thanks for looking at it!

> Why can't we have both, i.e., keep the caching as such and load the module on a cache miss?

Sure, I don't mind - it will *slightly* complicate the code, though. Does creating a `Tool`-object really involve any measurable performance penalty compared to a cache lookup?

PhilippWendler:
Note that the current state of this PR will break tool-info modules that expect that what is documented in BenchExec is always followed: first the methods that return general information about the tool are called, then the run-specific methods. While there is no explicit guarantee that methods like `executable()` are always called, several tool-info modules assume that, and if we change anything in this regard I would rather add this explicit guarantee instead of explicitly ruling it out.

Furthermore, by the same guarantees this kind of caching is safe, and I see no reason to change this and further complicate this (already quite complex) tool-info module.

lembergerth:
Why do you feel this PR makes the module more complicated? It removes state and duplicate code, so isn’t it a simplification?

And which behaviour will break? It seems I missed something important 😬 

> I would rather add this explicit guarantee instead of explicitly ruling it out.

This would make the reuse of benchexec tool-info modules more complicated per se (they were/are used by tbf, symbiotic and CoVeriTeam, at least), but it would also make handling any existing bugs due to state more straightforward - so that would be awesome.

PhilippWendler:
> Why do you feel this PR makes the module more complicated? It removes state and duplicate code, so isn’t it a simplification?

After fixing the broken behavior, the code will be necessarily more complex than before.

> And which behaviour will break? It seems I missed something important :grimacing:

Any tool-info module that stores state after detecting the executable or the version of the tool and needs this state for `determine_result` will be broken.

> > I would rather add this explicit guarantee instead of explicitly ruling it out.
> 
> This would make the reuse of benchexec tool-info modules more complicated per se (they were/are used by tbf, symbiotic and CoVeriTeam, at least), but it would also make handling any existing bugs due to state more straightforward - so that would be awesome.

Done in a77c6c5c.

lembergerth:
Took me a while, but now I understand - the issue with this PR is that I create new `Tool`-objects, so no state is stored between `executable` and `determine_result` for the sub-tools.

Thanks for clarifying the documentation so fast!

PR #740, opened 2021-06-28
WIP: Remove object state for symbiotic tool-info module

lembergerth:
Currently, Symbiotic stores the tool version as state in the tool-info module's `Tool` object when `Tool.executable(..)` is called.
This version is later used in `determine_result` to figure out how to interpret the output of Symbiotic.

This means that `determine_result` only works if `executable(..)` was called before on the same object, which may not always be the case.

Symbiotic outputs the version information at the beginning of its
output. This PR uses  that to determine the version, instead of having to call `executable(..)` before.

But this PR has some disadvantages, too: it means that Symbiotic must commit itself to keeping the tool version in the tool output and in a certain format, and string matching is error-prone, compared to calling `self.version`.

An alternative would be to get the executable at the beginning of `determine_result` to query the version number, but I was not sure about the correct way to get a `tool_locator`.

@mchalupa Mentioning you as the maintainer of Symbiotic.

### DIFF
--- a/benchexec/tools/symbiotic.py
+++ b/benchexec/tools/symbiotic.py
@@ -6,6 +6,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import re
+
 import benchexec.result as result
 from benchexec.tools.template import ToolNotFoundException
 
@@ -30,7 +32,6 @@ class Tool(OldSymbiotic):
             # this may be the old version of Symbiotic
             executable = OldSymbiotic.executable(self, tool_locator)
 
-        self._version = self.version(executable)
         return executable
 
     def program_files(self, executable):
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Note that BenchExec explicitly provides the guarantee that each `Tool` instance will never be used for two different tool installations or similar cases where the result of `version()` or `executable()` could change per call. So this caching is safe.

If you are worried about `executable()` not being called, this will also never happen and I am fine with adding this to the documented guarantees. Depriving the tool-info module of the chance to rely on the `ToolLocator` does not seem useful.

So I see no reason to change this and complicate the code.

Only if @mchalupa thinks that the new code is an improvement per se for the tool modul this could be merged. But I don't think this is the case, because that regex looks really brittle (who wants to guarantee now that Symbiotic will never use anything else than `-dev` and `-pre`,  like `-beta`, `-rc`, etc. will always have three numbers before that, ...?).

lembergerth:
I agree with everything - but third party tools do not adhere to these undocumented expectations, so tool-info modules break if used outside of this workflow.

So strict documentation/guarantees would be a perfect solution.

PhilippWendler:
> I agree with everything - but third party tools do not adhere to these undocumented expectations, so tool-info modules break if used outside of this workflow.

If other tools decide to use tool-info modules, we can't and won't stop them, but it is their responsibility to ensure that everything works. We cannot put this burden on the tool-info modules, first because there are many more tool-info modules than there are tools that use them, and second because I cannot know what assumptions other tools have and so it would never work that tools adhere to some additional external and undocumented rules, because I cannot check and test these rules when reviewing new tool-info modules.

> So strict documentation/guarantees would be a perfect solution.

Done in a77c6c5c.

lembergerth:
> I cannot know what assumptions other tools have and so it would never work that tools adhere to some additional external and undocumented rules

Of course, they should just adhere to the documentation.

> Done in a77c6c5.

Thanks a lot!

mchalupa:
I am a bit late, so just to confirm: I fully support how this PR was resolved.

PR #747, opened 2021-09-07
Make the vcloud Jar location configurable

mmuesly:
Running the vcloud client from benchexec works like a charm, but the vcloud Jar needs to be the same as deployed in the vcloud. Allowing the configuration of the used vcloud Jar file makes it easier for us to run the infrastructure on our machines.

### DIFF
--- a/contrib/vcloud-benchmark.py
+++ b/contrib/vcloud-benchmark.py
@@ -60,13 +60,16 @@ class VcloudBenchmark(VcloudBenchmarkBase):
     """
 
     def load_executor(self):
-        download_required_jars(self.config)
-
         import vcloud.benchmarkclient_executor as executor
 
-        executor.set_vcloud_jar_path(
-            os.path.join(_ROOT_DIR, "lib", "vcloud-jars", "vcloud.jar")
-        )
+        if self.config.vcloud_jar:
+            print(self.config.vcloud_jar)
+            executor.set_vcloud_jar_path(self.config.vcloud_jar)
+        else:
+            download_required_jars(self.config)
+            executor.set_vcloud_jar_path(
+                os.path.join(_ROOT_DIR, "lib", "vcloud-jars", "vcloud.jar")
+            )
 
         logging.debug(
             "This is vcloud-benchmark.py (based on benchexec %s) "
--- a/contrib/vcloud/vcloudbenchmarkbase.py
+++ b/contrib/vcloud/vcloudbenchmarkbase.py
@@ -85,6 +85,13 @@ class VcloudBenchmarkBase(benchexec.benchexec.BenchExec):
             type=str,
             help="Specify files or paths that shall also be transferred and be made available to the run in the cloud.",
         )
+        vcloud_args.add_argument(
+            self.get_param_name("cloudJar"),
+            dest="vcloud_jar",
+            metavar="FILE_OR_PATH",
+            type=str,
+            help="Use this vcloud jar instead of downloading it from the ivy repository.",
+        )
 
     def get_param_name(self, pname):
         return "--v" + pname


PhilippWendler:
Is a command-line parameter the most convenient way of specifying the location of the JAR? One would have to specify this always in your situation.

Furthermore, users who provide the JAR manually would also need to take care of providing all the dependency JARs in their correct version, which seems quite tedious and error-prone. So maybe it would be better to think about how to make the automatic download work for other users as well? For example, should we allow to customize the URL of the Ivy repository?

mmuesly:
I don't know, whether this is a one fits all solution and am happy to drop it in favor of any better approach. Getting the chance to specify the JAR to use on the command line made it possible for us, to run the tasks smoothly in CI and I wanted to share this changes. I only learned the way of deploying the vcloud in a single big JAR that includes all dependencies. We then manage deployed versions by versioning the big JAR. Dependency management has not been a problem for us of far.

An invocation of BenchExec already requires a longer list of parameters and some of them like the  `--master` parameter has to be set always to make things work. Adding a mode that parses a given config file first that allows to push all of the invocation parameters into this config file would really be a game changer for my work with BenchExec, but at the moment, I have shell scripts hiding the complexity of the invocation. I might help to restructure the invocation scripts in the direction of working with config files, if desired. Adding another parameter felt more like the current way of configuring BenchExec following the docs in the code base.

Configuring only the Ivy repository might help some users, but I don't have an Ivy repository up and running. Instead, I version the big JARs in a local Maven repository integrated to a local GitLab instance and provide docker images for running the BenchExec client on local machines with all required dependencies deployed to the image driving the clients for my ci runs.

As said, I have no strong opinions on the final design of the solution and am open for discussing other approaches.

PhilippWendler:
> An invocation of BenchExec already requires a longer list of parameters and some of them like the `--master` parameter has to be set always to make things work. Adding a mode that parses a given config file first that allows to push all of the invocation parameters into this config file would really be a game changer for my work with BenchExec,

That already exists (for a long time), cf. the [docs](https://github.com/sosy-lab/benchexec/blob/master/doc/benchexec.md#starting-benchexec).

> Adding another parameter felt more like the current way of configuring BenchExec following the docs in the code base.

Sure, this is correct in general. However, for something that always needs to be given, it might be worth considering alternatives (or ideally an automatic solution).

> Configuring only the Ivy repository might help some users, but I don't have an Ivy repository up and running. Instead, I version the big JARs in a local Maven repository integrated to a local GitLab instance and provide docker images for running the BenchExec client on local machines with all required dependencies deployed to the image driving the clients for my ci runs.

Ah, I hadn't remembered the "big" JAR, that of course solves the dependency problem.

Isn't Docker inconvenient because you have to be root, though? (question not related to this PR)

> As said, I have no strong opinions on the final design of the solution and am open for discussing other approaches.

Well, if you wouldn't benefit from a more automatic solution, it becomes less important. But maybe the VerifierCloud developers already have a plan on how to solve this in general for other users as well, or at least want to think about this? So I would like to ask @MartinSpiessl first.

mmuesly:
Thank you, I will reconsider, if I use these config files more intensively. But it feels like it is comparable to what we do with using standard bash syntax to pass the flags to the jobs at the moment.

I was rather relating to a property file located in the `home/...` or `/etc/...` dir that allows to configure these always present flags in a central place for a user or a machine. It is possible to implement it the way, that python checks this config file first and allows overwrites by command line flags, if any are present.

> Isn't Docker inconvenient because you have to be root, though? (question not related to this PR)

Running Docker images is no problem as the image is only running the benchmark client, which is basically a Java application that can be run by any user. So, I can run the client on an unprivileged docker runner in our GitLab instance. It becomes a longer running task, but apart from that the job is not different to unit testing. All of the computations that require cgroup access are run by the worker in the cloud with corresponding privileges. I only need a privileged runner that has root access for building the docker images in a docker-in-docker runner. I am happy to give you a short demo of our setup, if you are interested in the details.

Okay, I will wait for an answer of the VerifierCloud developers.

MartinSpiessl:
We are talking about you using your own vcloud master instance, right? Otherwise there would be no way of knowing which version of the jar will work with the sosy-lab master except looking into the ivy repository

Do you require to build the jar from source, or is this just a convenience factor for your? I mean, would it be acceptable if we improved how we make the jar available to cater your use case?

> but I don't have an Ivy repository up and running.
  
Why can't you use the sosy-lab ivy repository then? Especially when you are dependent on our vcloud instance, this does not add any more dependencies on external services. If you need the big jar we could consider providing that for you as well or you can use ivy to resolve the dependencies for you (I hope licensing allows us to publish the big jar, @PhilippWendler?).

> But maybe the VerifierCloud developers already have a plan on how to solve this in general for other users as well, or at least want to think about this?

All use cases I know of so far were dependent on the sosy-lab ivy repository. Dependency management done via ant. I am not sure which other solution would help in the "general use case".  Release via gitlab? Or maven? We also had no other master instances running so far which would make it necessary to provide the big jar for other parties. If it helps we could provide that one in the sosy-lab ivy or via the gitlab release feature.

I think from a software engineering perspective, one problem is that we (roughly) depend on running the same version for vcloud benchmark client and server. Replacing the underlying exchange protocol to something that is more backwards compatible and extensible would be the way to go here. Then you would ideally only need to update your jar if you need some newer feature or bug fix. But this is a huge refactoring and currently not planned. Ideally you also keep updating also the jar that the client uses, because there might be bug fixes nonetheless. So the fact the we need the same versions is not actually the problem here I guess.

mmuesly:
Yes, I am running my own master.

I was not aware that there are public releases in the sosy-lab ivy by now for the VerifierCloud. Awesome to hear that. Given the way BenchExec is packaged along with the VerifierCloud at the moment, I still need to build my own versions until this gets configurable in the VerifierCloud as part of the deployment process.

If you think the long term goal is the distribution over the sosy-lab ivy repository, it is legit to classify my use case as a single corner case that will not get a broader audience and close this PR.

PhilippWendler:
> I was not aware that there are public releases in the sosy-lab ivy by now for the VerifierCloud.

This has always been the case. But the latest version in the Ivy repository will always correspond to the version that our VerifierCloud instance currently needs. This may or may not be the version that other instances currently need. So I don't see how this would help right now?

PhilippWendler:
Closing as indicated by the submitter.

PR #752, opened 2021-09-23
Add tool module for Crux

RyanGlScott:
This adds a tool module for [Crux](https://crux.galois.com/).

See also https://gitlab.com/sosy-lab/sv-comp/bench-defs/-/merge_requests/218.

### DIFF
--- a/benchexec/tools/crux.py
+++ b/benchexec/tools/crux.py
@@ -1,0 +1,87 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+import re
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for Crux (https://crux.galois.com/).
+    """
+
+    REQUIRED_PATHS = ["."]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("crux-llvm-svcomp-driver.sh")
+
+    def name(self):
+        return "Crux"
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options += ["--svcomp-spec", task.property_file]
+        data_model_param = get_data_model_from_task(
+            task, {ILP32: "32bit", LP64: "64bit"}
+        )
+        if data_model_param:
+            options += ["--svcomp-arch", data_model_param]
+        return [executable] + options + list(task.input_files)
+
+    def version(self, executable):
+        s = self._version_from_tool(executable)
+        return s[s.find("version:") :]
+
+    def determine_result(self, run):
+        override_pat = re.compile(
+            'No implementation or override found for pointer: "(.+?)"'
+        )
+
+        for line in run.output:
+            # There are still a good number of functions for which Crux lacks
+            # overrides (see, for example, issue #187). Rather than reporting
+            # FALSIFIED for such programs (which will dock us points), we will
+            # instead conservatively return UNKNOWN, which doesn't lose (or
+            # gain) points. To make it more obvious which programs are UNKNOWN
+            # due to failing overrides versus incomplete goals, we include the
+            # name of the failing override in parentheses after UNKNOWN, which
+            # will show up in benchexec's reports.
+            m = override_pat.search(line)
+            if m:
+                return result.RESULT_UNKNOWN + "(no override: " + m.group(1) + ")"
+            # Crucible does not currently support inline assembly
+            elif "unsupported LLVM value: ValAsm" in line:
+                return result.RESULT_UNKNOWN + "(inline assembly)"
+            # Crucible does not currently support translating `long double`s
+            # (https://github.com/GaloisInc/crucible/issues/810)
+            elif "unsupported LLVM value: ValFP80" in line:
+                return result.RESULT_UNKNOWN + "(long double)"
+            elif "Verification result: VERIFIED" in line:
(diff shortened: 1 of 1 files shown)


benchexec/tools/crux.py
```diff
@@ -0,0 +1,86 @@
...
+            elif "Verification result: FALSIFIED" in line:
+                return result.RESULT_FALSE_PROP
+            elif "Verification result: UNKNOWN" in line:
+                return result.RESULT_UNKNOWN + "(incomplete)"
+            elif "Verification result: ERROR" in line:
+                return result.RESULT_ERROR
+        return result.RESULT_UNKNOWN
+
+    def program_files(self, executable):
+        return [executable] + self.REQUIRED_PATHS
```
PhilippWendler, line 86: As is, this method doesn't really make sense, because `self.REQUIRED_PATHS` is empty. So, first `REQUIRED_PATHS` should be set. If everything in the tool directory is expected to be relevant for execution, then `["."]` can be used as value, otherwise simply give a list of relevant files and directories.

But then, the current implementation of `program_files` does not work, because it needs to return paths that are relevant to the current working directory (and paths in `REQUIRED_PATHS` of course are not, because constants cannot know what the current working directory is). But the default implementation of `program_files` already handles this by assuming that the paths in `REQUIRED_PATHS` are relative to the path of `executable`. So just delete this method.

benchexec/tools/crux.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options += ["--svcomp-spec", task.property_file]
+        data_model_param = get_data_model_from_task(
+            task, {ILP32: "32bit", LP64: "64bit"}
+        )
+        if data_model_param:
+            options += ["--svcomp-arch", data_model_param]
+        return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 33: Does your tool actually support both tasks with more than one input file and tasks without an input file? If not, please use `task.input_files` or `task.single_input_file` accordingly, which will provide proper error messages for the user.

benchexec/tools/crux.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def version(self, executable):
+        s = self._version_from_tool(executable)
+        return s[s.find("version:"):]
+
+    def determine_result(self, run):
+        override_pat = re.compile("No implementation or override found for pointer: \"(.+?)\"")
```
PhilippWendler, line 40: Code formatting:
```suggestion
        override_pat = re.compile(
            'No implementation or override found for pointer: "(.+?)"'
        )
```

PhilippWendler (changes requested):
Thanks! I just have a few smaller suggestions and requests for improvements.

benchexec/tools/crux.py
RyanGlScott: Crux does support more than one input file, but I don't think it would produce meaningful results if it were not given _any_ input files. (Does that ever happen?) I tried `task.single_input_file`, but that caused strange results like:

```
Minimal command line:
        “['./crux-llvm-svcomp-driver.sh', 'I', 'N', 'P', 'U', 'T', '.', 'F', 'I', 'L', 'E']”
```

When I tried `task.input_files`, it worked as expected:

```
Minimal command line:
        “['./crux-llvm-svcomp-driver.sh', 'INPUT.FILE']”
```

As a result, I ended up choosing `task.input_files`. Let me know if this is correct.

benchexec/tools/crux.py
```diff
@@ -0,0 +1,86 @@
...
+                return result.RESULT_FALSE_TERMINATION
+            elif "Verification result: FALSIFIED (unreach-call)" in line:
+                return result.RESULT_FALSE_REACH
+            elif "Verification result: FALSIFIED" in line:
+                return result.RESULT_FALSE_PROP
+            elif "Verification result: UNKNOWN" in line:
+                return result.RESULT_UNKNOWN + "(incomplete)"
+            elif "Verification result: ERROR" in line:
+                return result.RESULT_ERROR
+        return result.RESULT_UNKNOWN
```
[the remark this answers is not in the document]
RyanGlScott: Indeed, I more-or-less guessed what should go here, since I've never been able to reach this line in practice. I'm quite happy to use `result.RESULT_ERROR` here instead.

benchexec/tools/crux.py
RyanGlScott: Hm, I don't recall why exactly I opted to override this. I suppose I was paranoid since the Crux bindist includes a separate `bin/` directory where the executables live, so I wanted to be absolutely sure the files under `bin/` were picked up. But as you've pointed out, I didn't even implement `REQUIRED_PATHS` correctly, so I don't think this paranoia was necessary. Everything still works after deleting `program_files`, so I went with that approach.

benchexec/tools/crux.py
PhilippWendler: > Crux does support more than one input file, but I don't think it would produce meaningful results if it were not given _any_ input files. (Does that ever happen?)

BenchExec supports such tasks, but of course this happens only if the user configures this (not in SV-COMP).

> I tried `task.single_input_file`, but that caused strange results like:
> 
> ```
> Minimal command line:
>         “['./crux-llvm-svcomp-driver.sh', 'I', 'N', 'P', 'U', 'T', '.', 'F', 'I', 'L', 'E']”
> ```

This is just because `single_input_file` is not a list of strings, but a single string (which in Python is a list of characters). One would have to append it to the list of arguments instead of concatenating it. But it doesn't matter, if your tool supports more than one input file, then `task.input_files` is better anyway.

benchexec/tools/crux.py
PhilippWendler: It works locally, but for SV-COMP `REQUIRED_PATHS` is necessary. So please add this with either `["."]` or a list of paths, depending on what fits better.

RyanGlScott: I'm OK with everything every file (both at the top level and under `bin`) being included, so I've added `REQUIRED_PATHS = ["."]`.

PhilippWendler (approved):

PR #753, opened 2021-09-29
Add remaining tests for contrib/mergeBenchmarkSets.py

### LINKED ISSUE #693: Tests for contrib/mergeBenchmarkSets.py

SvenUmbricht:
Reapplies the changes from PR #718. Closes #693.

### DIFF
--- a/contrib/mergeBenchmarkSets.py
+++ b/contrib/mergeBenchmarkSets.py
@@ -10,7 +10,7 @@
 import sys
 
 import argparse
-from decimal import Decimal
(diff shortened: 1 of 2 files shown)


MartinSpiessl (approved):
I checked the changes:
- Formatting is OK
- reuse still OK
- coverage is now very good
- tests make sense and check things that should not break in case benchexec changes

So it is fine from my side. I left one remark where some more documentation in the source would be nice, but not necessary for this to get merged. At second thought I will just commit the comments to this branch myself to address this, that is much easier.

PR #754, opened 2021-09-30
Add files via upload

Anthonysdu:
Adding a new tool info-module for Java witness validator

### DIFF
--- a/benchexec/tools/WitForJBMC.py
+++ b/benchexec/tools/WitForJBMC.py
@@ -1,0 +1,32 @@
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+ def executable(self, tool_locator):
+  return tool_locator.find_executable("Wit4JBMC.py")
+ def name(self):
+  return "Wit4JBMC"
+ def cmdline(self, executable, options, task, rlimits):
+  return [executable] + options + list(task.input_files_or_identifier)
+ def determine_result(self, run): 
+  output = run.output
+  validation = 'unknown'
+  for line in output:
+   if 'Exception' in line:
+    if 'AssertionError' in line:
+     validation = 'false'
+    else:
+     validation = 'unknown'
+    break
+   else:
+    validation = 'true'
+
+  if validation == 'false':
+   status = result.RESULT_FALSE_PROP
+  #print(exit_code)
+  elif validation == 'true':
+   status = result.RESULT_TRUE_PROP
+  else:
+   status = result.RESULT_UNKNOWN
+  return status


benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
```
PhilippWendler, line 5: Please add some documentation about the tool as [suggested here](https://github.com/sosy-lab/benchexec/commit/bd82323567b2c8ebade40b8778460de1dda53308#diff-564acd0f58dd345e8d161cb791d6af53), i.e., at least name and URL. An example is [here](https://github.com/sosy-lab/benchexec/blob/bd82323567b2c8ebade40b8778460de1dda53308/benchexec/tools/cpachecker.py).

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
+import benchexec.tools.template
```
PhilippWendler, line 1: Please add a header with copyright and license information in the REUSE format (cf. other files for examples), without it we cannot merge it.

If you have no further preference, we would appreciate it if you assign the copyright to Dirk Beyer and we can use Apache 2.0 as the license.

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
...
+import benchexec.result as result
+
+
```
PhilippWendler, line 4: Please format the source code with [black](https://pypi.org/project/black).

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
...
+class Tool(benchexec.tools.template.BaseTool2):
+ def executable(self, tool_locator):
+  return tool_locator.find_executable("Wit4JBMC.py")
+ def name(self):
+  return "Wit4JBMC"
+ def cmdline(self, executable, options, task, rlimits):
```
PhilippWendler, line 10: This module does not override `version`, which means that users will not see in the result tables which tool version was benchmarked. This can be confusing and hinders reproducibility.

(Note that SV-COMP actually requires a proper version number for each tool.)

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+ def executable(self, tool_locator):
+  return tool_locator.find_executable("Wit4JBMC.py")
```
PhilippWendler, line 7: Is the Python script the only file that your tool needs, or does it consist of other files and directories as well?

If the latter, for SV-COMP it is necessary to add a `REQUIRED_PATHS` field to this class with a list of all required paths. If everything in the same directory as `Wit4JBMC.py` is required, then `REQUIRED_PATHS = ["."]` suffices, otherwise one can define a list of files and subdirectories there as well.

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
...
+ def name(self):
+  return "Wit4JBMC"
+ def cmdline(self, executable, options, task, rlimits):
+  return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 11: Does your tool support tasks without any input file? If not, it would be better to use `task.input_files` because this will produce a nice error message for the user in case of a task without input files.

benchexec/tools/WitForJBMC.py
```diff
@@ -0,0 +1,32 @@
...
+   else:
+    validation = 'true'
+
+  if validation == 'false':
+   status = result.RESULT_FALSE_PROP
+  #print(exit_code)
+  elif validation == 'true':
+   status = result.RESULT_TRUE_PROP
+  else:
+   status = result.RESULT_UNKNOWN
```
PhilippWendler, line 31: I would suggest to return `RESULT_UNKNOWN` only if the tool actually gave up itself, and `RESULT_ERROR` in other cases. This will make it easier to distinguish crashes, timeouts, etc. from unknown results, and will allow BenchExec to add more information about the problem in case the tool terminates abnormally.

PhilippWendler (changes requested):
Thanks!

I have a few comments and suggestions, could you please work on them?

benchexec/tools/WitForJBMC.py
Anthonysdu: Hi, the python script is the only executable program. However, an additional python package named NetworkX should be installed and configured, should I download the NetworkX and add it to the archive and the REQUIRED_PATHS? Thanks!

PhilippWendler: For SV-COMP, you can request Ubuntu packages to be installed, and there is one for [networkx](https://packages.ubuntu.com/focal/python3-networkx). If this works for you, it would be the easiest way. Otherwise you have to package it.

Anthonysdu: Thank you, I have fixed these issues, should I make a new pull request?

PhilippWendler: In order to keep the discussion in one place, it would be better to update this PR instead. Just push your commits to your `Anthonysdu-patch-1` branch on your GitHub fork.

benchexec/tools/WitForJBMC.py
PhilippWendler: Thanks for adding the license and the copyright!

However, we require the use of the [REUSE format](https://reuse.software/) for license headers in this project, because these are machine readable. Please update the header accordingly (just like in all other files in this repo).

benchexec/tools/WitForJBMC.py
```diff
@@ -1,32 +1,63 @@
...
+    """
+    Tool info for Wit4JBMC
+    (https://github.com/Anthonysdu/MSc-project/blob/main/jbmc/Wit4JBMC.py).
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("Wit4JBMC.py")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
```
PhilippWendler, line 32: I noticed that the tool in your linked repository does not support a version yet.

Should we delay merging this PR until the way how to retrieve a version number from your tool is finalized, implemented, and tested, in order to avoid the potential for another round of changes in the near future?

benchexec/tools/WitForJBMC.py
```diff
@@ -1,32 +1,63 @@
...
+                    validation = "false"
+                else:
+                    validation = "unknown"
+                break
+            else:
+                validation = "true"
+
+        if validation == "false":
+            status = result.RESULT_FALSE_PROP
+        # print(exit_code)
```
PhilippWendler, line 55:
```suggestion
```
Please remove commented code.

PhilippWendler (changes requested):
I just noticed now that your file is named `WitForJBMC.py`, which is different from the typical naming of tool-info modules. Would you like to rename it to `wit4jbmc.py` to be consistent?

PhilippWendler:
Now everything in the tool-info module uses `wit4java`, was this intended? Currently this does not match what is present in your tool's repo, for example the name of the Python file is wrong.

Note that there was no request to rename your tool or your Python script. `Wit4Java` is fine there. We just wanted to have a consistent naming convention for the tool-info module. So feel free to revert the renaming except for the filename of the tool-info module.

PR #759, opened 2021-10-04
Tool-info module for SESL

SpencerL-Y:


benchexec/tools/sesl.py
```diff
@@ -0,0 +1,42 @@
...
+        return "SESL"
+
+    def version(self, executable):
+        return self._version_from_tool(executable, arg="--version")
+
+    def cmdline(self, executable, options, task, rlimits):
+        options += ["-t", "--sh-mem-leak", "--add-line-info"]
+        if task.property_file:
+            options += ["--svcomp-property", task.property_file]
+        options += list(task.input_files_or_identifier)
```
PhilippWendler, line 30: Does your tool accept tasks with more than one input file or without any input file? If not, I suggest to use `task.input_files` or `task.single_input_file`, because these will produce proper error messages for the user in case an unsupported task is used.

benchexec/tools/sesl.py
```diff
@@ -0,0 +1,42 @@
...
+    def determine_result(self, run):
+        if run.output.any_line_contains("INVALID_DEREF"):
+            return result.RESULT_FALSE_DEREF
+        elif run.output.any_line_contains("INVALID_FREE"):
+            return result.RESULT_FALSE_FREE
+        elif run.output.any_line_contains("Memtrack"):
+            return result.RESULT_FALSE_MEMTRACK
+        elif run.output.any_line_contains("Memcleanup"):
+            return result.RESULT_FALSE_MEMCLEANUP
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 42: I suggest to return `result.RESULT_ERROR` here. `UNKNOWN` is for cases where the tool intentionally gave up, whereas `ERROR` is for all other cases (timeouts, crashes, etc.). If you can distinguish `UNKNOWN` from other cases, then of course feel free to add another case for it.

PhilippWendler (changes requested):
Thanks! I just have a few minor suggestions.

SpencerL-Y:
Thanks for your suggestions Philipp, the tool-info module for sesl has been changed accordingly. Please give it a look : )

benchexec/tools/sesl.py
```diff
@@ -0,0 +1,54 @@
...
+    https://spencerl-y.github.io/SESL/
+    """
+
+    REQUIRED_PATHS = [
+        "bin",
+        "libs",
+        "scripts",
+        "libs/slah/lib",
+        "libs/z3/lib",
+        "sesl-svcomp.sh",
```
PhilippWendler, line 24:
```suggestion
```
These lines are actually redundant, because you have `libs` in there anyway. And the main executable itself is automatically added.

PhilippWendler (approved):
Looks good, thanks! We can merge.

Just a minor cleanup is possible if you wish.

PR #765, opened 2021-10-13
Add tool-info module for Locksmith

vesalvojdani:
This is a simple tool info module. I have tested this on our local instance of benchexec and it seems to work. I will create a pull request of my own compiled binary of Locksmith shortly. I will also create a pull request to the locksmith source repo with changes needed to build it and these sv-comp files. This should make it very easy for the original authors to compete if they desire; otherwise, I'm ready to compete with Locksmith, the gold standard in race detection.

### DIFF
--- a/benchexec/tools/locksmith.py
+++ b/benchexec/tools/locksmith.py
@@ -8,6 +8,7 @@
 import benchexec.tools.template
 import benchexec.result as result
 
+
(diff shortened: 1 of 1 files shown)


vesalvojdani:
Argh, the name is indeed wrong. I entered this via the github interface...

PR #772, opened 2021-10-15
Add symbiotic-witch tool

mchalupa:
A violation witness checker based on Symbiotic.

### DIFF
--- a/benchexec/tools/symbiotic-witch.py
+++ b/benchexec/tools/symbiotic-witch.py
@@ -1,0 +1,22 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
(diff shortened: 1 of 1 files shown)


mchalupa:
> Thanks!
> 
> My main question is: This tool is based on symbiotic, and it's tool-info module is close to the one for symbiotic. It is possible to use inheritance for tool-info modules in order to avoid the duplicate code (the tool-info modules for the family of Ultimate tools does this, for example).
> 
> Do you expect that further changes to the tool-info module of symbiotic would also apply to symbiotic-witch? E.g., if symbiotic changes its output format, would this also mean that the output of symbiotic-witch changes in the same way? Then inheritance should be used.
> 
> If symbiotic-witch is more like a fork of symbiotic and might develop in a different way, then it is better to avoid the inheritance.

At this moment, the tool is quite coupled with Symbiotic, so it is reasonable to use inheritance. I'll do that.

PR #782, opened 2021-11-01
Update fusebmc.py

kaled-alshmrany:
Added `REQUIRED_PATHS_TESTCOMP22` + Fixed minor things

### DIFF
--- a/benchexec/tools/fusebmc.py
+++ b/benchexec/tools/fusebmc.py
@@ -15,40 +15,19 @@ class Tool(benchexec.tools.template.BaseTool2):
     This class serves as tool adaptor for FuSeBMC (https://github.com/kaled-alshmrany/FuSeBMC)
     """
 
-    REQUIRED_PATHS_TESTCOMP20 = ["esbmc", "esbmc-wrapper.py", "my_instrument"]
-    REQUIRED_PATHS_TESTCOMP21 = [
-        "esbmc",
-        "fusebmc.py",
-        "FuSeBMC_inustrment/FuSeBMC_inustrment",
-        "fusebmc_output",
-        "map2check-fuzzer",
-    ]
-    REQUIRED_PATHS_TESTCOMP22 = [
-        "FuSeBMC_engines",
-        "FuSeBMC_FuzzerLib",
-        "FuSeBMC_GoalTracerLib",
-        "FuSeBMC_inustrument/FuSeBMC_inustrment",
-        "fusebmc_output",
-        "FuSeBMC_SeedGenLib",
-        "FuSeBMC_TCGen",
-        "fusebmc_util",
-        "fusebmc.py",
-        "fusebmc_getinput_from_testcase.py",
-        "fusebmc_seed_gen.py",
-    ]
     def executable(self, tool_locator):
         return tool_locator.find_executable("fusebmc.py")
-    
+
     def working_directory(self, executable):
         executableDir = os.path.dirname(executable)
         return executableDir
-    
+
     def version(self, executable):
         return self._version_from_tool(executable, "-v")
-    
+
     def name(self):
         return "FuSeBMC"
-    
+
     def cmdline(self, executable, options, task, rlimits):
         data_model_param = get_data_model_from_task(task, {ILP32: "32", LP64: "64"})
         if data_model_param and "--arch" not in options:
@@ -59,27 +38,17 @@ class Tool(benchexec.tools.template.BaseTool2):
             + options
             + [task.single_input_file]
         )
-    
-    def program_files(self, executable):
-        """
-        Determine the file paths to be adopted
-        """
-        if self._version == 21:
-            paths = self.REQUIRED_PATHS_TESTCOMP21
-        elif self._version > 21:
-            paths = self.REQUIRED_PATHS_TESTCOMP22
-        return self._program_files_from_executable(executable, paths)
-    
+
     def determine_result(self, run):
         status = result.RESULT_UNKNOWN
         
         if run.output.any_line_contains("DONE"):
             status = result.RESULT_DONE
-            
+
         if status == result.RESULT_UNKNOWN:
             if run.was_timeout:
                 status = "TIMEOUT"
             elif not run.output.any_line_contains("Unknown"):
                 status = "ERROR"
-                
+
         return status
--- a/benchexec/tools/fusebmc.py
+++ b/benchexec/tools/fusebmc.py
@@ -5,11 +5,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import benchexec.tools.esbmc as esbmc
-from benchexec.tools.template import ToolNotFoundException
+import os
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import benchexec.tools.template
+import benchexec.result as result
 
-
-class Tool(esbmc.Tool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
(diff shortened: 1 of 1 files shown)


kaled-alshmrany:
Updated fusebmc.py

kaled-alshmrany:
Dear @PhilippWendler , May I ask you whether you could help with this request?

dbeyer:
I would like to suggest to remove the whole REQUIRED_PATHS business entirely.

kaled-alshmrany:
@dbeyer May I ask you to check this update and let me know if it works fine with no errors. And, if it works fine, could you please run our tool "pre-run" to review our improvements.

benchexec/tools/fusebmc.py
```diff
@@ -5,41 +5,50 @@
...
-import benchexec.tools.esbmc as esbmc
-from benchexec.tools.template import ToolNotFoundException
+import os
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import benchexec.tools.template
+import benchexec.result as result
 
-
-class Tool(esbmc.Tool):
+class Tool(benchexec.tools.template.BaseTool2):
```
PhilippWendler, line 13: Please apply the formatting of our code formatter `black` (two blank lines before the class, no trailing whitespace).

benchexec/tools/fusebmc.py
```diff
@@ -5,41 +5,50 @@
...
+        )
+
+    def determine_result(self, run):
+        status = result.RESULT_UNKNOWN
+        
+        if run.output.any_line_contains("DONE"):
+            status = result.RESULT_DONE
+
+        if status == result.RESULT_UNKNOWN:
+            if run.was_timeout:
```
PhilippWendler, line 49: This is redundant, BenchExec will set the timeout status anyway.

benchexec/tools/fusebmc.py
```diff
@@ -5,41 +5,50 @@
...
+                status = "TIMEOUT"
+            elif not run.output.any_line_contains("Unknown"):
```
PhilippWendler, line 51: Wouldn't the logic be easier if `status` is initialized to `ERROR` and set to `UNKNOWN`/`DONE` if the respective strings occur in the output?

benchexec/tools/fusebmc.py
```diff
@@ -5,41 +5,50 @@
...
-        "esbmc",
-        "fusebmc.py",
-        "FuSeBMC_inustrment/FuSeBMC_inustrment",
-        "fusebmc_output",
-        "map2check-fuzzer",
-    ]
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("fusebmc.py")
+
+    def working_directory(self, executable):
```
PhilippWendler, line 21: I wonder why setting the working directory is now necessary when it was not necessary before?

Note that requiring a specific working directory is not a problem for BenchExec, but might be inconvenient for your users.

benchexec/tools/fusebmc.py
```diff
@@ -5,41 +5,50 @@
...
-        elif self._version > 20:
-            paths = self.REQUIRED_PATHS_TESTCOMP21
-        return self._program_files_from_executable(executable, paths)
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = get_data_model_from_task(task, {ILP32: "32", LP64: "64"})
+        if data_model_param and "--arch" not in options:
+            options += ["--arch", data_model_param]
+        return (
+            [executable]
+            + ["-p", task.property_file]
```
PhilippWendler, line 37: A property file is not guaranteed to exist in general. So it would be nicer for users if you check first whether it is `None` and raise an `UnsupportedFeatureException` in this case, instead of letting BenchExec crash later.

However, I notice that no property file was required before, and if the tool-info module would be changed to required a property file, this would mean it becomes incompatible with older tool versions, which is not nice for users and hinders reproducibility. Thus even better would be to use the property file here only if specified.

PhilippWendler (commented):
Thanks! I have some questions and comments, could you please address them?

benchexec/tools/fusebmc.py
kaled-alshmrany: I guess because before was inherited from esbmc, and now we want it to be independent of esbmc.

benchexec/tools/fusebmc.py
kaled-alshmrany: ok

benchexec/tools/fusebmc.py
kaled-alshmrany: Agree with that. My idea was to make it similar to what esbmc has.

kaled-alshmrany (commented):
We made these changes because @dbeyer faced an issue running our tool. And then I noted this is because of the change esbmc made in their file in benchexec. So I decide to stop this inheritance because any changes on the esbmc file will affect fusebmc. So it would be better if we treat fusebmc independently of esbmc.

Here is the issue Dirk faced:
https://gitlab.com/sosy-lab/test-comp/archives-2022/-/issues/2

benchexec/tools/fusebmc.py
PhilippWendler: Right, I overlooked this.

kaled-alshmrany (commented):
Is there anything else I can do or I missed? Sorry to ask, but I am not familiar with this work.

PhilippWendler:
> We made these changes because @dbeyer faced an issue running our tool. And then I noted this is because of the change esbmc made in their file in benchexec. So I decide to stop this inheritance because any changes on the esbmc file will affect fusebmc. So it would be better if we treat fusebmc independently of esbmc.
> 
> Here is the issue Dirk faced: [gitlab.com/sosy-lab/test-comp/archives-2022/-/issues/2](https://gitlab.com/sosy-lab/test-comp/archives-2022/-/issues/2)

Note that the problem discussed there was just a temporary syntax problem that was fixed a few minutes later. So this alone is not a valid reason for splitting the tool-info modules.

The question is: Does fusebmc continue to be based on ESBMC in the future and will for example its output change in the same way that ESBMC' output changes? Then inheritance is better. Otherwise, if the development of fusebmc will probably diverge from ESBMC, then separate tool-info modules are better.

kaled-alshmrany:
I would suggest separating tool-info modules.

PhilippWendler:
Ok. Do you still want to improve your tool-info module and consider some of my suggestions? At least the formatting needs to be fixed, then we can in principle merge this.

kaled-alshmrany (commented):
Yes please

kaled-alshmrany:
@PhilippWendler , I addressed the comments and I believe it is ready to merge.

benchexec/tools/fusebmc.py
```diff
@@ -45,10 +51,4 @@ def determine_result(self, run):
         if run.output.any_line_contains("DONE"):
             status = result.RESULT_DONE
 
-        if status == result.RESULT_UNKNOWN:
-            if run.was_timeout:
-                status = "TIMEOUT"
-            elif not run.output.any_line_contains("Unknown"):
```
PhilippWendler: It would be good if the tool-info module would still be able to recognize the difference between "unknown" (tool gave up because it could not compute an answer) and "error" (unexpected problem occurred, bug in the tool, crash, etc.). Only the check for timeout was redundant.

Of course, if fusebmc never outputs `Unknown` then the current logic is fine.

kaled-alshmrany: I would say fusebmc never outputs Unknown.

kaled-alshmrany (commented):
@PhilippWendler , I guess now It is ready to merge.

PR #786, opened 2021-11-17
Improve infers determine_result

matthiaskettl:


benchexec/tools/infer.py
```diff
@@ -34,11 +34,15 @@ def cmdline(self, executable, options, task, rlimits):
...
+        if not run.output:
+            return result.RESULT_ERROR
+        last_line = run.output[-1]
+        if last_line.startswith("Result:"):
+            run_result = last_line.split(":", maxsplit=2)[-1]
+            if run_result == "true":
+                return result.RESULT_TRUE_PROP
+            if run_result.startswith("false"):
+                return run_result
+            if run_result == "error":
```
PhilippWendler, line 46: I would consider it safer to keep the distinction between "error" and "unknown" as before: If the tool outputs "unknown", then return this, else "error". Otherwise if the tool just crashes with some other error message, segfaults, etc., the result would be "unknown".

Or is there a specific reason why this needs to be changed? (The commit message does not explain this.)

matthiaskettl: You are right. I changed it.

PR #788, opened 2021-11-20
Update brick.py

Taeyang123456:


PhilippWendler:
Having to update the URL everytime a new version of Brick is released does not seem really convenient. Is there no general webpage of Brick that could be linked instead?

Taeyang123456:
In order to not confuse qualification process, I uploads the URL because we use [brick-public-repo](https://github.com/brick-tool-dev/BRICK-2.0) when register for SV-COMP2022, which I think cannot be changed now.

I thinks it's a good idea to have a general webpage for Brick so that we don't need to upload URL every time, and I think I'll take your advice. Thank you!

PhilippWendler:
Ok, so please create a new PR once you have that.

Note that the URL here in the tool-info module is not related to SV-COMP and does not have anything to do with SV-COMP's qualification process. But given that the existing URL is not more helpful than the new one, I merge this PR such that the new URL is there.

PR #789, opened 2021-11-22
Conditions for setting memory and time limit fixed in Vampire tool

mezpusz:


benchexec/tools/vampire.py
```diff
@@ -32,15 +32,15 @@ def environment(self, executable):
 
     def cmdline(self, executable, options, task, rlimits):
         if "-t" not in options and "--time_limit" not in options:
-            if rlimits.walltime:
```
PhilippWendler, line 35: I would suggest the following alternatives (in order of preference) as these would probably make the code better understandable in the future:

1. Remove branching completely and use `options += ["-t", f"{rlimits.walltime or 0}s"]` (with explaining comment). This is the usual way for fallback values in Python and avoids the duplicate code.
2. Swapping branches instead of changing condition.
3. Use `if not rlimits.walltime:` (more "Pythonic" than comparison with `None`).

benchexec/tools/vampire.py
```diff
@@ -32,15 +32,15 @@ def environment(self, executable):
...
+            if rlimits.walltime is None:
                 # Default timeout of Vampire is 60s,
                 # so we set value of 0 explicitly (means unlimited)
                 options += ["-t", "0"]
             else:
                 options += ["-t", f"{rlimits.walltime}s"]
 
         if "-m" not in options and "--memory_limit" not in options:
-            if rlimits.memory:
```
PhilippWendler, line 43: Here I would suggest to just keep the `if rlimits.memory:` branch and omit the empty branch without limit (but keep the comment).

PhilippWendler (commented):
This makes sense, but given the mistake that has already occurred in this part of the code, I would suggest to improve the code further at this occasion in order to maximize its understandability and make it least error prone. I have added some suggestions.

mezpusz:
Hey, thanks for the quick review and the suggestions!

PR #790, opened 2021-11-23
Improve tool-info module of frama-c-sv

MartinSpiessl:
This adds the following improvements for frama-c-sv:

- We want to just return unknown and not the last line
in the tool output in case the tool process ended before
the final line with the result is printed
- We want to be able to pass options to the tool via
  the benchmark definition xml

### DIFF
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -5,6 +5,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import benchexec.result
 import benchexec.tools.template
 from collections.abc import Mapping
 
@@ -34,7 +35,12 @@ class Tool(benchexec.tools.template.BaseTool2):
             cmd += ["--property", task.property_file]
         if isinstance(task.options, Mapping) and "data_model" in task.options:
             cmd += ["--datamodel", task.options["data_model"]]
+        cmd += options
         return cmd
 
     def determine_result(self, run):
-        return run.output[-1].split(":", maxsplit=2)[-1]
+        lastline = run.output[-1]
+        if lastline.startswith("INFO:RESULT:"):
+          return lastline.split(":", maxsplit=2)[-1]
+        else:
+          return benchexec.result.RESULT_UNKNOWN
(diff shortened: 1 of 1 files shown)


benchexec/tools/frama-c-sv.py
```diff
@@ -34,7 +35,12 @@ def cmdline(self, executable, options, task, rlimits):
...
+        cmd += options
         return cmd
 
     def determine_result(self, run):
-        return run.output[-1].split(":", maxsplit=2)[-1]
+        lastline = run.output[-1]
+        if lastline.startswith("INFO:RESULT:"):
+          return lastline.split(":", maxsplit=2)[-1]
+        else:
+          return benchexec.result.RESULT_UNKNOWN
```
PhilippWendler, line 46: Should this be "unknown" or "error"?

Unknown is for cases where the tool gave up, error is for all kinds of unexpected situations. This looks more like the latter?

Please also note the incorrect indentation in this and the other branch.

MartinSpiessl: Thx, I thought  I ran black on this change, but apparently not. Still think 4 spaces for python is too much indentation, but there is no way to argue with black :sweat_smile:. I ran balck on the file to make sure it is formatted correctly now.

I think `unknown` is the right approach here. Usually this only happens if we are interrupted because of a timeout. I checked this locally already and if we return `unknown` then these results are just marked as `timeout` in the benchexec table. Not sure what would happen if we put error, then it would probably put it as `timeout(error)` or so, which is not what we want.

PhilippWendler: In case of timeout, BenchExec will use `TIMEOUT` as result no matter whether the tool-info module returned `UNKNOWN` or `ERROR`. So timeout cases are not relevant in the consideration here.

According to what you describe, there are no cases where the tool just gives up, so `ERROR` would indeed be better. Otherwise if the tool crashes due to some bug, segfault, etc., there would be `UNKNOWN` in the table instead of `ERROR` (with more specific information added by BenchExec if available).

MartinSpiessl: Sure, I will change it to ERROR then, thanks for the explanation!

Issue #776, opened 2021-10-19
Failed to configure container ... overlay mount for '/' failed


PhilippWendler:
On Ubuntu 18.04 with kernel 4.15.0-159 and on Ubuntu 20.04 with kernel 5.4.0-89-generic (and likely similarly recent versions) BenchExec fails by default with the error
```
CRITICAL - Failed to configure container: [Errno 22] Creating overlay mount for '/' failed: Invalid argument. Please use other directory modes, for example '--read-only-dir /'.
```
Avoiding the use of overlay for `/` (e.g., with `--read-only-dir /`) indeed makes it work. Overlay mode for other directories works.

According to the changelog ([here for 18.04](https://launchpad.net/ubuntu/bionic/+source/linux/+changelog)) the respective patch is very likely the one named "ovl: prevent private clone if bind mount is not allowed", which is originally [here](https://github.com/torvalds/linux/commit/427215d85e8d1476da1a86b8d67aceb485eb3631) and was backported to older kernels as well.

We have to investigate what exactly is the condition that makes it fail for `/` and lets it work for other directories, and then consider either reporting this as a regression or find a workaround. As the patch is a security fix (announcements from [Ubuntu](https://ubuntu.com/security/CVE-2021-3732), [Debian](https://security-tracker.debian.org/tracker/CVE-2021-3732)), we might need the latter.

PhilippWendler:
For those affected by this issue, you can install [BenchExec 3.25](https://github.com/sosy-lab/benchexec/releases/tag/3.25) and [fuse-overlayfs](https://github.com/containers/fuse-overlayfs) 1.10 or newer (available for example in Ubuntu 24.04) and then the default directory works again without the need for `--read-only-dir /` or similar.

younghojan:
Hi Philipp! Even though this issue has already been resolved, I realized that I never thoroughly investigated how it arose in the first place. I’m a bit confused. The commit [ovl: prevent private clone if bind mount is not allowed](https://github.com/torvalds/linux/commit/427215d85e8d1476da1a86b8d67aceb485eb3631) prohibits calling `clone_private_mount()` on a mount that contains a locked submount. Why does this lead to all mounts containing non-leaf mount points being treated as `lowerdir` by the overlay mount?

PhilippWendler:
[...]

PR #791, opened 2021-11-23
Support for cgroupsv2

### LINKED ISSUE #133: Support for cgroup v2 - testers needed
Linux 4.5 has added version 2 of cgroup ([documentation](https://www.kernel.org/doc/Documentation/cgroup-v2.txt)). Currently it is not ready (`cpu` controller is only documented but not yet merged, no sign of `cpuset` and `freezer` controllers), but when it is usable, BenchExec will need several adjustments to be able to work with cgroup-v2:
- `cgroup.procs` needs to be used instead of `tasks` (with granularity of processes instead of threads, but this is no problem for us).
- Children of current cgroup cannot be used while current cgroup contains processes: BenchExec needs to create a child cgroup for itself and move all processes from current cgroup there before creating cgroups for runs, or get as parameter a designated empty benchmarking cgroup.
- Controllers need to be enabled for each child cgroup in `cgroup.subtree_control` file, if we do not want to force the user to do this manually.
- `/proc/<PID>/cgroup` does not show controller name for cgroup-v2 hierarchy.
- There can be multiple v1 hierarchies and a v2 hierarchy, with controllers split between them (not sure if we want to support this).
- `cpuacct` controller has been merged into `cpu` controller.
- Files for reading measurements and writing limits have changed.
- Notification about OOM events needs to be done via the `memory.events` file.
- It might be worth adding memory protection (i.e., memory loosely "guaranteed" for a cgroup) with the `memory.low` file.
- Reading peak memory usage is supported starting in Linux 5.19.

### LINKED ISSUE #251: Use systemd API to create cgroup
Currently we create cgroups ourselves, but the kernel is moving in the direction of allowing only a single process to manipulate the cgroup hierarchy. This would be systemd in practice. So we should start migrating to use systemd's dbus API for creating cgroups, and later on for all cgroup accesses.

Documentation on this is available [here](https://www.freedesktop.org/wiki/Software/systemd/ControlGroupInterface/), we should be able to use a "[scope](https://www.freedesktop.org/software/systemd/man/systemd.scope.html)". We can also look at [systemd-run](https://manpages.debian.org/stretch/systemd/systemd-run.1.en.html) to see how this can be done.

Because scopes can be created by ordinary users, this even has the advantage that in the future we might be able to install and use BenchExec on systems as regular users without any special permissions, which would be great.

globin:
This adds support for cgroupsv2 to BenchExec. Support for cgroupsv1 is kept as is.

There are currently open TODOs:
- [x] Add use of cgroup namespace (allows proper nesting).
- [x] Better error messages and documentation.
- [ ] Automated testing (CI).

However, testing is already possible and would be highly appreciated. The actual functionality should be working fine.

There are two ways to let BenchExec use cgroupsv2:
- Start `benchexec`/`runexec` inside their own cgroup (as the only process). On systems with systemd, this can be done by prefixing the command line with:

      systemd-run --user --scope --slice=benchexec -p Delegate=yes ...

- Let BenchExec create their own cgroup. This requires systemd and the `pystemd` Python package installed. On Ubuntu, for example install `python3-pystemd`.

Please test and provide information in a comment here:
- What did work and what not?
- What problems did you encounter?
- What is your system (distribution, kernel version, systemd version)?


PhilippWendler:
Thanks!

Could you please keep the instructions in the PR updated regarding what is necessary to run it?
On my fresh Ubuntu 21.10 it works out-of-the-box without `systemd-run`, with and without `python3-pystemd` installed. Nice!

globin:
Updated to catch the case where benchexec is executed from within a cgroup that contains other tasks that theoretically is writable. This can occur for example when executed in a gnome-session, as this runs in a slice beneath `user@.session` instead of a top-level `session-xyz.scope`.

Also updated the description to inform of executing without need for systemd-run when having pystemd installed.

PhilippWendler:
If I run tests with `systemd-run --user --scope -p Delegate=yes python3 setup.py test`, I get a linear nesting of cgroups:
```
├── runexec_main_js04eizh
│  ├── runexec_main_p3_3o8eb
│  │  ├── runexec_main_6q0udc79
│  │  │  ├── runexec_main_tobkz5ct
│  │  │  │  ├── runexec_main_o6lp952l
│  │  │  │  │  ├── runexec_main_5cfk082c
│  │  │  │  │  │  ├── runexec_main_3bieg1sq
│  │  │  │  │  │  │  ├── runexec_main_76qrlftf
│  │  │  │  │  │  │  │  ├── runexec_main_4_2h6vl0
│  │  │  │  │  │  │  │  │  ├── runexec_main_a7q53ry3
│  │  │  │  │  │  │  │  │  │  ├── runexec_main_6rfgftd0
│  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_qld_exd_
│  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_z03s_npi
│  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_hs_xmhqd
│  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_6m3223il
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_lupmnbum
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main__d6wwh0p
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_z2pga0bb
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_2xda936c
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_uhfz0a1y
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_w8oynzu0
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_244jm9__
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_gy11lvrp
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_femz_3u1
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_su_lv0hx
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_gi6zxpup
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_v05irz6r
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_r7nnv93d
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_0vqbhdlv
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_h3e723o1
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_ct0zes2a
│  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  ├── runexec_main_4wc5k3j7
```
This breaks because the nesting gets too deep.

benchexec/runexecutor.py
```diff
@@ -328,7 +330,7 @@ def _init_cgroups(self):
         """
         This function initializes the cgroups for the limitations and measurements.
         """
-        self.cgroups = find_my_cgroups()
+        self.cgroups = Cgroups.from_system()
```
PhilippWendler, line 333: This call (because it uses the `initial_cgroup=False` default value) will always allocate a new cgroup and move the current process into it.

First, I think this is not good API because a call like `Cgroups.from_system()` looks as if it just creates a `Cgroups` object, it should not have side effects. Second, I think this is likely connected to the problem of unwanted nesting as discussed in https://github.com/sosy-lab/benchexec/pull/791#issuecomment-1061992192, because the nested cgroups are named `runexec_main_` so they result from a call to `create_fresh_child_cgroup()` with `move_to_child=True`, and this is the only case where this happens except for all callers of `move_to_scope()`.

I guess the only meaningful way to handle cgroup creation in `RunExecutor` is allow controlling it from the outside. `RunExecutor` cannot know whether it should create a child cgroup and move into it, or whether the current process was already moved into a child and we should use our current parent cgroup. We could handle this by adding a parameter with the `Cgroups` instance to `RunExecutor`. If given, we should use this instance and not do anything with the current process. BenchExec (and the main method of `runexec` could then prepare this). If the instance is not given, we can for backwards compatibility create it and move the process if necessary. How does that sound?

benchexec/runexecutor.py
```diff
@@ -571,11 +557,11 @@ def _setup_cgroup_time_limit(
             return timelimitThread
         return None
 
-    def _setup_cgroup_memory_limit(self, memlimit, cgroups, pid_to_kill):
+    def _setup_cgroup_memory_limit_thread(self, memlimit, cgroups, pid_to_kill):
         """Start memory-limit handler.
         @return None or the memory-limit handler for calling cancel()
         """
-        if memlimit is not None:
+        if memlimit is not None and cgroups.version == 1:
```
PhilippWendler, line 564: Is it correct to not use `KillProcessOnOomThread` with cgroups v2? The kernel docs for `memory.max` only claim that it invokes the OOM Killer, but what does this mean precisely?

For example what we do not want is that just some process of the cgroup is killed, we want to terminate the whole run.

If this indeed works as desired, an explaining comment should be added.

.gitlab-ci.yml
```diff
@@ -43,6 +43,7 @@ stages:
     # Give $PRIMARY_USER permission to create cgroups
     - test/for_each_of_my_cgroups.sh chgrp $PRIMARY_USER
     - test/for_each_of_my_cgroups.sh chmod g+w $PRIMARY_USER
+    - apt install -y libsystemd-dev python-dev
```
PhilippWendler, line 46: TODO for me: Add packages to `Dockerfile` on `main` and publish new Docker images.

Also, is there a specific reason why the `python3-pystemd` package was not used?

.gitlab-ci.yml
```diff
@@ -114,8 +115,10 @@ pytype:
   stage: test
   image: python:3.7
   before_script:
+    - apt update
+    - apt install -y libsystemd-dev python-dev
```
PhilippWendler, line 119: Probably better to switch to our Docker image for tests here.

benchexec/cgroupsv2.py
```diff
@@ -0,0 +1,365 @@
...
+        stats_map = {s[0]: s[1] for s in (s.split("=") for s in cpu_some_stats)}
+
+        return float(stats_map["total"]) / 1_000_000
+
+    def read_io_pressure(self):
+        io_stats = dict(self.get_key_value_pairs(self.IO, "pressure"))
+        io_some_stats = io_stats["some"].split(" ")
+        stats_map = {s[0]: s[1] for s in (s.split("=") for s in io_some_stats)}
+
+        return float(stats_map["total"]) / 1_000_000
```
PhilippWendler, line 324: Duplicate code here could be simplified.

PhilippWendler (commented):
Thanks!

I have some comments and questions that I added.

Would you please indicate in how far you would be willing to work on this? Just so that I know and can plan accordingly, without us duplicating work.

Further general TODOs:
- Documentation (installation, etc.)
- Check completeness of API docs for cgroups modules.
- `runexec` should get support for moving itself to a scope as well, I see no reason to restrict this functionality to `benchexec`.

PhilippWendler:
The next work items are to test all possible situations and further improve the error handling (e.g., having error messages that explain the user exactly what the problem is and what they should do).

Furthermore, we must not forget to port 87e2bcb679e885025337f07ce515980f299c2bd6 and a1515702ff0510be35da1abfdab884d59a6a253c for cgroups v2.

.gitlab-ci.yml
PhilippWendler: Done in 45c81500f4901be8527a690600377ed97efb15e8.

.gitlab-ci.yml
PhilippWendler: Done in f1fc88fd148a6cc3ed834f8be5224efc9c1fa035.

benchexec/runexecutor.py
```diff
@@ -296,6 +295,9 @@ def print_optional_result(key, unit="", format_fn=str):
     print_optional_result("memory", "B")
     print_optional_result("blkio-read", "B")
     print_optional_result("blkio-write", "B")
+    print_optional_result("total-cpu-pressure-some", "s")
+    print_optional_result("total-io-pressure-some", "s")
+    print_optional_result("total-memory-pressure-some", "s")
```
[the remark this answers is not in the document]
PhilippWendler: Implemented in 327cd01fb785b0cd966995842a791cf408e81201.

benchexec/cgroupsv2.py
PhilippWendler: Done in 9f1029c7d77ecba6e721004ca4a0fe3c9c9df77d.

benchexec/runexecutor.py
```diff
@@ -341,28 +343,33 @@ def _init_cgroups(self):
...
+                    "or container mode. Please enable at least one of them."
+                )
 
-        self.cgroups.require_subsystem(MEMORY)
-        if MEMORY not in self.cgroups:
+        self.cgroups.require_subsystem(self.cgroups.MEMORY)
+        if self.cgroups.MEMORY not in self.cgroups:
             logging.warning("Cannot measure memory consumption without memory cgroup.")
         else:
+            # FIXME
```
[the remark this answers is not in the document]
PhilippWendler: Implemented in 590ac51cd6087e3dd7fc0318f70319e92b02b5a5.

benchexec/runexecutor.py
```diff
@@ -1064,57 +1051,36 @@ def _get_cgroup_measurements(self, cgroups, ru_child, result):
...
-                            bytes_written += int(bytes_amount)
-                    except ValueError:
-                        pass  # There are irrelevant lines in this file with a different structure
-                result["blkio-read"] = bytes_read
-                result["blkio-write"] = bytes_written
+                result["memory"] = max_mem_usage
+
+            oom_count = cgroups.read_oom_count()
+            if oom_count:
+                result["oom"] = oom_count
```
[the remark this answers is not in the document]
PhilippWendler: This is actually the number of killed processes due to OOM, not the OOM count. But it is not really interesting for users, so since f2839b7d5e15c6c0571e6260a877b13a5a9dbffe we do not return it.

benchexec/runexecutor.py
PhilippWendler: Previously it did not work as required, but the kernel has a feature that does what we need, so since 6938ef83a8381e34d99859d5a4ba52a40f738afc this is fine.

lorenzleutgeb:
With `cgroups2`, the [systemd unit `benchexec-cgroups`](https://github.com/sosy-lab/benchexec/blob/38b27bd4ce1f182f68735d20a6da3dc75b0d73cb/debian/benchexec-cgroup.service) is not required anymore, because BenchExec can use pystemd to set things up, right?

https://github.com/sosy-lab/benchexec/blob/38b27bd4ce1f182f68735d20a6da3dc75b0d73cb/debian/benchexec-cgroup.service#L12-L13

PhilippWendler:
> With `cgroups2`, the [systemd unit `benchexec-cgroups`](https://github.com/sosy-lab/benchexec/blob/38b27bd4ce1f182f68735d20a6da3dc75b0d73cb/debian/benchexec-cgroup.service) is not required anymore, because BenchExec can use pystemd to set things up, right?

Yes, exactly.

lorenzleutgeb:
I am trying to package BenchExec for NixOS (see #920). I decided to go for the `cgroupsv2` branch directly, and would like to report.

Packaging went quite smoothly, there's just a complaint that `libseccomp` cannot be found, which I'll fix.

<details>
<summary><code>runexec --debug --no-container echo Test</code> complains that "Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available." cpuset is listed in `/proc/cgroups` (see below) and I am not sure what else I have to configure here.</summary>

<pre>
% runexec --debug --no-container echo Test
2023-10-03 12:57:47,760 - DEBUG - This is runexec 3.18-dev.
2023-10-03 12:57:47,765 - DEBUG - Analyzing /proc/mounts and /proc/self/cgroup to determine cgroups.
2023-10-03 12:57:47,766 - DEBUG - Available Cgroups: {'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'pids': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'io': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope')}
2023-10-03 12:57:47,803 - DEBUG - Process moved to a fresh systemd scope: benchexec_Ay4STzwJJH4.scope
2023-10-03 12:57:47,803 - DEBUG - Analyzing /proc/mounts and /proc/self/cgroup to determine cgroups.
2023-10-03 12:57:47,803 - DEBUG - Available Cgroups: {
  'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope'),
  'pids':   PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope'),
  'cpu':    PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope'),
  'io':     PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope'),
  'kill':   PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope'),
  'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope')
}
2023-10-03 12:57:47,803 - DEBUG - Available Cgroups: {
  'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchexec_process_33siw073'),
  'kill':   PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchexec_process_33siw073'),
  'cpu':    PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchexec_process_33siw073')
}
2023-10-03 12:57:47,803 - WARNING - Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available.
2023-10-03 12:57:47,804 - INFO - Starting command echo Test
2023-10-03 12:57:47,804 - INFO - Writing output to output.log
2023-10-03 12:57:47,804 - DEBUG - Setting up cgroups for run.
2023-10-03 12:57:47,804 - DEBUG - Available Cgroups: {
  'kill':   PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk'),
  'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk'),
  'pids':   PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk'),
  'cpu':    PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk'),
  'io':     PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk'),
  'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk')
}
2023-10-03 12:57:47,804 - DEBUG - Created cgroups {PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_Ay4STzwJJH4.scope/benchmark_fpoc5nqk')}.
2023-10-03 12:57:47,804 - DEBUG - Using additional environment {}.
2023-10-03 12:57:47,806 - DEBUG - Starting process.
2023-10-03 12:57:47,806 - DEBUG - Executing run with $HOME and $TMPDIR below /tmp/BenchExec_run_adf9q116.
2023-10-03 12:57:47,809 - DEBUG - Waiting for process echo with pid 15091
2023-10-03 12:57:47,810 - DEBUG - energy measurement output:
2023-10-03 12:57:47,810 - DEBUG - energy measurement output: cpu_count=1
2023-10-03 12:57:47,810 - DEBUG - energy measurement output: duration_seconds=0.001045
2023-10-03 12:57:47,810 - DEBUG - energy measurement output: cpu0_package_joules=0.046875
2023-10-03 12:57:47,810 - DEBUG - energy measurement output: cpu0_core_joules=0.050537
2023-10-03 12:57:47,810 - DEBUG - energy measurement output: cpu0_dram_joules=0.001770
2023-10-03 12:57:47,810 - DEBUG - Process terminated, exit code 0.
2023-10-03 12:57:47,811 - DEBUG - Getting cgroup measurements.
2023-10-03 12:57:47,911 - DEBUG - Resource usage of run: walltime=0.0024268529999744715, cputime=0.001914, cgroup-cputime=0.001432, memory=1032192
2023-10-03 12:57:47,911 - DEBUG - Cleaning up cgroups.
2023-10-03 12:57:47,911 - DEBUG - Cleaning up temporary directory /tmp/BenchExec_run_adf9q116.
2023-10-03 12:57:47,912 - DEBUG - Size of logfile 'output.log' is 100 bytes, size limit disabled.
starttime=2023-10-03T12:57:47.807575+02:00
returnvalue=0
walltime=0.0024268529999744715s
cputime=0.001432s
memory=1032192B
blkio-read=0B
blkio-write=0B
pressure-cpu-some=0.000002s
pressure-io-some=0s
pressure-memory-some=0s
cpuenergy=0.046875J
cpuenergy-pkg0-core=0.050537J
cpuenergy-pkg0-dram=0.001770J
cpuenergy-pkg0-package=0.046875J
</pre>
</details>

<details>
<summary><code>containerexec --debug bash</code> complains with "Failed to configure container: [Errno 22] Creating overlay mount for <code>/</code> failed: Invalid argument. Please use other directory modes, for example <code>--read-only-dir /</code>."</summary>
<pre>
% containerexec --debug bash
2023-10-03 13:04:50,824 - DEBUG - This is containerexec 3.18-dev.
2023-10-03 13:04:50,824 - INFO - Starting command bash
2023-10-03 13:04:50,825 - DEBUG - Cannot use overlay mode for /var/lib/lxcfs/proc because it is not a directory. Using read-only mode instead.
2023-10-03 13:04:50,825 - INFO - LXCFS is not available, some host information like the uptime leaks into the container.
2023-10-03 13:04:50,825 - DEBUG - Available Cgroups: {}
2023-10-03 13:04:50,825 - DEBUG - Starting process.
2023-10-03 13:04:50,826 - DEBUG - Parent: child process of RunExecutor with PID 16396 started.
2023-10-03 13:04:50,826 - DEBUG - Child: child process of RunExecutor with PID 16396 started
2023-10-03 13:04:50,827 - DEBUG - Failed to make b'/tmp/BenchExec_run_5qp72igi/mount/home/benchexec' a bind mount: [Errno 2] mount(b'/tmp/BenchExec_run_5qp72igi/mount/home/benchexec', b'/tmp/BenchExec_run_5qp72igi/mount/home/benchexec', None, 4096, None) failed: No such file or directory
2023-10-03 13:04:50,827 - DEBUG - Mounting '/' as overlay
2023-10-03 13:04:50,827 - DEBUG - [Errno 16] umount(b'/tmp/BenchExec_run_5qp72igi/mount/') failed: Device or resource busy
2023-10-03 13:04:50,827 - DEBUG - Creating overlay mount: target=b'/tmp/BenchExec_run_5qp72igi/mount/', lower=b'/', upper=b'/tmp/BenchExec_run_5qp72igi/temp/', work=b'/tmp/BenchExec_run_5qp72igi/overlayfs/1'
2023-10-03 13:04:50,827 - CRITICAL - Failed to configure container: [Errno 22] Creating overlay mount for '/' failed: Invalid argument. Please use other directory modes, for example '--read-only-dir /'.
2023-10-03 13:04:50,828 - DEBUG - Waiting for process bash with pid 16396
2023-10-03 13:04:50,828 - DEBUG - Parent: child process of RunExecutor with PID 16396 terminated with return value 128.
2023-10-03 13:04:50,828 - DEBUG - Process terminated, exit code 0.
2023-10-03 13:04:50,828 - DEBUG - Cleaning up temporary directory.
2023-10-03 13:04:50,828 - ERROR - execution in container failed, check log for details
  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 942, in _start_execution_in_container
    grandchild_pid = int(os.read(from_grandchild, 10))
ValueError: invalid literal for int() with base 10: b''

During handling of the above exception, another exception occurred:

  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 296, in main
    result = executor.execute_run(
  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 469, in execute_run
    pid, result_fn = self._start_execution(
  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 538, in _start_execution
    result = self._start_execution_in_container(
  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 946, in _start_execution_in_container
    check_child_exit_code()
  File "/nix/store/cmm6inc9n3q8yv23hjpydbhsnr2c45jf-benchexec-unstable-2023-09-05/lib/python3.10/site-packages/benchexec/containerexecutor.py", line 899, in check_child_exit_code
    raise BenchExecException(
benchexec.BenchExecException: execution in container failed, check log for details
Cannot execute bash: execution in container failed, check log for details.
</pre>
</details>

Fiddling around with the arguments a bit lead me to:

<details>
<summary><code>runexec --debug --read-only-dir / --hidden-dir /home --dir /tmp $(readlink -f $(which bash)) -- -c "echo Test"</code> which still complains, but I am not sure how bad the situation is. Maybe you can tell me? Another issue seems to be that my <code>/run/wrappers/bin/cpu-energy-meter</code> is not found because <code>/run</code> is mounted as hidden. How to best resolve that?</summary>
<pre>
2023-10-03 13:14:13,568 - DEBUG - This is runexec 3.18-dev.
2023-10-03 13:14:13,568 - INFO - LXCFS is not available, some host information like the uptime leaks into the container.
2023-10-03 13:14:13,568 - DEBUG - Available Cgroups: {}
2023-10-03 13:14:13,573 - DEBUG - Analyzing /proc/mounts and /proc/self/cgroup to determine cgroups.
2023-10-03 13:14:13,573 - DEBUG - Available Cgroups: {'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'io': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'pids': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope'), 'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope')}
2023-10-03 13:14:13,601 - DEBUG - Process moved to a fresh systemd scope: benchexec_O5DHPeGtQQ0.scope
2023-10-03 13:14:13,601 - DEBUG - Analyzing /proc/mounts and /proc/self/cgroup to determine cgroups.
2023-10-03 13:14:13,601 - DEBUG - Available Cgroups: {'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope'), 'io': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope'), 'pids': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope'), 'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope')}
2023-10-03 13:14:13,601 - DEBUG - Available Cgroups: {'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchexec_process_1y4d9uty'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchexec_process_1y4d9uty'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchexec_process_1y4d9uty')}
2023-10-03 13:14:13,602 - WARNING - Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available.
2023-10-03 13:14:13,602 - INFO - Starting command /nix/store/kxkdrxvc3da2dpsgikn8s2ml97h88m46-bash-interactive-5.2-p15/bin/bash -c 'echo Test'
2023-10-03 13:14:13,602 - INFO - Writing output to output.log
2023-10-03 13:14:13,602 - DEBUG - Setting up cgroups for run.
2023-10-03 13:14:13,602 - DEBUG - Available Cgroups: {'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn'), 'io': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn'), 'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn'), 'pids': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn')}
2023-10-03 13:14:13,602 - DEBUG - Created cgroups {PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn')}.
2023-10-03 13:14:13,602 - DEBUG - Using additional environment {}.
2023-10-03 13:14:13,604 - DEBUG - Starting process.
2023-10-03 13:14:13,605 - DEBUG - Parent: child process of RunExecutor with PID 18724 started.
2023-10-03 13:14:13,605 - DEBUG - Child: child process of RunExecutor with PID 18724 started
2023-10-03 13:14:13,606 - DEBUG - Failed to make b'/tmp/BenchExec_run_itv8bmlc/mount/home/benchexec' a bind mount: [Errno 2] mount(b'/tmp/BenchExec_run_itv8bmlc/mount/home/benchexec', b'/tmp/BenchExec_run_itv8bmlc/mount/home/benchexec', None, 4096, None) failed: No such file or directory
2023-10-03 13:14:13,606 - DEBUG - Mounting '/' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/nix' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/nix/store' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/proc' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/sys' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/sys/kernel/security' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/sys/fs/cgroup' as read-only
2023-10-03 13:14:13,606 - DEBUG - Mounting '/sys/fs/pstore' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/sys/firmware/efi/efivars' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/sys/fs/bpf' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/sys/kernel/debug' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/sys/fs/fuse/connections' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/sys/kernel/config' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/dev' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/dev/pts' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/dev/shm' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/dev/mqueue' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/dev/hugepages' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/run' as hidden
2023-10-03 13:14:13,607 - DEBUG - [Errno 16] umount(b'/tmp/BenchExec_run_itv8bmlc/mount/run') failed: Device or resource busy
2023-10-03 13:14:13,607 - DEBUG - Mounting '/tmp' as hidden
2023-10-03 13:14:13,607 - DEBUG - Mounting '/home' as hidden
2023-10-03 13:14:13,607 - DEBUG - [Errno 16] umount(b'/tmp/BenchExec_run_itv8bmlc/mount/home') failed: Device or resource busy
2023-10-03 13:14:13,607 - DEBUG - Mounting '/boot' as read-only
2023-10-03 13:14:13,607 - DEBUG - Mounting '/tmp' as hidden
2023-10-03 13:14:13,607 - DEBUG - Mounting '/run' as hidden
2023-10-03 13:14:13,608 - DEBUG - Mounting '/home' as hidden
2023-10-03 13:14:13,608 - WARNING - Could not load libseccomp2, please install it for improved container isolation (libseccomp.so.2: cannot open shared object file: No such file or directory).
2023-10-03 13:14:13,609 - DEBUG - Parent: executing /nix/store/kxkdrxvc3da2dpsgikn8s2ml97h88m46-bash-interactive-5.2-p15/bin/bash in grand child with PID 18725 via child with PID 18724.
2023-10-03 13:14:13,610 - DEBUG - Available Cgroups: {'cpu': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht'), 'io': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht'), 'kill': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht'), 'memory': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht'), 'freeze': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht'), 'pids': PosixPath('/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/benchexec_O5DHPeGtQQ0.scope/benchmark_z9eptppn/delegate_hbj_mrht')}
2023-10-03 13:14:13,612 - DEBUG - Waiting for signals
2023-10-03 13:14:13,613 - DEBUG - Child: process /nix/store/kxkdrxvc3da2dpsgikn8s2ml97h88m46-bash-interactive-5.2-p15/bin/bash terminated with exit code 0.
2023-10-03 13:14:13,614 - DEBUG - energy measurement output:
2023-10-03 13:14:13,614 - DEBUG - energy measurement output: cpu_count=1
2023-10-03 13:14:13,614 - DEBUG - energy measurement output: duration_seconds=0.000017
2023-10-03 13:14:13,614 - DEBUG - 0 output files matched the patterns and were transferred.
2023-10-03 13:14:13,614 - DEBUG - Waiting for process /nix/store/kxkdrxvc3da2dpsgikn8s2ml97h88m46-bash-interactive-5.2-p15/bin/bash with pid 18724
2023-10-03 13:14:13,615 - DEBUG - Parent: child process of RunExecutor with PID 18724 terminated with return value 0.
2023-10-03 13:14:13,615 - DEBUG - Process terminated, exit code 0.
2023-10-03 13:14:13,615 - DEBUG - Getting cgroup measurements.
2023-10-03 13:14:13,716 - DEBUG - Resource usage of run: walltime=0.002302315000179078, cputime=0.0026539999999999997, cgroup-cputime=0.002127, memory=835584
2023-10-03 13:14:13,716 - DEBUG - Cleaning up cgroups.
2023-10-03 13:14:13,716 - DEBUG - Cleaning up temporary directory /tmp/BenchExec_run_itv8bmlc.
2023-10-03 13:14:13,717 - DEBUG - Size of logfile 'output.log' is 183 bytes, size limit disabled.
starttime=2023-10-03T13:14:13.611688+02:00
returnvalue=0
walltime=0.002302315000179078s
cputime=0.002127s
memory=835584B
blkio-read=0B
blkio-write=0B
pressure-cpu-some=0.000002s
pressure-io-some=0s
pressure-memory-some=0s
</pre>
</details>

The state of the system is as follows:

<details><summary><code>cat /proc/cgroups</code></summary>
<pre>
#subsys_name    hierarchy   num_cgroups enabled
cpuset  0   140 1
cpu 0   140 1
cpuacct 0   140 1
blkio   0   140 1
memory  0   140 1
devices 0   140 1
freezer 0   140 1
net_cls 0   140 1
perf_event  0   140 1
net_prio    0   140 1
hugetlb 0   140 1
pids    0   140 1
rdma    0   140 1
misc    0   140 1
debug   0   140 1
</pre>
</details>

<details><summary><code>grep -E 'cgroup|overlay' /proc/filesystems</code></summary>
<pre>
nodev   cgroup
nodev   cgroup2
nodev   overlay
</pre>
</details>

<details><summary><code>gunzip --stdout /proc/config.gz | grep -E 'CONFIG_(OVERLAY_FS|USER_NS)='</code></summary>
<pre>
CONFIG_USER_NS=y
CONFIG_OVERLAY_FS=m
</pre>
</details>

<details><summary><code>lsmod | grep -E 'cpuid|msr|overlay'</code></summary>
<pre>
overlay               163840  0
intel_rapl_msr         20480  0
intel_rapl_common      28672  1 intel_rapl_msr
msr                    16384  0
cpuid                  16384  0
</pre>
</details>

<details><summary><code>uname -a</code></summary>
<pre>
uname -a
Linux [...] 6.1.44 #1-NixOS SMP PREEMPT_DYNAMIC Tue Aug  8 18:03:51 UTC 2023 x86_64 GNU/Linux
</pre>
</details>

PhilippWendler:
> I am trying to package BenchExec for NixOS (see #920). I decided to go for the `cgroupsv2` branch directly, and would like to report.

Thanks a lot!

> `runexec --debug --no-container echo Test` complains that "Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available." Note that cpuset is listed in `/proc/cgroups` (see below) and I am not sure what else I have to configure here.

This sounds like the kernel problem from 6cbd9fc5722cf763d141fbb5d1449f194d481b67. If you force `runexec` to require `cpuset`, for example with `--cores 0`, you should see a better error message. Maybe I should also improve the warning that is printed if `cpuset` is missing but not required.

> `containerexec --debug bash` complains with "Failed to configure container: [Errno 22] Creating overlay mount for `/` failed: Invalid argument. Please use other directory modes, for example `--read-only-dir /`."

Yes, this is a well-known kernel regression: #776 

> Fiddling around with the arguments a bit lead me to:
> `runexec --debug --read-only-dir / --hidden-dir /home --dir /tmp $(readlink -f $(which bash)) -- -c "echo Test"` which still complains, but I am not sure how bad the situation is. Maybe you can tell me?

`--dir` should be optional, though, right? Except if you are currently in a hidden directory, then you have to specify it, of course. Or, if you need the current directory, you can use an overlayfs for it.

I seen only one remaining warning, which is the one discussed above.

> Another issue seems to be that my `/run/wrappers/bin/cpu-energy-meter` is not found because `/run` is mounted as hidden. Note that it was found with `--no-container`. How to best resolve that?

The last log of `runexec` shows that it successfully used `cpu-energy-meter`.
Do you really have binaries in `/run`? If you need them, you can make them un-hidden with for example `--overlay-dir /run/wrappers/bin`, or we can add such logic to BenchExec if this is some NixOS rule.

lorenzleutgeb:
> > `runexec --debug --no-container echo Test` complains that "Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available." Note that cpuset is listed in `/proc/cgroups` (see below) and I am not sure what else I have to configure here.
> 
> This sounds like the kernel problem from [6cbd9fc](https://github.com/sosy-lab/benchexec/commit/6cbd9fc5722cf763d141fbb5d1449f194d481b67). If you force `runexec` to require `cpuset`, for example with `--cores 0`, you should see a better error message. Maybe I should also improve the warning that is printed if `cpuset` is missing but not required.
>

Yes, you're right. I'd prefer the workaround to be printed even for the warning, but that's a matter of taste. Also, I had to run a few commands (see below) in alternation with `runexec`. It would be helpful to print them all at once, but I don't know whether that's possible:

```shell
echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/cgroup.subtree_control
echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.subtree_control
echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/cgroup.subtree_control
echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/cgroup.subtree_control
```

Now the warning is gone! And it [looks like this is fixed in kernel 6.6](https://github.com/torvalds/linux/commit/0a67b847e1f06a70a7b560b69a06b3c78d3e72f0).

> > `containerexec --debug bash` complains with "Failed to configure container: [Errno 22] Creating overlay mount for `/` failed: Invalid argument. Please use other directory modes, for example `--read-only-dir /`."
> 
> Yes, this is a well-known kernel regression: #776
> 
> > Fiddling around with the arguments a bit lead me to:
> > `runexec --debug --read-only-dir / --hidden-dir /home --dir /tmp $(readlink -f $(which bash)) -- -c "echo Test"` which still complains, but I am not sure how bad the situation is. Maybe you can tell me?
> 
> `--dir` should be optional, though, right? Except if you are currently in a hidden directory, then you have to specify it, of course. Or, if you need the current directory, you can use an overlayfs for it.

I invoked `runexec` in a subdirectory of `/home`, so when I added `--hidden-dir /home` it complained that the directory did not exist. That's why I added `--dir /tmp`.

> I seen only one remaining warning, which is the one discussed above.
> 
> > Another issue seems to be that my `/run/wrappers/bin/cpu-energy-meter` is not found because `/run` is mounted as hidden. Note that it was found with `--no-container`. How to best resolve that?
> 
> The last log of `runexec` shows that it successfully used `cpu-energy-meter`.

Hmm yeah, sorry. I was confused.

> Do you really have binaries in `/run`?

Yes. Some background: In NixOS, the `cpu-energy-meter` binary is somewhere in `/nix/store/*-cpu-energy-meter-*/bin/cpu-energy-meter` and has no capabilities (for sandboxing builds, I believe). So there's another binary at `/run/wrappers/bin/cpu-energy-meter` configured via [`security.wrappers.cpu-energy-meter`](https://search.nixos.org/options?query=security.wrappers.<name>) that *does* have the `CAP_SYS_RAWIO` and will execute `/nix/store/*-cpu-energy-meter-*/bin/cpu-energy-meter`.

So, it seems that all warnings are resolved, CPU Energy Meter, LXCFS, and libseccomp are available. One thing that NixOS reviewers will ask is why I want to package an unreleased version. Are there any plans on releasing BenchExec with support for cgroups2? Do you need any more data or reports?

PhilippWendler:
> > > `runexec --debug --no-container echo Test` complains that "Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available." Note that cpuset is listed in `/proc/cgroups` (see below) and I am not sure what else I have to configure here.
> > 
> > 
> > This sounds like the kernel problem from [6cbd9fc](https://github.com/sosy-lab/benchexec/commit/6cbd9fc5722cf763d141fbb5d1449f194d481b67). If you force `runexec` to require `cpuset`, for example with `--cores 0`, you should see a better error message. Maybe I should also improve the warning that is printed if `cpuset` is missing but not required.
> 
> Yes, you're right. I'd prefer the workaround to be printed even for the warning, but that's a matter of taste. Also, I had to run a few commands (see below) in alternation with `runexec`. It would be helpful to print them all at once, but I don't know whether that's possible:
> 
> ```shell
> echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/cgroup.subtree_control
> echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.subtree_control
> echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/cgroup.subtree_control
> echo +cpuset | sudo tee /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/benchexec.slice/cgroup.subtree_control
> ```

Interesting. In my tests I think that systemd did that automatically once the initial command was run. But as long as BenchExec at least prints the next step each time it is run, I guess I will keep it that way because detecting what exactly to run upfront could be difficult.

> Now the warning is gone! And it [looks like this is fixed in kernel 6.6](https://github.com/torvalds/linux/commit/0a67b847e1f06a70a7b560b69a06b3c78d3e72f0).

Ah, nice! It is really hard to keep track of patches floating around in Linux land. I guess I can add a recommendation for Linux 6.6.

> > `--dir` should be optional, though, right? Except if you are currently in a hidden directory, then you have to specify it, of course. Or, if you need the current directory, you can use an overlayfs for it.
> 
> I invoked `runexec` in a subdirectory of `/home`, so when I added `--hidden-dir /home`

Which btw. is not strictly necessary, but does improve how the container's system config looks.

> it complained that the directory did not exist. That's why I added `--dir /tmp`.

Right. The typical alternative would be to use something like `--overlay-dir .` or `--overlay-dir ~`.

> > Do you really have binaries in `/run`?
> 
> Yes. Some background: In NixOS, the `cpu-energy-meter` binary is somewhere in `/nix/store/*-cpu-energy-meter-*/bin/cpu-energy-meter` and has no capabilities (for sandboxing builds, I believe). So there's another binary at `/run/wrappers/bin/cpu-energy-meter` configured via [`security.wrappers.cpu-energy-meter`](https://search.nixos.org/options?query=security.wrappers.%3Cname%3E) that _does_ have the `CAP_SYS_RAWIO` and will execute `/nix/store/*-cpu-energy-meter-*/bin/cpu-energy-meter`.

Ah, thanks for the explanation.

> > [...] or we can add such logic to BenchExec if this is some NixOS rule.
> 
> I don't think that's necessary.

Ok, as you prefer. Maybe add `--read-only-dir /run/wrappers/bin` to the package documentation or so?

> So, it seems that all warnings are resolved, CPU Energy Meter, LXCFS, and libseccomp are available. One thing that NixOS reviewers will ask is why I want to package an unreleased version. Are there any plans on releasing BenchExec with support for cgroups2? Do you need any more data or reports?

The most important reason was that I didn't get any feedback from others so far, so I wanted to do more internal tests. But thanks to you, I think I can just go ahead and release a version with cgroupsv2 and tell people to still treat it as somewhat experimental in the release notes. Ah, and I have to write more docs... So I hope I can do a release this week (but then without the polishings discussed in the last comments).

lorenzleutgeb:
> [...] I think I can just go ahead and release a version with cgroupsv2 and tell people to still treat it as somewhat experimental in the release notes. Ah, and I have to write more docs... So I hope I can do a release this week (but then without the polishings discussed in the last comments).

OK, that's great news. I am not in a rush, so please take your time and feel free to polish things up before releasing the coming weeks. I'll wait.

PhilippWendler:
> > > `runexec --debug --no-container echo Test` complains that "Cgroup subsystem cpuset is not available. Please make sure it is supported by your kernel and available." Note that cpuset is listed in `/proc/cgroups` (see below) and I am not sure what else I have to configure here.
> > 
> > 
> > This sounds like the kernel problem from [6cbd9fc](https://github.com/sosy-lab/benchexec/commit/6cbd9fc5722cf763d141fbb5d1449f194d481b67). If you force `runexec` to require `cpuset`, for example with `--cores 0`, you should see a better error message. Maybe I should also improve the warning that is printed if `cpuset` is missing but not required.
> 
> Yes, you're right. I'd prefer the workaround to be printed even for the warning, but that's a matter of taste.

I thought about that more now, and decided to completely remove the warning about missing `cpuset` if it is not required. If there is a core limit, we still print the error message with the workaround, and `benchexec.check_cgroups` now does it as well.

lorenzleutgeb:
> @lorenzleutgeb You might want to add this to the Nix package as well.

Instead of modifying `user@`, I believe it should be possible to only modify [`user@UID`](https://www.freedesktop.org/software/systemd/man/254/user@.service.html) for the UIDs of the users wanting to use BenchExec, right? I'd like to minimize the impact of BenchExec on the overall system, that's why I would prefer localizing the change for individual users. I'll test in a VM when I have time.

PhilippWendler:
Yes, this is correct. I decided to not do this in order to not overcomplicate things, because for example when installing a Debian package it is not easy to answer the question "which users should be allowed to do this" and we cannot even use a regular group or a single file, but have to do this individually for every single allowed user.

Furthermore, delegating cgroups v2 should be safe in principle (even Podman [considers](https://github.com/containers/podman/discussions/14322) enabling this by default for containers). Also cf. https://github.com/systemd/systemd/commit/b8df7f8629cb310beac982a4779b27eabe5362c6.

But if Nix allows you to do this more easily, you can do so.

lorenzleutgeb:
> But if Nix allows you to do this more easily, you can do so.

Yes. The following expression with the free variable `username`, will set `Delegate=yes` appropriately. The module I wrote for BenchExec can then take a list of usernames or UIDs to be configured.

```nix
systemd.slices.user-${builtins.toString config.users.users.${username}.uid}.serviceConfig.Delegate = "yes";
```

PR #793, opened 2021-12-03
Fix result files folder of benchmarkclient execut.

MartinSpiessl:
We want to put the result files into the same exact folder
as benchexec would do. We have all the necessary information
here so no need to hack together this filename manually, we
can just reuse run.result_files_folder. This will also ensure
that this will work if that folder should ever be named differently
by BenchExec.

### DIFF
--- a/contrib/aws/awsexecutor.py
+++ b/contrib/aws/awsexecutor.py
@@ -515,9 +515,7 @@ def handleCloudResults(benchmark, output_handler, start_time, end_time):
             raw_path = run.log_file[: -len(".log")]
             dirname, filename = os.path.split(raw_path)
             aws_files_directory = raw_path + ".files"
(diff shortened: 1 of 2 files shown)


PhilippWendler (commented):
Did not test, but looks as it always should have been like this.

The same code has been copy-pasted to `contrib/aws/awsexecutor.py`, would you mind copying the fix as well? For consistency it would be nice to have this in the same PR or even commit.

PhilippWendler:
FTR: Of course, in the long term the VerifierCloud should just put the files in the correct place. This whole place was just a hack to make it consistent while the VerifierCloud still uses a different directory structure (cf. [CPAchecker commit](https://gitlab.com/sosy-lab/software/cpachecker/-/commit/9255af0fc957cf40bd439481d6f5e7f43ce14b30).

PR #798, opened 2022-01-11
Update result matching for frama-c-sv.

SvenUmbricht:
We recently changed the output format of frama-c-sv which causes
Benchexec to report all new run results as error.
This commit adjusts result determination to match the new ouput
format.

### DIFF
--- a/benchexec/tools/frama-c-sv.py
+++ b/benchexec/tools/frama-c-sv.py
@@ -42,5 +42,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         lastline = run.output[-1]
         if lastline.startswith("INFO\tRESULT\t"):
             return lastline.split("\t", maxsplit=3)[-1]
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Changing this as suggested would mean that one no longer can use recent BenchExec versions together with older frama-c-sv versions (e.g., for reproducing previous results). Would it be possible to adjust the matching in order to support both versions?

In principle it could be done by determining the frama-c-sv version and then using a different matching code per version, but in this case I guess it is easier to just make the matching more generic and accept both `:` and `\t`? One could for example just keep the old `if` and add the new `if`.

SvenUmbricht:
That's a very good point. And indeed, using both the old and the new `if` would work because the output from any version of frama-c so far can fulfill at most one of the conditions, i.e. there is no ambiguity as to which `maxsplit` to use.

PR #803, opened 2022-01-14
Add additional column for witnesslint

SvenUmbricht:
As discussed in **https://github.com/sosy-lab/sv-witnesses/issues/55**, the witness type of the checked witness should be extracted into a new column.

@PhilippWendler I added a function to extract the witness type from the linter output, but how to create the new column? Looking at how it is done for CPAchecker, it seems that the columns have to be declared in the benchmark xml. Is it enough then to just implement `get_value_for_output` here?

### DIFF
--- a/benchexec/tools/witnesslint.py
+++ b/benchexec/tools/witnesslint.py
@@ -27,6 +27,12 @@ class Tool(benchexec.tools.template.BaseTool2):
         version_string = self._version_from_tool(executable)
         return version_string.partition("version")[2].strip().split(" ")[0]
 
+    def get_witness_type(self, run):
+        for line in run.output:
+            if line.startswith("Witness Type:"):
+                return line.split(" ", maxsplit=2)[-1]
+        return None
+
     def determine_result(self, run):
         exit_code = run.exit_code.value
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Yes, the crucial point is that the user has to request this column with a `<column>` tag (either in the benchmark definition or the table definition) and that the tool-info module has to implement `get_value_for_output`.

When the user defines such a tag, they have to provide a string (we call it pattern), and this pattern is passed to `get_value_for_output` (BenchExec does not actually do anything itself with this pattern and it does not have to follow a specific pattern language). The tool-info module has to use this string to decide what value it returns. For example, you could return the witness type if the user requests `witness_type`. Or you add a more general feature where the user can specify something like `witness.foo` and you return the `foo` field of the witness. Much is possible here, BenchExec does not care. For CPAchecker, we let the user specify arbitrary strings and the tool-info module searches for them in the log and returns the value printed after it, if it finds one. This strategy combined with https://github.com/sosy-lab/sv-witnesses/pull/56 would for example allow the user to request `Witness File`, `Producer`, etc. without having to implement each of them in the tool-info module. But you might also want to allow different patterns, whatever suits you.

PhilippWendler (approved):

PR #811, opened 2022-02-26
Add PDF icon

dbeyer:
I found it difficult to find our paper that describes BenchExec on the pages. So I added a PDF icon, in order to emphasize that there is a PDF document available.

@PhilippWendler What do you think?


PhilippWendler:
I prefer to not have the logo there. It breaks the line hight and spacing and thus really looks out of place. And even if the size is adjusted such that it looks appropriate on my screen, it will never look fine on all systems depending on the choice of the font size etc.

Furthermore, when there is a link with a PDF icon, I expect it to point to a PDF, but this is not the case. These links just point to a regular webpage (where you can download a PDF). But if one right-clicks and uses "Save link target" to download the PDF, it won't work as expected.

And it clutters the `doc/images` directory that contains images about BenchExec with unrelated stuff.

So all in all the benefit does not seem worth it.

PhilippWendler:
It is also inconsistent because we also have other links that point to PDFs (even real direct links to PDFs, not indirect ones like here) that do not get the PDF icon in this PR. But adding a PDF icon to every PDF link would add even more clutter and distraction.

PR #813, opened 2022-03-01
Adding an option to use a fresh cache for ivy

Sowasvonbot:
When you use vcloud-benchmark.py in multiple threads at the same time and ivy has to download jars, it produces some errors. With the command line option --fresh-ivy-cache, every execution of vcloud-benchmark.py uses a new cache folder for ivy.

### DIFF
--- a/contrib/vcloud-benchmark.py
+++ b/contrib/vcloud-benchmark.py
@@ -51,9 +51,8 @@ def download_required_jars(config):
     # Provide temporary directory
     if config.freshIvyCache:
         temp_dir = tempfile.mkdtemp()
-        print(temp_dir)
         cmd += ["-cache", str(temp_dir)]
-
+        logging.debug(f"Using fresh ivy cache: {temp_dir}")
     try:
         # install vcloud jar and dependencies
         subprocess.run(
--- a/contrib/vcloud-benchmark.py
+++ b/contrib/vcloud-benchmark.py
@@ -10,6 +10,8 @@
 import logging
 import os
 import sys
+import shutil
+import tempfile
 import urllib.request
 import subprocess
 
@@ -46,12 +48,22 @@ def download_required_jars(config):
         cmd += ["-warn"]
     cmd += ["-retrieve", "lib/vcloud-jars/[artifact](-[classifier]).[ext]"]
 
-    # install vcloud jar and dependencies
-    subprocess.run(
-        cmd,
-        cwd=_ROOT_DIR,
-        shell=vcloudutil.is_windows(),  # noqa: S602
-    )
+    # Provide temporary directory
+    if config.freshIvyCache:
+        temp_dir = tempfile.mkdtemp()
+        print(temp_dir)
+        cmd += ["-cache", str(temp_dir)]
+
+    try:
+        # install vcloud jar and dependencies
+        subprocess.run(
+            cmd,
+            cwd=_ROOT_DIR,
+            shell=vcloudutil.is_windows(),  # noqa: S602
+        )
+    finally:
+        if "temp_dir" in locals():
+            shutil.rmtree(temp_dir)
 
 
 class VcloudBenchmark(VcloudBenchmarkBase):
--- a/contrib/vcloud/vcloudbenchmarkbase.py
+++ b/contrib/vcloud/vcloudbenchmarkbase.py
@@ -85,6 +85,12 @@ class VcloudBenchmarkBase(benchexec.benchexec.BenchExec):
             type=str,
             help="Specify files or paths that shall also be transferred and be made available to the run in the cloud.",
         )
+        vcloud_args.add_argument(
+            "--fresh-ivy-cache",
+            dest="freshIvyCache",
+            action="store_true",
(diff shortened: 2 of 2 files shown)


contrib/vcloud-benchmark.py
```diff
@@ -46,12 +48,21 @@ def download_required_jars(config):
...
+        logging.debug(f"Using fresh ivy cache: {temp_dir}")
+    try:
+        # install vcloud jar and dependencies
+        subprocess.run(
+            cmd,
+            cwd=_ROOT_DIR,
+            shell=vcloudutil.is_windows(),  # noqa: S602
+        )
+    finally:
+        if "temp_dir" in locals():
```
PhilippWendler, line 64: This is quite ugly (and error-prone when renaming the variable). Ideally we would use `with`, alternatively at least use a regular variable access.

contrib/vcloud-benchmark.py
```diff
@@ -46,12 +48,21 @@ def download_required_jars(config):
...
+            shutil.rmtree(temp_dir)
```
PhilippWendler, line 65: Better use `tempfile.TemporaryDirectory` and its `cleanup()` function instead of manual implementation.

contrib/vcloud-benchmark.py
```diff
@@ -46,12 +48,21 @@ def download_required_jars(config):
...
-    subprocess.run(
-        cmd,
-        cwd=_ROOT_DIR,
-        shell=vcloudutil.is_windows(),  # noqa: S602
-    )
+    # Provide temporary directory
+    if config.freshIvyCache:
+        temp_dir = tempfile.mkdtemp()
+        cmd += ["-cache", str(temp_dir)]
+        logging.debug(f"Using fresh ivy cache: {temp_dir}")
```
PhilippWendler, line 55: Is that log message relevant? The directory will soon be deleted anyway.

Furthermore, for log statements please do not use f-strings, pass the parameters individually in order to allow lazy string formatting.

contrib/vcloud/vcloudbenchmarkbase.py
```diff
@@ -85,6 +85,12 @@ def add_vcloud_args(self, vcloud_args):
             type=str,
             help="Specify files or paths that shall also be transferred and be made available to the run in the cloud.",
         )
+        vcloud_args.add_argument(
+            "--fresh-ivy-cache",
```
PhilippWendler, line 89: The name and description of this option seem to be written from the point of view of the implementation, not of the user: The point is to not use the Ivy cache, the fact that Ivy still needs a temporary directory is an irrelevant detail for the user.

So the option should be named something like "no ivy cache" and the description should mention that this is useful for avoiding clashes due to concurrent access to the Ivy cache.

contrib/vcloud-benchmark.py
```diff
@@ -46,12 +48,21 @@ def download_required_jars(config):
...
-    # install vcloud jar and dependencies
-    subprocess.run(
-        cmd,
-        cwd=_ROOT_DIR,
-        shell=vcloudutil.is_windows(),  # noqa: S602
-    )
+    # Provide temporary directory
+    if config.freshIvyCache:
+        temp_dir = tempfile.mkdtemp()
+        cmd += ["-cache", str(temp_dir)]
```
PhilippWendler, line 54: Why `str()` here?

Furthermore, it is nicer to indicate the purpose of the directory in case that cleanup does not work, so `prefix` or `suffix` should be set. Otherwise users wonder where these directories come from.

PhilippWendler (changes requested):

contrib/vcloud-benchmark.py
Sowasvonbot: Changed it to use `TemporaryDirectory` and the deletion is now done with its `cleanup()`

contrib/vcloud-benchmark.py
Sowasvonbot: Removed the log message

contrib/vcloud-benchmark.py
Sowasvonbot: Using a suffix now.
I used `str()`, because I need a `str` or `Path-like object` at this point. Switched to `TemporaryDirectory.name`

contrib/vcloud-benchmark.py
Sowasvonbot: Problem with `with` is the conditional `ContextManager`. Python 3.7 introduced the [nullcontext](https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext), which would be helpful here, but afaik 3.6 is the minimum python version for benchexec.

Using regular variable access now.

contrib/vcloud-benchmark.py
Sowasvonbot: Didn't think about the lazy loading, thanks for the reminder

contrib/vcloud/vcloudbenchmarkbase.py
Sowasvonbot: Changed the parameter to `no-ivy-cache`. Also changed the help text. I hope it's now more understandable

PhilippWendler (approved):

Issue #242, opened 2017-05-13
Feature request: LaTeX output


mdangl:
Feature request:

BenchExec already has some cool output in CSV and HTML format.

As a researcher who wants to use BenchExec to enhance their paper with reproducible experiments,
I would like to directly include the output of my BenchExec results into my papers using LaTeX macros.

For example, I would like to be able to ``\include`` a file of LaTeX macros,
each of which represents one of the summary cells of the tables produced by the table generator.
There could even be different macros for different units, e.g.
a macro for time in seconds and another one for time in hours.
If there are different categories, there should be macros for individual categories and for overall.

Of course, there would have to be some smart conversion of identifiers to allowed LaTeX macro names;
LaTeX does not allow all characters for macros that are allowed for BenchExec run definitions.

PhilippWendler:
I have added a prototype for this in 566cd8a. It outputs only statistics for number of results and CPU time, but for these, all values from the table footers are available. There are total values, values for each category (correct/incorrect/error), and for each tuple of category and status.

Conversion into different units should simply be handled inside TeX, because we cannot now upfront what units are desired, and users might want to use the same value with different units.
Macros for individual categories are not directly possible because the result XML files do not contain enough information, but users can easily work around this by running the script on each of the individual category result files that BenchExec produces.

The conversion to LaTeX macro names is easy, we can simply remove all non-letter characters and convert to camel-case to make the name easier to read.

What do you think of this?

Further open questions:
- Integrate into table-generator or keep separate?
- Can we wrap the numbers in some siunitx command automatically? The commands should stay usable in tables with `S` columns.
- Add unit to command name? (Would be easier if integrated in table-generator.)

mdangl:
Cool!

> Integrate into table-generator or keep separate?

Yes, I think that would be beneficial. As you already said, it would be easier to have the unit feature there.
Also, it would then probably be more likely that new outputs added to the tables are also available as macros consistently.

> Can we wrap the numbers in some siunitx command automatically? The commands should stay usable in tables with S columns.

We could probably, for each value, output two macros: One wrapped, one plain.

PhilippWendler:
>> Can we wrap the numbers in some siunitx command automatically? The commands should stay usable in tables with S columns.
> 
> We could probably, for each value, output two macros: One wrapped, one plain.

[...]

PR #816, opened 2022-03-08
Moving the texgen to the tablegenerator 

Sowasvonbot:
Moves the generation of the tex file from `statitistics-tex.py` to `table-generator`.

Related to https://github.com/sosy-lab/benchexec/issues/242#issuecomment-1018488682


PhilippWendler:
Yes, something like this is likely required. It is common to for example report results on all safe/unsafe programs separately.

I wouldn't be too fond of an option for this, though, given that it would only accept the TeX output, which could be confusing. Maybe just always output these statistics as well?

benchexec/tablegenerator/test_util.py
```diff
@@ -98,3 +98,31 @@ def test_print_decimal_float(self):
...
+            3: "III",
+            4: "IV",
+            9: "IX",
+            14: "XIV",
+            21: "XXI",
+            49: "XLIX",
+            99: "XCIX",
+            849: "DCCCXLIX",
+            3000: "MMM",
+            3333: "MMMCCCXXXIII",
```
PhilippWendler, line 113: Thanks for writing tests, but I think this test could get more values, right?
For example, all numbers <= 10, something like 999, etc.

benchexec/tablegenerator/__init__.py
```diff
@@ -60,7 +64,7 @@
 
 DEFAULT_OUTPUT_PATH = "results"
 
-TEMPLATE_FORMATS = ["html", "csv"]
+TEMPLATE_FORMATS = ["html", "csv", "tex"]
```
PhilippWendler, line 67: With this change, I think table-generator automatically produces all three formats, does it?

I am not sure whether the tex output is useful enough to have it as default and would leave it disabled. So we can have another list like `DEFAULT_TEMPLATE_FORMATS` and use it as default in `write_table`.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
         out.write("\n")
 
 
+def write_tex_command_table(
```
PhilippWendler, line 1389: Given that this part adds ~200 lines to an already long module, I guess a separate module is better.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+    out,
+    run_sets: List[RunSetResult],
+    stats: List[List[ColumnStatistics]],
+    **kwargs,
+):
+    bench_name_set = set()
```
PhilippWendler, line 1395: The abbreviation `bench_name` is so far never used in BenchExec. Can you please use a more common term like `benchmark_name`? (Also in all other occurrences of `bench_name`, please.)

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+    for benchmark in run_sets:
+        bench_name_formatted = LatexCommand.format_command_part(
+            benchmark.attributes.get("benchmarkname")
+        )
+        if bench_name_formatted in bench_name_set:
+            logging.error(
+                "Duplicated formatted benchmark name %s detected. Benchmark names must be unique for Latex"
```
PhilippWendler, line 1402: Hm, I think this would be a common situation. As we can see it does already complicate our tests if this is not supported.

First, the benchmark name alone does not need to be unique, does it? Only the combination of benchmark and run set need to be unique AFAIU.
Second, can't we simply add a suffix with a number if the names are identical?

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+            )
+            return
+        bench_name_set.add(bench_name_formatted)
+
+    header = textwrap.dedent(
+        """\
+        % The following definition defines a command for each value.
+        % The command name is the concatenation of the first six arguments.
+        % To override this definition, define \\StoreBenchExecResult with \\newcommand before including this file.
+        % Arguments: benchmark name, run-set name, category, status, column name, statistic, value
```
PhilippWendler, line 1414: I still need to think about whether we should add more arguments here, for cases where users overwrite `StoreBenchExecResult` with their own command and could make use of it.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+            logging.error(
+                "Duplicated formatted benchmark name %s detected. Benchmark names must be unique for Latex"
+                "\nSkipping writing to file %s"
+                % (bench_name_formatted, kwargs["title"] + ".tex")
+            )
+            return
+        bench_name_set.add(bench_name_formatted)
+
+    header = textwrap.dedent(
+        """\
```
PhilippWendler, line 1410: A raw string could be useful here such that `\` does not need to be duplicated?

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+        return "\\StoreBenchExecResult{%s}{%s}{%s}{%s}{%s}{%s}" % (
+            self.bench_name,
+            self.runset_name,
+            self.column_title,
+            self.column_category,
+            self.column_subcategory,
+            self.stat_type,
+        )
+
+    def __get_command_formatted(self, value: str) -> str:
```
PhilippWendler, line 1552: AFAIK double underscore prefixes are discouraged in Python.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+        """Formats the command with all parts and appends the value
+
+        To use a custom format for the value, for example
+            \\StoreBenchExecResult{some}{stuff}...{last_name_part}{\\textbf{value}}
+        format the value and give it to this function
+        """
+        if not value:
+            logging.warning(
+                "Trying to print latex command without value! Using 0 as value for command:\n %s"
```
PhilippWendler, line 1561: We must not dump wrong values. If there is no value to print, then we cannot invent the number `0` (why `0` and not `42`?). So please make it print an empty value.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+        used_column_titles = set()
+        for column, column_stats in zip(run_set.columns, stat_list):
+            column_title = (
+                column.display_title if column.display_title else column.title
+            )
+            if column_title in used_column_titles:
+                logging.warning(
+                    "Detected already used %s! "
+                    "Columns should be unique, please consider changing the name or displayTitle of this column \n"
+                    "Skipping column %s for now" % (column_title, column_title)
```
PhilippWendler, line 1480: I don't think we should silently skip columns. This will just mean that in practice users will overlook the warning and just use the commands, and they get the values of the first of these columns without knowing about it.

Either this should be a real error, or we just print identical column names, or we make the columns unique by attaching a suffix number.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+        self.value = decimal.Decimal(value)
+        self.value *= scale_factor
+        return self
+
+    def to_latex_raw(self) -> str:
+        """Prints latex command with raw value (e.g. only number, no additional latex command)."""
+        return self.__get_command_formatted(util.print_decimal(self.value))
+
+    def __repr__(self):
+        return "\\StoreBenchExecResult{%s}{%s}{%s}{%s}{%s}{%s}" % (
```
PhilippWendler, line 1543: Typically we prefer to use f-strings in BenchExec.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+
+            used_column_titles.add(column_title)
+            current_command.set_command_part("column_title", column_title)
+
+            column_statistic_to_latex_command(
+                current_command, column_stats, latex_commands, **column.__dict__
+            )
+
+    out.write(header)
+    out.write("\n".join(command.to_latex_raw() for command in latex_commands))
```
PhilippWendler, line 1492: I would say that the trailing `%` on each line should be produced here, because it is not part of the command, we produce it because we want to add line breaks (and we add them here).

Furthermore, this probably creates a file with no trailing line break, right? I think it would be better to add one.
So
```python
for command in latex_commands:
    out.write(command.to_latex_raw())
    out.write("%\n")
```

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+        Args:
+            value: The new command value
+            scale_factor: Scaling factor for the new value
+            value_data: Remaining unused value_data
+
+        Returns:
+            This LatexCommand
+        """
+        self.value = decimal.Decimal(value)
+        self.value *= scale_factor
```
PhilippWendler, line 1535: Did you check whether this makes sense? Does the file really contain the same values as the HTML tables? I would say that the statistics values all have the scale factor already applied and now would be wrong. But it needs to be checked.

benchexec/tablegenerator/__init__.py
```diff
@@ -1382,10 +1386,201 @@ def write_head_line(
...
+                % self
+            )
+            value = "0"
+        return str(self) + "{%s}%%" % value
+
+    @staticmethod
+    def format_command_part(name: str) -> str:
+        name = re.sub("[0-9]+", util.number_to_roman_string, name)
+
+        name = re.sub("[^a-zA-Z]", "-", name)
```
PhilippWendler, line 1571: I think `re.split` could be used to avoid having to use the workaround of going via `-`.

PhilippWendler (changes requested):
Thanks! I have some comments, but most are minor adjustments.

benchexec/tablegenerator/util.py
```diff
@@ -279,6 +280,94 @@ def normalize_line_endings(text):
...
+    Args:
+        number: An integer, string, or Match object from the re module
+
+    Returns:
+        A string which represents the given number in roman number format.
+    """
+
+    # Can't check for Match, because python 3.6 doesn't have re.Match
+    if hasattr(number, "group"):
+        number = number.group()
```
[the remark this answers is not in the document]
Sowasvonbot: Changed in f831296c77f05521293aee8cf0b7cb5c0c9193df

benchexec/tablegenerator/util.py
```diff
@@ -279,6 +280,94 @@ def normalize_line_endings(text):
...
+        100: "C",
+        50: "L",
+        10: "X",
+        5: "V",
+        1: "I",
+    }
+
+    output_string = ""
+
+    # Subtracting the highest number
```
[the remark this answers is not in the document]
Sowasvonbot: Yep, that's better:
96588ae41d7f4c76decc53c1c9df90c0edee379e

benchexec/tablegenerator/util.py
```diff
@@ -279,6 +280,94 @@ def normalize_line_endings(text):
...
+    max_number = max(roman_numbers)
+    while number >= max_number:
+        output_string += roman_numbers[max_number]
+        number -= max_number
+
```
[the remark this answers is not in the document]
Sowasvonbot: I tried to explain it with an example (eeac5dfa1c9d7691eeb85a023a8bada9c6835095)

benchexec/tablegenerator/__init__.py
Sowasvonbot: Moved it in 903c860d6150d11d8eee6523b7377a86ab8954ee to `textable.py` (named "inspired" by `htmltable.py`)

benchexec/tablegenerator/__init__.py
Sowasvonbot: In this case, f-Strings are quite ugly. A f-String would look like:

`\\StoreBenchExecResult{{{variable}}}{{{variable2}}}...`

benchexec/tablegenerator/__init__.py
Sowasvonbot: Implemented in 42b03db25526167ff1cd679df4b8ff39d3e94e7c

benchexec/tablegenerator/__init__.py
Sowasvonbot: Changed in 2a9c8c7ef71d28e1c6c5d18aaca2e397020b80ad

benchexec/tablegenerator/__init__.py
Sowasvonbot: Changed in d2f303215667a9d3972231e7a4bd4c711c8fa874

Unfortunately I can't use `textwrap.dedent` with raw strings

benchexec/tablegenerator/__init__.py
Sowasvonbot: Of course, don't know why I missed the common naming.

a715231cf9a9859ef595ffb684fc2bbca9a944c6

benchexec/tablegenerator/__init__.py
Sowasvonbot: Adding a suffix now and printing a warning (in case the user wants to know, why there are numbers now)
9f66832bb7d3f82af68c55e5885cbf133245d3cb

benchexec/tablegenerator/__init__.py
Sowasvonbot: Changed in ce7c5a04214476457d2c1481a8067708e5485305

benchexec/tablegenerator/__init__.py
Sowasvonbot: Now adding suffix to benchmark names if duplication gets detected (5f69e1d5efccd4fb22978b250f1c81f315f7692c)

The second parameter in the command isn't the `run-set name` anymore. I have generated some latex files from my own benchexec runs and the `run-set name` was ridicolously long (About 200 chars). Therefore I choosed to use the `displayName` of the run, but forgot to change some doc (did this in commit 5f69e1d5efccd4fb22978b250f1c81f315f7692c now).

Please let me know, if this is okay. I first tried to limit the amount of chars in the `run-set name`, but this would be some arbitrary number (like 24, which I used for some time). I don't think, this is a good approach either. 

Another solution would be to choose the `displayName` if defined, otherwise choose `run-set name`

benchexec/tablegenerator/__init__.py
Sowasvonbot: Added the module constand in 54467c5996b5cc0ed8f1a8b48da2af5c6bc4dbfe

If you want to test the tex output, you have to `"tex"` in this list now. I am not sure if this is what you meant.

Sowasvonbot: This dows not work, because `tablegenerator` generates the tex files, if not explicitly said otherwise. Therefore the tests are failing, because more files get generated than expected

PhilippWendler: I want it such that `table-generator` does not produce the TeX output by default, only if you request it with `-f tex`. But the tests should create all formats and we want to check the expected results for all formats.

Sowasvonbot: Makes sense. Changed it in ef88080b81f8d237d422ee09573b7fe1da215588

I'm adding the `-f` arguments in `generate_tables_and_check_produced_files` now to the arguments, if none is present, to generate the tex files for the tests

PhilippWendler (changes requested):
A few points that are general or where I did not find the place that causes them:

- The statistics for all results are currently named "Total", but we renamed this in the HTML table to "all results" recently. So I think we should also use "All" here.
- We need to make the unit usable for users (if present)! I guess we could add another value named "Unit" in addition to "Min", "Max", "Avg" (because all these have the same as unit, except for "Stdev").
- The sum statistics value should also get a non-empty name, i.e., "Sum". But for cases where the value does not indicate a sum, but a count (i.e., the status column), it should be "Count" instead.
- I think the score should be dumped not as a kind of category, but as a kind of value, i.e. `{All}{}{Score}` instead of `{Score}{}{}`.
- After thinking more about it, I would prefer to not name the output format `tex` but something like `statistics-tex` or `tex-statistics`. The reason for this is that it does not contain the table itself, so for example the formats `csv` and `tex` have completely different content, not just a different format. And we might want to add "real" tex output later (with an actual table), so it is better to keep the format name `tex` free.
- For the same reason, how about using `.statistics.tex` as file-name suffix instead of `.tex`?
- And because of this I would also say that `textable.py` could be `texstatistics.py`.

Sorry for the number of change requests, but now when we move the functionality from some side script into proper BenchExec, we really need to think about what makes sense as output. Changing it later would create compatibility problems, so we should strive to get it right.

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+            column_title += str(used_column_titles[column_title])
+
+        used_column_titles[column_title] += 1
+        current_command.set_command_part("column_title", column_title)
+
+        for command in _column_statistic_to_latex_command(
+            current_command, column_stats, **column.__dict__
+        ):
+            yield command
```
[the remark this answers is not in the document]
Sowasvonbot: Yep, worked. Didn't know that :D
(87d627bc1784829d397fe0bfd41db553851293cd)

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+        return self
+
+    def to_latex_raw(self) -> str:
+        """Prints latex command with raw value (e.g. only number, no additional latex command)."""
+        return self._get_command_formatted(util.print_decimal(self.value))
+
+    def __repr__(self):
+        return "\\StoreBenchExecResult{%s}{%s}{%s}{%s}{%s}{%s}" % (
+            self.benchmark_name,
+            self.display_name,
```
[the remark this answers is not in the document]
Sowasvonbot: I had a benchmarkset with a combination of different runsets. The runset name in the table looked like this:
```
3-complete.[SV-COMP21_unreach-call.ConcurrencySafety-Main; SV-COMP21_unreach-call.ReachSafety-Arrays; SV-COMP21_unreach-call.ReachSafety-BitVectors; SV-COMP21_unreach-call.ReachSafety-Combinations; SV-COMP21_unreach-call.ReachSafety-ControlFlow; SV-COMP21_unreach-call.ReachSafety-ECA; SV-COMP21_unreach-call.ReachSafety-Floats; SV-COMP21_unreach-call.ReachSafety-Heap; SV-COMP21_unreach-call.ReachSafety-Loops; SV-COMP21_unreach-call.ReachSafety-ProductLines; SV-COMP21_unreach-call.ReachSafety-Recursive; SV-COMP21_unreach-call.ReachSafety-Sequentialized; SV-COMP21_unreach-call.ReachSafety-XCSP; SV-COMP21_unreach-call.SoftwareSystems-AWS-C-Common-ReachSafety; SV-COMP21_unreach-call.SoftwareSystems-DeviceDriversLinux64-ReachSafety; SV-COMP21_unreach-call.SoftwareSystems-DeviceDriversLinux64Large-ReachSafety; SV-COMP21_unreach-call.SoftwareSystems-uthash-ReachSafety; SV-COMP21_unreach-call]
```
In the html it was fine, I could ignore it. But in the tex file, it was horrible.
The table is available [here](https://www.cip.ifi.lmu.de/~kleinertt/evalution/finished_cloud_runs/3-complete.table.html)

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def write_tex_command_table(
+    out,
+    run_sets: List,
+    stats: List[List[ColumnStatistics]],
+    **kwargs,
+):
+    # Copy run_sets to modify the benchmarknames
+    run_sets = copy.deepcopy(run_sets)
```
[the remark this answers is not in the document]
Sowasvonbot: No need to copy, found another solution in 86ac91187f0f34ef4e0df99f39a4a79ea2ef5374

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # Check for duplicated benchmarkname + displayName
+    # Add suffix to benchmarkname if duplication is detected
```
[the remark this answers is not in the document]
Sowasvonbot: Implemented in 86ac91187f0f34ef4e0df99f39a4a79ea2ef5374

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+    benchmark_name_dict = defaultdict(int)
+    for benchmark in run_sets:
```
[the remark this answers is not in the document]
Sowasvonbot: Changed in 86ac91187f0f34ef4e0df99f39a4a79ea2ef5374

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+            benchmark.attributes["displayName"]
+        )
+
+        combined_command_name = benchmark_name_formatted + display_name_formatted
+
+        if combined_command_name in benchmark_name_dict:
+            logging.warning(
+                "Duplicated formatted benchmark name + displayName %s detected. "
+                "The combination of names must be unique for Latex"
+                "\nAdding suffix %s to benchmark name",
```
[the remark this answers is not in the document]
Sowasvonbot: Removed newlines in c8c062ad8c85c80b7f612b83ee7cb18dd51f4cfd

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+
+    # Saves used columns and their amount
+    used_column_titles = defaultdict(int)
+    for column, column_stats in zip(run_set.columns, stat_list):
+        column_title = column.display_title if column.display_title else column.title
+        if column_title in used_column_titles:
+            used_column_titles[column_title] += 1
+            logging.warning(
+                "Detected already used %s! "
```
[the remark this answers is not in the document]
Sowasvonbot: Changed in c8c062ad8c85c80b7f612b83ee7cb18dd51f4cfd

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+            value = ""
+        return str(self) + "{%s}" % value
+
+    @staticmethod
+    def format_command_part(name: str) -> str:
+        def regex_match_to_roman_number(match) -> str:
+            return util.number_to_roman_string(match.group())
+
+        name = re.sub("[0-9]+", regex_match_to_roman_number, name)
```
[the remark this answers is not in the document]
Sowasvonbot: Using lambda now: 86ac91187f0f34ef4e0df99f39a4a79ea2ef5374

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def _get_command_formatted(self, value: str) -> str:
+        """Formats the command with all parts and appends the value
+
+        To use a custom format for the value, for example
+            \\StoreBenchExecResult{some}{stuff}...{last_name_part}{\\textbf{value}}
+        format the value and give it to this function
+        """
+        if not value:
+            logging.warning(
```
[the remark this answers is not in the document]
Sowasvonbot: Removed it in b8d5e16312e4dd1f39f95e8bdae373b3099c17f9

benchexec/tablegenerator/textable.py
```diff
@@ -0,0 +1,219 @@
...
+        Args:
+            value: The new command value
+            scale_factor: Scaling factor for the new value
+            value_data: Remaining unused value_data
+
+        Returns:
+            This LatexCommand
+        """
+        self.value = decimal.Decimal(value)
+        self.value *= scale_factor
```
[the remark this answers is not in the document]
Sowasvonbot: Implemented in: 2c0780dffdd003ada6a390504b49731f11cc4d99, 9065daa7355d45d711a5c8e4e3d3da117be484d5

PhilippWendler:
> ```
> The statistics for all results are currently named "Total", but we renamed this in the HTML table to "all results" recently. 
> So I think we should also use "All" here.
> ```
> 
> Do you know, where the renaming happens? Maybe I can use the same code.

There is no code for this, we just [replaced](https://github.com/sosy-lab/benchexec/commit/5aab7d7c8de74aa22e178bddf56408037e2d9f96#diff-bfc1d7cc794bfb08a5c41fd16ba7cb75a3f62a9dd9dc2887119eb37f0171ccf8) the human-readable label that is visible in the table, but internally the key for this is still `total`.

> ```
> The sum statistics value should also get a non-empty name, i.e., "Sum". 
> But for cases where the value does not indicate a sum, but a count (i.e., the status column), it should be "Count" instead.
> ```
> 
> Atm always `Sum` is used. Is there a clever way to filter out `Counts`, e.g. to decide, which column wants `Count` instead of `Sum`?

We have [three cases](https://github.com/sosy-lab/benchexec/blob/7c240242155573bf885bf34fc35843c70fde83ff/benchexec/tablegenerator/statistics.py#L131-L140) regarding the column type here: text columns have no statistics, status columns have counts, and numeric columns have sums.

> ```
> I think the score should be dumped not as a kind of category, but as a kind of value, i.e. {All}{}{Score} instead of {Score}{}{}.
> ```
> 
> From `simple-table-with-scaling.table.statistics.tex`: `\StoreBenchExecResult{Test}{}{Status}{Score}{}{Sum}{0}%` Is this the correct format, except for `Sum`? I could remove the `stat_name`, if the current `column` is `Score`.

I think it should be `\StoreBenchExecResult{Mathsat}{}{Status}{All}{}{Score}{0}`. The point that I want to communicate is that this is the score for *all* results.

Btw.: GitHub does not support ` ```quote` unfortunately.

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+
+        name = re.split("[^a-zA-Z]", name)
+
+        name = "".join(util.cap_first_letter(word) for word in name)
+
+        return name
+
+
+class DuplicationHolder:
```
PhilippWendler, line 101: It is weird that `create_duplication_dictionary` returns a data structure that contains a field `used` that is not related at all to `create_duplication_dictionary` and only there because it happens to be useful for the one caller of `create_duplication_dictionary`.

And the field `duplication` seems redundant because these objects will only be used after being stored in a dict under the very same key.

The types of `DuplicationHolder` and `create_duplication_dictionary` force `pre_process` to return a `str`, but this seems unnecessary.

But anyway, I also think this is over-engineered. Inside `write_tex_command_table` there could simply exist two (local) dicts `name_count` and `current_name_index` or similarly. The former can probably be created in a one-liner using [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter). The latter could be a `defaultdict(int)` or so.

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,295 @@
...
+            # "v is None" instead of "if not v" used to allow number 0
+            if v is None:
+                continue
+            command.set_command_part("stat_type", k)
+            yield command.set_command_value(
+                column.format_value(value=v, format_target="csv")
+            )
+        if column.source_unit:
+            yield command.set_command_part("stat_type", "unit").set_command_value(
+                column.unit
```
PhilippWendler, line 294: Here there is a discrepancy between `column.source_unit` and `column.unit`?

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,295 @@
...
+            column_parts.append("")
+
+        # Some colum_categories use _ in their names, that's why the column_category is the
+        # whole split list except the last word
+        command.set_command_part(
+            "column_category",
+            "".join(
+                util.cap_first_letter(column_part) for column_part in column_parts[0:-1]
+            ),
+        ).set_command_part("column_subcategory", column_parts[-1])
```
PhilippWendler, line 282: Due to how the line breaking looks in Python style and due to the fact that here is never a possibility to chain more than two calls in one place, I think that the chaining here makes the code less readable than simply repeating `command` in a few more places.

The combination of chaining with `yield` is also not really nice because it puts the actions in the wrong order (`yield` happens last but is written first).

So please remove chaining for `set_command_part` and `set_command_value`.

PhilippWendler (changes requested):
Thanks for already improving this so much!

benchexec/tablegenerator/test_integration/expected/big-table.table.statistics.tex
```diff
@@ -0,0 +1,579 @@
...
+% Arguments: benchmark name, run-set name, category, status, column name, statistic, value
+\providecommand\StoreBenchExecResult[7]{\expandafter\newcommand\csname#1#2#3#4#5#6\endcsname{#7}}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Score}{}{Sum}{0}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Total}{}{Sum}{482}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Correct}{}{Sum}{338}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Correct}{True}{Sum}{196}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Correct}{False}{Sum}{142}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{Correct}{Unconfirmed}{Sum}{0}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{CorrectUnconfirmed}{True}{Sum}{0}%
+\StoreBenchExecResult{IntegrationPredicateAnalysisI}{}{Status}{CorrectUnconfirmed}{False}{Sum}{0}%
```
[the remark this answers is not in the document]
Sowasvonbot: Resolved in 7db8aad18b2494e84388e9a5971bb7f0f7430fb6

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: Implemented in bec6eb346df6fa1f9f880c3e33e94a525d90735b

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: Fixed in bec6eb346df6fa1f9f880c3e33e94a525d90735b

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: Rewritten in 23dd10982bd030937483d49364bf8c95ff8f011c

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,295 @@
+import copy
```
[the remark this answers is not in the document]
Sowasvonbot: Added in 310efc4952c2f878bf5359fc2a31b4a9350071ff

Is there no `reuse`-check?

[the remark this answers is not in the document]
PhilippWendler: Unfortunately, GitLab CI does not work for PRs from forks on GitHub. In the future, you can create branches in the BenchExec repo and work there.

benchexec/tablegenerator/test_integration/expected/big-table.table.statistics.tex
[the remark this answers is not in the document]
PhilippWendler: But the category "unconfirmed" still appears everywhere, right?

We need some logic like [this](https://github.com/sosy-lab/benchexec/blob/7c240242155573bf885bf34fc35843c70fde83ff/benchexec/tablegenerator/htmltable.py#L331-L334).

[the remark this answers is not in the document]
Sowasvonbot: Tried to address this in https://github.com/sosy-lab/benchexec/pull/816/commits/01863ca7cd9983b6cf021de09bfcdcaf6aa17e02

benchexec/tablegenerator/test_integration/expected/smt.diff.statistics.tex
```diff
@@ -0,0 +1,67 @@
...
+\StoreBenchExecResult{Mathsat}{}{Status}{Wrong}{}{Sum}{0}%
+\StoreBenchExecResult{Mathsat}{}{Status}{Wrong}{True}{Sum}{0}%
+\StoreBenchExecResult{Mathsat}{}{Status}{Wrong}{False}{Sum}{0}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Sum}{0.155786}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Min}{0.009091}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Max}{0.125825}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Avg}{0.05192866666666666666666666667}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Median}{0.02087}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Stdev}{0.05247340450382671923076363801}%
+\StoreBenchExecResult{Mathsat}{}{Cputime}{Total}{}{Unit}{s}%
```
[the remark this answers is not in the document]
PhilippWendler: Right, it matches what we have in the [HTML table](https://sosy-lab.github.io/benchexec/example-table/svcomp-simple-cbmc-cpachecker.table.html#/). Status is `0` because there are 0 results of the respective category, but other statistics are shown as `-` in order to show that this is the sum over the empty set instead of just a sum that happens to be 0.

For the LaTeX output, keeping the explicit status values with count 0 might be useful for some users that want to confirm that they did not overlook anything (they could add the equivalent to "count for category wrong is 0" to their LaTeX code, for example), so we'll keep it.

benchexec/tablegenerator/textable.py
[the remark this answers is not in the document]
Sowasvonbot: Forgot to answer this thread

I'm using `name` for the run-set name now (https://github.com/sosy-lab/benchexec/pull/816/commits/4b13d113c540d1421b68b2ff5446f3c7b68c79d1)

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,346 @@
...
+            "".join(util.cap_first_letter(column_part) for column_part in column_list),
+        )
+        command.set_command_part("status", status)
+
+        for k, v in stat_value.__dict__.items():
+            # "v is None" instead of "if not v" used to allow number 0
+            if v is None:
+                continue
+
+            if k in "sum":
```
PhilippWendler, line 331: This should probably use `==`?

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,346 @@
...
+            value: The new command value
+        Returns:
+            This LatexCommand
+        """
+        if value is None:
+            value = ""
+        self.value = str(value)
+        return self
+
+    def to_latex_raw(self) -> str:
```
PhilippWendler, line 77: The methods here seem overly complicated.

`to_latex_raw` is a one-line method that just delegates to another method on the same instance, and is called only in two places by code in the same class. `_get_command_formatted` takes a parameter, but is always called with the same value for the parameter, which is taken from a field of the same class. So I think both methods could be merged into a single parameter-less method.

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,346 @@
...
+        It will be converted to string.
+
+        Args:
+            value: The new command value
+        Returns:
+            This LatexCommand
+        """
+        if value is None:
+            value = ""
+        self.value = str(value)
```
PhilippWendler, line 74: This is somewhat confusing. Do we assume that the value is already a string, or not?

If yes: The body of the method can be replaced with `self.value = value`.
If not: The type signature of the method needs to be adjusted.

The doc string is also similarly confusing: If the value already needs to be formatted, why is it converted to string?

benchexec/tablegenerator/statisticstex.py
```diff
@@ -0,0 +1,346 @@
...
+    def _get_command_formatted(self, value: str) -> str:
+        """Formats the command with all parts and appends the value
+
+        To use a custom format for the value, for example
+            \\StoreBenchExecResult{some}{stuff}...{last_name_part}{\\textbf{value}}
+        format the value and give it to this function
+        """
+        if not value:
+            value = ""
+        return str(self) + "{%s}" % value
```
PhilippWendler, line 120: Same question about the type of `value` as for `set_command_value`.

PhilippWendler (commented):
Thanks a lot!

Output seems to look fine now, just some code questions.

PhilippWendler:
Please merge `main` into the branch in order to fix CI.

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: Changed the type signature and the doc string (https://github.com/sosy-lab/benchexec/pull/816/commits/eba35a09e07ee1424c286749bdcfa013d04fa538)

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: Was removed in https://github.com/sosy-lab/benchexec/pull/816/commits/23f0487e2b7ca76c961eedbb00143a631faf5689

benchexec/tablegenerator/statisticstex.py
Sowasvonbot: I don't know, why I was hesitant with equality, changed it in https://github.com/sosy-lab/benchexec/pull/816/commits/c5b8eb2bb4b6a6bcde64a446f9535ebed785aa87

contrib/statistics-tex.py
```diff
@@ -22,6 +22,11 @@
 
 sys.dont_write_bytecode = True  # prevent creation of .pyc files
 
+number_convert = {
```
PhilippWendler, line 25: Are the changes to the old script intentional? If not, don't bother, because we will delete it anyway. Just wanted to make sure that the changes should not go to the new code.

PR #817, opened 2022-03-10
Implement `randomizeEnvSize`

charmoniumQ:
- Implement `randomizeEnvSize`
- Test `randomizeEnvSize`
- FIx outputhandler when env contains UTF-8 chars. My prompt, for example, contains an ANSI escape sequence.

### DIFF
--- a/benchexec/runexecutor.py
+++ b/benchexec/runexecutor.py
@@ -12,7 +12,9 @@ import errno
 import logging
 import multiprocessing
 import os
+import random
 import signal
+import string
 import subprocess
 import sys
 import threading
@@ -515,7 +517,7 @@ class RunExecutor(containerexecutor.ContainerExecutor):
         if environments.get("keepEnv", None) is not None:
             run_environment = {}
         else:
-            run_environment = os.environ.copy()
+            run_environment = {key: val for key, val in os.environ.items()}
         for key in environments.get("keepEnv", {}).keys():
             if key in os.environ:
                 run_environment[key] = os.environ[key]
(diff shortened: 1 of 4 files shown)


benchexec/runexecutor.py
```diff
@@ -515,14 +517,32 @@ def _setup_environment(self, environments):
...
         for key in environments.get("keepEnv", {}).keys():
             if key in os.environ:
                 run_environment[key] = os.environ[key]
         for key, value in environments.get("newEnv", {}).items():
             run_environment[key] = value
         for key, value in environments.get("additionalEnv", {}).items():
             run_environment[key] = os.environ.get(key, "") + value
+        if environments.get("randomizeEnvSize", None) is not None:
+            # Caller should consider setting random.seed(...) for improved
+            # (but not complete) determinism in the runtime.
```
PhilippWendler, line 530: This comment with a note for callers seems a little bit out-of-place, because developers of callers won't see it here.

Also I don't really understand the "should" here: Isn't the whole point of this feature the non-determinism? Why should one want to fix a seed? And if this should indeed be done, why can't we do this for the caller? If we don't do this, callers will just forget it, and callers of the command-line `runexec` will not have the chance to do this anyway.

PhilippWendler (commented):
Thanks for your contribution!

I have a few comments and questions regarding the code.

And could you please split the feature and the bugfix into separate PRs? Right now it is difficult to review, because for example it seems the tests are mixed for both contributions. And I do not fully understand the XML problem yet, so we need to discuss it, which would be better done separately.

PhilippWendler:
@charmoniumQ Do you have the intention of carrying on with this?

charmoniumQ:
Hello, I'll reopen this when I have more time.

PR #825, opened 2022-04-13
Coveriteam tool module update

Sowasvonbot:
Fix for recent changes in Coveriteam.
The data model moved to the inputs

### DIFF
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -30,7 +30,7 @@ class Tool(coveriteam.Tool):
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
         if data_model_param and "data-model" not in options:
-            options += ["--input", "data-model", data_model_param]
+            options += ["--input", "data-model=", data_model_param]
 
         if task.property_file:
             options += ["--input", "specification_path=" + task.property_file]
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -30,7 +30,7 @@ class Tool(coveriteam.Tool):
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
         if data_model_param and not any(
-            [option.startswith("data_model=") for option in options]
+            re.match("data_model *=", option) for option in options
         ):
             options += ["--input", "data_model=" + data_model_param]
 
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -29,8 +29,8 @@ class Tool(coveriteam.Tool):
         data_model_param = get_data_model_from_task(
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
-        if data_model_param and "--data-model" not in options:
-            options += ["--data-model", data_model_param]
+        if data_model_param and "data-model" not in options:
+            options += ["--input", "data-model", data_model_param]
 
         if task.property_file:
             options += ["--input", "specification_path=" + task.property_file]
--- a/benchexec/tools/coveriteam-verifier-validator.py
+++ b/benchexec/tools/coveriteam-verifier-validator.py
@@ -29,7 +29,9 @@ class Tool(coveriteam.Tool):
         data_model_param = get_data_model_from_task(
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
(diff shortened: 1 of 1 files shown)


benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -29,8 +29,8 @@ def cmdline(self, executable, options, task, rlimits):
         data_model_param = get_data_model_from_task(
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
-        if data_model_param and "--data-model" not in options:
-            options += ["--data-model", data_model_param]
+        if data_model_param and "data_model" not in options:
```
lembergerth, line 32: Wild guess: Isn't this wrong?

I expect 'options' to be a list similar to
`["myComposition.cvt", "--input", "data_model=ILP32"]`

So this would never contain "data_model".

Maybe you need something like this?
`if data_model_param and not any(opt.startswith("data_model=") for opt in options)`

Sowasvonbot: Ah true, thanks for the hint.
Provided a fix

lembergerth: Looks good 👍 Did you test it?

PhilippWendler: What exactly is actually the format that CoVeriTeam supports for these options? For example, if CoVeriteam supports `--input "data_model = foo"` (like CPAchecker does), then the new check would also not always match correctly.

Sowasvonbot: The normal use case is `--input data_model=ILP32`, but `--input data_model =ILP32` is possible too.

Leading whitespace before like ` ILP32` is leads to a crash

Sowasvonbot: I'm not sure now, how to check for an existing `data_model`. We can also check for `data_model =`, but I think we need the `=`, otherwise `data_model_program_one` would block the adding of the input too

PhilippWendler: > Leading whitespace before like `ILP32` leads to a crash.

Sounds like a bug to fix in CoVeriTeam ;-)

>  I'm not sure now, how to check for an existing data_model. We can also check for `data_model =`, but I think we need the `=`, otherwise `data_model_program_one` would block the adding of the input too

I agree. But quite likely CoVeriTeam accepts any number of spaces before the `=`. So it might need a regex like `^data_model\s*=` or `^data_model *=`.

benchexec/tools/coveriteam-verifier-validator.py
```diff
@@ -29,8 +29,10 @@ def cmdline(self, executable, options, task, rlimits):
         data_model_param = get_data_model_from_task(
             task, {ILP32: "ILP32", LP64: "LP64"}
         )
-        if data_model_param and "--data-model" not in options:
-            options += ["--data-model", data_model_param]
+        if data_model_param and not any(
+            [option.startswith("data_model=") for option in options]
```
PhilippWendler, line 33: A list comprehension is unnecessary here, just omit the `[]`.

benchexec/tools/coveriteam-verifier-validator.py
Sowasvonbot: Okay, I'm using regex now to detect the `data_model`. I tested it with multiple amount of spaces and it's working.
However I'm negating the `any` call and not the regex.

Sowasvonbot: Do you agree that `LP64` and `    LP64` are two different data models?

benchexec/tools/coveriteam-verifier-validator.py
Sowasvonbot: Ah okay, so I guess without list comprehension the stuff inside the any function is a generator?

benchexec/tools/coveriteam-verifier-validator.py
PhilippWendler: This is not a question for BenchExec or the tool-info module, and it is not about data models in particular. If CoVeriTeam handles `foo = bar` the same as `foo=bar`, when we need to do so as well.

CoVeriTeam needs to decide whether using `foo= bar` sets the value of `foo` to `bar` or to ` bar`.

Sowasvonbot: In Coveriteam atm
`foo =bar` and `foo=bar` is equivalent, but `foo = bar` is different

PhilippWendler (approved):

PR #834, opened 2022-06-28
Added tool name to QuantilePlot popup (table-generator)

leventeBajczi:
Looking up which tool a certain line on the QuantilePlot tab belongs to is tedious, if more than a few tools are compared at once. This PR adds the tool's name to the popup information card, which makes this process easier.


PhilippWendler:
Thanks, this is a nice idea!

In my test, the produced value was empty, though. But anyway, I would say that we should simply show the same value that is in the plot legend, i.e., the `display_title` attribute of the column. What do you think?

leventeBajczi:
That's weird, I don't know why it would be empty - but sure, the display_title might be a better fit, I don't know why I chose the `name` attribute. I'll fix this shortly.

leventeBajczi:
I've checked the code, and I don't really understand which `display_title` you were referring to. I see the following information under `this.props.tools[toolIdx]`:
```
{
   "benchmarkname":"prob-new",
   "block":"test",
   "columns":[
      {
         "display_title":"status",
         "max_width":6,
         "relevant_for_diff":false,
         "title":"status",
         "type":"status",
         "colIdx":0
      },
      {
         "display_title":"cputime",
         "max_width":7,
         "number_of_significant_digits":3,
         "relevant_for_diff":false,
         "title":"cputime",
         "type":"measure",
         "unit":"s",
         "colIdx":1
      },
      {
         "display_title":"walltime",
         "max_width":8,
         "number_of_significant_digits":3,
         "relevant_for_diff":false,
         "title":"walltime",
         "type":"measure",
         "unit":"s",
         "colIdx":2
      },
      {
         "display_title":"memory",
         "max_width":6,
         "number_of_significant_digits":3,
         "relevant_for_diff":false,
         "title":"memory",
         "type":"measure",
         "unit":"MB",
         "colIdx":3
      }
   ],
   "cores":"8",
   "cpu":"QEMU Virtual CPU version 2.5+",
   "cpuCores":"6",
   "date":"2022-06-27 18:52:48 CEST",
   "endtime":"2022-06-28T11:38:52.190693+02:00",
   "freq":"2600 MHz",
   "generator":"BenchExec 3.11-dev",
   "host":"cloud-13411",
   "memlimit":"15000 MB",
   "name":"prob-theta-NOLBE-NOBVI-TVI-EXACT-PRED_SPLIT-0-COARSEST-LOCAL-TRACE-1.0E-6-EXACT-unbounded-MAX.test",
   "niceName":"prob-theta-NOLBE-NOBVI-TVI-EXACT-PRED_SPLIT-0-COARSEST-LOCAL-TRACE-1.0E-6-EXACT-unbounded-MAX.test",
   "options":"--tvi --exact --domain PRED_SPLIT --limit 0 --refinable-selection COARSEST --precision-locality LOCAL --predicate-propagation TRACE --tolerance 1.0E-6 --exact --arithmetic-type unbounded --optim-type MAX",
   "os":"Linux-5.13.0-19-generic-x86_64-with-glibc2.29",
   "ram":"16780 MB",
   "starttime":"2022-06-28T11:38:38.830181+02:00",
   "timelimit":"360 s",
   "tool":"Theta",
   "toolmodule":"benchexec.tools.theta",
   "version":"prob-theta",
   "toolIdx":351,
   "scoreBased":false
}
```

In the QuantilePlot legend, the value that shows up is either the `name` or the `niceName` attribute, and I'd like the same to sho up for the popup card as well.

PhilippWendler:
Sorry, I had overlooked something: The column's `display_title` is only relevant for quantile plots that show all columns of a run set:
https://github.com/sosy-lab/benchexec/blob/927aff556e40562c79a823f033962404c83a4398/benchexec/tablegenerator/react-table/src/components/QuantilePlot.js#L277

For quantile plots that show a specific column for all run sets, the relevant info is given by `getRunSetName`:
https://github.com/sosy-lab/benchexec/blob/927aff556e40562c79a823f033962404c83a4398/benchexec/tablegenerator/react-table/src/components/QuantilePlot.js#L260-L263

So we either need `getRunSetName(this.props.tools[toolIdx])` if `this.state.isValue` or `colTitle` otherwise (`colTitle` is already the proper title of the column).

And we should not name this field `tool`, because it will show different things. Maybe `series`?
It would also be good if we change `info` to `task` while we are at it, because with more fields `info` is too generic.

So, like this (plus proper formatting):
```diff
-          info: rowName,
+          task: rowName,
+          series: this.state.isValue ? getRunSetName(this.props.tools[toolIdx]) : colTitle,
```
Would you mind updating the PR?

PhilippWendler:
Thanks a lot!

PR #835, opened 2022-07-13
Add a tool-info module for BTOR2C

nianzelee:
This PR intends to add a tool-info module for BTOR2C, a converter from BTOR2 models to C programs.


PhilippWendler:
Is the change to `abc.py` intended?

nianzelee:
> Is the change to `abc.py` intended?

Yes.

Should I create a different PR? It is just one line so I thought it would be okay to have this change together with the tool-info module.

PhilippWendler:
That's ok. Just since you didn't mention it it could also have been by accident.

PR #844, opened 2022-10-10
Add Mopsa tool-info

rmonat:


benchexec/tools/mopsa.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Tool info for Mopsa.
+    URL: https://gitlab.com/mopsa/mopsa-analyzer/
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("mopsa-sv-comp", subdir="bin/")
+
+    def version(self, executable):
+        # the provided utility will fail due to return code 2 in Mopsa
```
PhilippWendler, line 24: Maybe check for return codes here, too? I.e., provide a warning if the return code is not 2?

Furthermore, I would suggest to add error handling for cases like invalid JSON.

benchexec/tools/mopsa.py
```diff
@@ -0,0 +1,55 @@
...
+        if run.was_timeout:
+            return "TIMEOUT"
+        r = run.output.text
+        last = r.split("\n")
+        if last[-1] != '': last = last[-1]
+        else: last = last[-2]
+        if last.startswith("true"): return result.RESULT_TRUE_PROP
+        elif last.startswith("unknown"): return result.RESULT_UNKNOWN
+        elif last.startswith("ERROR"): return result.RESULT_ERROR + last[len("ERROR"):]
+        else: raise ValueError(last)
```
PhilippWendler, line 55: I would suggest to not crash here. The output of the tool can always be truncated, for example in case of out-of-memory, if the user presses Ctrl+C here etc. If the tool-info module then crashes, this will make all results unusable.

benchexec/tools/mopsa.py
```diff
@@ -0,0 +1,55 @@
...
+            cmd += ["--data_model", task.options.get("data_model")]
+        if task.property_file:
+            cmd += ["--property", task.property_file]
+        return cmd 
+
+    def determine_result(self, run):
+        if run.was_timeout:
+            return "TIMEOUT"
+        r = run.output.text
+        last = r.split("\n")
```
PhilippWendler, line 49: Note that `run.output` already is a sequence of output lines, no need to call `.text` (which concatenates them) and split them again. `run.output[-1]` works.

benchexec/tools/mopsa.py
```diff
@@ -0,0 +1,55 @@
...
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2022 Raphaël Monat
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+
+import re, json, subprocess
```
PhilippWendler, line 11: Please keep imports on separate lines. `re` is unused, btw.

PhilippWendler (changes requested):
Thanks!

I have some smaller suggestions for improvements. Could you afterwards please also apply the code formatter `black`?

benchexec/tools/mopsa.py
rmonat: You're right. I've changed the wrapper to make it compatible with `BaseTool2._version_from_tool`.

benchexec/tools/mopsa.py
rmonat: Fixed

benchexec/tools/mopsa.py
rmonat: Right, thank you. Fixed.

benchexec/tools/mopsa.py
rmonat: Cleaned. I also applied `black` as requested.

benchexec/tools/mopsa.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            r = r[-2]
+        if r.startswith("true"):
+            return result.RESULT_TRUE_PROP
+        elif r.startswith("unknown"):
+            return result.RESULT_UNKNOWN
+        elif r.startswith("ERROR"):
+            return result.RESULT_ERROR + r[len("ERROR") :]
+        else:
+            return result.RESULT_ERROR + f"(unknown: {last})"
```
PhilippWendler, line 53: `last` is undefined here, I guess you mean `r`?

PhilippWendler (approved):

PR #845, opened 2022-10-16
Add PIChecker tool-info

Lapulatos:
Add tool-info for PIChecker.

### DIFF
--- a/benchexec/tools/pichecker.py
+++ b/benchexec/tools/pichecker.py
@@ -1,0 +1,23 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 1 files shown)


benchexec/tools/pichecker.py
```diff
@@ -0,0 +1,23 @@
...
+class Tool(cpachecker.Tool):
+    """
+    Tool info for PIChecker.
+    URL: https://gitlab.com/Lapulatos/pichecker
+    """
+
+    REQUIRED_PATHS = list(cpachecker.Tool.REQUIRED_PATHS) + ["resources"]
+
+    def version(self, executable):
+        return "1.0"
```
PhilippWendler, line 20: Hard-coding a version number in this module is not a good idea. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

If this is not possible, please return an empty value.

PhilippWendler (changes requested):

benchexec/tools/pichecker.py
Lapulatos: Thank you for your suggestion!

PhilippWendler (approved):

PR #846, opened 2022-10-16
Limit number of workers for table-generator

dbeyer:
When I run table-generator on a machine with many processing units,
then table-generator creates too many workers, such that the operating system
complains with `OSError: [Errno 24] Too many open files`.

Therefore, let us limit the number of workers for table-generator to max. 256.

### DIFF
--- a/benchexec/tablegenerator/__init__.py
+++ b/benchexec/tablegenerator/__init__.py
@@ -1581,8 +1581,9 @@ def main(args=None):
         cpu_count = os.cpu_count() or 1
     except AttributeError:
         pass
-    # Use up to cpu_count*2 workers because some tasks are I/O bound.
-    parallel = concurrent.futures.ProcessPoolExecutor(max_workers=cpu_count * 2)
+    # Use up to cpu_count*2 workers because some tasks are I/O bound,
+    # but limit the number of worker to avoid too many open files.
+    parallel = concurrent.futures.ProcessPoolExecutor(max_workers=min(cpu_count * 2, 256))
 
     name = options.output_name
     outputPath = options.outputPath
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Hm, interesting. Do you have more than 256 columns? Because otherwise I would think that this does not even limit the number of processes in your cases, because we never submit more than that number of parallel tasks to the executor. So a limit that is so high should not make a difference.

In general, it is a difficult question what this limit should be.

dbeyer:
> Do you have more than 256 columns?

Example (not counted, but there are *many* columns):
https://sv-comp.sosy-lab.org/2022/results/results-verified/META_ReachSafety.table.html#/table

I saw with `ps` that it indeed had started more than 500 threads, each with a pipe or so.

> In general, it is a difficult question what this limit should be.

I do not feel the difficulty here. Using multi-threading is good because it improves the performance, but it is not necessary that it uses *all* processing units of a give very large machine.

With the limit of 256 the tool is at least not crashing anymore.  I would say a limit of 30 or 50 would also be very reasonable.

But you can also make up a formula that considers the limit for open files.

PhilippWendler:
For the record:
- BenchExec starts one parallel job per input file, which is typically way less than 256.
- BenchExec also starts one parallel job per column (for statistics computation), which can reach that high numbers, but doesn't open files. But the pipes that are necessary for communicating with the worker processes also count as file descriptors, and this is what probably reaches the limit. (I had not noticed this at first, so I want to write it down for the future.)

So it makes sense to reduce this limit, and we have to choose an arbitrary limit because we cannot reliably determine the number of all open files for our user. So let's pick 32.

PhilippWendler:
The logic from this PR was replaced in 1960c941c0c06b31018a44f8909b1c1de3214c28.

PR #847, opened 2022-10-20
Added veriabsl.py

SakshiAgrawal18:


PhilippWendler:
This is almost the same as the tool-info module for VeriAbs, except for the returned name. Even the executable that is started is the same. Wouldn't it be possible to just just the `veriabs` tool-info module instead?

SakshiAgrawal18:
Want to submit 2 tools, folder structure is same for both - VeriAbs and VeriAbsL. veriabsl.py is needed for VeriAbsL.

PhilippWendler:
A file named `veriabsl.py` is not referenced in your tool-info module, however?

If VeriAbsL is a different tool, then a respective tool-info module can be added to BenchExec. However, given the similarity of the tool-info modules I wonder about the relation of the tools and whether the tool-info modules should share code. For example, suppose that the command-line interface of VeriAbs would change in the future and chnages to its tool-info module are made. Would the same changes than also likely apply to VeriAbsL? If yes, then the modules should share code. Or do you think that the future development of VeriAbs and VeriAbsL might diverge and changes to one are unlikely to apply to the other?

priyankadarke:
@PhilippWendler, you are right. Currently, the code can be shared. VeriAbsL can extend from VeriAbs.

PR #851, opened 2022-10-21
Add VeriOover tool-info module

Mete0o0r:
Add VeriOover tool-info

### DIFF
--- a/benchexec/tools/verioover.py
+++ b/benchexec/tools/verioover.py
@@ -27,9 +27,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, task, rlimits):
-        if task.single_input_file:
-            options = options + ["-file", task.single_input_file]
-
+        options = options + ["-file", task.single_input_file]
         if task.property_file:
             options = options + ["-spec", task.property_file]
 
@@ -47,4 +45,3 @@ class Tool(benchexec.tools.template.BaseTool2):
                 status = result.RESULT_TRUE_PROP
 
         return status
-
--- a/benchexec/tools/verioover.py
+++ b/benchexec/tools/verioover.py
@@ -1,0 +1,47 @@
+
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    VeriOover
+    """
+
+    def name(self):
+        return "VeriOover"
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("VeriOover")
+    
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        if "-file" not in options:
+            options = options + ["-file"]
+
+        options = options + list(task.input_files_or_identifier)
+
+        if task.property_file:
+            options = options + ["-spec", task.property_file]
+
+        return [executable] + options
+
+
+    def determine_result(self, run):
+        # parse output
+        status = result.RESULT_UNKNOWN
+        print(run.output)
+        for line in run.output:
+            if "spec incorrect!" in line:
+                status = result.RESULT_FALSE_PROP
+            elif "spec unknow!" in line:
+                status = result.RESULT_UNKNOWN
+            elif "spec correct!" in line:
+                status = result.RESULT_TRUE_PROP
+
+        return status
+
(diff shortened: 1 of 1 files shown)


benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
...
+        if task.property_file:
+            options = options + ["-spec", task.property_file]
+
+        return [executable] + options
+
+
+    def determine_result(self, run):
+        # parse output
+        status = result.RESULT_UNKNOWN
+        print(run.output)
```
PhilippWendler, line 37: Please do not clutter the output of BenchExec with debugging statements.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
+
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    VeriOover
```
PhilippWendler, line 10: Please add some more information about the tool here, e.g., at least a link to its homepage.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
...
+        return tool_locator.find_executable("VeriOover")
+    
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        if "-file" not in options:
+            options = options + ["-file"]
+
+        options = options + list(task.input_files_or_identifier)
```
PhilippWendler, line 26: Is `-file` the parameter that the tool uses to receive the input file? If yes, then there should simply be no `-file` in options. There is no need for checking it here because it could be in a wrong position inside `options` anyway. Just add it unconditionally like with `-spec`.

Furthermore, does your tool support tasks that consist of more than one input file, or have no input file at all? If not, please do not use `task.input_files_or_identifier` but one of the other properties of `task` such that users get a nice error message.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
+
```
PhilippWendler, line 1: We need a license in order to be able to redistribute this.

So please provide a header with license and copyright information. It must use the machine-readable format of the [REUSE project](https://reuse.software). And it should contain the two lines
```
This file is part of BenchExec, a framework for reliable benchmarking:
https://github.com/sosy-lab/benchexec
```

We would prefer it if you assign the copyright to Dirk Beyer, in which case the Apache 2.0 license would apply. Then the header would be this:
```
# This file is part of BenchExec, a framework for reliable benchmarking:
# https://github.com/sosy-lab/benchexec
#
# SPDX-FileCopyrightText: 2007-2022 Dirk Beyer <https://www.sosy-lab.org>
#
```

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
...
+
+import benchexec.util as util
```
PhilippWendler, line 3: Please remove this unused import.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+    def determine_result(self, run):
+        # parse output
+        status = result.RESULT_UNKNOWN
+        print(run.output)
+        for line in run.output:
+            if "spec incorrect!" in line:
+                status = result.RESULT_FALSE_PROP
+            elif "spec unknow!" in line:
```
PhilippWendler, line 41: Is this typo intentional?

PhilippWendler (changes requested):

benchexec/tools/verioover.py
Mete0o0r: Thanks for pointing out these problems, we'll fix them soon.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,50 @@
...
+        return "VeriOover"
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("VeriOover")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.single_input_file:
```
PhilippWendler, line 30: This check is not necessary. `single_input_file` will automatically throw if there is not exactly one input file, and BenchExec will provide a nice error message to the user. And the returned string would never be empty anyway. So just add it to `options` unconditionally.

benchexec/tools/verioover.py
```diff
@@ -0,0 +1,50 @@
...
+        for line in run.output:
+            if "spec incorrect!" in line:
+                status = result.RESULT_FALSE_PROP
+            elif "spec unknown!" in line:
+                status = result.RESULT_UNKNOWN
+            elif "spec correct!" in line:
+                status = result.RESULT_TRUE_PROP
+
+        return status
+
```
PhilippWendler, line 50:
```suggestion
```
The code formatter would remove this empty line.

PhilippWendler (commented):
Thanks! Just two minor things remain.

PR #852, opened 2022-10-21
Add WASP-C tool-info

filipeom:
Test-Comp 2023 Submission 7869

### DIFF
--- a/benchexec/tools/wasp-c.py
+++ b/benchexec/tools/wasp-c.py
@@ -5,8 +5,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import os
-from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
 
@@ -18,10 +16,6 @@ class Tool(benchexec.tools.template.BaseTool2):
 
     def executable(self, tool_locator):
         return tool_locator.find_executable('wasp-c', subdir='bin')
-
-    def working_directory(self, executable):
-        executableDir = os.path.dirname(executable)
-        return executableDir
 
     def version(self, executable):
         return self._version_from_tool(executable, '-v')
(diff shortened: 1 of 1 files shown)


benchexec/tools/wasp-c.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    This class serves as tool adaptor for WASP-C (https://github.com/wasp-platform/wasp)
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable('wasp-c', subdir='bin')
+
+    def working_directory(self, executable):
+        executableDir = os.path.dirname(executable)
+        return executableDir
```
PhilippWendler, line 24: In general, I advise on writing tools in a way such that they work no matter from which directory they are called. This is much more convenient for users, less error-prone, etc. If your tool supports this, we can remove this method. Is this the case?

PhilippWendler (commented):

benchexec/tools/wasp-c.py
filipeom: Yes it does, I'll remove the method.

benchexec/tools/wasp-c.py
```diff
@@ -0,0 +1,46 @@
...
+        if task.property_file:
+            options = options + ["--property", task.property_file]
+        else:
+            raise benchexec.tools.template.UnsupportedFeatureException(
+                "property file is required"
+            )
+        return [executable] + options + [task.single_input_file]
+
+    def determine_result(self, run):
+        status = result.RESULT_UNKNOWN
```
PhilippWendler, line 37: Sorry, I had not noticed this before: I would suggest to return `RESULT_UNKNOWN` only if the tool actually gave up itself, and `RESULT_ERROR` in other cases. This will make it easier to distinguish crashes, timeouts, etc. from unknown results, and will allow BenchExec to add more information about the problem in case the tool terminates abnormally. For example, currently if your tool crashes hard (segfault or so) without the chance to print `WASP crashed`, it would be listed as "unknown" instead of an error.

Also, I see that you return "unknown" in case of timeouts, but BenchExec will overwrite this with "timeout" anyway.

In summary, the intended results would be:
- "timeout" in case of timeout (BenchExec does that automatically, no need to worry)
- "unknown" in case the tool gives up voluntarily
- "done" if everything was fine
- "error" or a different string in case the tool crashes

filipeom: Thank you so much for your help! I've updated the `determine_result` method return values accordingly. :slightly_smiling_face:

PR #853, opened 2022-10-21
Rename graves.py to graves-cpa.py. Add new graves.py

will-leeson:
We are submitting a new tool, Graves, which is based on coveriteam. To reduce confusion, we have moved graves.py to graves-cpa.py. graves.py now reflects a tool called graves, not graves-cpa

### DIFF
--- a/benchexec/tools/graves-cpa.py
+++ b/benchexec/tools/graves-cpa.py
@@ -1,0 +1,20 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.cpachecker as cpachecker
+
+
+class Tool(cpachecker.Tool):
+    """
+    Tool info for GraVeS.
+    URL: https://github.com/will-leeson/cpachecker
+    """
+
+    REQUIRED_PATHS = list(cpachecker.Tool.REQUIRED_PATHS) + ["resources"]
+
+    def name(self):
+        return "GraVeS"
--- a/benchexec/tools/graves.py
+++ b/benchexec/tools/graves.py
@@ -5,16 +5,37 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import benchexec.tools.cpachecker as cpachecker
+import benchexec.tools.template
 
 
-class Tool(cpachecker.Tool):
+class Tool(benchexec.tools.template.BaseTool2):
     """
-    Tool info for GraVeS.
-    URL: https://github.com/will-leeson/cpachecker
+    Using the tool info for CoVeriTeam: On-Demand Composition of Cooperative Verification Systems.
+    URL: https://gitlab.com/sosy-lab/software/coveriteam.
+
+    This class has 2 purposes:
+        1. to serve as an abstract class for specific coveriteam programs like verifiers, validators, etc.
+        2. to serve as the tool info module for any generic coveriteam program.
     """
 
-    REQUIRED_PATHS = list(cpachecker.Tool.REQUIRED_PATHS) + ["resources"]
+    REQUIRED_PATHS = [
+        "coveriteam",
+        "bin",
+        "lib",
+        "depends",
+        "predict",
+    ]
 
     def name(self):
-        return "GraVeS"
+        return "Graves"
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("graves.sh")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )


PhilippWendler:
Hm, this could be quite confusing for people. For example those who want to reproduce old competition results would need to change the benchmark definitions. I would prefer if the existing tool-info module `graves.py` would not be changed incompatibly. Could simply a new tool-info module be added? Could there be one tool-info module that supports both versions of your tool?

will-leeson:
What if we were to change the name from 'Graves' to 'Graves-SV'. It would probably be more appropriate, as Graves is a technique and this is simply an instantiation for C programs following SV-Comp specifications. Then, we could call it's files 'graves-sv.[py,xml,etc.]'.

PhilippWendler:
How you call your tool is up to you, of course.

I just would like to keep BenchExec backwards compatible and thus not change/rename the existing `graves.py` in an incompatible way.

will-leeson:
I have reverted graves-cpa file back to "graves.py". We have added a new file for our new submission under the name "graves-par.py". This will be our submission's name moving forward.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.coveriteam as coveriteam
+import benchexec.result as result
+from benchexec.tools.template import UnsupportedFeatureException
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
```
PhilippWendler, line 11: Please remove all unused imports.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+import re
+
+class Tool(coveriteam.Tool):
```
PhilippWendler, line 14: This is somewhat confusing. Currently, the class inherits from `coveriteam.Tool`, but it overwrites all methods from `coveriteam.Tool`, so this is useless. On the other hand, it contains two methods copied verbatim from `coveriteam.Tool` (`version` and `program_files`) and one copied verbatim from `coveriteam-verifier-validator`.

The main question here is: What is the relationship between CoVeriTeam and Graves-Par?
Is Graves-Par a variant of CoVeriTeam that will follow CoVeriTeam's development, or is it a tool that will be developed independently.
Suppose for example that for a new version of CoVeriTeam we make adjustments to its tool-info module, should the tool-info module of Graves-Par also get these adjustments automatically, or not?
And does the answer of this question apply to the base CoVeriTeam or to the specific coveriteam-verifier-validator instance?

Based on this it should be decided whether to use inheritance or not. If inheritance is preferable, then the duplicate code should be removed.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("graves.sh")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
```
PhilippWendler, line 40: It seems this was copied from `coveriteam.Tool` but as is likely contains a bug. For CoVeriTeam, the executable is in a subdirectory, so here `parent_dir=True` needs to be used, but for Graves-Par the executable is in the main directory, so now the paths apply to the wrong directory.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+    def cmdline(self, executable, options, task, rlimits):
+        """
+        Modified from the CoVeriTeam definition
```
PhilippWendler, line 45: This is not really a useful comment and actually misleading because other methods here are also copied and modified from the tool-info module for CoVeriTeam. Please remove.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+
+import benchexec.tools.coveriteam as coveriteam
+import benchexec.result as result
+from benchexec.tools.template import UnsupportedFeatureException
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import re
+
+class Tool(coveriteam.Tool):
+    """
+    Tool infor for Graves, a verifier selector based on Graph Neural Networks
```
PhilippWendler, line 16: Please add a URL for the tool.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,78 @@
...
+            options += [task.property_file]
+        else:
+            raise UnsupportedFeatureException(
+                "Can't execute Graves: "
+                "Specification is missing."
+            )
+
+        options += [task.single_input_file]
+
+        return [executable] + options
```
PhilippWendler, line 58: As I expect that you want to submit Graves-Par to SV-COMP: Are you sure that you do not need the data model?

PhilippWendler (changes requested):

benchexec/tools/graves-par.py
will-leeson: Graves-Par is a variant of coveriteam. We want to use their infrastructure and allow our selector to configure coveriteam. Inheritance would be preferred. I will make the appropriate changes.

benchexec/tools/graves-par.py
will-leeson: Added

benchexec/tools/graves-par.py
will-leeson: Correct. We initially tested on 32bit only

benchexec/tools/graves-par.py
will-leeson: Added a more useful comment

benchexec/tools/graves-par.py
will-leeson: I see. I have changed it

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,64 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.coveriteam_verifier_validator as coveriteam
```
PhilippWendler, line 8: Did you test this and does it work for you?
For me it produces `ImportError: No module named 'benchexec.tools.coveriteam_verifier_validator'`.

I guess something like https://github.com/sosy-lab/benchexec/blob/9cfa48ba87862d1bfa6da1c58b56583b8fdf7ff5/benchexec/tools/yogar-cbmc-parallel.py#L10 is needed.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,64 @@
...
+        if data_model_param and not any(
+            re.match("data_model *=", option) for option in options
+        ):
+            options += [data_model_param]
+
+        if task.property_file:
+            options += [task.property_file]
+        else:
+            raise UnsupportedFeatureException(
+                "Can't execute Graves: " "Specification is missing."
```
PhilippWendler, line 54: This could be one string.

benchexec/tools/graves-par.py
```diff
@@ -0,0 +1,64 @@
...
+    def name(self):
+        return "Graves-Par"
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("graves.sh")
+
+    def cmdline(self, executable, options, task, rlimits):
+        """
+        Graves-PAR takes in three arguments: the program looking to be verified,
+        a property file, and the data model. From their it forms its prediction
```
PhilippWendler, line 39: This comment does not match the implementation of this method, is this a bug?

Contrary to the comment, the method adds the data model, the property file, and the input file (in this order).

Note: I would say that relaying on a fixed order of positional command-line arguments (e.g. "first parameter is input file, second parameter is property file") is unusual, confusing, and error-prone. I would recommend to use named parameters for such cases.

benchexec/tools/graves-par.py
will-leeson: Fixed

benchexec/tools/graves-par.py
will-leeson: I agree. This is very brittle (and currently wrong). I have fixed the ordering, but we will make it flag based instead of positional in the future.

benchexec/tools/graves-par.py
will-leeson: Thanks. Fixed

PR #854, opened 2022-10-22
Add bubaak tool-info module

mchalupa:


benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+# SPDX-FileCopyrightText: 2022 Marek Chalupa marek.chalupa@ista.ac.at
+#
+# SPDX-License-Identifier: Apache-2.0
+
+""
+
+import benchexec
+import benchexec.result as result
+import benchexec.util as util
```
PhilippWendler, line 13: Please remove unused imports.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+# SPDX-FileCopyrightText: 2022 Marek Chalupa marek.chalupa@ista.ac.at
+#
+# SPDX-License-Identifier: Apache-2.0
+
+""
```
PhilippWendler, line 9: Please remove this unnecessary line.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+import benchexec.result as result
+import benchexec.util as util
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+
+def get_verdict(s: str):
+    return s[17:].strip()
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Info object for the tool Bubaak.
```
PhilippWendler, line 21: Can you please add some more documentation, e.g., at least a URL?

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options += ["-prp", task.property_file]
+
+        arch = get_data_model_from_task(task, {ILP32: "-32", LP64: "-64"})
+        if arch is not None and arch not in options:
+            options += [arch]
+
+        return [executable, "-sv-comp"]\
```
PhilippWendler, line 44: Will this hard-code a specific configuration of your tool? This would not be desired. BenchExec is not just a tool for SV-COMP, but a generic benchmarking tool, and people would also want to use it with other configurations. So all non-required configuration options should be removed here and but into the benchmark description.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+               + options\
+               + list(task.input_files_or_identifier)
```
PhilippWendler, line 46: Does your tool support tasks with more than one input file or without any input files? If not, please use `task.input_files` or `task.single_input_file` instead because these will produce a proper error message for the user if an unsupported task is used.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def determine_result(self, run):
+        if run.was_timeout:
+            return "timeout"
+
+        if run.output is None:
+            return "error (no output)"
```
PhilippWendler, line 53: This will never be true, not even for empty output.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+        if arch is not None and arch not in options:
+            options += [arch]
+
+        return [executable, "-sv-comp"]\
+               + options\
+               + list(task.input_files_or_identifier)
+
+    def determine_result(self, run):
+        if run.was_timeout:
+            return "timeout"
```
PhilippWendler, line 50: This check is not required, BenchExec will do that anyway. But you can keep it if you want to avoid executing the rest of the method in case of a timeout. In this case, please return `"TIMEOUT"`, though.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if run.output is None:
+            return "error (no output)"
+
+        for line in run.output:
+            if "SV-COMP verdict:" in line:
+                return get_verdict(line)
+
+        if run.returncode != 0:
+            return f"{result.RESULT_ERROR}: ret {returncode}"
```
PhilippWendler, line 60: `returncode` is undefined here.

In general, for consistency with BenchExec and other tools, I would suggest to use `ERROR(ret 1)` instead of `ERROR: ret 1`, but this is only a suggestion.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+        if run.returnsignal != 0:
+            return f"{result.RESULT_ERROR}: sig {returnsignal}"
```
PhilippWendler, line 62: Same as above.

benchexec/tools/bubaak.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 64: I would suggest to return `RESULT_UNKNOWN` only if the tool actually gave up itself, and `RESULT_ERROR` in other cases. This will make it easier to distinguish crashes, timeouts, etc. from unknown results, and will allow BenchExec to add more information about the problem in case the tool terminates abnormally.

PhilippWendler (changes requested):

benchexec/tools/bubaak.py
mchalupa: The tool supports multiple input files.

benchexec/tools/bubaak.py
mchalupa: Upps, my bad, sorry.

benchexec/tools/bubaak.py
mchalupa: Good point, forgot that this option is in the benchmark definition.

PhilippWendler (approved):

PR #855, opened 2022-10-22
Add tool-info module for mlb

kcanamgal:
Add mlb bench exec

### DIFF
--- a/benchexec/tools/mlb.py
+++ b/benchexec/tools/mlb.py
@@ -1,0 +1,45 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool):
+    """
+    Tool info for MLB
+    (https://github.com/MLB-SE/Experiment).
+    """
+
+    def executable(self):
+        return util.find_executable("run.sh")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, arg="-version")
+
+    def name(self):
+        return "MLB"
+
+    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
+        cmd = [executable]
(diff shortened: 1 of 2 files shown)


benchexec/tools/mlb.py
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.util as util
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool):
```
PhilippWendler, line 13: This is using our old and outdated API `BaseTool`. Please switch to `BaseTool2` and adjust your module accordingly. We have documented [hints](https://github.com/sosy-lab/benchexec/blob/main/doc/tool-integration.md#migrating-tool-info-modules-to-new-api) for this, and it should be not much effort.

Otherwise certain features will not work.

benchexec/tools/mlb.py
```diff
@@ -0,0 +1,45 @@
...
+        cmd = [executable]
+        if options:
+            cmd = cmd + options
+        if propertyfile:
+            cmd.append(propertyfile)
+        return cmd + tasks
+
+    def determine_result(self, returncode, returnsignal, output, isTimeout):
+        # parse output
+        status = result.RESULT_UNKNOWN
```
PhilippWendler, line 38: I would suggest to return `RESULT_UNKNOWN` only if the tool actually gave up itself, and `RESULT_ERROR` in other cases. This will make it easier to distinguish crashes, timeouts, etc. from unknown results, and will allow BenchExec to add more information about the problem in case the tool terminates abnormally.

PhilippWendler (changes requested):

benchexec/tools/mlb.py
```diff
@@ -0,0 +1,47 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# import benchexec.util as util
```
PhilippWendler, line 8: Please remove this unnecessary line. Otherwise the module looks fine.

PR #856, opened 2022-10-27
Update mlb.py

kcanamgal:


PhilippWendler:
Your PR includes a bunch of old commits that are irrelevant and should not be merged. Please create PRs from a current state of the target repository.

What is the goal of this PR? What are you attempting to fix? Please always explain this in the PR description.

The changes in this PR will not work and apparently were not [tested](https://github.com/sosy-lab/benchexec/blob/main/doc/tool-integration.md#testing-the-tool-integration), because the current state of the PR will inevitably trigger a failing assertion.

kcanamgal:
Problem is solved and I find that there is no need to change the .py file.  I will close the PR.

PR #857, opened 2022-10-27
Added data model options

SakshiAgrawal18:


PhilippWendler:
It seems this PR has a bunch of old commits with additional unwanted changes in it. Can you please clean it up by removing the old commits, potentially rebasing onto current BenchExec's main branch, and force pushing?

PhilippWendler:
I guess your problem is that you created this PR from `main`, which is by default a protected branch and prevents force pushing. You can [configure this](https://github.com/sosySakshiAgrawal18/benchexec/settings/branches). I would recommend to follow standard best practices and create branches for PRs in the future.

There is nothing preventing you from replacing a PR with a new one, but in general it is nicer for the target repo's maintainers to not have to deal with superfluous PRs and it keeps the history/discussions/etc. in one place.

PR #859, opened 2022-10-29
Fix user message

dbeyer:
Fix user message: Wrong term used in user message ('runset' does not exist in user's world)

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -565,7 +565,7 @@ class RunSet(object):
                 base = os.path.basename(run.identifier)
                 if base in sourcefilesSet:
                     logging.warning(
-                        "Input file with name '%s' appears twice in runset. "
+                        "Input file with name '%s' appears twice in run definition. "
                         "This could cause problems with equal logfile-names.",
                         base,
                     )


PhilippWendler:
The license that this PR attaches to the images does not match the existing license of the LaTeX sources, and the copyright string has deviating numbers. Because we own the copyright this is in principle legal, but it can be confusing to others and it could actually create license problems for others if they want to change the image.

Was this done on purpose or can be we use the same license and copyright info as for the sources?

dbeyer:
How about changing the license for the sources also to CC-BY? I think for papers and diagrams, CC-BY is preferred.

PhilippWendler:
That's right. Nobody ever touched these files except for us, so we can do this.

PhilippWendler:
I did the relicensing in 488ff4755a4d8940cfad11ba6cb2e0c6f9d54334. And I created 7397aee3dab6c6a8769d686cbd66b58285215ec5, which introduces SVGs instead of PDFs for the images, which also fulfills the purpose of making them viewable online while being a text format, which is nicer in git. So everything from this PR is implemented.

PR #860, opened 2022-10-29
Added veriabsl.py and data_model option in veriabs.py

SakshiAgrawal18:


PhilippWendler:
The title of this PR claims that in addition to the data-model option, a file named `veriabsl.py` is to be added, but no such file is actually added. This is the opposite of #857 which did have the file but where it seemed accidental because it was not present as described change.

Note also how #857 is a good example why one should not create PRs from the main branch of the fork (which this PR does again). Because the branch has changed in the meantime now it is no longer possible to retrospectively see what was part of the PR.

The current PR also has a confusing merge commit in it, please avoid this.

A PR that correctly does what is described in the title of this PR should have exactly two commits:
- one commit adding `veriabsl.py`
- one commit adding the data-model option
Such a PR should come from an appropriate branch in your fork and be based on a somewhat recent commit of our main branch.

When creating a PR, please check before what is going to be part of your PR! GitHub shows this right on the page where you are describing the PR. Everything else is a waste of time for everyone.

Also, please make sure to format the code with the formatter [Black](https://pypi.org/project/black/).

PR #861, opened 2022-10-31
Update toolinfos of gazer and theta

leventeBajczi:
This updates gazer-theta.py to use a new starter script, and theta.py to use new exit codes

### DIFF
--- a/benchexec/tools/gazer-theta.py
+++ b/benchexec/tools/gazer-theta.py
@@ -20,7 +20,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     REQUIRED_PATHS = [".."]
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("gazer_starter.py", subdir="scripts")
+        return tool_locator.find_executable("gazer-start.sh", subdir="scripts")
 
     def name(self):
         return "gazer-theta"
--- a/benchexec/tools/gazer-theta.py
+++ b/benchexec/tools/gazer-theta.py
@@ -20,7 +20,10 @@ class Tool(benchexec.tools.template.BaseTool2):
     REQUIRED_PATHS = [".."]
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("gazer-start.sh", subdir="scripts")
+        try:
+            return tool_locator.find_executable("gazer-start.sh")
+        except ToolNotFoundException:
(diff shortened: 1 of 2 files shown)


PhilippWendler:
The problem with the current state of the PR is that it would break backwards compatibility of BenchExec, because it could no longer be used to benchmark previous versions of these tools. Could you please rewrite it such that it supports both the new and old version?

I think it should not be too difficult. For `executable()` one can catch `ToolNotFoundException` and try a fallback (`map2check.py` does this, for example). For `determine_result` it seems that one could just keep all cases?

leventeBajczi:
@PhilippWendler I modified `gazer-theta.py` as you requested, it hopefully detects both versions (favoring the new one).
The other modifications (mainly the if -> elif change in `theta.py` and the new exitcodes) should not impose breaking changes for the old version of the executable, only the different "unknown/error" outcomes will be different. Is it acceptable like this?

PhilippWendler:
> @PhilippWendler I modified `gazer-theta.py` as you requested, it hopefully detects both versions (favoring the new one).

Thanks!

> The other modifications (mainly the if -> elif change in `theta.py` and the new exitcodes) should not impose breaking changes for the old version of the executable, only the different "unknown/error" outcomes will be different. Is it acceptable like this?

It would be acceptable, but why not just keep all of them? Would there be a disadvantage?

leventeBajczi:
The new system (with the new exitcodes) uses systematically increasing codes for similar errors (starting with 200, 210, 220 for these "groups"). Currently there would be no overlap, but in the future there might be (some of the old error codes are dangerously close). I think that moving forward we will only add codes to the list rather than overwrite old ones, but we also need to have a clean slate so that there are no compatibility issues down the line.

benchexec/tools/gazer-theta.py
```diff
@@ -20,7 +20,10 @@ class Tool(benchexec.tools.template.BaseTool2):
     REQUIRED_PATHS = [".."]
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("gazer_starter.py", subdir="scripts")
+        try:
+            return tool_locator.find_executable("gazer-start.sh")
+        except ToolNotFoundException:
```
PhilippWendler, line 25: This needs to be imported. Then the PR is ready to merge.

leventeBajczi: True, I forgot to add the include. Fixed in a26ad64.

PR #863, opened 2022-11-01
Update veriabs.py

SakshiAgrawal18:
This update is to change the parsing logic, in case unknown string is present before success or failure string it should not pick unknown string. So rather than checking for string in each line check for string in complete run.output.

### DIFF
--- a/benchexec/tools/veriabs.py
+++ b/benchexec/tools/veriabs.py
@@ -55,12 +55,10 @@ class Tool(benchexec.tools.template.BaseTool2):
         return [executable] + options + [task.single_input_file]
 
     def determine_result(self, run):
-        for line in run.output:
-            if "VERIABS_VERIFICATION_SUCCESSFUL" in line:
-                return result.RESULT_TRUE_PROP
-            elif "VERIABS_VERIFICATION_FAILED" in line:
-                return result.RESULT_FALSE_REACH
-            elif "NOT SUPPORTED" in line or "VERIABS_UNKNOWN" in line:
-                return result.RESULT_UNKNOWN
-
+        if run.output.any_line_contains("VERIABS_VERIFICATION_SUCCESSFUL"):
+            return result.RESULT_TRUE_PROP
+        elif run.output.any_line_contains("VERIABS_VERIFICATION_FAILED"):
+            return result.RESULT_FALSE_REACH
+        elif run.output.any_line_contains("VERIABS_UNKNOWN" or "NOT SUPPORTED"):
+            return result.RESULT_UNKNOWN
         return result.RESULT_ERROR


benchexec/tools/veriabs.py
```diff
@@ -55,12 +55,10 @@ def cmdline(self, executable, options, task, rlimits):
...
-            elif "VERIABS_VERIFICATION_FAILED" in line:
-                return result.RESULT_FALSE_REACH
-            elif "NOT SUPPORTED" in line or "VERIABS_UNKNOWN" in line:
-                return result.RESULT_UNKNOWN
-
+        if run.output.any_line_contains("VERIABS_VERIFICATION_SUCCESSFUL"):
+            return result.RESULT_TRUE_PROP
+        elif run.output.any_line_contains("VERIABS_VERIFICATION_FAILED"):
+            return result.RESULT_FALSE_REACH
+        elif run.output.any_line_contains("VERIABS_UNKNOWN" or "NOT SUPPORTED"):
```
PhilippWendler, line 62: This does not work. `"VERIABS_UNKNOWN" or "NOT SUPPORTED"` evaluates to `"VERIABS_UNKNOWN"`, so this line does not check for `NOT SUPPORTED`.

Anyway, I wonder why the change? It seems that the intention of the new code is the same as the old, so it would be only a refactoring, but the old code is not so bad that it would need to be changed. Feel free to keep it as is.

PhilippWendler: Or do you want to change the parsing logic with regards to what happens if two or more of these result strings occur in different lines in the output? (The behavior actually does change for such logs.)

In this case, it would be good to explain such a reason in the commit message, because otherwise the knowledge about this behavior would lost in the future. Actually, it would even be better to additionally have a comment describing the desired behavior in this method if it is important.

SakshiAgrawal18: Yes, this is for parsing logic. Have updated the commit comment, if that works.

PhilippWendler:
Unfortunately this was merged too early, the bug from https://github.com/sosy-lab/benchexec/pull/863#discussion_r1010340968 still exists. Can you please fix it?

PhilippWendler:
The script does not work the same way as earlier, because the earlier script did check for `NOT SUPPORTED` and the new one does not.

SakshiAgrawal18:
> The script does not work the same way as earlier, because the earlier script did check for `NOT SUPPORTED` and the new one does not.

Won't this "run.output.any_line_contains("VERIABS_UNKNOWN" or "NOT SUPPORTED")" will take care of both the strings?

PhilippWendler:
No, as explained in https://github.com/sosy-lab/benchexec/pull/863#discussion_r1010340968, `"VERIABS_UNKNOWN" or "NOT SUPPORTED"` evaluates to `"VERIABS_UNKNOWN"`, so this line does not check for `NOT SUPPORTED`.

SakshiAgrawal18:
> No, as explained in [#863 (comment)](https://github.com/sosy-lab/benchexec/pull/863#discussion_r1010340968), `"VERIABS_UNKNOWN" or "NOT SUPPORTED"` evaluates to `"VERIABS_UNKNOWN"`, so this line does not check for `NOT SUPPORTED`.

oh right. Script is returning ERROR for "NOT SUPPORTED" string. Need to have different checks for both of the strings.

PR #864, opened 2022-11-01
Add a pre-defined result value for TIMEOUT

dbeyer:
The return values for TIMEOUT are inconsistent in the tool-info modules.
E.g., `symbiotic4` calls it "timeout" and `seahorn` calls it "TIMEOUT".

Solution: Let us add a pre-defined value for this return value.

Once merged, I would rewrite all tool-info modules to use this predefined return value.

Users are expecting such a pre-defined value: https://github.com/sosy-lab/benchexec/commit/c9a22a7275abb0f393071c2253682e852366184d

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -31,9 +31,9 @@ WALLTIMELIMIT = "walltimelimit"
 _BYTE_FACTOR = 1000  # byte in kilobyte
 
 _ERROR_RESULTS_FOR_TERMINATION_REASON = {
-    "cputime": "TIMEOUT",
-    "cputime-soft": "TIMEOUT",
-    "walltime": "TIMEOUT",
+    "cputime": result.RESULT_TIMEOUT,
+    "cputime-soft": result.RESULT_TIMEOUT,
+    "walltime": result.RESULT_TIMEOUT,
     "memory": "OUT OF MEMORY",
     "killed": "KILLED",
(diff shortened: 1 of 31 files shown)


PhilippWendler:
Actually, most of these checks in the tool-info module are probably completely irrelevant, because BenchExec will overwrite the status to `TIMEOUT` anyway in case of a timeout. The `run.was_timeout` field only exists such that tool-info modules do not need to further analyze the run result in such a case and in order for the tool-info module being able to distinguish cases like "unexpected kill by signal 9" from "timeout and expected kill by signal 9". But as authors of tool-info modules keep adding code for handling timeout cases event though I tell them they don't need to, we can as well accept that and at least make it easy. So I support this.

The four occurrences of `"TIMEOUT"` in `benchexec/model.py` should then also be switched to the constant together with the tool-info modules, and history would be a little bit nicer if introduction of the constant and use of the constant would be in the same commit.

Do you want to finish this or shall I implement it?

dbeyer:
I can do it.

I did not touch https://github.com/sosy-lab/benchexec/blob/0caa86264c42f7761033a7897e0cd45843a6c7c3/benchexec/tools/theta.py#L56 because it looks like the authors want to see that it was a timeout from inside their tool.

PhilippWendler (approved):
Thanks!

PR #865, opened 2022-11-01
Update tool-info for PeSCo

cedricrupb:
The new version of PeSCo employs a new command line interface. 
To support the interface, I created a new tool-info module. The module should be backwards compatible.

If preferred, I could also create a second tool info module.


PhilippWendler (commented):
Looks good, but please format with Black.

PR #869, opened 2022-11-03
Add new result subtype for witnesslint

MartinSpiessl:
This makes use of the feature added to the witness linter via https://gitlab.com/sosy-lab/benchmarking/sv-witnesses/-/merge_requests/66

For the competition (esp. the new validator track) we want to be able to detect those cases where the tool produced a syntactically fine witness, but just of wrong type (e.g. a correctness witness for a task where we would expect a violation witness).


PhilippWendler:
@dbeyer Feel free to directly merge after you review, fine from my side.

dbeyer (approved):

PR #870, opened 2022-11-04
Change of status and category for mismatching witness type

dbeyer:
Change of status and category if a validator (the linter `WitnessLint`) reported an unexpected witness type.

Explanation:  Assume a verification task with expected verdict TRUE.
Suppose a verifier returns TRUE (proved program correct) but returns a witness of type `violation_witness`.
Further suppose that the violation witness does not describe any feasible path that violates the specification,
and a validator returns TRUE as well (correctly, because no violation found).
Then the verifier gets the score points, but did not provide a witness with the necessary invariants.

Therefore, we have to change the status and category in order to assign 0 score points.

### DIFF
--- a/contrib/mergeBenchmarkSets.py
+++ b/contrib/mergeBenchmarkSets.py
@@ -103,6 +103,12 @@ def get_witness_result(witness, verification_result):
             f"witness invalid ({status_from_verification})",
             result.CATEGORY_ERROR,
         )
+    # A mismatch of the witness type counts as error of the verifier.
+    if status_from_validation == "ERROR (unexpected witness type)":
+        return (
+            f"witness mismatch ({status_from_verification})",
+            result.CATEGORY_ERROR,
+        )
     # Other unconfirmed witnesses count as CATEGORY_CORRECT_UNCONFIRMED.
     if category_from_verification == result.CATEGORY_CORRECT:
         return status_from_verification, result.CATEGORY_CORRECT_UNCONFIRMED


dbeyer:
Depends on https://github.com/sosy-lab/benchexec/pull/869 and https://gitlab.com/sosy-lab/benchmarking/sv-witnesses/-/merge_requests/66

strejcek:
I understand the problem, but I'm not sure about the solution that asks for modifications of all witness validators. What happen if one validator does not report the mismatching error? I guess that this mismatch should be detected by witness linter (and witnesses should be passed to validators only if the linter agrees that they are syntactically correct and they match the verification result).

dbeyer:
> I'm not sure about the solution that asks for modifications of all witness validators.

This is possibly a misunderstanding: I did not ask validators for changes as solution to this problem discussed here. The mail about changes to the validators is solving a different problem: https://gitlab.com/sosy-lab/sv-comp/bench-defs/-/merge_requests/327

> I guess that this mismatch should be detected by witness linter

Yes, exactly, this is the solution, implemented here: https://gitlab.com/sosy-lab/benchmarking/sv-witnesses/-/merge_requests/66/diffs and https://github.com/sosy-lab/benchexec/pull/869/files

And this MR here is for changing the status and category for the score counting, in case the witness type does not match.

dbeyer:
With this explanation, do you agree?

contrib/mergeBenchmarkSets.py
```diff
@@ -103,6 +103,12 @@ def get_witness_result(witness, verification_result):
             f"witness invalid ({status_from_verification})",
             result.CATEGORY_ERROR,
         )
+    # A mismatch of the witness type counts as error of the verifier.
+    if status_from_validation == "ERROR (unexpected witness type)":
+        return (
+            f"witness mismatch ({status_from_verification})",
+            result.CATEGORY_ERROR,
+        )
```
MartinSpiessl, line 111: This is code duplication. The if condition in line 101 should just be extended, maybe even checking if satus_from_validation is one of the elements of a set with the two strings inside.

MartinSpiessl (changes requested):
Functionally OK, just the code duplication is not nice, consider fixing it.

contrib/mergeBenchmarkSets.py
strejcek: But the reported status is also different in these two cases: "witness invalid" vs. "witness mismatch". So I'm in favor of the original solution.

strejcek:
OK, now I understand and agree.

strejcek (approved):

contrib/mergeBenchmarkSets.py
dbeyer: Ok, let's keep then.

dbeyer: @MartinSpiessl Would you approve?

MartinSpiessl: `witness mismatch` is weird, this is not informative to the reader. It is not clear what is mismatching. I would propose to have `witness type mismatch`, which would also be consistent with how we call this in the witness linter code now.

PR #871, opened 2022-11-04
New tool-info modules for ABC and AVR

Po-Chun-Chien:
ABC and AVR are both hardware verification tools.
This PR adds a tool info-module for AVR and updates the existing module for ABC.


benchexec/tools/abc.py
```diff
@@ -24,7 +25,33 @@ def name(self):
         return "ABC"
 
     def cmdline(self, executable, options, task, rlimits):
-        return [executable] + options + [task.single_input_file]
+        return (
+            [executable]
+            + ["-c", "&r {}; &put".format(task.single_input_file)]
```
PhilippWendler, line 30: This is a somewhat unexpected way to call a tool, maybe a comment would be nice that explains why this is done?

And is this backwards compatible or does it change the behavior of the tool compared to the previous version?

Using f-strings would be nicer than `.format()`.

benchexec/tools/abc.py
```diff
@@ -24,7 +25,33 @@ def name(self):
...
+            + options
+        )
+
+    def determine_result(self, run):
+        """
+        @return: status of ABC after executing a run
+        """
+        if run.was_timeout:
+            return "TIMEOUT"
```
PhilippWendler, line 39: Please use `result.RESULT_TIMEOUT` instead of a literal.

But most of the time, this check is not actually needed because BenchExec sets the timeout result anyway. Only if you want to avoid returning any other irrelevant result below you need this.

Same for avr.py, of course.

benchexec/tools/abc.py
```diff
@@ -24,7 +25,33 @@ def name(self):
...
+        status = None
+        for line in run.output:
```
PhilippWendler, line 41: Is it by design that this loop does not return if a match is found and always checks all lines?

Same for avr.py.

benchexec/tools/abc.py
nianzelee, line 30: Is this general enough for all commands in ABC?
Remember that this tool-info module might be used for other applications besides verification.

Po-Chun-Chien: I changed the read method because the old ABC read method cannot process uninitialized registers correctly (see e1e2b06).

The circuit is stored as a GIA (the newer AIG representation) after `&read` (`&r`, the new read method).
Then `&put` create a copy the circuit and store it as an AIG (the old AIG representation).
Any following command can be executed no matter it needs a GIA or AIG.

However, this new `&read` can only take an AIGER file as input, whereas the old ABC `read` can consume other formats such as BLIF.

Po-Chun-Chien: If we want to support other file formats, we can also add a switch here based on the file extension of the input.

nianzelee: Yes, we want to support other file formats, too.
Could you add the switch and some documentation here?

benchexec/tools/abc.py
Po-Chun-Chien: I cannot find the definition of `RESULT_TIMEOUT` in `result.py`.

Po-Chun-Chien: Ahh, it is added just recently. I will merge the new changes first.

Po-Chun-Chien: Addressed in 0ef3c7f

benchexec/tools/abc.py
Po-Chun-Chien: Please have a look at c2ccbdd.

Po-Chun-Chien: > This is a somewhat unexpected way to call a tool, maybe a comment would be nice that explains why this is done?

I've added some documentation.

> And is this backwards compatible or does it change the behavior of the tool compared to the previous version?

Yes, this is backward compatible.

> Using f-strings would be nicer than `.format()`.

Done in c2ccbdd.

benchexec/tools/abc.py
Po-Chun-Chien: @nianzelee what is the intended behavior here?
Since we (usually) run only one analysis at a time. It does not matter if we return earlier or not.

nianzelee: I think we can return when a match is found.

Po-Chun-Chien: Addressed in e31c8f0 and d0aa31a

Po-Chun-Chien:
I just tested AVR and ABC locally with the tool info-module in this MR. They are working as expected.

PhilippWendler (approved):

nianzelee (approved):
I checked the changes, and they look good.

PR #873, opened 2022-11-06
Tool-info module for Mopsa: Shorten Status Message

dbeyer:
Remove large volume of log dumping from status message.

The variable `r` can contain huge amounts of log dumps.
The status message should be very brief, to be shown in a table column.

The command that is called is always available as first line in the log output,
and further information can be retrieved from the log output as well.
No need to put this into the status.

Example status string:
`OUT OF MEMORY (ERROR(unknown: running /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/bin/mopsa-c -config=c/cell-string-length-itv-congr.json -additional-stubs=c/sv-comp.c -use-stub=reach_error -no-warning -ccopt=-fbracket-depth=2048 -stub-ignore-case=malloc.failure,malloc.empty -c-check-signed-implicit-cast-overflow=false -c-check-unsigned-implicit-cast-overflow=false -c-check-unreachable-memory ../../sv-benchmarks/c/hardware-verification-bv/btor2c-lazymod.unknown_ridecore.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/assert.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/sys/socket.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/netinet/in.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/arpa/inet.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/math.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/stdio.c /tmp/vcloud_worker_vcloud-master_on_vcloud-master/run_dir_595f7703-0f2a-48bd-84c7-311f3fc051fb/bin/mopsa-cfsnv5euho/share/mopsa/stubs/c/libc/stdlib.c -format=json))`

### DIFF
--- a/benchexec/tools/mopsa.py
+++ b/benchexec/tools/mopsa.py
@@ -44,4 +44,4 @@ class Tool(benchexec.tools.template.BaseTool2):
         elif r.startswith("error"):
             return result.RESULT_ERROR + r[len("ERROR") :]
         else:
-            return result.RESULT_ERROR + f"(unknown: {r})"
+            return result.RESULT_ERROR + "(unknown)"


PhilippWendler:
@rmonat Do you agree with this change? If no, please provide an alternative implementation that ensures that the result of `determine_result` is not a long string.

rmonat:
Sorry for breaking the layout in benchexec's columns. This seem to only happen in case of out of memory terminations. Is there a way to detect it on `run`?

PhilippWendler:
Besides the exit code of the tool, you have access to the fact whether the run was terminated by BenchExec due to a violation of resource limits, or whether it terminated itself. However, the previous code just always dumped the last line of the log, which could likely be any possible other output in many more situations than just OOM, right?

So if you want to add something, then the tool-info module should specifically extract this information from the log. For example, other tool-info modules check whether something specific occurs in the log (like an indication of a crash with an exception, a failing assertion, a segfault, etc.) and then indicate this briefly in the status value.

PR #877, opened 2022-11-15
Legion/SymCC update

gernst:


benchexec/tools/legion-symcc.py
```diff
@@ -17,8 +18,10 @@ class Tool(benchexec.tools.template.BaseTool2):
     REQUIRED_PATHS = [
         "legion.sh",
         "Legion.py",
-        "__VERIFIER.c",
-        "ubuntu2004",
+        "Verifier.cpp",
+        "legion",
+        "lib",
+        "lib32",
```
PhilippWendler, line 24: I would suggest to merge the old and new list, such that it works with both version.

benchexec/tools/legion-symcc.py
```diff
@@ -41,5 +44,12 @@ def cmdline(self, executable, options, task, rlimits):
...
-                return line[len(identifier) :]
+                value = line[len(identifier) :]
+                return value.strip()
         return None
+
+    def determine_result(self, run):
+        for line in run.output:
+            if "reach_error() detected." in line:
+                return result.RESULT_FALSE_REACH
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 55: Is "unknown" desired for all other results? Previously it was "done".

The difference is supposed to be as follows: If the tool is given a task that can be answered with true/false (e.g., "does this property hold?"), then "unknown" is the answer for "don't know". If the tool is just given the task to do something (no actual answer expected), then "done" is the answer for "completed successfully".

benchexec/tools/legion-symcc.py
gernst: Uh, but missing required files gives an error, doesn't it? The current archive does not have the old files any more.

PhilippWendler: No, these are just patterns for file names, and as such non-matching patterns are ignored.

PR #886, opened 2022-12-09
Fix scoring and add new test cases for score computation 

matthiaskettl:
Fixes scoring and introduces 5 new unit tests based on real data.


contrib/mergeBenchmarkSets.py
```diff
@@ -98,7 +98,7 @@ def get_witness_result(witness, verification_result):
     if status_from_validation == status_from_verification:
         return status_from_verification, category_from_verification
     # An invalid witness counts as error of the verifier.
-    if status_from_validation == "ERROR (invalid witness syntax)":
+    if status_from_validation.startswith("ERROR"):
```
PhilippWendler, line 101: Please note my previous comment:

I think it is possible that a validator returns `ERROR` due to an internal error (e.g., because it cannot parse the program). In this case the witness shouldn't be labeled as invalid, right?

PhilippWendler (commented):
1. Please fix CI.
2. How does this "fix scoring"? In order for this to be reviewable now and understandable later on, I think this should be explained.
3. From what I currently understand about the problem and from the single change in this PR, I do not think it can be fixed by this change. Whether a validator returns "invalid witness syntax" or not should not be able nor necessary to influence scoring of those results where no validation run succeeds.

contrib/mergeBenchmarkSets.py
MartinSpiessl: Also note that the invalid witness syntax really should only be decided by the linter, not by some individual validator. If a validator falsely claims that the witness syntax is invalid, but all other validators do not do this, how do we detect/handle this? Especially cases where the linter says the witness is alright but a validator claims the syntax is invalid. Currently we have the following ensured by the current state of the tool-info modules:
```
benchexec/benchexec/tools $ grep invalid -rHn
fshell-witness2test.py:106:            status += " (invalid witness file)"
smack.py:85:            if run.output.any_line_contains("invalid pointer dereference"):
smack.py:87:            elif run.output.any_line_contains("invalid memory deallocation"):
ultimate.py:448:                    status += " (invalid witness file)"
cbmc.py:95:                # in this case an exception is expected as the XML is invalid
cpachecker.py:160:                    status += " (invalid witness file)"
cpachecker.py:173:                    status += " (invalid witness file)"
impara.py:88:                    # in this case an exception is expected as the XML is invalid
theta.py:64:                status = "ERROR (invalid parameter)"
witnesslint.py:44:            return result.RESULT_ERROR + " (invalid witness syntax)"
```

So only the witness linter is allowed to produce "invalid witness syntax", so we will never see this from another tool. Other validators will at most say "invalid witness file". This is an indirect way to filter for the linter's result only. While this is not ideal, I think it works for now. We need to take care though to ensure this in variant still holds after our changes, and people need to be aware about this (maybe add a comment that the linter is the only one to be expected to return this verdict).

MartinSpiessl (changes requested):
As advertised this MR really does 2 things:
1. fix scoring
2. add test cases

My review leads to the following conclusion on these two points:
1. I think the change to `mergeBenchmarkSets` needs to be clarified/improved, as Philipp already outlined. It is not clear how this fixes anything WITHOUT breaking other cases.
2. I gave detailed feedback on how the test cases could be improved. While this might look like overkill, I personally think it is worth it. Having the tests as a nicely readable high-level spec will make future maintenance much easier. This does not necessarily need to be addressed in this MR, we could also defer refactoring of the tests to later. Most important should be to fix the scoring, and to be reasonably sure that this fix actually improve things, for which the tests shall serve as evidence.

contrib/test_mergeBenchmarkSets/test_1/linter_1.xml
```diff
@@ -0,0 +1,50 @@
...
+for SV-COMP 2022, https://sv-comp.sosy-lab.org/2022/
+by dbeyer@wbeyer1.sosy.ifi.lmu.de
+on 2021-12-10T21:56+01:00
+based on the components
+git@gitlab.com:sosy-lab/sv-comp/archives-2022  git-describe: svcomp22-rc.1-0-g456ae18
+git@gitlab.com:sosy-lab/benchmarking/sv-benchmarks  git-describe: svcomp22-rc.2-0-g3ca99b0f41
+https://gitlab.com/sosy-lab/software/benchexec  git-describe: 3.10-7-g38a2f4c6
+git@gitlab.com:sosy-lab/benchmarking/competition-scripts  git-describe: svcomp22-rc.1-5-ga9e074e
+git@gitlab.com:sosy-lab/sv-comp/bench-defs.git  git-describe: svcomp22-rc.1-1-ga39fca2
+Archive: val_witnesslint.zip  DATE: 2021-11-14T13:58+01:00  SHA1: 5618535fa1...</description>
```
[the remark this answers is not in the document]
PhilippWendler: 2. All of `contrib/` stays out of the packages. And so the tests for stuff in `contrib/` should stay in `contrib/`.

PR #888, opened 2022-12-29
Double the score if a validator refutes or confirms a wrong witness

dbeyer:
ATTENTION: Do not squash because the commits are referenced in competition artifacts.

SV-COMP 2023 scoring schema is extended to double the score if a validator refutes or confirms a wrong witness.

Reference: https://doi.org/10.1007/978-3-031-22308-2_8 (page 171, last paragraph)

### DIFF
--- a/benchexec/result.py
+++ b/benchexec/result.py
@@ -111,10 +111,14 @@ class Property(collections.namedtuple("Property", "filename is_svcomp name")):
 
     __slots__ = ()  # reduce per-instance memory consumption
 
-    def compute_score(self, category, result):
+    def compute_score(self, category, result, is_invalid_witness=False):
         if not self.is_svcomp:
             return None
-        return _svcomp_score(category, result)
+        score = _svcomp_score(category, result)
+        if is_invalid_witness:
+            # If a validator refutes (confirms) a wrong witness, it gets double the score (reduction).
+            score *= 2
+        return score
 
     def max_score(self, expected_result):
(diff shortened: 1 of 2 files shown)


PhilippWendler:
There is a note warning that the commit hashes are important, but there are several force pushes to this branch. Is this intended and the current set of hashes the desired one?

dbeyer:
> There is a note warning that the commit hashes are important, but there are several force pushes to this branch. Is this intended and the current set of hashes the desired one?

First I thought that it is ugly to show others how I found my way to the final implementation of this, and I force-pushed.

But then I realized that I had ran experiments for SV-COMP that now have those intermediate commit hashes in the provenance section of the results XML files, and thus, I restored the history. It would be super-ugly for reproducibility if the SV-COMP results point to commit hashes that do not exist in the history of any branch or tag.

PhilippWendler:
Current state looks really nice! Would be fine to merge for me from a conceptual point of view.

dbeyer:
The corresponding change in the results post-processing are here:
https://gitlab.com/sosy-lab/benchmarking/competition-scripts/-/commit/8f0f55c3e9b772e11c89e396105136d5a7fcdff3

PhilippWendler (approved):

PhilippWendler:
Ready to merge for me, for you as well?

dbeyer:
Yes, ready to merge.

Issue #748, opened 2021-09-21
Use cache architecture for core assignment


PhilippWendler:
BenchExec currently supports machines with several CPUs, CPUs with several NUMA nodes, NUMA nodes with several physical cores, and physical cores with several virtual cores. However, modern large CPUs add another layer: Certain cores of a NUMA node can share L3 cache, but have a separate L3 cache from other cores on the same NUMA node.

Example architectures:
- http://developer.amd.com/wp-content/resources/56420.pdf (page 15)
- https://nas.nasa.gov/assets/nas/pdf/ams/2021/AMS_20210720_Hogan-ONeill.pdf (slides 11 and 17)

For performance and determinism, it is better if on such machines, core allocation takes the cache architecture into account, such that cores that share cache are assigned preferably to a run.

We can read this information from `/sys/devices/system/cpu/cpuX/topology/die_cpus_list` ([docs](https://www.kernel.org/doc/html/latest/admin-guide/abi-stable.html#file-srv-docbuild-lib-git-linux-stable-sysfs-devices-system-cpu)), which is next to the files we use for retrieving the other topology information.

We should cross-check with `hwloc-ls` and `hwloc-info` that we get it right.

PhilippWendler:
Actually, it seems that this is not what Linux calls the "die". We need some other way to distinguish such cases.

PR #892, opened 2023-01-19
Draft: Update core assignment algorithm in benchexec/resources.py

CGall42:
Referring to issue #748, the core assignment can now handle additional hierarchy layers (such as a shared L3 cache).
The addition of further layers can be implemented without knowing the exact topology of a machine - the hierarchy of the layers (CPUs, NUMA nodes, L3 caches, hyperthreading, etc) is determined by the algorithm.

### DIFF
--- a/benchexec/resources.py
+++ b/benchexec/resources.py
@@ -1,9 +1,4 @@
-# This file is part of BenchExec, a framework for reliable benchmarking:
-# https://github.com/sosy-lab/benchexec
-#
-# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
-#
-# SPDX-License-Identifier: Apache-2.0
+#resources.py
 
 """
 This module contains functions for computing assignments of resources to runs.
@@ -15,146 +10,147 @@ import logging
 import math
 import os
 import sys
+from functools import cmp_to_key
 
-from benchexec import cgroups
-from benchexec import util
+import cgroups
+import util
 
-__all__ = [
-    "check_memory_size",
-    "get_cpu_cores_per_run",
-    "get_memory_banks_per_run",
-    "get_cpu_package_for_core",
-]
-
-
-def get_cpu_cores_per_run(
-    coreLimit, num_of_threads, use_hyperthreading, my_cgroups, coreSet=None
-):
-    """
-    Calculate an assignment of the available CPU cores to a number
-    of parallel benchmark executions such that each run gets its own cores
-    without overlapping of cores between runs.
-    In case the machine has hyper-threading, this method tries to avoid
-    putting two different runs on the same physical core
-    (but it does not guarantee this if the number of parallel runs is too high to avoid it).
-    In case the machine has multiple CPUs, this method avoids
-    splitting a run across multiple CPUs if the number of cores per run
-    is lower than the number of cores per CPU
-    (splitting a run over multiple CPUs provides worse performance).
-    It will also try to split the runs evenly across all available CPUs.
-
-    A few theoretically-possible cases are not implemented,
-    for example assigning three 10-core runs on a machine
-    with two 16-core CPUs (this would have unfair core assignment
-    and thus undesirable performance characteristics anyway).
-
-    The list of available cores is read from the cgroup file system,
-    such that the assigned cores are a subset of the cores
-    that the current process is allowed to use.
-    This script does currently not support situations
-    where the available cores are asymmetrically split over CPUs,
-    e.g. 3 cores on one CPU and 5 on another.
-
-    @param coreLimit: the number of cores for each run
-    @param num_of_threads: the number of parallel benchmark executions
-    @param coreSet: the list of CPU cores identifiers provided by a user, None makes benchexec using all cores
-    @return a list of lists, where each inner list contains the cores for one run
-    """
+# prepping function, consider change of name
+def get_cpu_cores_per_run(  
+    coreLimit, num_of_threads, use_hyperthreading, my_cgroups, coreSet=None):
+    hierarchy_levels = []
     try:
-        # read list of available CPU cores
-        allCpus = util.parse_int_list(my_cgroups.get_value(cgroups.CPUSET, "cpus"))
+        # read list of available CPU cores (int)
+        allCpus_list = get_cpu_list(my_cgroups, coreSet)
 
-        # Filter CPU cores according to the list of identifiers provided by a user
-        if coreSet:
-            invalid_cores = sorted(set(coreSet).difference(set(allCpus)))
-            if len(invalid_cores) > 0:
-                raise ValueError(
-                    "The following provided CPU cores are not available: "
-                    + ", ".join(map(str, invalid_cores))
-                )
-            allCpus = [core for core in allCpus if core in coreSet]
+        # read & prepare hyper-threading information, filter superfluous entries
+        siblings_of_core = get_siblings_mapping (allCpus_list)
+        cleanList = []
+        for core in siblings_of_core:
+            if core not in cleanList:
+                for sibling in siblings_of_core[core]:
+                    if sibling != core:
+                        cleanList.append(sibling)
+        for element in cleanList:
+            siblings_of_core.pop(element)
+        # siblings_of_core will be added to hierarchy_levels list after sorting
 
-        logging.debug("List of available CPU cores is %s.", allCpus)
+        # read & prepare mapping of cores to L3 cache
+        # cores_of_L3cache = 
+        # hierarchy_levels.append(core_of_L3cache)
+        
+        # read & prepare mapping of cores to NUMA region
+        cores_of_NUMA_Region = get_NUMA_mapping (allCpus_list)
+        hierarchy_levels.append(cores_of_NUMA_Region)
 
-        # read mapping of core to memory region
-        cores_of_memory_region = collections.defaultdict(list)
-        for core in allCpus:
-            coreDir = f"/sys/devices/system/cpu/cpu{core}/"
-            memory_regions = _get_memory_banks_listed_in_dir(coreDir)
-            if memory_regions:
-                cores_of_memory_region[memory_regions[0]].append(core)
-            else:
-                # If some cores do not have NUMA information, skip using it completely
-                logging.warning(
-                    "Kernel does not have NUMA support. Use benchexec at your own risk."
-                )
-                cores_of_memory_region = {}
-                break
-        logging.debug("Memory regions of cores are %s.", cores_of_memory_region)
+        # read & prepare mapping of cores to group
+        # core_of_group =
+        #hierarchy_levels.append(core_of_group)
 
-        # read mapping of core to CPU ("physical package")
-        cores_of_package = collections.defaultdict(list)
-        for core in allCpus:
-            package = get_cpu_package_for_core(core)
-            cores_of_package[package].append(core)
-        logging.debug("Physical packages of cores are %s.", cores_of_package)
-
-        # select the more fine grained division among memory regions and physical package
-        if len(cores_of_memory_region) >= len(cores_of_package):
-            cores_of_unit = cores_of_memory_region
-            logging.debug("Using memory regions as the basis for cpu core division")
-        else:
-            cores_of_unit = cores_of_package
-            logging.debug("Using physical packages as the basis for cpu core division")
-
-        # read hyper-threading information (sibling cores sharing the same physical core)
-        siblings_of_core = {}
-        for core in allCpus:
-            siblings = util.parse_int_list(
-                util.read_file(
-                    f"/sys/devices/system/cpu/cpu{core}/topology/thread_siblings_list"
-                )
-            )
-            siblings_of_core[core] = siblings
-        logging.debug("Siblings of cores are %s.", siblings_of_core)
+        # read & prepare mapping of cores to CPU/physical package/socket?
+        cores_of_package = get_package_mapping (allCpus_list)
+        hierarchy_levels.append(cores_of_package)
+    
     except ValueError as e:
         sys.exit(f"Could not read CPU information from kernel: {e}")
-    return _get_cpu_cores_per_run0(
-        coreLimit,
-        num_of_threads,
-        use_hyperthreading,
+    
+    # generate sorted list of dicts in accordance with their hierarchy
+    def compare_hierarchy(dict1, dict2):
+        value1 = len(next(iter(dict1.values())))
+        value2 = len(next(iter(dict2.values())))
+        if value1 > value2:
+            return 1
+        if value1 < value2:
+            return -1
+        if value1 == value2:
+            return 0
+
+    # sort hierarchy_levels according to the dicts' corresponding unit sizes
+    hierarchy_levels.sort(key = cmp_to_key(compare_hierarchy))  #hierarchy_level = [dict1, dict2, dict3]
+    # add siblings_of_core at the beginning of the list
+    hierarchy_levels.insert(0, siblings_of_core)
+
+    # create v_cores
+    allCpus = {}
+    for cpu_nr in allCpus_list:
+        allCpus.update({cpu_nr: v_core(cpu_nr,[])})
+
+    for level in hierarchy_levels:                          # hierarchy_levels = [dict1, dict2, dict3]
+        for key in level:
+            for core in level[key]:
+                allCpus[core].memory_regions.append(key)    # memory_regions = [key1, key2, key3]
+
+
+    # call the actual assignment function
+    return assignmentAlgorithm (
+        coreLimit, 
+        num_of_threads, 
+        use_hyperthreading, 
         allCpus,
-        cores_of_unit,
         siblings_of_core,
+        hierarchy_levels
     )
 
+# define class v_core to generate core objects
+class v_core: 
+    #def __init__(self, id, siblings=None, memory_regions=None):
+    def __init__(self, id, memory_regions=None):
+        self.id = id
+        #self.siblings = siblings
+        self.memory_regions = memory_regions
+    def __str__(self):
+        return self.id + " " + self.memory_regions
 
-def _get_cpu_cores_per_run0(
-    coreLimit,
-    num_of_threads,
-    use_hyperthreading,
+# assigns the v_cores into specific runs
+def assignmentAlgorithm (
+    coreLimit,              
+    num_of_threads, 
+    use_hyperthreading, 
     allCpus,
-    cores_of_unit,
     siblings_of_core,
+    hierarchy_levels
 ):
     """This method does the actual work of _get_cpu_cores_per_run
     without reading the machine architecture from the file system
-    in order to be testable. For description, c.f. above.
-    Note that this method might change the input parameters!
-    Do not call it directly, call getCpuCoresPerRun()!
-    @param use_hyperthreading: A boolean to check if no-hyperthreading method is being used
+    in order to be testable.
+    @param coreLimit: the number of cores for each run
+    @param num_of_threads: the number of parallel benchmark executions
+    @param use_hyperthreading: boolean to check if no-hyperthreading method is being used
     @param allCpus: the list of all available cores
-    @param cores_of_unit: a mapping from logical unit (can be memory region (NUMA node) or physical package(CPU), depending on the architecture of system)
-                          to lists of cores that belong to this unit
-    @param siblings_of_core: a mapping from each core to a list of sibling cores including the core itself (a sibling is a core sharing the same physical core)
+    @param siblings_of_core: mapping from each core to list of sibling cores including the core itself
+    cores_of_L3cache,
+    cores_of_NUMA_Region,
+    core_of_group,
+    cores_of_package
     """
-    # First, do some checks whether this algorithm has a chance to work.
+    # First: checks whether the algorithm can & should work
+
+    # no HT filter: delete all but one the key core from siblings_of_core 
+    # delete those cores from all dicts in hierarchy_levels    
+    if not use_hyperthreading:
+        for core in siblings_of_core:
+            no_HT_filter = []
+            for sibling in siblings_of_core[core]:
+                if sibling != core:
+                    no_HT_filter.append(sibling)
+            for virtual_core in no_HT_filter:
+                siblings_of_core[core].remove(virtual_core)
+                region_keys = allCpus[virtual_core].memory_regions
+                i=1
+                while i < len(region_keys):
+                    hierarchy_levels[i][region_keys[i]].remove(virtual_core)
+                    i = i+1
+                allCpus.pop(virtual_core)
+
+    # compare number of available cores to required cores per run
     coreCount = len(allCpus)
     if coreLimit > coreCount:
         sys.exit(
-            f"Cannot run benchmarks with {coreLimit} CPU cores, "
-            f"only {coreCount} CPU cores available."
-        )
+                f"Cannot run benchmarks with {coreLimit} CPU cores, "
+                f"only {coreCount} CPU cores available."
+            )
+    
+    # compare number of available run to overall required cores
     if coreLimit * num_of_threads > coreCount:
         sys.exit(
             f"Cannot run {num_of_threads} benchmarks in parallel "
@@ -162,44 +158,9 @@ def _get_cpu_cores_per_run0(
             f"Please reduce the number of threads to {coreCount // coreLimit}."
         )
 
-    if not use_hyperthreading:
-        unit_of_core = {}
-        unused_cores = []
-        for unit, cores in cores_of_unit.items():
-            for core in cores:
-                unit_of_core[core] = unit
-        for core, siblings in siblings_of_core.items():
-            if core in allCpus:
-                siblings.remove(core)
-                cores_of_unit[unit_of_core[core]] = [
-                    c for c in cores_of_unit[unit_of_core[core]] if c not in siblings
-                ]
-                siblings_of_core[core] = [core]
-                allCpus = [c for c in allCpus if c not in siblings]
-            else:
-                unused_cores.append(core)
-        for core in unused_cores:
-            siblings_of_core.pop(core)
-        logging.debug(
-            "Running in no-hyperthreading mode, avoiding the use of CPU cores %s",
-            unused_cores,
-        )
 
-    unit_size = len(next(iter(cores_of_unit.values())))  # Number of units per core
-    if any(len(cores) != unit_size for cores in cores_of_unit.values()):
-        sys.exit(
-            "Asymmetric machine architecture not supported: "
-            "CPUs/memory regions with different number of cores."
-        )
-
-    core_size = len(next(iter(siblings_of_core.values())))  # Number of threads per core
-    if any(len(siblings) != core_size for siblings in siblings_of_core.values()):
-        sys.exit(
-            "Asymmetric machine architecture not supported: "
-            "CPU cores with different number of sibling cores."
-        )
-
-    all_cpus_set = set(allCpus)
+    # check if all HT siblings are available for benchexec
+    all_cpus_set = set(allCpus.keys())
     for core, siblings in siblings_of_core.items():
         siblings_set = set(siblings)
         if not siblings_set.issubset(all_cpus_set):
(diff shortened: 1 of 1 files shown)


benchexec/resources.py
```diff
@@ -1,9 +1,4 @@
-# This file is part of BenchExec, a framework for reliable benchmarking:
-# https://github.com/sosy-lab/benchexec
-#
-# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
-#
-# SPDX-License-Identifier: Apache-2.0
+#resources_new.py
```
PhilippWendler, line 1: Please restore the original comment with the copyright information.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
 import math
 import os
 import sys
+from functools import cmp_to_key
 
-from benchexec import cgroups
-from benchexec import util
+import cgroups
+import util
```
PhilippWendler, line 16: Please use the original way the imports are written and not relative imports. For example, right now the tests in CI cannot be executed due to this.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
 import math
 import os
 import sys
+from functools import cmp_to_key
```
PhilippWendler, line 13: I would suggest to just import `functools`, it seems not worth to import `cmp_to_key` directly into our namespace because we use it only once. And please put the import in alphabetical order.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
-from benchexec import util
+import cgroups
+import util
 
-__all__ = [
-    "check_memory_size",
-    "get_cpu_cores_per_run",
-    "get_memory_banks_per_run",
-    "get_cpu_package_for_core",
-]
```
PhilippWendler, line 27: Why is this removed? It declares which functions in this file are supposed to be called from outside.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
+    #def __init__(self, id, siblings=None, memory_regions=None):
+    def __init__(self, id, memory_regions=None):
+        self.id = id
+        #self.siblings = siblings
+        self.memory_regions = memory_regions
+    def __str__(self):
+        return str(self.id) + " " + str(self.memory_regions)
+
+# assigns the v_cores into specific runs
+def assignmentAlgorithm (
```
PhilippWendler, line 105: At many times, it is best if a function name contains a verb because it should describe what the function *does*. Can you please ensure this?

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
-    where the available cores are asymmetrically split over CPUs,
-    e.g. 3 cores on one CPU and 5 on another.
-
-    @param coreLimit: the number of cores for each run
-    @param num_of_threads: the number of parallel benchmark executions
-    @param coreSet: the list of CPU cores identifiers provided by a user, None makes benchexec using all cores
-    @return a list of lists, where each inner list contains the cores for one run
-    """
+# prepping function, consider change of name
+def get_cpu_cores_per_run(
```
PhilippWendler, line 19: Text explaining functions for callers should be doc strings (string at first line of the function) and not comments. This actually attaches the text with the explanation to the function in Python.

Please use this style for all functions, and extensively document them, their requirements, their guarantees, and their parameters and return values.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
     )
 
-
-def _get_cpu_cores_per_run0(
-    coreLimit,
-    num_of_threads,
-    use_hyperthreading,
+# define class v_core to generate core objects
+class v_core: 
+    #def __init__(self, id, siblings=None, memory_regions=None):
```
PhilippWendler, line 96: Please clean up the code and removed outdated code in comments (in other places as well).

benchexec/resources.py
```diff
@@ -209,145 +170,191 @@ def _get_cpu_cores_per_run0(
...
+                "CPUs/memory regions with different number of cores."
+            )
 
-    # Second, compute some values we will need.
-    unit_count = len(cores_of_unit)
-    units = sorted(cores_of_unit.keys())
+    # ( compute some values we will need.)
+
+    # coreLimit_rounded_up (int): recalculate # cores for each run accounting for HT
+    core_size =  len(next(iter(siblings_of_core.values()))) # Wert aus hierarchy_level statt siblings_of_core?
```
PhilippWendler, line 186: Please make sure everything is in English.

benchexec/resources.py
```diff
@@ -209,145 +170,191 @@ def _get_cpu_cores_per_run0(
...
-    units_per_run = int(math.ceil(coreLimit_rounded_up / unit_size))
-    if units_per_run > 1 and units_per_run * num_of_threads > unit_count:
+
+    # Choose hierarchy level for core assignment
+    chosen_level = 1
+    # move up in hierarchy as long as the number of cores at the current level is smaller than the coreLimit
+    # if the number of cores at the current level is as big as the coreLimit: exit loop 
+    while len (next(iter(hierarchy_levels[chosen_level].values()))) < coreLimit_rounded_up  and chosen_level < len (hierarchy_levels):
+        chosen_level = chosen_level+1
+    print("chosen_level = ",chosen_level)
```
PhilippWendler, line 197: Please do not use `print()`, this messes up the output of BenchExec.

Debug output can be made with `logging.debug()`, but it needs to be written in a way that is useful for (experienced) users and other developers of BenchExec to understand.

benchexec/resources.py
```diff
@@ -209,145 +170,191 @@ def _get_cpu_cores_per_run0(
...
+    # calculate runs per unit of hierarchy level i
+    runs_per_unit = int(math.floor(unit_size/coreLimit_rounded_up))
+    
+    # compare num of units & runs per unit vs num_of_threads
+    if len(hierarchy_levels[chosen_level]) * runs_per_unit < num_of_threads:
         sys.exit(
-            f"Cannot split runs over multiple CPUs/memory regions "
-            f"and at the same time assign multiple runs to the same CPU/memory region. "
-            f"Please reduce the number of threads to {unit_count // units_per_run}."
+            f" .........................."
```
PhilippWendler, line 209: I don't think this line of dots improves the output.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
+        hierarchy_levels
     )
 
-
-def _get_cpu_cores_per_run0(
-    coreLimit,
-    num_of_threads,
-    use_hyperthreading,
+# define class v_core to generate core objects
+class v_core:
```
PhilippWendler, line 95: Class names are typically written in CamelCase. Also, what does "v" mean? virtual?

We could simply use "Core", because some systems might not even have "virtual" cores.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
-
-    @param coreLimit: the number of cores for each run
-    @param num_of_threads: the number of parallel benchmark executions
-    @param coreSet: the list of CPU cores identifiers provided by a user, None makes benchexec using all cores
-    @return a list of lists, where each inner list contains the cores for one run
-    """
+# prepping function, consider change of name
+def get_cpu_cores_per_run(  
+    coreLimit, num_of_threads, use_hyperthreading, my_cgroups, coreSet=None):
+    hierarchy_levels = []
```
PhilippWendler, line 21: There definitively needs to be an extensive explanation of this crucial data structure. I would even suggest that there is a concrete example.

benchexec/resources.py
```diff
@@ -209,145 +170,191 @@ def _get_cpu_cores_per_run0(
                 f"of core {core} are not usable. "
                 f"Please always make all virtual cores of a physical core available."
             )
+    # check if all units of the same hierarchy level have the same number of cores
+    for index in range(len(hierarchy_levels)): # [dict, dict, dict, ...]
```
PhilippWendler, line 174: It seems here to use a for loop with a loop counter, but you only use the loop counter to access `hierarchy_levels[index]`. So you could also simply use a `for hierarchy_level in hierarchy_levels:` loop. In general, the latter style is preferred and should be always used when possible.

benchexec/resources.py
```diff
@@ -15,191 +10,157 @@
...
+        value2 = len(next(iter(dict2.values())))
+        if value1 > value2:
+            return 1
+        if value1 < value2:
+            return -1
+        if value1 == value2:
+            return 0
+
+    # sort hierarchy_levels according to the dicts' corresponding unit sizes
+    hierarchy_levels.sort(key = cmp_to_key(compare_hierarchy))  #hierarchy_level = [dict1, dict2, dict3]
```
PhilippWendler, line 69: An easier way to write this is to avoid `cmp_to_key` and just provide a function that returns an appropriate sorting key. Right now, `compare_hierarchy` looks up what should be compared (then lengths), then compares them, and then `cmp_to_key` undoes the comparison in order to provide some sort keys that `.sort()` can use. Just directly tell `.sort()` what it should compare on (and make sure to use `reverse` if necessary. Note that if the docs of `cmp_to_key` tell you to not use it in new code.

PhilippWendler (changes requested):
This is a preliminary review with some hints, mostly on code style and documentation. Please fix these issues and provide documentation, such that a full review becomes possible and makes sense.

As part of this (as first step actually), please format the source code with the formatter [black](https://pypi.org/project/black) and make sure to use it in the future for each commit. Consistent code formatting is a big help for readability.

And please also have a look at all the other [CI failures](https://gitlab.com/sosy-lab/software/benchexec/-/pipelines/748593523). After each commit these checks will run automatically, so always check whether CI is green for your most recent commit. The check `check-format` is fixed automatically if you use the code formatter, and `reuse` only complains about the missing copyright header. But `flake8` and `pytype` provide hints about potential errors in your code. And of course the `unit-tests` checks just execute our tests.

benchexec/resources.py
PhilippWendler: Right now it is `virtualCore`, which is fine but it should confirm to Python's naming convention for classes, which would mean `VirtualCore` with capital letter at the beginning.

benchexec/resources.py
CGall42: Done

benchexec/resources.py
CGall42: Renamed to `VirtualCore`

benchexec/resources.py
CGall42: Edited loop and corresponding asymmetry check in https://github.com/sosy-lab/benchexec/pull/892/commits/943aaebd910dd09583a0f5b44b0d4bd75b30fae2

benchexec/resources.py
```diff
@@ -209,109 +408,432 @@ def _get_cpu_cores_per_run0(
...
+    cpu_max_frequencies = collections.defaultdict(list)
+    for core in allCpus_list:
+        max_freq = int(
+            util.read_file(
+                f"/sys/devices/system/cpu/cpu{core}/cpufreq/cpuinfo_max_freq"
+            )
         )
+        cpu_max_frequencies[max_freq].append(core)
+    available_max_freq = sorted(cpu_max_frequencies.keys())
+    freq_threshold = available_max_freq[-1] * (1 - threshold)
```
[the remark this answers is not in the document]
CGall42: Implemented in https://github.com/sosy-lab/benchexec/pull/892/commits/ec4d37f4c9ad6af4d6a616c7c871dcf5bfa5d5fa

benchexec/resources.py
```diff
@@ -209,109 +408,432 @@ def _get_cpu_cores_per_run0(
...
+                "The following provided CPU cores are not available: "
+                + ", ".join(map(str, invalid_cores))
+            )
+        allCpus_list = [core for core in allCpus if core in coreSet]
+    allCpus_list = frequency_filter(allCpus, 0.05)
+    logging.debug("List of available CPU cores is %s.", allCpus)
+    return allCpus_list
+
+
+def frequency_filter(allCpus_list, threshold):
```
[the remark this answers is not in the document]
CGall42: Implemented in https://github.com/sosy-lab/benchexec/pull/892/commits/ec4d37f4c9ad6af4d6a616c7c871dcf5bfa5d5fa

Please advise whether the explanation is sufficient or which information is expected.

benchexec/resources.py
```diff
@@ -209,109 +408,432 @@ def _get_cpu_cores_per_run0(
...
+# Hilfsfunktion, um die nodes mit der geringsten Distance auszulesen
+def get_closest_nodes(distance_list):
+    smallest_distance = sys.maxsize
+    second_to_smallest = sys.maxsize
+    for distance in distance_list:
+        if distance < smallest_distance:
+            second_to_smallest = smallest_distance
+            smallest_distance = distance
+        if distance < second_to_smallest and distance != smallest_distance:
+            second_to_smallest = distance
```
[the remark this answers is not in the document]
CGall42: Implemented in https://github.com/sosy-lab/benchexec/pull/892/commits/a7128fd4eb4d55f3623a9ca6563da388d48946f1

benchexec/resources.py
```diff
@@ -57,149 +63,332 @@ def get_cpu_cores_per_run(
...
+
+def check_distribution_feasibility(
     coreLimit,
     num_of_threads,
     use_hyperthreading,
     allCpus,
-    cores_of_unit,
     siblings_of_core,
+    hierarchy_levels,
+    isTest=True,
```
PhilippWendler, line 248: The less production code is affected by specific needs of the test, the better. It is possible to do this if it is necessary, though. But here I don't really see a need: Couldn't the test simply check that the exception is thrown instead?

benchexec/resources.py
```diff
@@ -57,149 +63,332 @@ def get_cpu_cores_per_run(
...
+    sub_units_per_run = calculate_sub_units_per_run(
+        coreLimit_rounded_up, hierarchy_levels, chosen_level
+    )
+    print("sub_units_per_run = ", sub_units_per_run)
+    # number of nodes at subunit-Level / sub_units_per_run
+    if len(hierarchy_levels[chosen_level - 1]) / sub_units_per_run < num_of_threads:
+        if not isTest:
+            sys.exit(
+                f"Cannot split memory regions between runs. "
+                f"Please reduce the number of threads to {math.floor(len(hierarchy_levels[chosen_level-1]) / sub_units_per_run)}."
```
PhilippWendler, line 306: For long expressions like this (that exceed the line length) I guess it is better readable if the expression is stored in an additional variable first.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+
+    def compare_hierarchy_by_dict_length(level: HierarchyLevel):
+        """comparator function for number of elements in a dict's value list"""
+        return len(next(iter(level.values())))
+
+    hierarchy_levels.sort(key=compare_hierarchy_by_dict_length, reverse=False)
+    """sort hierarchy_levels (list of dicts) according to the dicts' value sizes"""
+
+    # add siblings_of_core at the beginning of the list to ensure the correct index
+    hierarchy_levels.insert(0, siblings_of_core)
```
PhilippWendler, line 152: I am wondering what the problem here could be? What would need to happen such that siblings land in the wrong place? And in any case like this (e.g., if there are siblings that have different L3 caches), wouldn't the allocation be meaningless anyway?

I guess it would be better to actually check for this instead of blindly inserting `siblings_of_core` in the first position. One could for example insert `siblings_of_core` normally, then sort, and then check that it is in the expected position.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+
+    # create VirtualCores
+    allCpus = {}
+    """creates a dict of VirtualCore objects from core ID list"""
```
PhilippWendler, line 156: Docstrings are supposed to be used at start of functions and describe the function. They are not useful inside functions, use regular comments instead.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+    for cpu_nr in allCpus_list:
+        allCpus.update({cpu_nr: VirtualCore(cpu_nr, [])})
```
PhilippWendler, line 158: Isn't this simply `allCpus = {cpu_nr: VirtualCore(cpu_nr, []) for cpu_nr in allCpus_list}`?

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+        coreLimit_rounded_up / len(hierarchy_levels[chosen_level - 1][0])
+    )
+    return sub_units_per_run
+
+
+def check_and_add_meta_level(
+    hierarchy_levels: List[HierarchyLevel], allCpus: Dict[int, VirtualCore]
+) -> None:
+    """
+    Adds a meta_level to hierarchy_levels to iterate through all cores (if necessary)
```
PhilippWendler, line 407: What is a `meta_level` and what purpose does it have? This term never occurs anywhere else.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+    """
+    if len(hierarchy_levels[-1]) > 1:
+        top_level_cores = []
+        for node in hierarchy_levels[-1]:
+            top_level_cores.extend(hierarchy_levels[-1][node])
+        hierarchy_levels.append({0: top_level_cores})
+        for cpu_nr in allCpus:
+            allCpus[cpu_nr].memory_regions.append(0)
```
PhilippWendler, line 415: It seems this adds the memory region `0` to every core. But isn't this blatantly wrong? Why is this here, without any explanation?

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+    def compare_hierarchy_by_dict_length(level: HierarchyLevel):
+        """comparator function for number of elements in a dict's value list"""
+        return len(next(iter(level.values())))
+
+    hierarchy_levels.sort(key=compare_hierarchy_by_dict_length, reverse=False)
+    """sort hierarchy_levels (list of dicts) according to the dicts' value sizes"""
+
+    # add siblings_of_core at the beginning of the list to ensure the correct index
+    hierarchy_levels.insert(0, siblings_of_core)
+
```
PhilippWendler, line 153: Wouldn't it be a good idea to check that every hierarchy level contains the same set of cores?

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+    return group_list  # [0 1 2 3]
+
+
+def get_cluster_id_for_core(core: int) -> int:
+    """Get the id of the cluster a core belongs to."""
+    return int(util.read_file(f"/sys/devices/system/cpu/cpu{core}/topology/cluster_id"))
+
+
+def get_cluster_mapping(allCpus_list: List[int]) -> HierarchyLevel:
+    cores_of_cluster = collections.defaultdict(list)  # Zuordnung DIE ID zu core ID
```
PhilippWendler, line 825: Please really make sure to not use German anywhere.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+def get_L3cache_mapping(allCpus_list: List[int]) -> HierarchyLevel:
+    cores_of_L3cache = collections.defaultdict(list)
+    try:
+        for core in allCpus_list:
+            L3cache = get_L3cache_id_for_core(core)
+            cores_of_L3cache[L3cache].append(core)
+    except FileNotFoundError:
+        cores_of_L3cache = {}
+        logging.error(
+            "Level 3 cache information not available at /sys/devices/system/cpu/cpuX/cache/cacheX"
```
PhilippWendler, line 910: Why is this an error? An error means that BenchExec cannot continue to work. Is this the case? If yes, then an exception should be thrown. If not, then it should not be an error message.

Is this something important for the user? If not, then a debug message should be used.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+        if NUMA_regions:
+            cores_of_NUMA_region[NUMA_regions[0]].append(core)
+            # adds core to value list at key [NUMA_region[0]]
+        else:
+            # If some cores do not have NUMA information, skip using it completely
+            logging.warning(
+                "Kernel does not have NUMA support. Use benchexec at your own risk."
+            )
+            cores_of_NUMA_region = {}
+            break
```
PhilippWendler, line 931: It is easier to understand what happens if simply `return {}` is used instead. There is already a log message above, no reason for having another debug message with an empty set of cores.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
-        # select the more fine grained division among memory regions and physical package
-        if len(cores_of_memory_region) >= len(cores_of_package):
-            cores_of_unit = cores_of_memory_region
-            logging.debug("Using memory regions as the basis for cpu core division")
-        else:
-            cores_of_unit = cores_of_package
-            logging.debug("Using physical packages as the basis for cpu core division")
+        # read & prepare mapping of cores to L3 cache
+        cores_of_L3cache = get_L3cache_mapping(allCpus_list)
+        hierarchy_levels.append(cores_of_L3cache)
```
PhilippWendler, line 104: Why are in some cases the return values of these methods checked, and in others not?

CGall42: The thought behind that was that some methods may return empty lists if not supported by the processor (e.g. NUMA) while there should always be a socket or a level 3 cache (even if it is not shared). 
This is changed in commit https://github.com/sosy-lab/benchexec/pull/892/commits/4b5c08435488c589cb62875273ce834abb56d688 : now all return values are checked.

benchexec/resources.py
CGall42: Done

benchexec/resources.py
CGall42: Changed into a debug message

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+    logging.debug("Siblings of cores are %s.", siblings_of_core)
+    return siblings_of_core
+
+
+def get_die_id_for_core(core: int) -> int:
+    """Get the id of the die a core belongs to."""
+    return int(util.read_file(f"/sys/devices/system/cpu/cpu{core}/topology/die_id"))
+
+
+def get_die_mapping(allCpus_list: List[int]) -> HierarchyLevel:
```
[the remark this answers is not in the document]
CGall42: Refactored code in commit https://github.com/sosy-lab/benchexec/pull/892/commits/4b5c08435488c589cb62875273ce834abb56d688

benchexec/resources.py
CGall42: Yes, it adds a `0` to the memory region list of each core, because on the highest hierarchy level that we just added, all cores belong to the unit with id `0` (see linie 413). will add an explanation in the next commit.

benchexec/resources.py
CGall42: Adds a meta_level or root_level which includes all cores to hierarchy_levels (if necessary).
This is necessary to iterate through all cores if the highest hierarchy level consists of more than one unit.
Also adds the identifier for the new level to the memory region of all cores in allCpus

benchexec/resources.py
CGall42: Looks like it. I personally can grasp the code with a "classic" for loop easier.

benchexec/resources.py
CGall42: A theoretic case where the sibling could land in the "wrong" place would be a topology without hyperthreading siblings and without shared L3 caches (each core has its own). Come to think about it, then it probably wouldnt make a difference.  
We now check whether there are identical levels and delete one of them anyway. 
The insert follows the thought of "better safe than sorry" and makes it easy to understand that hierarchy_levels[0] is the HT siblings.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+        # read & prepare hyper-threading information, filter redundant entries
+        siblings_of_core = get_siblings_mapping(allCpus_list)
+        cleanList = []
+        for core in siblings_of_core:
+            if core not in cleanList:
+                for sibling in siblings_of_core[core]:
+                    if sibling != core:
+                        cleanList.append(sibling)
+        for element in cleanList:
+            siblings_of_core.pop(element)
```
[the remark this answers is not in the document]
CGall42: the original siblings_of_core dict has two key-value pairs for the same HT siblings list: { 0: [0,128] } and { 128: [0,128] },
the second key-value pair is deleted. Due to the fact that one cannot delete an entry from a list one iterates through, the key is first added to the clean list and then deleted after the for loop terminates.

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+
+    if (
+        type(coreLimit) != int
+        or type(num_of_threads) != int
+        or type(use_hyperthreading) != bool
+    ):
+        sys.exit("Incorrect data type entered")
+
+    if coreLimit < 1 or num_of_threads < 1:
+        sys.exit("Only integers > 0 accepted for coreLimit & num_of_threads")
```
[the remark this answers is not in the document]
CGall42: It is supposed to intercept wrong input from users. 
How should this be changed? A different error message for each input? 
Or what do you refer to here?

benchexec/resources.py
CGall42: This comment exists twice, closing this one

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+                f"/sys/devices/system/cpu/cpu{core}/cpufreq/cpuinfo_max_freq"
+            )
+        )
+        cpu_max_frequencies[max_freq].append(core)
+    freq_threshold = max(cpu_max_frequencies.keys()) * (1 - threshold)
+    for key in cpu_max_frequencies:
+        if key < freq_threshold:
+            for core in cpu_max_frequencies[key]:
+                allCpus_list.remove(core)
+    return allCpus_list
```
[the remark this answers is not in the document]
CGall42: Changed into:
```
freq_threshold = max(cpu_max_frequencies.keys()) * (1 - threshold)
filtered_allCpus_list = []
    slow_cores = []
    for key in cpu_max_frequencies:
        if key >= freq_threshold:
            filtered_allCpus_list.extend(cpu_max_frequencies[key])
        else:
            slow_cores.extend(cpu_max_frequencies[key])
    logging.debug(f"Unused cores due to frequency more than {threshold*100}% below frequency of fastest core ({max(cpu_max_frequencies.keys())}): {slow_cores}")
    return filtered_allCpus_list
```

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+    This script does currently not support situations where the available cores are
+    asymmetrically split over CPUs, e.g. 3 cores on one CPU and 5 on another.
+
+    @param coreLimit:           the number of cores for each thread
+    @param num_of_threads:      the number of parallel benchmark executions
+    @param use_hyperthreading:  boolean to check if no-hyperthreading method is being used
+    @param coreSet:             the list of CPU core identifiers provided by a user,
+                                None makes benchexec using all cores
+    @return hierarchy_levels:   list of dicts of lists: each dict in the list corresponds to one topology layer
+                                and maps from the identifier read from the topology to a list of the cores belonging to it
```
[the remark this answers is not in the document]
CGall42: True, the function returns the core allocation list, at some point before, this is outdated => changed

benchexec/resources.py
```diff
@@ -15,191 +15,199 @@
...
+
 __all__ = [
     "check_memory_size",
     "get_cpu_cores_per_run",
     "get_memory_banks_per_run",
     "get_cpu_package_for_core",
 ]
 
+# typing defintions
+ListOfIntLists = List[List[int]]
```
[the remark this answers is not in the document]
CGall42: `ListOfIntLists` ist used only once:
```
def get_memory_banks_per_run(coreAssignment, cgroups) -> Optional[ListOfIntLists]:
    """Get an assignment of memory banks to runs that fits to the given coreAssignment,
    i.e., no run is allowed to use memory that is not local (on the same NUMA node)
    to one of its CPU cores."""
```
It may be longer, but it doesn't act up in CICD like the predecessor:
```
./benchexec/resources.py:943:6: TAE002 too complex annotation (4 > 3)
) -> Optional[List[List[int]]]:
```
Please suggest a useful name for it?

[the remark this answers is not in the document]
PhilippWendler: There is no need in complicating the code just to please CI. It would be valuable to assign it a type alias that has a meaningful name, but not `ListOfIntLists`. So if you don't know a better name, then keep it as is (it is pre-existing code, after all), and silence the warning.

benchexec/resources.py
```diff
@@ -15,313 +15,916 @@
...
-            invalid_cores = sorted(set(coreSet).difference(set(allCpus)))
-            if len(invalid_cores) > 0:
-                raise ValueError(
-                    "The following provided CPU cores are not available: "
-                    + ", ".join(map(str, invalid_cores))
-                )
-            allCpus = [core for core in allCpus if core in coreSet]
+        # read list of available CPU cores (int)
+        allowedCpus = get_cpu_list(my_cgroups)
+        allCpus_list = get_cpu_list(my_cgroups, coreSet)
```
PhilippWendler, line 87: I notice that this function is called twice, and think that this will lead to highly confusing log output, because the methods logs all cores, and so there will be two log messages with potentially different sets of cores, both claiming to be the available cores. Additionally, there are the log messages by the frequency filter.

I think this should be refactored such that the cores are read just once, logged just once, and the frequency filter is applied and logged just once.

benchexec/resources.py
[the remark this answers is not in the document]
PhilippWendler: This is better.

Btw.: Is the order of cores in `allCpus` relevant for the algorithm? This method will change it, maybe that is unwanted/problematic?

benchexec/resources.py
```diff
@@ -209,109 +398,444 @@ def _get_cpu_cores_per_run0(
...
-            ]
-
-        assert len(cores) == coreLimit, (
-            f"Wrong number of cores for run {run + 1} of {num_of_threads} "
-            f"- previous results: {result}, "
-            f"remaining cores per CPU/memory region: {cores_of_unit}, "
-            f"current cores: {cores}"
+                )
+                == 3
+            ):
```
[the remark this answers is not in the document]
CGall42: Done in commit https://github.com/sosy-lab/benchexec/pull/892/commits/4b5c08435488c589cb62875273ce834abb56d688

benchexec/resources.py
```diff
@@ -209,109 +398,444 @@ def _get_cpu_cores_per_run0(
...
-            )
+        raise Exception("More then one smallest distance")
+    if distance_list.count(second_to_smallest) == 1:
+        group_list.append(distance_list.index(second_to_smallest))
+    elif distance_list.count(second_to_smallest) > 1:
+        index = 0
+        for dist in distance_list:
+            if dist == second_to_smallest:
+                group_list.append(index)
+            index += 1
```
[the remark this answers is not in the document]
CGall42: Fixed in https://gitlab.com/sosy-lab/software/benchexec/-/pipelines/902433871
```
    if second_to_smallest != greatest_distance:
        index = 0
        for dist in distance_list:
            if dist == second_to_smallest:
                group_list.append(index)
            index += 1
```

benchexec/resources.py
```diff
@@ -209,109 +408,432 @@ def _get_cpu_cores_per_run0(
...
+        group_list.append(distance_list.index(smallest_distance))
     else:
-        if coreLimit_rounded_up * num_of_threads > len(allCpus):
-            assert coreLimit_rounded_up * runs_per_unit > unit_size
-            need_HT = True
-            logging.warning(
-                "The number of threads is too high and hyper-threading sibling cores need to be split among different runs, which makes benchmarking unreliable. Please reduce the number of threads to %s.",
-                len(allCpus) // coreLimit_rounded_up,
-            )
+        raise Exception("More then one smallest distance")
```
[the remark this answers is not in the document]
CGall42: Explanation extended in https://github.com/sosy-lab/benchexec/pull/892/commits/8dd1b260993ce40ce667e07d43c069d5cc9f224f

benchexec/resources.py
CGall42: Unfortunately, I don't understand your feedback here.
I wrote a "check feasibility" function that is used in 2 places:

1. to determine the highest possible coreLimit 
```
        i = coreLimit
            while i >= coreRequirement:
                if check_distribution_feasibility(
                    i,
                    num_of_threads,
                    allCpus,
                    siblings_of_core,
                    hierarchy_levels,
                    isTest=True,
                ):
                    break
                else:
                    i -= 1
```

2 . in the assignment, leading to a sys.exit if the input parameters dont add up

```
    check_distribution_feasibility(
        coreLimit,
        num_of_threads,
        allCpus,
        siblings_of_core,
        hierarchy_levels,
        isTest=False,
    )
```

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+    # Filter CPU cores according to the list of identifiers provided by a user
+    if coreSet:
+        invalid_cores = sorted(set(coreSet).difference(set(allCpus)))
+        if len(invalid_cores) > 0:
+            raise ValueError(
+                "The following provided CPU cores are not available: "
+                + ", ".join(map(str, invalid_cores))
+            )
+        allCpus_list = [core for core in allCpus if core in coreSet]
+    allCpus_list = frequency_filter(allCpus, 0.05)
```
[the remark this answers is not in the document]
CGall42: Extended the docstring of get_cpu_list:     
A filter is applied to make sure, all cores used for the benchmark run at the same clock speed (allowing a deviation of 0.05 (5%) from the highest frequency)

Additionally, the docstring of of frequency_filter: 

Filters the list of all available CPU cores so that only the fastest cores are used for the benchmark run.
Only cores with a maximal frequency within the distance of the defined threshold from the maximal frequency of the fastest core are added to the filtered_allCpus_list and returned for further use. (max_frequency of core) >= (1-threshold)*(max_frequency of fastest core).
All cores that are slower will not be used for the benchmark and displayed in a debug message.

@param allCpus_list: list of all cores available for the benchmark run
@param threshold: accepted difference (as percentage) in the maximal frequency of a core from the fastest core to still be used in the benchmark run
@return: filtered_allCpus_list with only the fastest cores

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
-        self.assertInvalid(2, 5)
-        self.assertInvalid(5, 2)
-        self.assertInvalid(3, 3)
+            x                       x
 
+    x   x       x   x       x   x      x    x
 
-class TestCpuCoresPerRun_singleCPU_HT(TestCpuCoresPerRun_singleCPU):
-    ht = True
+    x-  x-      x-  x-      x-  x-     x-  x-
```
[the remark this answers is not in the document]
CGall42:
```
    x : symbolizes a unit (package, NUMA, L3, core)
    - : visualizes that a core is there, but it is not available because
        use_hyperthreading is set to False
```

benchexec/resources.py
PhilippWendler: Ah, I think now I understand. `isTest` is not related to running this in unit tests?

I guess it would still be cleaner to always have it throw an exception and catch it in the caller instead of this parameter. The alternative would be to make it return `Optional[str]` with the error message and throw if the result is not None, but I don't think this is better.

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
+        self.t_unit_assertValid(8, self.eightCore_assignment)
 
 
-class TestCpuCoresPerRun_singleCPU(TestCpuCoresPerRun):
-    cpus = 1
-    cores = 8
-    ht = False
+class Test_Topology_P1_NUMA2_L8_C16_F(TestCpuCoresPerRun):
+    num_of_cores = 16
+    num_of_packages = 1
```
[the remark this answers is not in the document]
CGall42: The algorithm should be able to handle levels being non-existent, so we test that in some classes. It doesnt make sense to test all variation, that would make too many tests. 
A constant of 1 doesn't mean that a level is unused - we test the code with 1 and 2 packages, since with 2 packages, the code has to create a root level. 

Which core assignments should be identical?

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
+    twoCore_assignment = [[0, 2], [4, 6], [8, 10]]
+    threeCore_assignment = [[0, 2, 4]]  # ,[8,10,6]
+    fourCore_assignment = [[0, 2, 4, 6]]
 
-class TestCpuCoresPerRun_threeCPU(TestCpuCoresPerRun):
-    cpus = 3
-    cores = 5
-    ht = False
+    def test_threeCoresPerRun(self):
+        self.t_unit_assertValid(3, self.threeCore_assignment, 1)
```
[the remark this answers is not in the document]
CGall42: The test covers coreLimit 1, 2, and 4 automatically but 3 is an edgecase where the simple maxThread calculation used for the unittests doesn't return the right result, while the algorithm does it right. Therefore, the method is overwritten here. 
We discussed this via Mail (Jan 30, 2023)

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
+class Test_Topology_P1_NUMA2_L8_C16_T(TestCpuCoresPerRun):
+    num_of_cores = 16
+    num_of_packages = 1
+    num_of_NUMAs = 2
+    num_of_L3_regions = 8
+    num_of_hyperthreading_siblings = 2
+    use_hyperthreading = True
 
+    # expected results for different coreLimits
+    oneCore_assignment = [[x] for x in [0, 8, 2, 10, 4, 12, 6, 14]]
```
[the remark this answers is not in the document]
CGall42: I realized that some missunderstanding concerning the unit test arise from different core numbering an the way, the machine is built from the variables.  While there are 16 cores, halb of them are hyperthreading siblings. Since hyperthreading is off, our maximal number of threads is 8 for the one core assigment und 4 for the 2 core assigment.
I put the core id into the commets for clarification

benchexec/resources.py
CGall42: Done in commit https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0

benchexec/resources.py
```diff
@@ -210,118 +214,739 @@ def _get_cpu_cores_per_run0(
...
+def get_sub_unit_dict(
+    allCpus: Dict[int, VirtualCore], parent_list: List[int], hLevel: int
+) -> Dict[int, List[int]]:
+    child_dict = {}
+    for element in parent_list:
+        subSubUnitKey = allCpus[element].memory_regions[hLevel]
+        if subSubUnitKey in list(child_dict.keys()):
+            child_dict[subSubUnitKey].append(element)
+        else:
+            child_dict.update({subSubUnitKey: [element]})
```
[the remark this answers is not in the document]
CGall42: Done in commit https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0
Thanks, that's way better.

benchexec/resources.py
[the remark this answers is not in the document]
CGall42: See commit https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0

benchexec/test_core_assignment.py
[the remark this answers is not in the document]
CGall42: For better understanding, this is edited again in https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0 

    x : symbolizes a unit (package, NUMA, L3, core)
    - : visualizes that a core is there, but it is not available because
        use_hyperthreading is set to False
    int: core id

                         x

            x                       x

    x   x       x   x       x   x      x    x

    0-  2-     4-   6-     8-   10-   12-  14-

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
-class TestCpuCoresPerRun_singleCPU(TestCpuCoresPerRun):
-    cpus = 1
-    cores = 8
-    ht = False
+class Test_Topology_P1_NUMA2_L8_C16_F(TestCpuCoresPerRun):
+    num_of_cores = 16
+    num_of_packages = 1
+    num_of_NUMAs = 2
+    num_of_L3_regions = 8
+    num_of_hyperthreading_siblings = 2
```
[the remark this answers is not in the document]
CGall42: Done https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0

benchexec/test_core_assignment.py
```diff
@@ -21,14 +24,21 @@ def lrange(start, end):
 
 
 class TestCpuCoresPerRun(unittest.TestCase):
+    num_of_cores = 0
+    num_of_packages = 0
+    num_of_groups = 0
+    num_of_NUMAs = 0
+    num_of_L3_regions = 0
+    num_of_hyperthreading_siblings = 0
```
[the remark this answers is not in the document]
CGall42: I think that having the the hyperthreading_siblings per core is better for interpretation, because we also test 3 HT siblings per core. I could of course rename the variable, if you prefer that. 

Groups is just another topology layer, the name is lend from lstopo.

None vs 0 is done in commit https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0

benchexec/test_core_assignment.py
```diff
@@ -40,57 +50,133 @@ def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
...
+        siblings_of_core = defaultdict(list)
+        cores_of_L3cache = defaultdict(list)
+        cores_of_NUMA_Region = defaultdict(list)
+        cores_of_group = defaultdict(list)
+        cores_of_package = defaultdict(list)
+        hierarchy_levels = []
+
+        for cpu_nr in range(self.num_of_cores):
+            # package
+            if self.num_of_packages and self.num_of_packages != 0:
```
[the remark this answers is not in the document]
CGall42: Done in commit https://github.com/sosy-lab/benchexec/pull/892/commits/0689df517323f46784726a9fc5b3d74c0abcdad0

benchexec/resources.py
```diff
@@ -210,118 +214,739 @@ def _get_cpu_cores_per_run0(
...
+    if len(hierarchy_levels[-1]) > 1:
+        top_level_cores = []
+        for node in hierarchy_levels[-1]:
+            top_level_cores.extend(hierarchy_levels[-1][node])
+        hierarchy_levels.append({0: top_level_cores})
+        for cpu_nr in allCpus:
+            allCpus[cpu_nr].memory_regions.append(0)
+
+
+def get_sub_unit_dict(
```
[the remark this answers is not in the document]
CGall42: Documentation added

benchexec/resources.py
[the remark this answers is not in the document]
CGall42: Renamed `ListOfIntLists` into `_2DIntList`

benchexec/resources.py
CGall42: > Wouldn't it be a good idea to check that every hierarchy level contains the same set of cores?

That should be true implicitly, since we iterate through the allCpus_list to build the hierarchy levels

benchexec/resources.py
CGall42: Fixed

benchexec/resources.py
```diff
@@ -15,313 +15,918 @@
...
+                if any(
+                    len(set1.difference(set(s2))) == 0
+                    for s2 in hierarchy_levels[index + 1].values()
+                ):
+                    anyIdentical = True
+                allIdentical = allIdentical and anyIdentical
+            if allIdentical:
+                removeList.append(hierarchy_levels[index])
+    for hLevel in removeList:
+        hierarchy_levels.remove(hLevel)
```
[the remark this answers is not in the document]
CGall42: Moved into seperate function in commit https://github.com/sosy-lab/benchexec/pull/892/commits/54c8071e8ebd7475192d1a706c42cf6dcd412bd1

benchexec/resources.py
CGall42: I hope that all are done now ... let me know if you find a docstring missing.

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
     threeCore_assignment = None
     fourCore_assignment = None
     eightCore_assignment = None
-    use_ht = True
+    use_hyperthreading = True
+
+    """def test_singleThread(self):
+        # test all possible coreLimits for a single thread
+        self.t_unit_assertValid (1, self.oneCore_assignment)"""
```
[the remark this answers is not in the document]
CGall42: Deleted

benchexec/test_core_assignment.py
```diff
@@ -40,57 +50,133 @@ def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
...
+        for level in hierarchy_levels:  # hierarchy_levels = [dict1, dict2, dict3]
+            for key in level:
+                for core in level[key]:
+                    allCpus[core].memory_regions.append(key)
+
+        check_and_add_meta_level(hierarchy_levels, allCpus)
+
+        return allCpus, siblings_of_core, hierarchy_levels
+
+    def t_unit_assertValid(self, coreLimit, expectedResult, maxThreads=None):
```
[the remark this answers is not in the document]
CGall42: Renamed to `mainAssertValid` since it's the main test class

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
+    fiveCore_assignment = [[0, 1, 2, 3, 4]]  # ,[8,9,10,11,6]]
+    eightCore_assignment = [[0, 1, 2, 3, 4, 5, 6, 7]]
 
-class TestCpuCoresPerRun_dualCPU_no_ht(TestCpuCoresPerRun):
-    cpus = 2
-    cores = 8
-    ht = True
-    use_ht = False
+    def test_threeCoresPerRun(self):
+        self.t_unit_assertValid(5, self.fiveCore_assignment, 1)
```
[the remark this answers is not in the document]
CGall42: The test covers coreLimit 1, 2, and 4 automatically but 5 is an edgecase that we wanted to test here.
 ```
   def test_fiveCoresPerRun(self):
        self.mainAssertValid(5, self.fiveCore_assignment)
```

File: benchexec/test_core_assignment.py [hunk 54]
```diff
@@ -40,57 +50,133 @@ def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
...
+            siblings_of_core.pop(element)
+
+        for item in [
+            siblings_of_core,
+            cores_of_L3cache,
+            cores_of_NUMA_Region,
+            cores_of_package,
+            cores_of_group,
+        ]:
+            if len(item) > 0:
```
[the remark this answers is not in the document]
CGall42: Done

benchexec/test_core_assignment.py
```diff
@@ -40,57 +50,133 @@ def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
...
+        def compare_hierarchy_by_dict_length(level):
+            return len(next(iter(level.values())))
+
+        # sort hierarchy_levels (list of dicts) according to the dicts' corresponding unit sizes
+        hierarchy_levels.sort(key=compare_hierarchy_by_dict_length, reverse=False)
+        # add siblings_of_core at the beginning of the list
+
+        allCpus_list = list(range(self.num_of_cores))
+
+        for cpu_nr in allCpus_list:
```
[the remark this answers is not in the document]
CGall42: Fixed

benchexec/resources.py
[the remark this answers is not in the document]
CGall42: No, the order is not relevant here.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+
+    return is_feasible
+
+
+def calculate_chosen_level(
+    hierarchy_levels: List[HierarchyLevel], coreLimit_rounded_up: int
+) -> int:
+    """Calculates the hierarchy level necessary so that number of cores at the chosen_level is at least
+    as big as the cores necessary for one thread"""
+    chosen_level = 1
```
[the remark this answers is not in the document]
PhilippWendler: Likely related to fd58f753f5affe9b12a2f314be071fb1b7acb232.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+    hierarchy_levels: List[HierarchyLevel],
+) -> List[List[int]]:
+    """Actual core distribution method:
+    uses the architecture read from the file system by get_cpu_cores_per_run
+
+    Calculates an assignment of the available CPU cores to a number
+    of parallel benchmark executions such that each run gets its own cores
+    without overlapping of cores between runs.
+    In case the machine has hyper-threading, this method avoids
+    putting two different runs on the same physical core.
```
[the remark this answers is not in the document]
PhilippWendler: Our old test suite would have also found that, currently these tests are ignored but need to be re-enabled: deae4acc0f15066278bb233a1fc08fd5d1bbf58e

benchexec/test_core_assignment.py
```diff
@@ -100,601 +186,319 @@ def test_singleThread(self):
...
-        self.assertRaises(
-            SystemExit,
-            _get_cpu_cores_per_run0,
-            1,
-            1,
-            True,
-            [0],
-            {0: [0, 1]},
-            {0: [0, 1]},
-        )
```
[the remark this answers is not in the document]
PhilippWendler: Explained in message of ef13d2aa527c8b520d223b9d4c9729a2004bc659

benchexec/resources.py
```diff
@@ -57,149 +63,332 @@ def get_cpu_cores_per_run(
...
+                for sibling in siblings_of_core[core]:
+                    if sibling != core:
+                        cleanList.append(sibling)
+        for element in cleanList:
+            siblings_of_core.pop(element)
+        # siblings_of_core will be added to hierarchy_levels list after sorting
+
+        # read & prepare mapping of cores to L3 cache
+        cores_of_L3cache = get_L3cache_mapping(allCpus_list)
+        hierarchy_levels.append(cores_of_L3cache)
```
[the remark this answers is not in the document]
PhilippWendler: Implemented by 4a6414232aec50678971a9e3e7867fc079cee7fe.

benchexec/resources.py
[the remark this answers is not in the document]
PhilippWendler: Further improved in 4a6414232aec50678971a9e3e7867fc079cee7fe because the name was not fitting for the function.

benchexec/resources.py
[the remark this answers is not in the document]
PhilippWendler: Further improved in 4a6414232aec50678971a9e3e7867fc079cee7fe because the magic constant should be only in one place.

benchexec/test_core_assignment.py
```diff
@@ -40,57 +50,133 @@ def assertValid(self, coreLimit, num_of_threads, expectedResult=None):
...
-            _get_cpu_cores_per_run0,
+            get_cpu_distribution,
             coreLimit,
             num_of_threads,
-            self.use_ht,
+            self.use_hyperthreading,
             *self.machine(),
         )
 
+    def assertEqualResult(self, coreLimit, num_of_threads, expectedResult=None):
```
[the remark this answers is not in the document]
PhilippWendler: Deleted in 4a6414232aec50678971a9e3e7867fc079cee7fe.

benchexec/resources.py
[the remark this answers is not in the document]
PhilippWendler: Cleaned up by 2f3f9640ee20cce5a124e2903accd71f51b1238b.

benchexec/resources.py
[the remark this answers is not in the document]
PhilippWendler: Improved by 2f3f9640ee20cce5a124e2903accd71f51b1238b.

benchexec/resources.py
PhilippWendler: Implemented by 2f3f9640ee20cce5a124e2903accd71f51b1238b.

benchexec/resources.py
PhilippWendler: Improved by 2f3f9640ee20cce5a124e2903accd71f51b1238b.

benchexec/resources.py
PhilippWendler: Resolveed by a00ede6cd284523173d5f9debbe1f900cb2a77be.

benchexec/resources.py
```diff
@@ -210,118 +218,738 @@ def _get_cpu_cores_per_run0(
...
+def check_and_add_meta_level(
+    hierarchy_levels: List[HierarchyLevel], allCpus: Dict[int, VirtualCore]
+) -> None:
+    """
+    Adds a meta_level to hierarchy_levels to iterate through all cores (if necessary)
+    """
+    if len(hierarchy_levels[-1]) > 1:
+        top_level_cores = []
+        for node in hierarchy_levels[-1]:
+            top_level_cores.extend(hierarchy_levels[-1][node])
```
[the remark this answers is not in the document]
PhilippWendler: Removed by a00ede6cd284523173d5f9debbe1f900cb2a77be.

PR #893, opened 2023-01-21
More accurate score calculation for validation

dbeyer:
- for witnesses that are neither correct nor wrong, the validation score should be 0 (for example, if a witness has syntax errors)
- the score adjustment should also be done for the maximal score

### DIFF
--- a/benchexec/result.py
+++ b/benchexec/result.py
@@ -133,18 +133,27 @@ class Property(collections.namedtuple("Property", "filename is_svcomp name")):
 
     __slots__ = ()  # reduce per-instance memory consumption
 
-    def compute_score(
-        self, category, result, witness_category=WITNESS_CATEGORY_UNKNOWN
-    ):
+    def _adjust_score_for_witness_validation(self, score, witness_category):
+        if witness_category == WITNESS_CATEGORY_CORRECT:
+            # Score is already correctly set.
+            return score
+        elif witness_category == WITNESS_CATEGORY_WRONG:
+            # If a validator refutes (confirms) a wrong witness, then the score (reduction) is multiplied by a factor.
+            return score * _SCORE_FACTOR_WRONG_WITNESS
+        else:
+            # Discard result for a witness that is neither correct nor wrong.
+            return 0
+
+    def compute_score(self, category, result, witness_category=None):
         if not self.is_svcomp:
             return None
         score = _svcomp_score(category, result)
-        if witness_category == WITNESS_CATEGORY_WRONG:
-            # If a validator refutes (confirms) a wrong witness, then the score (reduction) is multiplied by a factor.
-            score *= _SCORE_FACTOR_WRONG_WITNESS
-        return score
+        if witness_category is None:
+            return score
+        else:
+            return self._adjust_score_for_witness_validation(score, witness_category)
 
-    def max_score(self, expected_result):
+    def max_score(self, expected_result, witness_category=None):
         """
         Return the maximum possible score for a task that uses this property.
         @param expected_result:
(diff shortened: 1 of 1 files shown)


benchexec/result.py
```diff
@@ -133,26 +133,39 @@ class Property(collections.namedtuple("Property", "filename is_svcomp name")):
...
             return None
         score = _svcomp_score(category, result)
-        if witness_category == WITNESS_CATEGORY_WRONG:
-            # If a validator refutes (confirms) a wrong witness, then the score (reduction) is multiplied by a factor.
-            score *= _SCORE_FACTOR_WRONG_WITNESS
-        return score
+        if witness_category is None:
+            return score
+        else:
+            return Property._adjust_score_witness(score, witness_category)
```
PhilippWendler, line 154: why not a regular method call?

benchexec/result.py
```diff
@@ -133,26 +133,39 @@ class Property(collections.namedtuple("Property", "filename is_svcomp name")):
 
     __slots__ = ()  # reduce per-instance memory consumption
 
-    def compute_score(
-        self, category, result, witness_category=WITNESS_CATEGORY_UNKNOWN
-    ):
+    def _adjust_score_witness(score, witness_category):
```
PhilippWendler, line 136: A "score witness" is not an existing concept, the method would better be named `adjust_score_for_witness_validation`.

benchexec/result.py
```diff
@@ -133,26 +133,39 @@ class Property(collections.namedtuple("Property", "filename is_svcomp name")):
...
             an ExpectedResult indicating whether the property is expected to hold for the task
         """
         if not self.is_svcomp or not expected_result:
             return None
-        return _svcomp_max_score(expected_result.result)
+        score = _svcomp_max_score(expected_result.result)
+        if witness_category is None:
+            return score
+        else:
+            return Property._adjust_score_witness(score, witness_category)
```
PhilippWendler, line 168: The maximum score for a task shouldn't depend on the category, right? It should depend on nothing from the run, actually, only on the data from the task. We also do not use the primary `category` here.

benchexec/result.py
dbeyer: Because it does not access anything from self.

benchexec/result.py
dbeyer: Yes, this is what I had in mind, and omitted 'for' and 'validation'.  Will change.

benchexec/result.py
dbeyer: The witness is part of the task.  A task for witness validation is (program, spec, witness).  It is referring to a category from the input.  It is like 'expected verdict'.

PhilippWendler: Hm, I hadn't realized. Maybe then `witness-category` is actually not such a good name? On the other hand, it does fit quite nicely.

benchexec/result.py
PhilippWendler: I would say that doesn't matter. In Python, static methods are use quite rarely and only where necessary typically (because one does not have an instance). For helper methods called from an instance method, instance methods are fine and shorter to write.

dbeyer: Done. And I added the second underscore for private methods.

benchexec/result.py
dbeyer: Done.

benchexec/result.py
PhilippWendler: Standard Python style is to use a single underscore ("*generally, double leading underscores should be used only to avoid name conflicts with attributes in classes designed to be subclassed*" in [PEP-8](https://peps.python.org/pep-0008/)). A double underscore has a technical effect that is rarely needed and not well-known, it should be used only where required.

dbeyer: Aha. Fixed.

PR #894, opened 2023-01-23
More accurate max-score calculation for validation

dbeyer:
We had adjusted the score by considering the witness-category [https://github.com/sosy-lab/benchexec/pull/888].
We were able to get the value using the `sourcefileTag`, i.e., from the XML:
https://github.com/sosy-lab/benchexec/blob/2731613721b8d2169d3e92ba26f9f71774fcdc3c/benchexec/tablegenerator/__init__.py#L884

This PR adds the similar score adjustment for the calculation of the maximal achievable score.
Here we have to take the value from the rows of the table.

### DIFF
--- a/benchexec/tablegenerator/htmltable.py
+++ b/benchexec/tablegenerator/htmltable.py
@@ -211,10 +211,27 @@ def _prepare_benchmark_setup_data(
     }
 
 
+def _get_column_index(column_name, rows):
+    if len(rows) == 0:
+        return None
+    columns = rows[0].results[0].columns
+    column_index = next(
+        (columns.index(c) for c in columns if c.title == column_name), None
+    )
+    return column_index
+
+
+def _get_column_value(column_index, row):
+    if column_index is None:
+        return None
+    return row.results[0].values[column_index]
+
+
 def _get_task_counts(rows):
     """Calculcate number of true/false tasks and maximum achievable score."""
     count_true = count_false = 0
     max_score = None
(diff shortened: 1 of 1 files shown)


PhilippWendler:
I am afraid that this is not so easy.

The current code only looks in `results[0]` for the column named `witness-category`.

First, this can mean that the code misses this column if it appears in a later group of columns.

Second, there can be more than row with the name `witness-category`, and there is nothing preventing them from having different values. The current code ignore this and just takes the first row with this name.

The only safe way to compute this value that I see is to check whether the column occurs in all run sets, check that the values all match, and only then use it.
We can fine-tune this to hide the score completely if there is at least one column named `witness-category` but the conditions above are not fulfilled.

PhilippWendler:
There is another problem:

The user can simply omit the column `witness-category` from the table (via the table definition). In this case, a wrong max score would be computed because the logic in `htmltable.py` would not see the column `witness-category` and compute the max score using the regular scoring scheme.

I think the only way to solve this is to compute the max score upfront and store it as a property of the `Row`. Maybe the max score should even become a part of the row id (which is used to uniquely identify rows)?

PhilippWendler:
> Maybe the max score should even become a part of the row id (which is used to uniquely identify rows)?

We discussed this and decided that the witness category should conceptually be part of the task id in table-generator anyway (if it exists). Then the score is easy to compute for each row.

This was implemented in 172a46779c76cb82220131a6af7266136cbf6baf and replaces this PR.

PR #904, opened 2023-03-05
bump setuptools to support latest PEPs

Saransh-cpp:
- `setuptools` `64.0.0` supports editable installs directly from `pyproject.toml` - https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6400
- An appropriate version of `wheel` is automatically installed while installing `setuptools` - https://github.com/pypa/setuptools/commit/f7d30a9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,9 +7,8 @@
 
 [build-system]
 requires = [
-    # Require versions that support our license files
-    'setuptools >= 42.0.0',
-    'wheel >= 0.32.0',
+    # Require versions that support our license files and the latest PEPs
+    'setuptools >= 64.0.0',
 ]
 build-backend = 'setuptools.build_meta'
 
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,5 @@ import warnings
 warnings.filterwarnings("default", module=r"^benchexec\..*")
 
 # This file is still required for compatibility with pip<19.0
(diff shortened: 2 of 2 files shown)


PhilippWendler:
Thanks, nice to be aware that this feature now exists. As long as we can't/don't want to remove `setup.py` completely, I don't think it is worth bumping the `setuptools` version, though.

For `wheel`: I don't see a guarantee in your link which version of `wheel` will be installed. We specifically need `>= 0.32.0` in order to produce license-compliant files. So I would keep this requirement.

Saransh-cpp:
Ah, fair and valid points.

I see that `benchexec` uses `python setup.py <something>` commands in its codebase. These commands should be avoided as the `setuptools` frontend is not actively maintained, but again, as pointed out by you, `benchexec` cannot get rid of `setup.py` until these commands are replaced with the latest commands (for instance - `python setup.py sdist bdist_wheel` -> `python -m build`, and so on).

I will be more than happy to update the `setup.py` commands to more modern commands.

Thanks for the quick feedback! I'll close this if the changes are not required at the moment.

PhilippWendler:
> I will be more than happy to update the `setup.py` commands to more modern commands.

That would be nice!

Saransh-cpp:
Great! I will close this one for now.

PR #905, opened 2023-03-06
Get rid of python setup.py commands

Saransh-cpp:
These should work, but I hope the CI will catch any errors if they don't.

PS: I don't have much experience with `nosetests`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,8 +17,8 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
   # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
   # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
-  - python -m pip install lxml
+  - python -m pip install lxml nose
   - python -m pip install --user .
 
 test_script:
-  - python -m nose --tests benchexec.tablegenerator
+  - nosetests --tests benchexec.tablegenerator
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,7 +17,7 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
   # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
   # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
-  - python -m pip install --user ".[setup_requires]"
+  - python -m pip install --user ".[dev]"
 
 test_script:
   - nosetests --tests benchexec.tablegenerator
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,4 +21,4 @@ install:
   - python -m pip install --user .
 
 test_script:
-  - python setup.py nosetests --test benchexec.tablegenerator
+  - nosetests --tests benchexec.tablegenerator
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,8 +17,7 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
   # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
   # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
-  - python -m pip install lxml
-  - python -m pip install --user .
+  - python -m pip install --user ".[setup_requires]"
 
 test_script:
   - nosetests --tests benchexec.tablegenerator
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,4 +21,4 @@ install:
   - python -m pip install --user .
 
 test_script:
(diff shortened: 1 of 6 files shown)


.appveyor.yml
```diff
@@ -17,8 +17,8 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
   # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
   # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
-  - python -m pip install lxml
+  - python -m pip install lxml nose
```
Saransh-cpp, line 20: Should `nose` be installed here like it is being done in the travis pipeline?

```
pip install nose --no-binary :all:
```

PhilippWendler: This would be necessary for Python > 3.10.

Why is a manual installation of `nose` here in this PR? Shouldn't this be handled by the `pip install .`?

Saransh-cpp: A bit of debugging and it works!

The issue -
- `nose` was installed as a `setup_requires` dependency, which means that whenever a `python setup.py <something>` was executed, `nose` and `lxml` were installed automatically.
- This however does not happen when one installs the package using `pip install .`. In this case, only the true requirements, that is, `PyYAML` is installed.

> In previous versions of setuptools, this used to be accomplished with the setup_requires keyword but is now considered deprecated in favor of the [PEP 517](https://peps.python.org/pep-0517/) style described above. To peek into how this legacy keyword is used, consult our [guide on deprecated practice (WIP)](https://setuptools.pypa.io/en/latest/deprecated/index.html).

Solution -
- I have moved `lxml` and `nose` to a new optional dependency group named `"dev"`. This group can now be installed as - `pip install ".[dev]"`.

PhilippWendler: Thanks! This makes sense and I like the solution. It is certainly much clear to keep test dependencies out of `setup_requires`.

Would you mind updating the [docs](https://github.com/sosy-lab/benchexec/blob/main/doc/DEVELOPMENT.md?plain=1#L33) as well?

release.sh
```diff
@@ -83,9 +83,11 @@ pushd "$TEMP3/benchexec"
...
 # avoid the wheel on PyPi for nose, it does not work on Python 3.10
 pip install nose --no-binary :all:
-pip install -e "."
-python setup.py nosetests
-python setup.py sdist bdist_wheel
+# install build if it is not installed by default (it usually is)
+pip install build
+pip install -e ".[dev]"
+python -m nose
+python -m build
```
PhilippWendler, line 90: There is also an explanation of what `release.sh` does in `doc/DEVELOPMENT.md`, could you please update it with the reference to `build` and explain that it builds source dist and wheel for us?

.appveyor.yml
```diff
@@ -17,8 +17,7 @@ install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
   # The package installation via our setup.py file uses easy_install, which fails to correctly install lxml.
   # As some Python environments don't have lxml preinstalled, install it here to avoid errors during the execution in those cases.
-  - python -m pip install lxml
-  - python -m pip install --user .
+  - python -m pip install --user ".[dev]"
 
 test_script:
-  - python setup.py nosetests --test benchexec.tablegenerator
+  - python -m nose --tests benchexec.tablegenerator
```
PhilippWendler, line 23: Just to confirm your proposal: This is a better way to run the tests. :+1: 

Not only is it required for getting rid of `setup.py` eventually, the [test output](https://ci.appveyor.com/project/sosy-lab/benchexec/builds/46428757/job/0pvyg2nm59g37xb3#L40) is also not cluttered anymore with installation stuff as [before](https://ci.appveyor.com/project/sosy-lab/benchexec/builds/46351006/job/r035xgrtp3q2i7yf#L40).

PhilippWendler (commented):
Nice, looks good!

Besides updating the docs, there is just one thing left: `.gitlab-ci.yml` also needs to be updated in the same style (install `[dev]` and run `-m nose`). [GitLab CI](https://gitlab.com/sosy-lab/software/benchexec/-/pipelines) is actually our main CI since Travis does not work reliably anymore, it just doesn't support running on PRs from external forks unfortunately. But when the PR gets updated I will push it to an internal branch so that we can see the results.

.appveyor.yml
Saransh-cpp: I did not compare the test output with a previous build, nice!

Saransh-cpp:
Thanks for the review! I've updated the docs and the GitLab CI. This should hopefully work. Please let me know if I missed something.

PhilippWendler (approved):
[GitLab CI](https://gitlab.com/sosy-lab/software/benchexec/-/jobs/3887676597) also works as expected.

Thanks a lot for these suggestions! Keeping up with best practices of Python packaging is unfortunately not an easy task if one doesn't do a lot of Python packaging, so this help is really welcome!

PR #908, opened 2023-03-28
Draft: Migrate test_cgroups.py and test_benchmark_definiton.py to pytest framework

Naman-Priyadarshi:
I have successfully created a small prototype for `test_cgroups.py` which works for pytest.  The test was running successfully on my Linux machine.

### DIFF
--- a/benchexec/test_cgroups_pytest.py
+++ b/benchexec/test_cgroups_pytest.py
@@ -7,56 +7,51 @@
 
 import subprocess
 import pytest
+
 from benchexec import check_cgroups
 
-class TestCheckCgroups:
-    def execute_run_extern(self, *args, **kwargs):
-        try:
-            return subprocess.check_output(
-                args=["python3", "-m", "benchexec.check_cgroups"] + list(args),
-                stderr=subprocess.STDOUT,
-                universal_newlines=True,
-                **kwargs,
-            )
-        except subprocess.CalledProcessError as e:
-            if e.returncode != 1:  # 1 is expected if cgroups are not available
-                print(e.output)
-                raise e
 
-    def test_extern_command(self, caplog):
-        """Test external command."""
-        with caplog.at_level("DEBUG"):
-            self.execute_run_extern()
+def execute_run_extern(*args, **kwargs):
+    try:
+        return subprocess.check_output(
+            args=["python3", "-m", "benchexec.check_cgroups"] + list(args),
+            stderr=subprocess.STDOUT,
+            universal_newlines=True,
+            **kwargs,
+        )
+    except subprocess.CalledProcessError as e:
+        if e.returncode != 1:  # 1 is expected if cgroups are not available
+            print(e.output)
+            raise e
 
-    def test_simple(self, caplog):
-        """Test check_cgroups.main() with --no-thread."""
-        with caplog.at_level("DEBUG"):
-            try:
-                check_cgroups.main(["--no-thread"])
-            except SystemExit as e:
-                # expected if cgroups are not available
-                pytest.skip(str(e))
+def test_extern_command():
+    execute_run_extern()
 
-    def test_threaded(self, caplog):
-        """Test check_cgroups.main() with threading."""
-        with caplog.at_level("DEBUG"):
-            try:
-                check_cgroups.main([])
-            except SystemExit as e:
-                # expected if cgroups are not available
-                pytest.skip(str(e))
+def test_simple():
+    try:
+        check_cgroups.main(["--no-thread"])
+    except SystemExit as e:
+        # expected if cgroups are not available
+        pytest.skip(e)
 
-    def test_thread_result_is_returned(self, caplog):
-        """
-        Test that an error raised by check_cgroup_availability is correctly
-        re-raised in the main thread by replacing this function temporarily.
-        """
-        tmp = check_cgroups.check_cgroup_availability
-        try:
-            check_cgroups.check_cgroup_availability = lambda wait: exit(1)
+def test_threaded():
+    try:
+        check_cgroups.main([])
+    except SystemExit as e:
+        # expected if cgroups are not available
+        pytest.skip(e)
 
-            with caplog.at_level("DEBUG"), pytest.raises(SystemExit):
-                check_cgroups.main([])
+def test_thread_result_is_returned():
+    """
+    Test that an error raised by check_cgroup_availability is correctly
+    re-raised in the main thread by replacing this function temporarily.
+    """
+    tmp = check_cgroups.check_cgroup_availability
+    try:
+        check_cgroups.check_cgroup_availability = lambda wait: exit(1)
 
-        finally:
-            check_cgroups.check_cgroup_availability = tmp
+        with pytest.raises(SystemExit):
+            check_cgroups.main([])
+
+    finally:
+        check_cgroups.check_cgroup_availability = tmp
--- a/benchexec/test_cgroups_pytest.py
+++ b/benchexec/test_cgroups_pytest.py
@@ -1,0 +1,60 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import subprocess
+import pytest
+
+from benchexec import check_cgroups
+
+logging.disable(logging.CRITICAL)
+
+class TestCheckCgroups:
+    def execute_run_extern(self, *args, **kwargs):
+        try:
+            return subprocess.check_output(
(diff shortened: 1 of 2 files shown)


benchexec/test_cgroups_pytest.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import subprocess
+import pytest
+
+from benchexec import check_cgroups
+
+logging.disable(logging.CRITICAL)
```
PhilippWendler, line 14: This seems like it might change behavior compared to the previous code: So far it was executed right before the tests in class `TestCheckCgroups`, now it is executed at import time. Does this make a difference, e.g., if tests from more than one module are run and they all change the log level? Or does pytest ensure that the module is imported right before it is used or that such changes are otherwise isolated?

It seems that pytest has some [logging-specific features](https://docs.pytest.org/en/7.1.x/how-to/logging.html), maybe it is even possible to improve our log handling in tests compared to the existing situation?

benchexec/test_cgroups_pytest.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TestCheckCgroups:
```
PhilippWendler, line 16: Is it idiomatic for pytest to use classes for tests?

benchexec/test_cgroups_pytest.py
Naman-Priyadarshi: In pytest, each test function is run in a separate isolated environment, which means that changes to global state, such as log levels, made by one test function do not affect other test functions. So if tests from more than one module are run and they all change the log level, these changes will not affect each other.

However, if the log level is changed in a fixture that is shared across multiple tests, then the change to the log level may affect all the tests that use that fixture. In such cases, it's important to ensure that the fixture is designed to handle such global state changes in a way that is appropriate for the specific use case.

Therefore, if we change the log level in a module that has already been imported, those changes will persist across subsequent test runs unless you reset the log level manually. To avoid this, we can either reset the log level at the end of each test, or you can import the module inside a fixture that is scoped to the test function, so that the module is re-imported and its code is re-executed for each test.

benchexec/test_cgroups_pytest.py
Naman-Priyadarshi: There are no strict rules regarding organizing tests into modules vs classes. It's all a personal choice in my opinion

benchexec/test_cgroups_pytest.py
PhilippWendler: So, does this mean that we need a fixture here?

Actually, it would of course be nice if we do get the logging output if we want to debug the test. But it should not clutter the usual output. So far the workaround was to disable logging in the tests. But if we can improve this that would be great!

benchexec/test_cgroups_pytest.py
PhilippWendler: What do you think for this case?

Naman-Priyadarshi:
In this modified code, I have created a new fixture called `enable_logging` that enables logging for the specific test function that it is applied to. It sets the logging level to ``logging.DEBUG``, which means that logging messages with severity ``DEBUG`` and above will be output.

The fixture also adds a finalizer function that disables logging after the test has completed, ensuring that logging is only enabled for the duration of the test.

To use this fixture, we simply add it as an argument to each test function that requires logging. When pytest runs the test, it will automatically call the fixture function and pass it as an argument to the test function. This will allow us to enable logging for specific tests without cluttering the output of other tests.

benchexec/test_cgroups_pytest.py
Naman-Priyadarshi: As this test is comparatively small in size, it would be better to use classes according to me

PhilippWendler:
> In this modified code, I have created a new fixture called `enable_logging` that enables logging for the specific test function that it is applied to. It sets the logging level to `logging.DEBUG`, which means that logging messages with severity `DEBUG` and above will be output.

How does this look for the developer? How will they see the log?

> To use this fixture, we simply add it as an argument to each test function that requires logging. When pytest runs the test, it will automatically call the fixture function and pass it as an argument to the test function. This will allow us to enable logging for specific tests without cluttering the output of other tests.

Is there a way to achieve this without having to add it to every single test manually?

> The fixture also adds a finalizer function that disables logging after the test has completed, ensuring that logging is only enabled for the duration of the test.

Actually, this is not my main concern. If we have a decent handling of log output for tests, we can just use it for all tests and do not need to worry about fiddling with log levels at all. Only if we do not have a decent handling of log outputs for tests and need to disable logging, then we need to make sure it is actually disabled for each test.

Naman-Priyadarshi:
>
> How does this look for the developer? How will they see the log?
>
There are multiple ways a developer can do this
1. Use the -s option: When running pytest from the command line, they can use the -s option to enable stdout output capture. This will allow them to see any logging messages that were printed to the console during the test run.
2. Use the caplog fixture: Pytest provides a fixture called caplog that allows to capture logging output from tests. We can use it to access the log records emitted by the logging module, and then perform assertions on them to ensure that the expected log messages were produced. 
> 
> Is there a way to achieve this without having to add it to every single test manually?
>
Yes! We can achieve this task by defining a custom Pytest plugin. These plugins will allow us to extend Pytest's functionality by defining custom hooks that are called at various points during the test running process. We can define a custom plugin that sets up logging configuration for specific tests based on configuration options or markers, and then use the `pytest_plugins` list in a `conftest.py` file to load the plugin automatically for all tests in a directory or project.

PhilippWendler:
How would that caplog solution look like? Can you show a concrete example? Or tell me how to try it out?

Naman-Priyadarshi:
So, I tried adding the caplog solution to our test and used `pytest-logger` package.

We can use these command to read the caplog text file:
`pytest test_cgroups_pytest.py --log-file=<path-to-log-file>` 

When we run our tests with this command, pytest will capture all log messages and save them to the specified file. We can then open the file in a text editor to read the log messages.

PhilippWendler:
So, I have now invested some time into playing around with this and reading the [tutorial](https://docs.pytest.org/en/latest/how-to/logging.html), and found the following:

- By default, pytest does not show test output except for failing tests, and thus our reason for manually changing the log levels in tests is gone. This is good, and we can just get rid of that.
- By default, pytest *does* show test output for failing tests, which is a nice and convenient improvement compared to the current state. It can still be disabled via command line or config file if this is ever necessary.
- The log level for what is being captured and shown for failing tests can be easily set via `pyproject.toml`. This means we do not need to clutter all our tests with fixtures for logging.
- `--log-file` works but it will just concatenate all logs from all tests with no distinction, which is rarely useful. Much nicer is `-o log_cli=true` or `--log-cli-level=LEVEL`.
- Format of log messages can be improved for all these cases using config options.

So in summary, pytest is really nice in this regard, we can and want to get rid of all manual logging configs in tests, and we might want to set some nice logging format for pytest in `pyproject.toml`.

Naman-Priyadarshi:
Hi! sorry for being inactive, my current semester is ending soon so I'm busy with project submissions and other stuff! will work on this as soon as I get time.

Naman-Priyadarshi:
> So, I have now invested some time into playing around with this and reading the [tutorial](https://docs.pytest.org/en/latest/how-to/logging.html), and found the following:
> 
>     * By default, pytest does not show test output except for failing tests, and thus our reason for manually changing the log levels in tests is gone. This is good, and we can just get rid of that.
> 
>     * By default, pytest _does_ show test output for failing tests, which is a nice and convenient improvement compared to the current state. It can still be disabled via command line or config file if this is ever necessary.
> 
>     * The log level for what is being captured and shown for failing tests can be easily set via `pyproject.toml`. This means we do not need to clutter all our tests with fixtures for logging.
> 
>     * `--log-file` works but it will just concatenate all logs from all tests with no distinction, which is rarely useful. Much nicer is `-o log_cli=true` or `--log-cli-level=LEVEL`.
> 
>     * Format of log messages can be improved for all these cases using config options.
> 
> 
> So in summary, pytest is really nice in this regard, we can and want to get rid of all manual logging configs in tests, and we might want to set some nice logging format for pytest in `pyproject.toml`.

Hi @PhilippWendler, I was wondering which format should we use and should we implement the caplog solution or not?

PhilippWendler:
I see no benefit in `caplog` so far. As for the format, it could simply be based on our standard log format.

Naman-Priyadarshi:
Made the required changes and migrated test_benchmark_definition.py too

Naman-Priyadarshi:
Hi @PhilippWendler! I was wondering if I should open a new PR for next workloads or add more commits to this PR? (P.S. Currently working on the next files to migrate)

PhilippWendler:
I think it makes sense to see this as a prototype, and create a fresh PR for the actual changes. We don't need the commits for trying out. And of course, we need the changes to the tests in the original files and not in additional new files.

PhilippWendler:
Will be superseded by other PRs.

PR #909, opened 2023-04-03
Add tool-info module for VeriFast

MartinSpiessl:
This adds a basic tool-info module for VeriFast.
I am not the author of VeriFast, but use this tool for reasearch purposes.


benchexec/tools/verifast.py
```diff
@@ -0,0 +1,41 @@
...
+                return result.RESULT_UNKNOWN
+            elif "Callee must have contract" in line:
+                return result.RESULT_ERROR + "(underspec)"
+            elif line.endswith(": dead code"):
+                return result.RESULT_ERROR + "(dead code)"
+            elif line.endswith(": Parse error."):
+                return result.RESULT_ERROR + "(parse error)"
+            elif ": Cannot prove condition." in line:
+                return result.RESULT_ERROR + "(cannot prove condition)"
+        return result.RESULT_ERROR + "(unknown)"
```
PhilippWendler, line 41: I would suggest to just return `RESULT_ERROR` here. The `(unknown)` does not add information.

benchexec/tools/verifast.py
```diff
@@ -0,0 +1,41 @@
...
+    def name(self):
+        return "VeriFast"
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def determine_result(self, run):
+        for line in run.output:
+            line = line.strip()
+            if line.startswith("0 errors found"):
```
PhilippWendler, line 29: It seems this would also match something like `10 errors found`. Is it possible to strengthen this check, e.g., by avoiding `startswith`? It might also be a good idea to strengthen the other checks where possible.

PhilippWendler: Hm, not sure what I thought here. Of course it does not match `10 errors found` *due to* the `startswidth`. I think I misread as `endswith`. Sorry.

PhilippWendler:
Is merging this still desired or should it be closed?

PR #910, opened 2023-04-03
Add tool-info module for VerCors

MartinSpiessl:
This adds a basic tool-info module for VerCors.
I am not the author of VerCors, but use that tool for research purposes.


benchexec/tools/vercors.py
```diff
@@ -0,0 +1,39 @@
...
+                return result.RESULT_TRUE_PROP
+            elif "Parsing failed" in line:
+                return result.RESULT_ERROR + "(Parsing failed)"
+            elif "Exception in thread" in line:
+                return result.RESULT_ERROR + "(Exception)"
+            elif "is not supported" in line:
+                return result.RESULT_ERROR + "(Unsupported)"
+            elif line.startswith("[ERROR]") or "An error condition was reached" in line:
+                return result.RESULT_ERROR
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 39: This returns `unknown` for cases where the tool output cannot be parsed successfully, which would be the case for example in case of a crash. However, `unknown` is meant for cases where the tool gave up. Is it possible to distinguish these cases from errors/crashes?

PhilippWendler:
Is merging this still desired or should it be closed?

PhilippWendler:
Closing due to lack of response.

PR #921, opened 2023-10-09
Added Fizzer tool

trtikm:


benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,148 @@
+import benchexec.tools.template
+import benchexec.result as result
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import os
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for sbt-fizzer.
```
PhilippWendler, line 8: Is there more information that you could give about the tool here, like its homepage?

benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,148 @@
+import benchexec.tools.template
```
PhilippWendler, line 1: We need a machine-readable license and copyright declaration, and would prefer if the license is Apache 2.0. Could you please add a header like in other BenchExec files?

benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Determine a version string for this tool, if available.
+        Do not hard-code a version in this function, either extract the version
+        from the tool or do not return a version at all.
+        There is a helper function `self._version_from_tool`
+        that should work with most tools, you only need to extract the version number
+        from the returned tool output.
+        @return a (possibly empty) string
+        """
+        return ""
```
PhilippWendler, line 42: This method does not return the tool's version, which means that users will not see in the result tables which tool version was benchmarked. This can be confusing and hinders reproducibility.

benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,148 @@
...
+        needs to know all files belonging to a tool
+        (to transport them to a cloud service, for example).
+        Returns a list of files or directories that are necessary to run the tool,
+        relative to the current directory.
+        The default implementation returns a list with the executable itself
+        and all paths that result from expanding patterns in self.REQUIRED_PATHS,
+        interpreting the latter as relative to the directory of the executable.
+        @return a list of paths as strings
+        """
+        return [executable] + self._program_files_from_executable(executable, ["lib", "lib32", "tools"], parent_dir=False)
```
PhilippWendler, line 56: Please overwrite only those methods that you actually need to overwrite. Instead of duplicating the code from `program_files`, you can simply declared a fields `REQUIRED_PATHS = ["lib", "lib32", "tools"]`.

benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,148 @@
...
+        All paths passed to this method (executable and fields of task)
+        are either absolute or have been made relative to the designated working directory.
+        @param executable: the path to the executable of the tool (typically the result of executable())
+        @param options: a list of options, in the same order as given in the XML-file.
+        @param task: An instance of of class Task, e.g., with the input files
+        @param rlimits: An instance of class ResourceLimits with the limits for this run
+        @return a list of strings that represent the command line to execute
+        """
+        return (
+            [executable, "--input_file", os.path.relpath(task.single_input_file, os.getcwd())] +
```
PhilippWendler, line 73: Please remove the `relpath` here. BenchExec takes care of path handling, and the working directory of the tool can be different from what `getcwd()` returns.

PhilippWendler (changes requested):
Thanks!

Please address the comments and format the code with [black](https://pypi.org/project/black/).

trtikm:
Requested changes are available.

benchexec/tools/sbt-fizzer.py
```diff
@@ -0,0 +1,158 @@
...
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import os
```
PhilippWendler, line 11: Please remove the unused import. Then I can merge it.

dbeyer:
@trtikm Could you please rename to fizzer.py, if that is the correct name of the tool?

trtikm:
The file is renamed.

PhilippWendler:
Hm, what is actually the name of the tool? Given that the `name()` method returns `sbt-fizzer` and that the GitHub Repo has the same name I would say that `sbt-fizzer.py` is also the better name of the module. Or do you plan to rename your tool to `fizzer`?

Btw., `name()` is supposed to return the human-readable tool name, and according to the readme on https://github.com/staticafi/sbt-fizzer this seems to be `SBT-Fizzer`. Why not return that?

trtikm:
The proper name of the tool is just Fizzer. So, the method name() now returns "fizzer".

PhilippWendler:
If the name of the tool is "Fizzer", then `name()` should return just that, not "fizzer".

(Side note: I would find it confusing that the GitHub repo is named differently than the tool, that the readme of that repo (consistently!) uses a different name for the tool as well, and that the executable also has that other name, and I would strongly recommend to make all that consistent. Otherwise people will for example google for "Fizzer" and not noticed that this "SBT-Fizzer" repo is supposed to be the same tool.)

Please format again with black and remove the unused import.

trtikm:
Requested changes are available.

PR #929, opened 2023-10-25
Added support for custom columns in ESBMC

rafaelsamenezes:
This adds support for using regex to extract more information from the run.

### DIFF
--- a/benchexec/tools/esbmc.py
+++ b/benchexec/tools/esbmc.py
@@ -10,6 +10,8 @@ import re
 from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
+import decimal
+
 
 class Tool(benchexec.tools.template.BaseTool2):
     """
@@ -69,10 +71,10 @@ class Tool(benchexec.tools.template.BaseTool2):
                 status = "ERROR"
 
         return status
-    
+
     def get_value_from_output(self, output, identifier):
         regex = re.compile(identifier)
-        matches = list()
+        matches = []
 
         # Match first element of each line
         for line in output:
@@ -82,9 +84,6 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         if len(matches) == 1:
             return matches[0]
-        elif len(matches) > 1:
-            total = 0
-            for x in matches:
-                total = total + float(x)
-            return str(total)
+        if len(matches) > 1:
+            return str(sum(decimal.Decimal(value) for value in matches))
         return None
--- a/benchexec/tools/esbmc.py
+++ b/benchexec/tools/esbmc.py
@@ -6,10 +6,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import re
 from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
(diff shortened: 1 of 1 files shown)


benchexec/tools/esbmc.py
```diff
@@ -6,11 +6,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import re
 from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
 import benchexec.tools.template
 import benchexec.result as result
 
-
```
PhilippWendler, line 13: Please keep this line (or format with `black`).

benchexec/tools/esbmc.py
```diff
@@ -69,3 +69,22 @@ def determine_result(self, run):
                 status = "ERROR"
 
         return status
+
```
PhilippWendler, line 72: Please avoid trailing whitespace (or format with `black`).

benchexec/tools/esbmc.py
```diff
@@ -69,3 +69,22 @@ def determine_result(self, run):
                 status = "ERROR"
 
         return status
+    
+    def get_value_from_output(self, output, identifier):
+        regex = re.compile(identifier)
+        matches = list()
```
PhilippWendler, line 75:
```suggestion
        matches = []
```
slightly more "pythonic"

benchexec/tools/esbmc.py
```diff
@@ -69,3 +69,22 @@ def determine_result(self, run):
...
+            match = regex.search(line.strip())
+            if match and len(match.groups()) >= 1:
+                matches.append(match.group(1))
+
+        if len(matches) == 1:
+            return matches[0]
+        elif len(matches) > 1:
+            total = 0
+            for x in matches:
+                total = total + float(x)
```
PhilippWendler, line 88: If you care a lot about precision, you could use the `decimal` package here, which implements arbitrary-precision decimals. And you can keep the value as is when returning without `str()`, BenchExec will handle this appropriately.

The whole `elif` body part could also be replaced with
`return sum(float(value) for value in matches)`
(equivalently with `decimal.Decimal(value)` instead of `float`.

PhilippWendler:
Nice to see this feature become more used!

benchexec/tools/esbmc.py
rafaelsamenezes: I did the change to use `decimal.Decimal`. However, if I don't do the cast, I will get `AttributeError: 'decimal.Decimal' object has no attribute 'lower'` at `benchexec/tablegenerator/util.py", line 157` at the end.

benchexec/tools/esbmc.py
rafaelsamenezes: Applied black

benchexec/tools/esbmc.py
rafaelsamenezes: Done

benchexec/tools/esbmc.py
rafaelsamenezes: Applied black

benchexec/tools/esbmc.py
PhilippWendler: Sorry, I had misremembered. But it is a nice feature to have, so I added it in e61f7b103365f0f9d9b36aca619394bc5cdb0bbf and changed your tool-info module accordingly.

PR #931, opened 2023-10-27
Add too-info module for JCWIT

Chriszai:
Add too-info module for JCWIT

### DIFF
--- a/benchexec/tools/jcwit.py
+++ b/benchexec/tools/jcwit.py
@@ -1,0 +1,69 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+from benchexec.tools.template import ToolNotFoundException
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+
+    """
+    Tool info for jcwit
+    (https://github.com/Chriszai/JCWIT).
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("jcwit.py")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return "jcwit"
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + list(task.input_files)
+
+    def determine_result(self, run):
+        output = run.output
+        validation = "unknown"
+        for line in output:
+            if "Exception" in line:
+                if "AssertionError" in line:
+                    validation = "false"
+                else:
+                    validation = "unknown"
+                    
+            if "Witness result: Unknown" in line:
+                validation = "unknown"
+                break
+            elif "Witness result: False" in line:
+                validation = "false"
+                break
+            else:
+                if "Witness validation: False" in line:
+                    validation = "false"
+                    break
+                elif "Witness validation: Unknown":
+                    validation = "unknown"
+                    break
+                else:
+                    validation = "true"
+
+        if validation == "false":
+            status = result.RESULT_FALSE_PROP
+
+        elif validation == "true":
+            status = result.RESULT_TRUE_PROP
+
+        elif validation == "unknown":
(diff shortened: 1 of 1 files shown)


benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+                break
+            else:
+                if "Witness validation: False" in line:
+                    validation = "false"
+                    break
+                elif "Witness validation: Unknown":
+                    validation = "unknown"
+                    break
+                else:
+                    validation = "true"
```
PhilippWendler, line 56: Having nested branching inside an `else:` like this is useless, you can just move the branches out to the outer if. And that will make it more clear that your logic sets  `validation = "true"` for almost every line (once the other bug is fixed), which does not seem intended.

benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+                validation = "unknown"
+                break
+            elif "Witness result: False" in line:
+                validation = "false"
+                break
+            else:
+                if "Witness validation: False" in line:
+                    validation = "false"
+                    break
+                elif "Witness validation: Unknown":
```
PhilippWendler, line 52: This condition is wrong, it always evaluates to `True`.

benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+        for line in output:
+            if "Exception" in line:
+                if "AssertionError" in line:
+                    validation = "false"
+                else:
+                    validation = "unknown"
+                    
+            if "Witness result: Unknown" in line:
+                validation = "unknown"
+                break
```
PhilippWendler, line 44: Setting a variable to a string here, then `break` and then compare that variable to the string again in order to set another variable to another value just in order to return that last value is overly complex and makes the control flow here difficult to understand. Why not just do `return result.RESULT_FALSE_PROP`? (Same for the similar cases below.)

If you want to make sure that having `Exception` in the output returns in an error result (which is understandable), then the current logic also does not fulfill this and needs to be changed (no `break`s for example).

benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def determine_result(self, run):
+        output = run.output
+        validation = "unknown"
+        for line in output:
+            if "Exception" in line:
+                if "AssertionError" in line:
+                    validation = "false"
+                else:
+                    validation = "unknown"
```
PhilippWendler, line 40: If the tool crashes, this should typically not be hidden in result `RESULT_UNKNOWN` but reflected with `RESULT_ERROR`. `RESULT_UNKNOWN` is for cases where the tool voluntarily gives up (e.g., because of a conceptually incomplete method).

It also seems weird that any kind of `AssertionError` produces `RESULT_FALSE_PROP`. Certainly a tool crash with `AssertionError` is not (always) indicating that the validation result is `FALSE`?

benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return "jcwit"
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + list(task.input_files)
+
+    def determine_result(self, run):
+        output = run.output
```
PhilippWendler, line 33: This variable is just used once and seems useless, could be inlined.

benchexec/tools/jcwit.py
```diff
@@ -0,0 +1,69 @@
...
+        return tool_locator.find_executable("jcwit.py")
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def name(self):
+        return "jcwit"
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + list(task.input_files)
```
PhilippWendler, line 30: Does your tool actually support tasks with more than one input file? If not, I would recommend to use `task.single_input_file`.

Is is correct that your tool does not need to know the data model of the C programs it gets? Typically this is required in order to be able to interpret the program semantics correctly.

PhilippWendler (changes requested):
Please format the source code with [black](https://pypi.org/project/black), as there are inconsistent line endings.

Chriszai:
My tool supports the task of multiple input files, it can be multiple files or a directory. I have made some changes, the new too-info module is available now.

PR #934, opened 2023-10-30
Added KLEEF tester

Columpio:


PhilippWendler:
It looks as if KLEEF is a fork of KLEE and the too-info module is copied from KLEE's.

What is the plan for KLEEF in the future? Will it be developed independently and might diverge from KLEE, such that for example also the output format or command-line parameters will look differently in the future?
Or is KLEEF more light an extension of KLEE that will regularly be updated to use the current version of KLEE and will thus always behave similarly with respect to inputs and outputs?

Or put differently: If for example in the future KLEE has some change in its output that will required the tool-info module to be adapted, will those changes also apply to KLEEF and the KLEEF tool-info module would also need to be adapted, or not?
If the answer is that yes, changes to tool-info modules will likely always be needed to be made in both modules, then the class in `kleef.py` should inherit from the one in `klee.py` and overwrite only those methods where changes are needed. We want to avoid redundant code where possible.

Columpio:
KLEEF is developed independently for quite a long time (several years) by a big team, and we do not have plans to stop the development in next 5 years. The output format of KLEEF already diverges from original KLEE and will diverge more and more in the future. The main reason for that is that KLEEF has some big features which KLEE has not (like trace reproduction), which greatly change both input and output formats.
So, KLEEF will not behave similarly to KLEE with respect to inputs and outputs.

Speaking about your last question, if KLEE changes its format, it will not affect KLEEF, as we follow a completely independent development path, so our format will more probably diverge more and more from the one the latest KLEE has.

PhilippWendler:
Thanks for the explanation! I was just wondering because for example the current tool-info module seems to indicate that KLEEF currently identifies as KLEE when called with `--version` and also for error messages. Will this stay as is or do you intend to change it soon? (Then we could wait until this can be reflected in the tool-info module).

Columpio:
Thank you for your suggestions. I have changed our `--version` output and force-pushed here a new version of the script. Unfortunately, we cannot change `KLEE:` mentions in the output right now as some industrial software using KLEEF relies on it.

benchexec/tools/kleef.py
```diff
@@ -0,0 +1,92 @@
...
+
+import benchexec.result as result
+import benchexec.tools.template
+import benchexec.model
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for KLEEF (https://github.com/UnitTestBot/klee).
```
PhilippWendler, line 16: I noticed that this URL is different from the URL below. What is the correct webpage of KLEEF?

PR #935, opened 2023-10-30
Added ULTIMATE TestGen tool-info module

MaxBarth95:


benchexec/tools/ultimatetestgenerator.py
```diff
@@ -0,0 +1,155 @@
...
+        "AutomizerCTestGeneratorCoverError.xml",
+        "Ultimate",
+        "Ultimate.ini",
+        "Ultimate.py",
+        "z3",
+        "mathsat",
+    ]
+
+    def name(self):
+        return "ULTIMATE TestGen"
```
PhilippWendler, line 55: What will the name of this tool be in the future (once it is merged in Ultimate's repository)?
- `TestGen` has returned here and in the class doc above?
- `TestGeneration` as the name of the branch?
- `TestGenerator` as the name of this file?

The name here and the name of the tool-info module should both be based on the correct tool name.

benchexec/tools/ultimatetestgenerator.py
```diff
@@ -0,0 +1,155 @@
...
+                return result.RESULT_TRUE_PROP
+            elif line.startswith("UNKNOWN"):
+                return result.RESULT_UNKNOWN
+            elif line.startswith("ERROR"):
+            	if not line.startswith("ERROR: Caught known exception: Unsupported non-linear arithmetic"):
+                    status = result.RESULT_ERROR
+                    if line.startswith("ERROR: INVALID WITNESS FILE"):
+                        status += " (invalid witness file)"
+                    return status
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 155: It seems that this is a lot of code copied from `ultimate.py`, with some small changes.

Do (some of) those change make sense to have in `ultimate.py` in general? I guess for example that this is the case for the check for non-linear arithmetic, or at least it wouldn't hurt to have in `ultimate.py`?

Where changes cannot be upstreamed: Can those methods be overwritten in a way that most of the code does not need to be copied, but is called with super?

PhilippWendler (changes requested):

benchexec/tools/ultimatetestgenerator.py
MaxBarth95: The tool name is Ultimate TestGen, short for Ultiamte TestGenerator. Should i rename the file to ultimatetestgen.py ?

PhilippWendler: If the tool name is "TestGen", then the name of the tool-info module should reflect this, yes.

benchexec/tools/ultimatetestgenerator.py
MaxBarth95: I dont want to iterate twice over the logfile and i think it is better if I return the non-linear arithmetic error anyway. I removed return values that are not used in test-case generation.

PhilippWendler: But this still uses duplicate code, and if Ultimate changes its output, then both `ultimate.py` and the new code here needs to be updated, which is easily forgotten. Can't there be one method that supports parsing the output for all the use cases?

MaxBarth95: I changed `ultimate.py`, Daniel gave his ok with this change. Any further suggestions?

PhilippWendler: Thanks, this is already much nicer, partially because now we can see what the actual difference is.

And it seems that the only difference in output parsing is the additional `line.startswidth("DONE")`. So I am wondering, would it really hurt to move that condition up into `ultimate.py`? It doesn't look at if this would break parsing other Ultimate variant outputs, and it would simplify the whole code even more. You could also choose to make the check stricter than `startswith("DONE")` if you fear it might match something else accidentally.

benchexec/tools/ultimate.py
```diff
@@ -333,10 +333,19 @@ def program_files(self, executable):
 
     def determine_result(self, run):
         if any(arg for arg in run.cmdline if "--spec" == arg or ".prp" in arg):
-            return self._determine_result_with_property_file(run)
+            return self._determine_result_with_property_file(self, run)
```
PhilippWendler, line 336: A method call with `self.f(self)` is just weird. Same for declaring `@staticmethod` with a `self` parameter. While it technically works, I see no reason to use it. A method intended for overwriting should simply not be declared `@staticmethod`.

Note that this and the whole other changes around the methods here would go away if the parsing logic for TestGen is simply inlined.

benchexec/tools/ultimatetestgenerator.py
MaxBarth95: You are right, @danieldietsch and I decided to change `ultiamte.py` back and added the condition there.

benchexec/tools/ultimate.py
```diff
@@ -66,6 +66,9 @@ class UltimateTool(benchexec.tools.template.BaseTool2):
     def __init__(self):
         self.java = None
 
+    def project_url(self):
+        return "https://github.com/ultimate-pa/ultimate"
```
danieldietsch, line 70: This URL should be https://www.ultimate-pa.org

danieldietsch (changes requested):

benchexec/tools/ultimate.py
MaxBarth95: Changed it, but then it is the wrong url in sys-lab:main, too

benchexec/tools/ultimatetestgenerator.py
PhilippWendler: Nice! Thanks being willing to iterate on this with me. I am confident that the current solution will create less maintenance effort in the future.

PR #936, opened 2023-10-30
Add tool-info module for SWAT

felixmaechtle:
Added SWAT to tools for the SV-COMP 2024 Verification competition.
Doi of the tool: `10.5281/zenodo.10054904`

### DIFF
--- a/benchexec/tools/swat.py
+++ b/benchexec/tools/swat.py
@@ -1,0 +1,43 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.result as result
+from benchexec.tools.template import BaseTool2
+
+
+class Tool(BaseTool2):
+    """
+    Tool info module for SWAT a tool ensemble for the dynamic symbolic execution of modern Java.
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("run-swat.sh")
+
+    def version(self, executable):
+        return '42'
+
+    def name(self):
+        return "SWAT"
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [executable] + options
+        if task.property_file:
+            cmd.append(task.property_file)
+        return cmd + list(task.input_files)
+
+    def determine_result(self, run):
+        status = result.RESULT_ERROR
+        if run.output.any_line_contains("== ERROR-UNREACH-CALL"):
+            status = result.RESULT_FALSE_REACH
+        elif run.output.any_line_contains("== ERROR"):
+            status = result.RESULT_FALSE_PROP
+        elif run.output.any_line_contains("== OK"):
+            status = result.RESULT_TRUE_PROP
+        elif run.output.any_line_contains("== DONT-KNOW"):
+            status = result.RESULT_UNKNOWN
+
+        return status
--- a/benchexec/tools/swat.py
+++ b/benchexec/tools/swat.py
@@ -18,10 +18,6 @@ class Tool(BaseTool2):
     def executable(self, tool_locator):
         return tool_locator.find_executable("run-swat.sh")
 
-#    def version(self, executable):
-#        return '42'
-#        #return self._version_from_tool(executable, arg="-v")
-
     def name(self):
         return "SWAT"
 
@@ -32,7 +28,6 @@ class Tool(BaseTool2):
         return cmd + list(task.input_files)
 
     def determine_result(self, run):
-
         if run.output.any_line_contains("== ERROR-UNREACH-CALL"):
             return result.RESULT_FALSE_REACH
         elif run.output.any_line_contains("== ERROR"):
--- a/benchexec/tools/swat.py
+++ b/benchexec/tools/swat.py
@@ -11,14 +11,16 @@ from benchexec.tools.template import BaseTool2
 
 class Tool(BaseTool2):
     """
-    Tool info module for SWAT a tool ensemble for the dynamic symbolic execution of modern Java.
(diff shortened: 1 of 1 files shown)


benchexec/tools/swat.py
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.result as result
+from benchexec.tools.template import BaseTool2
+
+
+class Tool(BaseTool2):
+    """
+    Tool info module for SWAT a tool ensemble for the dynamic symbolic execution of modern Java.
+    """
```
PhilippWendler, line 15: Is there also a tool webpage that could be linked here? "SWAT" is unfortunately a very generic name that seems impossible to look up via a web search. Even a search for "swat tool ensemble for the dynamic symbolic execution of modern Java" does not seem to find your tool, only GDart?!

benchexec/tools/swat.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("run-swat.sh")
+
+    def version(self, executable):
+        return '42'
```
PhilippWendler, line 21: Hard-coding a version number in this module is not a good idea. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

If this is not possible, please remove the method completely. (But note that if you want to participate in SV-COMP or Test-Comp, AFAIK the rules there require a proper version number.)

benchexec/tools/swat.py
```diff
@@ -0,0 +1,43 @@
...
+        if run.output.any_line_contains("== ERROR-UNREACH-CALL"):
+            status = result.RESULT_FALSE_REACH
+        elif run.output.any_line_contains("== ERROR"):
+            status = result.RESULT_FALSE_PROP
+        elif run.output.any_line_contains("== OK"):
+            status = result.RESULT_TRUE_PROP
+        elif run.output.any_line_contains("== DONT-KNOW"):
+            status = result.RESULT_UNKNOWN
+
+        return status
```
PhilippWendler, line 43: The `status` variable seems superfluous here. You could just use `return` in every branch.

benchexec/tools/swat.py
```diff
@@ -0,0 +1,43 @@
...
+        return '42'
+
+    def name(self):
+        return "SWAT"
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [executable] + options
+        if task.property_file:
+            cmd.append(task.property_file)
+        return cmd + list(task.input_files)
```
PhilippWendler, line 30: Does your tool support tasks that consist of more than one input file? If yes, this is fine, otherwise please use `task.single_input_file` such that users immediately get a proper error.

PhilippWendler (changes requested):

benchexec/tools/swat.py
felixmaechtle: Hi, it was a rather late decision to participate, so we have not set up a website or a proper repository. 
Our plan is to do that by the final deadline on the 22nd. But we don't have one for the pre-run. 

If it's mandatory, I can set up a quick website on our institute's homepage, but for now it would be rather empty.

PhilippWendler: If you already know what the URL of the tool webpage will be in the future, you can put it here already. Otherwise please file another PR later to update it.

benchexec/tools/swat.py
felixmaechtle: It should be no problem to add a proper version flag.  
However, would it be possible to leave it as it is for the pre-run? I have tested the current submission a few times and have already uploaded it to Zenodo. I am afraid that any last minute changes would break it.

But I can assure you that we will fix this for the final submission.

felixmaechtle: I've just read the other comments. I'll just remove / comment the function for the Pre-Run, if that is okay with you?

PhilippWendler: A hard-coded version number is not acceptable for BenchExec. A missing one is suboptimal, but is acceptable if the tool does not have a version number.

Everything else is not the concern of BenchExec, and would need to be discussed elsewhere if necessary.

benchexec/tools/swat.py
felixmaechtle: As we are in the Java context, we retrieve multiple arguments, i.e. the path to the common folder and the SUT itself. 
Upon refractoring I get the following error: `Tool does not support tasks with more than one input file`. So I assume this is correct.

benchexec/tools/swat.py
felixmaechtle: I've added a link to the institutes website as a first draft :)

We will do an additional PR until the 22nd!

felixmaechtle:
I've removed the version number for the pre-run of SV-Comb, however, we will do another PR with a correct version number, i.e. dynamic, and a link to our tool after the pre-run.

benchexec/tools/swat.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def determine_result(self, run):
+
```
PhilippWendler, line 35: Our code formatter would remove this empty line. Please run [Black](https://pypi.org/project/black/) or remove it manually.

benchexec/tools/swat.py
```diff
@@ -0,0 +1,45 @@
...
+    SWAT is currently being developed by the Institute for IT Security at the University of Luebeck ( https://www.its.uni-luebeck.de/en/institute.html ).
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("run-swat.sh")
+
+#    def version(self, executable):
+#        return '42'
+#        #return self._version_from_tool(executable, arg="-v")
+
```
PhilippWendler, line 24: Just remove this for now. The line with `return '42'` does not make sense at all, and the whole comment does not conform to our code formatter, so would need to be reformatted otherwise.

benchexec/tools/swat.py
felixmaechtle: Blank line is removed.

PR #940, opened 2023-11-01
FDSE: add tool info module

passengerZ:


benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("fdse")
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def version(self, executable):
+        return "fdse-testcomp"
```
PhilippWendler, line 27: Hard-coding a version number in this module is not a good idea. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

If this is not possible, please remove the method completely. (But note that AFAIK Test-Comp does require a proper version number.)

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for FDSE.
+    https://github.com/zbchen/FDSE
```
PhilippWendler, line 15: This link doesn't exist?

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("fdse")
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
```
PhilippWendler, line 24: Why are you overwriting this? `REQUIRED_PATHS` is empty, so this has no effect.

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def version(self, executable):
+        return "fdse-testcomp"
+
+    def cmdline(self, executable, options, task, rlimits):
+        new_option = ["--testcomp"]
```
PhilippWendler, line 30: Is this an option that is always required to run your tool together with this tool-info module, or is this a configuration option that would not be necessary in other circumstances, e.g., outside of Test-Comp? Then please leave it out and put it in the benchmark definition for Test-Comp.

BenchExec is a general benchmarking framework and its use is not restricted to Test-Comp, so we want to keep the code general here.

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+        This method always needs to be overridden.
+        If the tool gave a result, this method needs to return one of the
+        benchexec.result.RESULT_* strings.
+        Otherwise an arbitrary string can be returned that will be shown to the user
+        and should give some indication of the failure reason
+        (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
+        """
+        for line in run.output:
+            if "[BUG]" in line:
+                if line.find("ASSERTION FAIL!") != -1:
```
PhilippWendler, line 54: All the uses of `find()` could be replaced by `str in line`.

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+                return result.RESULT_DONE
+        return result.RESULT_UNKNOWN
+
+    def get_value_from_output(self, run):
+        status = result.RESULT_UNKNOWN
+
+        if run.output.any_line_contains("DONE"):
+            status = result.RESULT_DONE
+
+        return status
```
PhilippWendler, line 72: This is wrong, because `get_value_from_output` has a different purpose than returning a status value, and it even needs to have a different number of parameters. If you do not want the use case of `get_value_from_output`, please remove it.

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,72 @@
...
+        for line in run.output:
+            if "[BUG]" in line:
+                if line.find("ASSERTION FAIL!") != -1:
+                    return result.RESULT_FALSE_REACH
+                elif line.find("Index out-of-bound") != -1:
+                    return result.RESULT_FALSE_DEREF
+                elif line.find("overflow") != -1:
+                    return result.RESULT_FALSE_OVERFLOW
+                else:
+                    return f"ERROR ({run.exit_code.value})"
```
PhilippWendler, line 61: Providing `exit_code.value` is something that BenchExec would do as well anyway, if you just return `ERROR`. And the implementation in BenchExec works better in case there is no exit value (e.g., because the tool was killed due to timeout). So I suggest to return `RESULT_ERROR` here.

PhilippWendler (changes requested):

PhilippWendler:
Please do not create multiple useless PRs. If you want to update a PR, keep it open and push to the respective branch.

benchexec/tools/fdse.py
passengerZ: Oh!  It's my fault, thank you so much.

benchexec/tools/fdse.py
passengerZ: Sorry, I forgot to delete this function. I will finish cleaning up soon

benchexec/tools/fdse.py
passengerZ: Good idea !

benchexec/tools/fdse.py
passengerZ: This link is currently private.

benchexec/tools/fdse.py
passengerZ: Your suggestion is very correct, I will fix it soon.

benchexec/tools/fdse.py
passengerZ: This function will be remove soon.

benchexec/tools/fdse.py
passengerZ: I got it. and I also think it's a good idea.

benchexec/tools/fdse.py
PhilippWendler: Is there a public URL for your tool? Or will this be made public?

Please remove the "(now is private)" comment from the code, it doesn't make sense to have it there because you would only need to file another pull request once the page becomes online.

benchexec/tools/fdse.py
```diff
@@ -0,0 +1,50 @@
...
+                    return result.RESULT_FALSE_REACH
+                elif "Index out-of-bound" in line:
+                    return result.RESULT_FALSE_DEREF
+                elif "overflow" in line:
+                    return result.RESULT_FALSE_OVERFLOW
+                else:
+                    return result.RESULT_ERROR
+            if "Done : End analysis" in line:
+                return result.RESULT_DONE
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 50: If your tool does not manage to produce any of the above output, does this really indicate "unknown" (i.e., tool gave up). Wouldn't it be an error? (Consider cases like crashes, seg faults, OOM, etc.)

PhilippWendler:
Please comment once you have pushed something and think the PR is ready, otherwise I will not notice.

benchexec/tools/fdse.py
passengerZ: Hello, my tool is currently only intended to participate in the coverage of the testcomp, so I have removed the output classification.

passengerZ: The capture error is currently based on the signal provided by the system, so it has not yet been able to accurately distinguish the error.

PhilippWendler: Sorry, I don't really understand what you mean with "capture error". If you want to get the exit value of your program, you can do so via `run.exit_code`.

If you just want to distinguish between "done" and "error", this is also fine. Then just please change `RESULT_UNKNOWN` to `RESULT_ERROR` in order to do so.

passengerZ: Thank you for your suggestion. Now the output can be divided into three types: done, timeout, and error.

PR #942, opened 2023-11-01
add tool-info module of aise for SV-COMP 2024

ZhenWang233:


benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.result as result
+import benchexec.tools.template
+import benchexec.model
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for AISE.
```
PhilippWendler, line 15: Please add a URL for your tool here. AISE seems to be a quite common abbreviation and it seems difficult to find your tool via Google.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("aise", subdir="bin")
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
```
PhilippWendler, line 24: Overwriting this method in this way is useless unless you define `REQUIRED_PATHS`. So just remove it.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 27: Does your tool support tasks with several input files? Does your tool support tasks without any input file at all? If yes, you can keep it, otherwise please use one of the other properties than `task.input_files_or_identifier`.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def name(self):
+        return "AISE"
+
+    def version(self, executable):
+        return "AISE 1.0.0"
```
PhilippWendler, line 33: Hard-coding a version number in this module is not a good idea. If people run a newer version of your tool with BenchExec they will get a wrong version number in the output, which could be very confusing.

If this is not possible, please remove the method completely. (But please note that SV-COMP and Test-Comp AFAIK require a proper version number.)

Also, the version number should not duplicate the tool name.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+        This method always needs to be overridden.
+        If the tool gave a result, this method needs to return one of the
+        benchexec.result.RESULT_* strings.
+        Otherwise an arbitrary string can be returned that will be shown to the user
+        and should give some indication of the failure reason
+        (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
+        """
+        for line in run.output:
+            if line.startswith("KLEE: ERROR: "):
+                if line.find("ASSERTION FAIL:") != -1:
```
PhilippWendler, line 47: You can just replace all uses of `find()` with `needle in haystack`, e.g., `"ASSERTION FAIL:" in line`. This is easier to read.

And you could easily merge these two `if`s into one.

Same below.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_value_from_output(self, lines, identifier):
+        # search for the text in output and get its value,
+        # stop after the first line, that contains the searched text
+        for line in lines:
+            if (
+                line.startswith("KLEE: done: ") or line.startswith("AISE: done: ")
+            ) and line.find(identifier + " = ") != -1:
+                startPosition = line.rfind("=") + 2
+                return line[startPosition:].strip()
```
PhilippWendler, line 85: This could be simplified with something like `return line.rsplit("=", maxsplit=1)[-1].strip()` (at least I think this is what you want, right?).

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+            elif line.find("Number of total " + identifier) != -1:
+                endPosition = line.find(" Number")
+                return line[0:endPosition].strip()
```
PhilippWendler, line 88: Are you sure this is correct? Are you really looking for `Number ...` first and then for ` Number`? What would an example output line for this look like? (Hint: it is a good idea to add such an example here as a comment for documentation).

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+        and should give some indication of the failure reason
+        (e.g., "CRASH", "OUT_OF_MEMORY", etc.).
+        """
+        for line in run.output:
+            if line.startswith("KLEE: ERROR: "):
+                if line.find("ASSERTION FAIL:") != -1:
+                    return result.RESULT_FALSE_REACH
+        for line in run.output:
+            if line.startswith("CLAM ERROR:"):
+                return f"ERROR ({run.exit_code.value})"
```
PhilippWendler, line 51: Providing `exit_code.value` is something that BenchExec would do as well anyway, if you just return `ERROR`. And the implementation in BenchExec works better in case there is no exit value (e.g., because the tool was killed due to timeout). So I suggest to return `RESULT_ERROR` here.

Same below.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,89 @@
...
+            if line.startswith("KLEE: WARNING"):
```
PhilippWendler, line 52: This level of `if`s with the `startswith` calls could use `elif`. It wouldn't change the semantics, but be more obvious to read.

PhilippWendler (changes requested):

PhilippWendler:
Please do not create multiple useless PRs. If you want to update a PR, keep it open and push to the respective branch.

benchexec/tools/aise.py
ZhenWang233: Ok, I will add a URL.

benchexec/tools/aise.py
ZhenWang233: Ok, I will remove it.

benchexec/tools/aise.py
ZhenWang233: No, it just need single input file. I fixed it.

benchexec/tools/aise.py
ZhenWang233: Ok, I added a parameter to return version.

benchexec/tools/aise.py
ZhenWang233: Ok, I fixed it.

benchexec/tools/aise.py
ZhenWang233: Sorry for my poor code, I removed the  "get_value_from_output" function. it is just used in my local experiment.

benchexec/tools/aise.py
ZhenWang233: Ok, i fixed it.

benchexec/tools/aise.py
ZhenWang233: Ok, I fixed it

benchexec/tools/aise.py
PhilippWendler: There is no need to remove the function, it is fine to have it if it does something useful.

benchexec/tools/aise.py
```diff
@@ -0,0 +1,67 @@
...
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for AISE(https://github.com/ZhenWang233/AISE).
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("aise", subdir="bin")
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + [task.single_input_file]
```
PhilippWendler, line 22: Just noticed: If you want to participate in SV-COMP'24, are you sure you do not need additional information from the task, in particular the data model?

ZhenWang233: Thank you for your suggestion, but I want to keep it in the current version. I used a hard-code to built-in a property (unreachcall, ILP32) in our tool. Because we only focus on a subcategory"ReachSafety-Loops" in SV-COMP'24, their data models are same, so built-in a property can work in most cases. I known it is not a good handling, but changing it need some time and testing.

PR #943, opened 2023-11-01
Update tool-info module for PredatorHP

versokova:


.gitignore
```diff
@@ -17,3 +17,4 @@
 node_modules/
 /contrib/vcloud/lib/ivy*.jar
 /contrib/vcloud/lib/vcloud-jars
+.DS_Store
```
PhilippWendler, line 20: Please don't mix other changes into your PR.

benchexec/tools/predatorhp.py
```diff
@@ -23,7 +23,7 @@ def name(self):
         return "PredatorHP"
 
     def version(self, executable):
-        return self._version_from_tool(executable, use_stderr=True)
+        return self._version_from_tool(executable)
```
PhilippWendler, line 26: Does this mean that the new code would not work anymore for old versions of PredatorHP? We aim to keep backwards-compatibility here, because other people might use this for benchmarking older versions. So if the answer is yes, then please change the code to support both old and new versions.

versokova: No possible. I don't want to use python2 - is old and need to be installed. I googled it and if you know how to force python3 to print to stderr I'd appreciate it.

.gitignore
versokova: OK. Sorry.

benchexec/tools/predatorhp.py
PhilippWendler: Note that in no way I wanted to imply you have to change your tool. If it prints to stderr, this is fine. But the code here should then support both current and old versions of your tool, this is all.

PR #944, opened 2023-11-01
Improve the tool-info modules of ABC and AVR

Po-Chun-Chien:
Changes include:

- add `REQUIRED_PATHS` attributes for ABC and AVR
- parse version of ABC
- pass resource limits to AVR's command lines

### DIFF
--- a/benchexec/tools/avr.py
+++ b/benchexec/tools/avr.py
@@ -38,7 +38,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         """
         if run.was_timeout:
             return result.RESULT_TIMEOUT
-        for line in run.output:
+        for line in run.output[::-1]:
             # skip the lines that do not contain verification result
             if not line.startswith("Verification result:"):
                 continue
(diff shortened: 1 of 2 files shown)


benchexec/tools/avr.py
```diff
@@ -15,13 +15,21 @@ class Tool(benchexec.tools.template.BaseTool2):
     URL: https://github.com/aman-goel/avr
     """
 
+    REQUIRED_PATHS = ["build/"]
```
PhilippWendler, line 18: A directory with this name looks as if it contains build-time stuff, which is typically not required at runtime. Are you sure that this is not the case here?

benchexec/tools/avr.py
```diff
@@ -15,13 +15,21 @@ class Tool(benchexec.tools.template.BaseTool2):
...
+
     def executable(self, tool_locator):
         return tool_locator.find_executable("avr.py")
 
     def name(self):
         return "AVR"
 
     def cmdline(self, executable, options, task, rlimits):
+        from math import ceil
```
PhilippWendler, line 27: I don't see a reason why this needs to be a local import, so please make it a standard module-level import.

Po-Chun-Chien: Done in a37525b.

benchexec/tools/avr.py
Po-Chun-Chien: Yes, the build script of AVR by default places the compiled executables under `build/` and the intermediate object code (`*.o`) under `src/`.

PR #946, opened 2023-11-01
Create concurrentwitness2test.py

leventeBajczi:
We plan to submit a new validator for violation witnesses in the ConcurrencySafety category. This PR adds support for this new tool to benchexec.

### DIFF
--- a/benchexec/tools/concurrentwitness2test.py
+++ b/benchexec/tools/concurrentwitness2test.py
@@ -1,0 +1,43 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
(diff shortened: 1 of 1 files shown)


benchexec/tools/concurrentwitness2test.py
```diff
@@ -0,0 +1,42 @@
...
+            elif "Verdict: Unknown error" in line:
+                status = result.RESULT_ERROR
+            elif "Verdict: Incompatible witness" in line:
+                status = result.RESULT_ERROR + "(Incompatible witness)"
+            elif "Verdict: Parsing failed" in line:
+                status = result.RESULT_ERROR + "(Parsing failed)"
+            elif "Verdict: Compilation error" in line:
+                status = result.RESULT_ERROR + "(Compilation error)"
+
+        return status
```
PhilippWendler, line 42: Is it intended that the last verdict line wins? If you expect there to be only one, you could just return early and avoid the status variable.

leventeBajczi: Fixed in new commit.

PR #947, opened 2023-11-01
Adds tool info module for owic

filipeom:
Tool info module for owic participating in Test-Comp 24 (Easychair submission 2394)

Artifact:
 
[DOI](https://doi.org/10.5281/zenodo.10064291)

cc @zapashcanon

### DIFF
--- a/benchexec/tools/owic.py
+++ b/benchexec/tools/owic.py
@@ -1,0 +1,56 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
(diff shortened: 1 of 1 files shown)


benchexec/tools/owic.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("owic", subdir="bin")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, "--version")
+
+    def name(self):
+        return "owic"
```
PhilippWendler, line 26: Is the tool name "owi" or "owic"? Here the proper tool name should be returned.

And is the spelling "owi", "OWI", or "Owi"? I can find all three of these spellings here and in your readme.

filipeom: The tool executable is named `owic`. But the display name should be "OWI" as specified in the benchmark and category definitions https://gitlab.com/sosy-lab/test-comp/bench-defs/-/merge_requests/114. Will update this accordingly, thanks!

PR #948, opened 2023-11-01
Add tool-info module for Rizzer

adamstafa:
This pull request adds support for the tool [Rizzer](https://github.com/staticafi/sbt-fizzer/tree/rizzer). It's based off of Fizzer merged in #921, hence the structure is very similar.

### DIFF
--- a/benchexec/tools/rizzer.py
+++ b/benchexec/tools/rizzer.py
@@ -5,18 +5,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import benchexec.tools.template
-import benchexec.result as result
-from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import benchexec.tools.fizzer as fizzer
 
-
-class Tool(benchexec.tools.template.BaseTool2):
+class Tool(fizzer.Tool):
     """
     Tool info for rizzer.
     https://github.com/staticafi/sbt-fizzer/tree/rizzer
     """
 
-    REQUIRED_PATHS = ["lib", "lib32", "tools"]
+    REQUIRED_PATHS = fizzer.REQUIRED_PATHS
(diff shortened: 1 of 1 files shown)


PhilippWendler:
If the tool is based on Fizzer, does this mean that they will also share future development? I.e., if the output of Fizzer changes and the tool-info module needs to be updated, is it likely that the tool-info module of Rizzer will also need to be updated as well?

If yes, then inheritance should be used to avoid the duplicate code and the associated maintenance burden.

adamstafa:
Both tools currently have a separate branch in the repository and will diverge a bit in the near future. We plan to merge them at some point, but until then, it is easier to have an independent tool-info module for each tool.

PhilippWendler:
Do you plan in letting them diverge with respect to their inputs and outputs, i.e., those things that would reflect anything here in the tool-info module? Especially if you plan anyway on integrating them, I expect this to not be the case and thus the common code in the tool-info modules should be refactored out.

adamstafa:
The output will likely change for both tools. Fizzer's output has been changing a lot in the past few months and Rizzer will probably output some information regarding potential crashes of Klee that is used.

PhilippWendler:
First, note that once this is merged here and used, if you update your tool's output later we will expect to keep backwards compatibility of the tool-info module with old versions of the tool. This is in order to let people in the future with newer BenchExec replicate benchmarks with old tool versions. So consider this when making changes, maybe you can reduce the maintenance burden that comes with this.

Second, as long as it is unclear what will change, and given the fact that the current code needs to stay there anyway for backwards compatibility, it seems clear that having that much duplicated code is not good practice and will make maintenance more effort. Thus please refactor.

adamstafa:
I've changed the module so it inherits Fizzer's module. Is this alright?

PR #949, opened 2023-11-02
sv-comp  tool info module for new tool Proton

rmetta:
Proton is a new tool participating in SV-COMP 2024 in the termination category (only). This is the tool info module required for benchexec. Please merge.

### DIFF
--- a/benchexec/tools/owic.py
+++ b/benchexec/tools/owic.py
@@ -1,0 +1,56 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    This class serves as tool adaptor for OWI (https://github.com/OCamlPro/owi)
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("owic", subdir="bin")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, "--version")
+
+    def name(self):
+        return "OWI"
+
+    def cmdline(self, executable, options, task, _):
+        if task.property_file:
+            options = options + ["--property", task.property_file]
+        else:
+            raise benchexec.tools.template.UnsupportedFeatureException(
+                "property file is required"
+            )
+
+        data_model_param = get_data_model_from_task(task, {ILP32: "32", LP64: "64"})
+        if data_model_param and "--arch" not in options:
+            options += ["--arch", data_model_param]
+
+        return [executable] + options + [task.single_input_file]
+
+    def determine_result(self, run):
+        if run.exit_code == 1:
+            return result.RESULT_ERROR
+
+        for line in run.ouput:
+            if "Assert failure" in line:
+                return result.RESULT_FALSE_REACH
+            elif "ERROR" in line:
+                return result.RESULT_ERROR
+            elif "All OK" in line:
+                return result.RESULT_DONE
+
+        if run.was_timeout:
+            return result.RESULT_TIMEOUT
+        return result.RESULT_UNKNOWN


PhilippWendler:
What is the relation between CBMC and Proton? More specifically, if for example the output of CBMC changes and the tool-info module needs to be updated, is it likely that the tool-info module of Proton will also need to be updated as well?

If yes, then inheritance should be used to avoid the duplicate code and the associated maintenance burden.

I also don't really understand the code at the top of the file. Why would a new tool-info module be created next year? Note that this tool-info module here for BenchExec is independent of SV-COMP. Even though it will be used in SV-COMP, it can and will likely be used by other people as well to benchmark Proton in other use cases.

rmetta:
Hi.

There is no module related dependency between Proton and Cbmc. Proton,
instruments a given C program, with assertions for checking
non-termination. and then invokes cbmc. We are a bit late into the
development and are pressed for time, And so, we just took cbmc's tool
definition module (and other related files), and modified these files
as appropriate for proton.

We chose the year as 2024 as it is SV-COMP "2024".  In this needs to
be changed to "2023", please let us know, and we will update
accordingly.

Best regards,
Ravi

PhilippWendler:
I am not sure what you mean with "module related dependency". But it seems that the output of Proton will always be the output of CBMC, right?

So we do not want to duplicate the code from CBMC's tool-info module for the typical reasons that duplicate code is bad. Please make the tool-info module for Proton inherit from the one for CBMC.

rmetta:
The proton script calls cbmc executable (after appropriate annotations to
the source code), checks the cbmc output and accordingly reports an output.
Right now, it is same as cbmc, but we do intend to change this in future.
Is this OK?

Best regards,
Ravi

rmetta:
The proton script calls cbmc executable (after appropriate annotations to the source code), checks the cbmc output and accordingly reports an output. Right now, after checking CBMC's results, we are outputng "FALSE(termination)", which CBMC doesn't.. We are sorry for any confusion we used the cbmc's tool info py module  as a template. We have now changed it, to reflect only termination., and also our error codes

PhilippWendler:
First, note that if you change your output in the future, you will need to update the tool-info module here in a backwards compatible manner and keep the support for parsing the output of old versions of your tool. Otherwise people could no longer benchmark these versions of your tool with a future BenchExec version.
As this is typically somewhat complex and adds maintenance burden, you might want to consider if you can avoid that (or make the switch immediately).

For the current output of the tool, I still see no reason why the code here should be duplicated from the tool-info module of CBMC instead of being reused.

rmetta:
> First, note that if you change your output in the future, you will need to update the tool-info module here in a backwards compatible manner and keep the support for parsing the output of old versions of your tool. Otherwise people could no longer benchmark these versions of your tool with a future BenchExec version. As this is typically somewhat complex and adds maintenance burden, you might want to consider if you can avoid that (or make the switch immediately).
> 
> For the current output of the tool, I still see no reason why the code here should be duplicated from the tool-info module of CBMC instead of being reused.

So, in the tool info module "proton.py" (copied from cbmc.py), we need to specifiy that our bash-script "proton" needs to be called, and the error messages displayed should be as per what we need.  How to do this via cbmc's tool definition module? Also, we have updated the proton.py file to tr
[proton.py.txt](https://github.com/sosy-lab/benchexec/files/13237333/proton.py.txt)
y to avoid any confusion we may have cause. We have attached this here. Please take a look at and kindly let us know if this will do.  If this attached one is ok, we will cancel this merge request and upload this one

PhilippWendler:
> So, in the tool info module "proton.py" (copied from cbmc.py), we need to specifiy that our bash-script "proton" needs to be called, and the error messages displayed should be as per what we need. How to do this via cbmc's tool definition module?

I mentioned already that you can inherit from CBMC's tool-info module. And of course you can then overwrite the necessary methods.

Actually, I am now wondering whether the current code even makes sense for Proton. Does Proton have the same two kinds of output as CBMC, with XML and without XML? Does it support the `--xml-ui` parameter? Do you want to support the XML output in this tool-info module? If not, there is a lot of completely useless code here. If that is removed, there might not even remain that much of duplicated code. So first it would be good to get an understanding of this.

rmetta:
> > So, in the tool info module "proton.py" (copied from cbmc.py), we need to specifiy that our bash-script "proton" needs to be called, and the error messages displayed should be as per what we need. How to do this via cbmc's tool definition module?
> 
> I mentioned already that you can inherit from CBMC's tool-info module. And of course you can then overwrite the necessary methods.
> 
> Actually, I am now wondering whether the current code even makes sense for Proton. Does Proton have the same two kinds of output as CBMC, with XML and without XML? Does it support the `--xml-ui` parameter? Do you want to support the XML output in this tool-info module? If not, there is a lot of completely useless code here. If that is removed, there might not even remain that much of duplicated code. So first it would be good to get an understanding of this.

Thanks for pointhing this out. There are indeed many options that we did not neet. So, we have re-written the tool info module to align with our requirement. We have attached the same here. Request you to please take a look at it and let us know if this is better.

[proton.py.txt](https://github.com/sosy-lab/benchexec/files/13237725/proton.py.txt)

rmetta:
Sorry, we just updated our local (forked) branch discarding the earlier commit., and this request got closed.  We appreciate your   help in this matter, and we hope that new proton.py file works as intended without dependency on cbmc

PhilippWendler:
It seems that in this case you need to create a new PR.

PR #952, opened 2023-11-02
Proton tool info module proton.py

rmetta:
Rewrote the Proton tool info module, based on the feedback from Philipp Wendler.

### DIFF
--- a/benchexec/tools/proton.py
+++ b/benchexec/tools/proton.py
@@ -12,7 +12,7 @@ import benchexec.result as result
 
 class Tool(benchexec.tools.template.BaseTool2):
     """
-    PROTON -- intended to be shared in future  at : https://github.com/kumarmadhukar/term
+    PROTON --- https://github.com/kumarmadhukar/term
     """
 
     def executable(self, tool_locator):
@@ -39,7 +39,6 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         status = result.RESULT_ERROR
         if run.exit_code.value in [0, 10] and len(output) > 0:
-            # SV-COMP mode
             result_str = output[-1].strip()
 
             if result_str == "TRUE":
--- a/benchexec/tools/proton.py
+++ b/benchexec/tools/proton.py
@@ -1,0 +1,70 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    PROTON
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("proton")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, use_stderr=True)
+
+    def name(self):
+        return "PROTON"
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options = options + ["--propertyFile", task.property_file]
+
+        data_model_param = get_data_model_from_task(task, {ILP32: "32", LP64: "64"})
+        if data_model_param and data_model_param not in options:
+            options += [data_model_param]
+
+        self.options = options
+
+        return [executable] + options + list(task.input_files_or_identifier)
+
+    def determine_result(self, run):
+        output = run.output
+
+        if run.exit_code.value in [0, 10]:
+            status = result.RESULT_ERROR
+            if len(output) > 0:
+                # SV-COMP mode
+                result_str = output[-1].strip()
+
+                if result_str == "TRUE":
+                    status = result.RESULT_TRUE_PROP
+                elif "FALSE" in result_str:
+                    if result_str == "FALSE(termination)":
+                        status = result.RESULT_FALSE_TERMINATION
+                    else:
+                        status = result.RESULT_FALSE_REACH
+                elif "UNKNOWN" in output:
+                    status = result.RESULT_UNKNOWN
+
+        elif run.exit_code.value == 64 and "Usage error!" in output:
+            status = "INVALID ARGUMENTS"
+
+        elif run.exit_code.value == 6 and "Out of memory" in output:
+            status = "OUT OF MEMORY"
+
+        elif run.exit_code.value == 6 and "SAT or SMT checker error: out-of-memory or internal-error" in output:
+            status = "OUT-OF-MEMORY or INTERNAL-ERROR"
+
+        else:
+            status = result.RESULT_ERROR
+
+        return status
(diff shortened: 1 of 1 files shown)


rmetta:
We have updated the proton.py by fixing the error in "Check code format".

benchexec/tools/proton.py
```diff
@@ -0,0 +1,70 @@
...
+        self.options = options
+
+        return [executable] + options + list(task.input_files_or_identifier)
+
+    def determine_result(self, run):
+        output = run.output
+
+        if run.exit_code.value in [0, 10]:
+            status = result.RESULT_ERROR
+            if len(output) > 0:
```
PhilippWendler, line 44: To make the code easier to read, I guess one would move this condition into the `if run.exit_code.value in [0, 10]:` line. Then one gets rid of one nesting of `if`s.

benchexec/tools/proton.py
```diff
@@ -0,0 +1,70 @@
...
+        elif run.exit_code.value == 6 and "Out of memory" in output:
+            status = "OUT OF MEMORY"
+
+        elif run.exit_code.value == 6:
+            status = "OUT-OF-MEMORY or INTERNAL-ERROR"
+
+        else:
+            status = result.RESULT_ERROR
+
+        return status
```
PhilippWendler, line 70: Why not have a `return` everywhere instead of the `status` variable? Seems easier to follow that way.

benchexec/tools/proton.py
```diff
@@ -0,0 +1,70 @@
...
+                        status = result.RESULT_FALSE_TERMINATION
+                    else:
+                        status = result.RESULT_FALSE_REACH
+                elif "UNKNOWN" in output:
+                    status = result.RESULT_UNKNOWN
+
+        elif run.exit_code.value == 64 and "Usage error!" in output:
+            status = "INVALID ARGUMENTS"
+
+        elif run.exit_code.value == 6 and "Out of memory" in output:
```
PhilippWendler, line 61: These conditions with `in output` will check if there is a line in the output that contains exactly this string (and nothing else). Is this intended?

Especially for `UNKNOWN` I am wondering because for `FALSE` and `TRUE` only the last line is checked, and for `FALSE` even with a substring match.

benchexec/tools/proton.py
```diff
@@ -32,35 +32,29 @@ def cmdline(self, executable, options, task, rlimits):
...
-                elif "FALSE" in result_str:
-                    if result_str == "FALSE(termination)":
-                        status = result.RESULT_FALSE_TERMINATION
-                    else:
-                        status = result.RESULT_FALSE_REACH
-                elif "UNKNOWN" in output:
-                    status = result.RESULT_UNKNOWN
+        status = result.RESULT_ERROR
+        if run.exit_code.value in [0, 10] and len(output) > 0:
+            # SV-COMP mode
```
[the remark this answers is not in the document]
rmetta: Fixed both the above and committed. Thanks a lot, for the rigorous review.

benchexec/tools/proton.py
PhilippWendler: I would like to get confirmation from you that this inconsistency between e.g. `TRUE` and `UNKNOWN` is indeed intendend.

rmetta:
We have fixed this now and committed the fixed version. Actually, we had
different error strings in our implementation of Proton. Now, made the
error codes in sync with our implementation.

Best regards,
Ravi

benchexec/tools/proton.py
```diff
@@ -0,0 +1,65 @@
...
+                status = result.RESULT_FALSE_TERMINATION
+
+            elif "UNKNOWN" in output:
+                status = result.RESULT_UNKNOWN
+
+            elif "INTERNAL-ERROR" in result_str:
+                status = "OUT-OF-MEMORY or INTERNAL-ERROR"
+
+            elif "INCONCLUSIVE" in result_str:
+                status = "INCONCLUSIVE"
```
PhilippWendler, line 57: What is the distinction between `INCONCLUSIVE` and `UNKNOWN`? `RESULT_UNKNOWN` is for cases where the tool decided to not produce a result (e.g., because of conceptual limitations). All non-standard results (such as `INCONCLUSIVE`) would be treated by BenchExec as an error result and for example printed in the color for errors or subsumed under errors for the purpose of filtering results. If you do not want this, consider returning `RESULT_UNKNOWN` here.

benchexec/tools/proton.py
```diff
@@ -0,0 +1,65 @@
...
+                status = result.RESULT_TRUE_PROP
+
+            elif "FALSE(termination)" in result_str:
+                status = result.RESULT_FALSE_TERMINATION
+
+            elif "UNKNOWN" in output:
+                status = result.RESULT_UNKNOWN
+
+            elif "INTERNAL-ERROR" in result_str:
+                status = "OUT-OF-MEMORY or INTERNAL-ERROR"
```
PhilippWendler, line 54: Is the `OUT-OF-MEMORY` an oversight here? Also note that such long result strings make the result tables not so convenient to look at. So if you can distinguish between these two cases, I would recommend to do so. Or abbreviate it?

benchexec/tools/proton.py
PhilippWendler: Seems this was overlooked?

benchexec/tools/proton.py
PhilippWendler: Hm, with the current code there is still inconsistency with regards wo `... in result_str`, `result_str == ...` and `... in output`, all three of which have different semantics. Is this intended?

Did you [test this](https://github.com/sosy-lab/benchexec/blob/main/doc/tool-integration.md#testing-the-tool-integration)?

Note that for future maintainers of this logic it might be useful to have example output lines of the tool here, if it is not as trivial as just a line consisting only of the specific string.

rmetta:
Thanks, yet again. We removed the checking of output and committed. Right
now. the only strings output (as the last output line) by proton are:-
TRUE, FALSE(termination), INTERNAL-ERROR and INCONCLUSIVE. We tested proton
bash script at our end, and it seems to work fine.

Best regards,
Ravi

PhilippWendler:
Ok, but your comment seems to indicate that `INCONCLUSIVE` is really used as a kind of "result not known" and the code should return `RESULT_UNKNOWN`, right?

Although, do you think that you will fulfill conditions like `run.exit_code.value == 64 and "Usage error!" in output` in future versions of Proton? If not, please remove them.

Given all this back and forth and the fact that it now does not look at all anymore like CBMC's output: Is checking for the exit code still relevant?

rmetta:
1. We have two checks in Proton: one for non-termination (nt-check) and one
for termination (t-check). If both of the checks terminate within their
respective time slice budget, but both disagree: that is nt-check reports
non-termination and t-check reports termination, then we report
INCONCLUSIVE. Even though our nt-check is sound, our current tooling has
issues (bugs, as well as unsupported complex pointers) and so does our
technique (especially in the case of arrays) , due to which the nt-check
may give a wrong answer. Where as when Proton reports UNKNOWN, it means
that neither t-check, nor nt-check, were successful.  So, this distinction
is important for us.
2. Cleaned up the error code issue mentioned. And, we are now relying only
on strings for outputs.

Committed the new version.

Thanks and best regards,
Ravi

PhilippWendler:
Thanks for the explanation! Then indeed having `INCONCLUSIVE` as a result of category "error" makes sense.

The new code is also much easier to understand.

rmetta:
Thanks for your rigorous and timely review. It was a big help. Highly
appreciated.

Best regards,
Ravi

PR #953, opened 2023-11-02
Add tool info for validator Witch

ayazip:
Add tool info for YAML violation witness validator Witch

### DIFF
--- a/benchexec/tools/witch.py
+++ b/benchexec/tools/witch.py
@@ -1,0 +1,22 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+# SPDX-FileCopyrightText: 2021 Marek Chalupa
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from benchexec.tools.symbiotic import Tool as SymbioticTool
+
+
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Can't you use https://github.com/sosy-lab/benchexec/blob/main/benchexec/tools/symbiotic-witch.py?

ayazip:
@PhilippWendler 
I could, but I am not sure how to then specify that  Witch should use this module. The verifiers have this info in the fm-tools repository, but the validators don't have this. 

I guess that it would it be enough to put "symbiotic-witch" as the tool name in the benchmark definition file for Witch, but I am not sure. Please confirm that this is the case; if it is, I will change it and close this PR.

PhilippWendler:
> @PhilippWendler I could, but I am not sure how to then specify that Witch should use this module. The verifiers have this info in the fm-tools repository, but the validators don't have this.

For questions about the FM-Tools-Repository, please ask there. But AFAIK know a tool listed there can point to its tool-info module with the `benchexec_toolinfo_module` key. For example, CPAchecker does so (and it is both a verifier and a validator).

> I guess that it would it be enough to put "symbiotic-witch" as the tool name in the benchmark definition file for Witch, but I am not sure. Please confirm that this is the case; if it is, I will change it and close this PR.

The `tool` attribute of the `benchmark` tag in the benchmark definition determines which tool-info module BenchExec uses, yes.

ayazip:
Great, thanks!

PR #954, opened 2023-11-02
Add relay-sv, wrapper for analyzing a single file with RELAY

vesalvojdani:
RELAY is a race detection tool optimized for very large programs. And can perform summary based analysis in parallel. The results on the SV-COMP race detection category on my internal runs was -56 points, so I'm not sure it makes sense to have this part of the contest. As it might be a useful tool for race detection of larger programs, I will make my build available.

### DIFF
--- a/benchexec/tools/relay-sv.py
+++ b/benchexec/tools/relay-sv.py
@@ -1,0 +1,39 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for RELAY adapted for BenchExec.
+    URL: https://github.com/vesalvojdani/relay-sv
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("run_relay.sh")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, line_prefix="Version: ")
+
+    def name(self):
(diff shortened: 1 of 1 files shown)


benchexec/tools/relay-sv.py
```diff
@@ -0,0 +1,39 @@
...
+        return tool_locator.find_executable("run_relay.sh")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, line_prefix="Version: ")
+
+    def name(self):
+        return "relay-sv"
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable, *options, *task.input_files]
```
PhilippWendler, line 28: Does the tool support more than one input file? If not, please use `task.single_input_file`.

benchexec/tools/relay-sv.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def determine_result(self, run):
+        status = result.RESULT_ERROR
```
PhilippWendler, line 31: The status variable is not really useful here, why not use `return` everywhere?

benchexec/tools/relay-sv.py
```diff
@@ -0,0 +1,39 @@
...
+        if run.output:
+            if run.output.any_line_contains("Possible race"):
+                status = result.RESULT_UNKNOWN
+            elif run.output.any_line_contains("Fatal error"):
+                status = result.RESULT_ERROR
+            else:
+                status = result.RESULT_TRUE_PROP
```
PhilippWendler, line 38: Does any non-empty output really indicate result `TRUE`? This would for example also result in `TRUE` for crashes, seg faults, etc.

benchexec/tools/relay-sv.py
vesalvojdani: The tool does, but the preprocessing has to be done manually in some complicated ways, so for now, the wrapper will only support a single file. I will fix this and the other comments. I'll do a test run on my machine, and then push it here. Thanks!

PR #955, opened 2023-11-02
Add tool-info module for TracerX-Del and TracerX-WP

ArpitaDutta:
This pull request adds support for the TracerX tool with two different interpolation methods.

### DIFF
--- a/benchexec/tools/tracerx-wp.py
+++ b/benchexec/tools/tracerx-wp.py
@@ -17,7 +17,7 @@ class Tool(benchexec.tools.template.BaseTool2):
     """
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("tracerx", subdir="bin")
+        return tool_locator.find_executable("tracerx-wp", subdir="bin")
 
     def program_files(self, executable):
         return self._program_files_from_executable(
@@ -56,7 +56,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return [executable] + options + list(task.input_files_or_identifier)
 
     def name(self):
-        return "TracerX-Del"
+        return "TracerX-WP"
 
     def determine_result(self, run):
         """


benchexec/tools/tracerx.py
```diff
@@ -56,7 +56,7 @@ def cmdline(self, executable, options, task, rlimits):
         return [executable] + options + list(task.input_files_or_identifier)
 
     def name(self):
-        return "Tracer-X"
+        return "TracerX-Del"
```
PhilippWendler, line 59: Hm, this would mean that if somebody benchmarks a previous version of TracerX, they would get these benchmarks labeled with TracerX-Del from now on. Would that be correct, or would it be confusing/wrong?

Maybe it is better to just have one tool-info module for all variants of TracerX?

ArpitaDutta: Yes, this year we are participating with two different versions of the TracerX System.
1. TracerX with Deletion interpolation
It is our more stable system. In the previous Test-Comp competitions, this system was represented as TracerX but this year we are using TracerX-Del for it. 
2. TracerX with Weakest PreCondition interpolation
This is another choice of TracerX. We represent this system as TracerX-WP


Both systems have their separate binaries.

PhilippWendler: Note that you do not need two tool-info modules in BenchExec in order to participate with different variants of your tool in SV-COMP. (For example, CPAchecker has participated several times in different variants and has always ever used only one tool-info module.) And the submission name for the competition can be chosen independently from the tool name that is returned here. Have a look in the CoVeriTeam columns in [this table](https://sv-comp.sosy-lab.org/2023/results/results-verified/unreach-call.ReachSafety-ControlFlow.table.html#/) for an example of how this could look (the first row can be chosen in the SV-COMP submission, the "Tool" row is what is defined here.

The only thing that requires you to have a different tool-info module is if you have two different executables, but actually even for this you could in principle a fallback like "look for `tracerx` and if it does not exist, look for `tracerx-wp`, if you prefer this.

Just, if you do have two tool-info modules, then please use inheritance to avoid the duplicated code.

ArpitaDutta: Yes, we have two tool-info modules and two separate binaries for the two systems.
We are participating in Test-Comp 2024.

ArpitaDutta: Also, like to add that this statement is also different in both files and it is called the main file for the systems.

In tracerx-wp.py

        return tool_locator.find_executable("tracerx-wp", subdir="bin")

In tracerx.py

        return tool_locator.find_executable("tracerx", subdir="bin")

PhilippWendler: Then please update the PR as requested.

ArpitaDutta:
May I request you accept this merge request as it is as the file modification using inheritance does not seem to be straightforward right now?

PhilippWendler:
To me it looks like this would be pretty straightforward actually, more than for other cases where we already have tool-info modules that use inheritance. Just let the class in `tracerx-wp.py` inherit from the one in `tracerx.py` and overwrite the two methods.

ArpitaDutta:
I am just worrying that the class name in tracerx-wp.py is as same as in  tracerx.py.

If I change the class name in tracerx-wp.py to inherit properties from tracerx.py it may create some side effects in other steps of execution.

PhilippWendler:
No, there is no problem here. Lots of existing tool-info modules do exactly this. Have a look at `graves.py` and `cpachecker.py`, for example.

ArpitaDutta:
I have updated the PR as requested.

PR #958, opened 2023-11-08
Make task copy&paste easier in table-generator HTML by adding left margin

lembergerth:
This MR adds a small margin to the left of table-generator HTML's tasks column.
This makes it easier to select the task name with click+drag, because it adds some more space that is not the clickable link.
The added space is 1ex to make it scale for people that need a large font.

Example (the new margin is the not-highlighted part of the column, left of the highlighting):

Screenshot from 2023-11-08 14-43-59


PhilippWendler:
Good idea, and thanks for the exemplary PR with the screenshot!

PR #967, opened 2023-11-16
Metaval svcomp24

marian-lingsch:
While previously SV-COMP allowed the selection of the backend tool for MetaVal based on the property in the benchmark definition files, this changed with SV-COMP24.
Due to this, the selection of Backend Verifier needs to be moved either into MetaVal or into the benchmark definition module.
Since the BenchExec tool info module contains some necessary preprocessing to determine the backend tool to be executed, using the ToolLocator, it is probably for the best to implement the algorithm selection inside it.

### DIFF
--- a/benchexec/tools/metaval.py
+++ b/benchexec/tools/metaval.py
@@ -77,6 +77,20 @@ class Tool(benchexec.tools.template.BaseTool2):
         ), "we expect that all wrapped tools extend BaseTool2"
         return tool.determine_result(run)
 
+    def _select_verifier_from_property(self, property: str):
+        if "unreach-call" in property:
+            return "cpachecker"
+        elif "valid-memcleanup" in property:
+            return "symbiotic"
+        elif "valid-memsafety" in property:
+            return "symbiotic"
+        elif "no-overflow" in property:
+            return "ultimateautomizer"
+        elif "termination" in property:
+            return "ultimateautomizer"
+        else:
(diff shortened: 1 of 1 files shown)


PhilippWendler:
First, a property is identified by it's file content. I would strongly advise against relying on file names of property files anywhere, they are not defined and could change at any point.

Second, is it a good idea to implement strategy selection in the tool-info module? Consider that this makes it more difficult for everyone to use your tool outside of BenchExec, or even with `runexec` alone. It also makes it difficult to update/improve the tool if one also has to update the tool-info module. Even just trying out MetaVal with a different strategy selection, for example for developing a future version, will be difficult and error-prone. Worse, if the strategy selection is hard-coded here, this basically means that it will not be able to change, because for BenchExec it is important to keep backwards compatibility for transparency and ease of reproducibility. This means that it is fine if for example MetaVal 2.0 uses a different strategy selection that MetaVal 1.0 and thus produces completely different results. It is not acceptable, though, if MetaVal 1.0 with BenchExec 3.19 uses a different strategy selection than MetaVal 1.0 with BenchExec 3.20. The only way out of this solution would be to implement some logic like "if version of MetaVal is 1.0, then apply this strategy selection, else if version of MetaVal is 2.0, then apply this strategy selection, etc." But this would be too tedious, error-prone, and intransparent. So overall, I think such code belongs into the tool and I cannot accept it here.

PR #973, opened 2023-11-22
Fix tool info module proton

dbeyer:
First I implemented the minimal fix for the bug, then I implemented what I think was the intention, and then I simplified.

### DIFF
--- a/benchexec/tools/proton.py
+++ b/benchexec/tools/proton.py
@@ -64,4 +64,4 @@ class Tool(benchexec.tools.template.BaseTool2):
         elif "UNRECOGNIZED PROPERTY" in result_str:
             return "UNSUPPORTED PROPERTY SPECIFIED"
 
-        return "UNKNOWN ERROR"
+        return result.RESULT_ERROR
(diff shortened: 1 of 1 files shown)


benchexec/tools/proton.py
```diff
@@ -38,34 +38,30 @@ def cmdline(self, executable, options, task, rlimits):
...
         elif "INCONCLUSIVE" in result_str:
-            status = "INCONCLUSIVE"
+            return "INCONCLUSIVE"
 
         elif "UNRECOGNIZED PROPERTY" in result_str:
-            status = "UNSUPPORTED PROPERTY SPECIFIED"
+            return "UNSUPPORTED PROPERTY SPECIFIED"
 
-        return status
+        return "UNKNOWN ERROR"
```
PhilippWendler, line 67: This string doesn't seem to provide additional value compared to the standard solution:
```suggestion
        return result.RESULT_ERROR
```

PhilippWendler (commented):
Thanks!

I do like the refactoring, the logic is now much better understandable.

rmetta:
> @rmetta Do you agree with this?

Thank you so much, Dirk and and and Philipp. I agree with the fix.

PR #978, opened 2023-11-24
Add way to output reason for status to LIV

MartinSpiessl:
This allows the tool to output a status like
`false(unreach-call)` instead of just `false`,
or `unknown(<reason>)` to specify a reason for why it was `unknown`.

### DIFF
--- a/benchexec/tools/liv.py
+++ b/benchexec/tools/liv.py
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
+import re
 
 import benchexec.result
 from benchexec.tools.sv_benchmarks_util import ILP32, LP64, get_data_model_from_task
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Note that BenchExec will currently treat `unknown(...)` as a result of category ERROR, not of category UNKNOWN. That means you get the category color magenta and it will be classified under ERROR for filters etc. Is that what you want or do you want to keep `unknown` and just extend the status for `false`? For `true` the situation is the same, only `RESULT_TRUE_PROP` counts, and here it is more important, of course.

MartinSpiessl:
Unfortunately `false(foo)` would still count as a (potentially) wrongly rejected witness in the validator track, so extending the `false` case does not work. 

> For true the situation is the same, only RESULT_TRUE_PROP counts, and here it is more important, of course.

Currently not planning to output anything in brackets behind `true`, though this tool-info rewrite would allow the tool to do this.

My use-case is that I want to show that the result is `unknown`, but actually we would have rejected the witness if we assume a stricter witness semantics, hence the status is `unknown(false)` currently. In result tables I have seen things like `TIMEOUT(true)` or so, I would like a similar mechanism for adding such a "hint" to `unknown` by the tool itself.

I guess I could also add a custom column to the tables to indicate this, but these are not available e.g. in the SV-COMP result tables iirc.

PhilippWendler:
> Unfortunately `false(foo)` would still count as a (potentially) wrongly rejected witness in the validator track, so extending the `false` case does not work.

Yes, if you want "unknown" and not "false" you need "unknown". My sentence was misleading, sorry. I did not want to suggest to you to return "false" instead. I wanted to suggest to only apply the "()" thing in the "false" case in the tool-info module, and not for "unknown".

> > For true the situation is the same, only RESULT_TRUE_PROP counts, and here it is more important, of course.
> 
> Currently not planning to output anything in brackets behind `true`, though this tool-info rewrite would allow the tool to do this.
> 
> My use-case is that I want to show that the result is `unknown`, but actually we would have rejected the witness if we assume a stricter witness semantics, hence the status is `unknown(false)` currently. In result tables I have seen things like `TIMEOUT(true)` or so, I would like a similar mechanism for adding such a "hint" to `unknown` by the tool itself.

I understand, and it makes sense. I just want to inform you that you can return "unknown(false)" but it would be colored like an ERROR and not like an UNKNOWN, and it would not be shown if someone selects category "unknown" in the drop-down filter.

It does make sense to extend BenchExec's category matching to handle "unknown()" like "unknown", but I don't want to rush this this morning.

So it is up to you if you prefer "unknown" with category "unknown" or "unknown(false)" with category "error" for now.

MartinSpiessl:
> It does make sense to extend BenchExec's category matching to handle "unknown()" like "unknown", but I don't want to rush this this morning.

Yeah sure, did not expect you to do this. For me the `unknown(false)` works, even if it is colored as error, this is not a concern currently as long as it is not interpreted as `false` or `true` then I am happy

PR #979, opened 2023-11-24
Fix owic tool-info module

filipeom:
Fixes owic's tool-info module https://gitlab.com/sosy-lab/test-comp/test-comp-2024/-/issues/14#note_1665295781


dbeyer (approved):
Look good and I tested it. It works now.

PR #980, opened 2023-11-26
Update concurrentwitness2test.py with better result determination

leventeBajczi:
This PR adds support for a wider set of statuses as the output of the benchmarks, while preserving backwards compatibility with older logs. 

Instead of separate if-branches to check for statuses beginning with `Verdict: `, we combine them into one branch.


dbeyer:
@PhilippWendler Is it possible to make a new release 3.19.1 after this merge?

PR #981, opened 2023-12-09
Fix an 'IndexError: list index out of range' in tool-info module for WitnessLint

dbeyer:


PhilippWendler:
I don't think `EXCEPTION` is a good status for the case `if len(run.output) == 0` (which btw. is equivalent to `if run.output`), but neither do I think that it is a good status for the existing `if "witnesslint finished" not in run.output[-1]`. Some generic `ERROR` seems to be better than mixing these cases into `EXCEPTION`. @marian-lingsch Can you please improve this?

marian-lingsch:
> @marian-lingsch Can you please improve this?

Tried to do it. See https://github.com/sosy-lab/benchexec/pull/982
Feedback is very welcome!

PR #982, opened 2023-12-11
Refactor witness output interpretation

marian-lingsch:
In order to make it easier to understand the output of the witness linter we sepparate the exception case from the case where there was no output


PhilippWendler:
Does `"witnesslint finished" not in run.output[-1]` really imply an `EXCEPTION` result?

marian-lingsch:
Thank you very much for your comments. I tried to address both concerns with the last two commits by returning an Error per default and splitting the error that witnesslint did not finish from the case where we know there has been an exception.
Feedback is always very welcome!

PhilippWendler (approved):

PR #987, opened 2024-02-01
Added Coccinelle and Vfit to tools

justfeelix:
Coccinelle repo: https://github.com/coccinelle/coccinelle?search=1

### DIFF
--- a/benchexec/tools/coccinelle.py
+++ b/benchexec/tools/coccinelle.py
@@ -15,7 +15,6 @@ from benchexec.tools.template import ToolNotFoundException, BaseTool2
 class Tool(BaseTool2):
     """
     Tool info for Coccinelle
-    (https://github.com/coccinelle/coccinelle).
     """
 
     REQUIRED_PATHS = ["standard.h" "ocaml" "standard.iso" "ocaml"]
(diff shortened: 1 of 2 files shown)


benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+import subprocess
+
+import benchexec.result as result
+from benchexec.tools.template import ToolNotFoundException, BaseTool2
+
+
+class Tool(BaseTool2):
+    """
+    Tool info for Coccinelle
+    (https://github.com/coccinelle/coccinelle).
```
PhilippWendler, line 18: Can you please adapt this to our more recent style of having the URL in a `project_url()` method? (Same as for almost all existing tool-info modules.)

benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+        dir_name = os.path.dirname(exe)
+        logging.debug("Looking in %s for spatch", dir_name)
+        for _, dir_names, file_names in os.walk(dir_name):
+            if "spatch" in file_names:
+                return exe
+            break
+        msg = (
+            f"ERROR: Did find a spatch in {os.path.dirname(exe)} "
+            f"but no 'spatch' directory besides it"
+        )
```
PhilippWendler, line 37: Can you please explain a little bit about what this is supposed to do and why?

How does the existence of any file with name `spatch` somewhere deep inside the directory hierarchy next to the executable have an effect on running Coccinelle?

Why does the code look for files, but the error message is about a directory?

Furthermore, I think this check is completely redundant, because it will never fail, because there is always a file named `spatch` - the executable file itself that is searched above.

benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+        raise ToolNotFoundException(msg)
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def cmdline(self, executable, options, task, resource_limits):
+        return [executable] + options + list(task.input_files) + [task.property_file]
```
PhilippWendler, line 44: Does this tool support more than one input file? If not, please use `task.single_input_file` in order to produce an appropriate error message.

What is used as a property file here?

Note that the current code will lead to a crash if no property file is given, because then `task.property_file` will be `None`.

benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def program_files(self, executable):
+        paths = self.REQUIRED_PATHS
+        return [executable] + self._program_files_from_executable(executable, paths)
```
PhilippWendler, line 48: There seems to be no reason to have this method, the default implementation does the same.

benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def determine_result(self, run):
+        if run.exit_code != 0:
```
PhilippWendler, line 51: This condition is always true, because `run.exit_code` is an instance of `ProcessExitCode`, not a number.

Note that you properly do not need this method at all. The default always returns `RESULT_DONE`, and then BenchExec will overwrite this with `ERROR ($exitcode)` if the exit code is not zero, which seems to be what you want.

benchexec/tools/coccinelle.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @staticmethod
+    def _version_from_tool(executable):
+        version_output = subprocess.check_output(
+            [executable, "--version"], universal_newlines=True
+        )
+        for line in version_output.splitlines():
+            if line.startswith("spatch version"):
+                return line.split("spatch version")[1].strip()
+        return ""
```
PhilippWendler, line 64: It does not make sense to overwrite this method. If you want to change the code for determining the version, overwrite `version()`.

I think you can even use the standard implementation of `_version_from_tool` with `line_prefix` parameter. If you really need to call the tool yourself, please add proper error handling.

PhilippWendler (changes requested):
All comments apply to both tool-info modules, of course.

benchexec/tools/coccinelle.py
justfeelix: Thanks for the informative comments!
So the reason I am using `task.property_file` is because Coccinelle needs a .cocci template file in order to inject a task with a fault. And I am using the tool with Coveriteam and pass the template as a property file. 
I have modified the code now to consider a case of `None` in `task.property_file`.

justfeelix:
Thanks for the informative reply. I have adjusted the tools to adapt to the format of the rest of the tools.
Let me know if something else needs to be modified to conform with the rest of the tools.

PhilippWendler (commented):
Just very minor things left.

benchexec/tools/coccinelle.py
PhilippWendler: Thanks.

I am not sure whether the .cocci template really fits as a property file or whether it would be easier to understand if you just pass it as an input parameter of the tool (in `options`), but for the current state of the tool-info module it doesn't matter, so I am fine with it.

PhilippWendler (approved):

Issue #875, opened 2022-11-07
New process architecture for benchexec in container mode


PhilippWendler:
Currently, the architecture of `benchexec` in the `localexecution.py` module is such that it starts a number of worker threads depending on `--numOfThreads` and each of the worker threads repeatedly executes runs using `RunExecutor`. In container mode, execution of a run involves calling `clone()` to create a subprocess.

This architecture has several problems:
- Glibc's `clone()` is not safely usable in processes with more than one thread and can produce deadlocks (#656). We currently have a workaround for this, but it is not a full solution.
- Since Python 3.8, the API documentation [requests that subprocesses are only created from the main thread](https://docs.python.org/3/c-api/init.html#fork-and-threads), not from worker threads as we do. We have not encountered problems related to this so far, but it could happen.
- With a high (3-digit) number of threads we have not seen the expected throughput. While no profiling has been performed yet it is plausible that this is caused by the fact that all the pre- and postprocessing of runs (e.g., log analysis, writing results) is performed in Python threads, of which only one can be active at the same time due to the GIL.

(Note that all these problems do not affect users of `runexec` / `containerexec`, where the run execution is started from a single-threaded process.)

So in the long term it would probably be good to change this architecture. There at least two potential solutions:
- Switch from worker threads to worker processes just like the multiprocessing module. Each worker process would be single threaded.
- Have one designated (single-threaded) subprocess that is created in the beginning and whose sole responsibility is to spawn all further subprocesses on request. (Android uses this and calls it the Zygote process.)

Instead of `clone()` one can also use `unshare()` and `os.fork()` for creating a container, which should be safer, but due to the way how `unshare()` works with PID namespaces this would involve yet another process per run and probably complicate process handling even more than any of the other alternatives.

Things to consider:
- Whether and how this affects and works for cases where `benchexec` is not called as a command-line tool, but executed as part of a larger Python program (that may have created threads before `benchexec` is even loaded).
- The subprocess that we start for each run needs to be cloned from a process that already has all the required modules loaded, because this process is inside the container for the run and might not have access to the Python interpreter's files on disk.
- How communication is possible with the process that hosts the tool-info module if more than one worker process needs to communicate, or if each worker process should also get its own separate process with an instance of the tool-info module.
[...]

PR #990, opened 2024-02-12
Add tool_pid argument to parent_setup_fn

charmoniumQ:
The container implementation is easy. Just pass the pid we already got to `parent_setup_fn`. They already wait for the parent to signal readiness.

However, the non-conainer implementation does not already have this wait, so I had to add it in. Like the container case, the child inherits the read-end of a pipe through which the parent sends a byte when they are ready.

There are two ways I thought of to implement the wait:
1. Wait in `preexec_fn`. The naive approach creates a deadlock while executing the `subprocess.Popen(...)`. That is not obvious to me, since the [`preexec_fn` gets executed](https://github.com/python/cpython/blob/a82fbc13d0e352b9af7d7ffbef4bc04cf635f07f/Modules/_posixsubprocess.c#L793) *after* the [`fork`](https://github.com/python/cpython/blob/a82fbc13d0e352b9af7d7ffbef4bc04cf635f07f/Modules/_posixsubprocess.c#L923). It may have something to do with the GIL?
2. Run a "wrapper" program that waits and then exec's the tool. Replace `subprocess.run(cmd)` with `subprocess.run(["python", "-c", "do_wait; os.execv(cmd)"]):

I used the second strategy because the first one was deadlocked. However, this creates a few test failures, both relate to the non-container case, where we replace a direct 

1. Something is wrong with `test_input_is_redirected_from_stdin` probably because the semantics of how file-descriptors are inherited is changed slightly..
4. It doesn't detect when the command doesn't exist in quite the same way. `subprocess.run(...)` generates with an exception in the parent process. `subprocess.run(["python", "-c", "os.execv(...)"])` generates an exception, but not in the parent process.

I think I am going try debuggin the deadlock in `preexec_fn` instead.


PhilippWendler:
Using a subprocess here is not possible. It adds complexity and a lot of additional possible failure modes and things that can go wrong. And the overhead of this subprocess would be included in the measurement, but BenchExec takes great care to ensure that only a minimal amount of setup and teardown code is included in the measurement.

Doing anything sophisticated inside `preexec_fn` is also difficult. The documentation already says that one should not use it if the application is multi-threaded, although we are currently having no problem with that. But if you start putting more code into it, especially code that is related to threading, such as waiting, I would not be surprised if it breaks.

So our goal is to avoid additional risks of failure and complexity in this area of BenchExec as far as possible. Right now this PR seems to add something that would be unused inside BenchExec itself and only be of use for potential external extensions of BenchExec, is this correct? In this case we would need a very strong argument for accepting the risk and maintenance effort related to this.

charmoniumQ:
What about making the feature "PID gets passed `parent_setup_fn`" only available in container-execution mode, where the implementation is much simpler (the main branch already knows the PID of the grandchild at that point)?

The argument for implementation is that eBPF code can likely not be invoked inside the container, but if we have the PID, we can easily invoke outside of the container.

PhilippWendler:
If we can add this with negligible risk and maintenance effort, which seems to be the case according to your description, then yes, I would be ok with that.

Medium to long term a lot of this might change due to #875 and because I want to switch from PID tracking to PID FDs as soon as we can ignore kernels that are too old for PID FDs. So I won't promise to keep this forever, but certainly as long as it doesn't hinder us.

charmoniumQ:
The new commit only implements passing PID in `ContainerExecutor._start_execution_in_container` which is minimally invasive. I update the docstrings to refer to the new feature.

benchexec/containerexecutor.py
```diff
@@ -647,6 +647,9 @@ def grandchild():
                 # Signal readiness to parent by sending our PID
                 # and wait until parent is also ready
                 os.write(to_parent, str(my_outer_pid).encode())
+                logging.debug(
```
PhilippWendler, line 650: While these log statements may be useful during development of this PR, I don't think they are that useful further on, and especially I think they clutter the output too much (something like `runexec --debug` is also commonly used to debug problems with the system setup, so it is good if there is not too much of the purely internal and developer-only debug messages).

I would suggest to remove all of the added log messages.

benchexec/baseexecutor.py
```diff
@@ -75,8 +92,8 @@ def _start_execution(
         parent_cleanup_fn,
     ):
         """Actually start the tool and the measurements.
-        @param parent_setup_fn a function without parameters that is called in the parent process
-            immediately before the tool is started
+        @param parent_setup_fn a function that is called in the parent process
+            immediately before the tool is started. The keyword-arguments passed may differ in subclasses.
```
PhilippWendler, line 96: I think beyond that we need documentation what parameters may actually be passed, either here or in the respective classes.

benchexec/baseexecutor.py
```diff
@@ -49,7 +49,24 @@ def handle_basic_executor_options(options, parser):
...
+        """
+        self.child_setup_fn = (
+            child_setup_fn if child_setup_fn is None else util.dummy_fn
+        )
+        self.parent_setup_fn = (
+            parent_setup_fn if parent_setup_fn is None else util.dummy_fn
+        )
+        self.parent_cleanup_fn = (
+            parent_cleanup_fn if parent_cleanup_fn is None else util.dummy_fn
+        )
```
PhilippWendler, line 69: Why this change?

benchexec/containerexecutor.py
```diff
@@ -974,11 +977,18 @@ def check_child_exit_code():
             if use_cgroup_ns:
                 cgroups = cgroups.create_fresh_child_cgroup_for_delegation()
 
+            # Do parent setup
+            logging.debug("Calling parent_setup_fn)")
+            parent_setup = parent_setup_fn(
+                child_pid=child_pid,
+                grandchild_pid=grandchild_pid,
+            )
```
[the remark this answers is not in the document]
charmoniumQ: Got it; thanks

benchexec/containerexecutor.py
charmoniumQ: Removed

benchexec/baseexecutor.py
charmoniumQ: I wasn't sure where to pass in these functions. I opted to pass them into `RunExecutor.execute_run` instead.

charmoniumQ:
It's actually not too hard to have the child let the parent know its pid in their setup function. The hard part (which I am not going to do) is having the child wait for the parent's setup function to be done before continuing.

Having the parent get the child's PID makes the API consistent. Clients won't have to change their `parent_setup_fn` when they switch on/off `use_namespaces`.

benchexec/runexecutor.py
```diff
@@ -745,6 +749,13 @@ def execute_run(
             if files_size_limit < 0:
                 sys.exit(f"Invalid files-size limit {files_size_limit}.")
 
+        self.parent_setup_fn = (
+            util.dummy_fn if parent_setup_fn is None else parent_setup_fn
+        )
+        self.parent_cleanup_fn = (
+            util.dummy_fn if parent_cleanup_fn is None else parent_cleanup_fn
+        )
```
PhilippWendler, line 757: Why store this in attributes? AFAIS the only use is in `_execute`, which is directly called from this method.

benchexec/containerexecutor.py
```diff
@@ -976,7 +976,10 @@ def check_child_exit_code():
 
             # start measurements
             cgroups.add_task(grandchild_pid)
-            parent_setup = parent_setup_fn()
+            parent_setup = parent_setup_fn(
+                child_pid=child_pid,
+                grandchild_pid=grandchild_pid,
```
PhilippWendler, line 981: Do we need both? `child_pid` and the fact that there are two processes are really implementation details, and we shouldn't expose them in a public API. Would it be enough if we simply pass the PID of the benchmarked process?

benchexec/runexecutor.py
```diff
@@ -869,6 +881,8 @@ def postParent(preParent_result, exit_code, base_path):
             if exit_code.value not in [0, 1]:
                 _get_debug_output_after_crash(output_filename, base_path)
 
+            self.parent_cleanup_fn(parent_setup, exit_code, base_path)
```
PhilippWendler, line 884: This does not conform to the documented contract of `parent_cleanup_fn`. Also, passing `exit_code` seems fine, but I would hesitate to pass `base_path`.

benchexec/runexecutor.py
```diff
@@ -648,6 +650,8 @@ def execute_run(
         @param files_size_limit: None or maximum size of files that may be written.
         @param error_filename: the file where the error output should be written to (default: same as output_filename)
         @param write_headers: Write informational headers to the output and the error file if separate (default: True)
+        @param parent_setup_fn: A function that gets called in the parent process before the tool is allowed to run (default: noop fn). When used with use_namespaces=True, the function recieves child_pid and grandchild_pid.
+        @param parent_cleanup_fn: A function that gets called after the tool finishes running with the return value of parent_setup_fn.
```
PhilippWendler, line 654: We should think about how to make it easy to ensure compatibility in the future here. For example, what can we do such that we can add more information in the future?

benchexec/runexecutor.py
```diff
@@ -829,19 +840,20 @@ def _execute(
                     # Disable energy measurements because we use only parts of a CPU
                     packages = None
 
-        def preParent():
+        def preParent(**kwargs):
```
PhilippWendler, line 843: I think we should use proper parameters here. This is called only internally, so we do not need to pass through an unknown number of parameters.

benchexec/baseexecutor.py
```diff
@@ -88,14 +88,21 @@ def _start_execution(
...
+
             # Do some other setup the caller wants.
             child_setup_fn()
 
             # put us into the cgroup(s)
             pid = os.getpid()
             cgroups.add_task(pid)
 
+            os.write(to_parent, str(pid).encode())
+            os.close(to_parent)
```
PhilippWendler, line 104: If we decide to keep this, it would need to move above the `cgroups.add_task(pid)` call, because that starts the measurement.

But considering the different semantics to the container case, and the fact that we do not have a use case for it, I would tend to omit it.

PhilippWendler:
@charmoniumQ Are you still interested in continuing with this?

charmoniumQ:
No, I don't have the time any more.

Issue #986, opened 2024-01-29
Remove support for Python 3.8 and 3.9


PhilippWendler:
As discussed [previously](https://github.com/sosy-lab/benchexec/issues/717) we will remove support for Python 3.7 in one of the next releases. I suspect this will not be critical because all currently supported versions of Ubuntu or Debian ship Python 3.8 or newer. We should still wait for at least one more release and announce the upcoming change in the changelog.

---

As a heads-up, we plan to **remove support for Python 3.8** once Ubuntu 20.04 goes out of support, i.e., in **April 2025**. At that point **we might even drop support for Python 3.9** and jump directly to requiring Python 3.10.

Please comment here if either of these plans are problematic for you.

YashGaur000:
I think let it be as it is for now. Benchexe should remove support for Python 3.8 once Ubuntu 20.04 goes out of support, in 2025. And jumping directly to 3.10 version of python will be a great decision.

PhilippWendler:
BenchExec 3.22 was just released and will be the last version supporting Python 3.7.

PhilippWendler:
For the record: Python 3.10 would finally allow us to implement https://github.com/sosy-lab/benchexec/issues/433#issuecomment-654926492, so the current plan is to indeed start requiring Python 3.10 for BenchExec in 2025.

PhilippWendler:
For anyone subscribed to this issue: We plan to go ahead with this and release one more version of BenchExec with support for Python 3.8 and 3.9 in the next week(s). After that we will raise the required Python version to 3.10, which means that future BenchExec versions will not run anymore for example on Debian Bullseye and Ubuntu 20.04.

We currently have no plans to remove support for Python 3.10 before May 2027.

PR #995, opened 2024-02-18
Adding SLURM-compatibility to Benchexec

leventeBajczi:
I've added experimental support for the SLURM workload manager for HPC as a new `contrib/` entry. I have no idea whether this contribution is interesting for other users of Benchexec, but for Hungarian researchers, we can request access to Komondor (https://hpc.kifu.hu/hu/komondor) for research projects, and it uses SLURM as a way of scheduling jobs across its nodes. 

As an overview of what I've implemented (I based these modifications on the AWS integration):

1. A new benchmarking entrypoint `contrib/slurm-benchmark.py`, which enables SLURM-compatibility via the flag `--slurm`
2. A new executor `contrib/slurm/slurmexecutor.py`, which takes a benchmark and runs it using `srun`, possibly in a `singularity` container (for dependency management). 

Currently, most of the stuff works (besides keeping resulting files except for the log, but that is not that hard to implement from here), but I feel like it is very easy for the integration to break. I'm wondering whether it is worth putting in some extra effort to make this more production-ready. I'm eagerly waiting for some feedback on this; I'd be happy to work on it further if there is interest.


PhilippWendler:
I am pretty sure that there are also other people who would like to use it, so this is welcome!

I know about Slurm, but do not have an installation here, so I can't test or maintain this. But we have the `contrib/` directory precisely for sharing such additional components outside of the supported Benchxec, and I would like to merge this here. Ideally you would be willing to somewhat maintain in the future?

The general architecture of this PR matches what we do in similar cases and is certainly what I would have recommended.

So besides what you think should be added, I would focus mostly on some documentation.

Can you explain a little bit more on the effect of using Slurm for users who are familar with BenchExec? I assume, for example, that stuff like core allocations and measurements are handled purely by Slurm, so users wouldn't get what they know from BenchExec, but instead what Slurm provides?

Are there any particular requirements beyond a standard Slurm installation or do users need to do something specific?

Ideally these questions together with an overview of supported/unsupported features of BenchExec would be discussed in a readme.

PhilippWendler:
I don't have any particular testing suggestions. If you want to test more tools, you could relatively easily download the SV-COMP participants and try them? All data required for this is [online](https://gitlab.com/sosy-lab/benchmarking/fm-tools), but I wouldn't require this as condition for merging this PR.

leventeBajczi:
I've added some documentation, and also update this PR's status from draft to Open. Is there anything else you'd like me to do before we merge? If there are some problems after we merge and people start using this, I'll be happy to help sort these out -- but I feel like it is good enough for now (it is good enough for us).

leventeBajczi:
I've added myself to the list of contributors in the main README, feel free to revert that commit (26d1933) if that is not applicable to such "external" contributions. Also, I added contact info to the top of the new README in the slurm directory, so that people can tag/contact me if they encounter issues.

README.md
```diff
@@ -122,6 +122,7 @@ Contributors:
 - [Thomas Stieglmaier](https://stieglmaier.me/)
 - [Martin Yankov](https://github.com/marto97)
 - [Ilja Zakharov](https://github.com/IljaZakharov)
+- [Levente Bajczi](https://github.com/leventeBajczi)
```
PhilippWendler, line 125: You definitively should be listed here! I would just ask you to insert the name in the alphabetical ordering.

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+SPDX-FileCopyrightText: Critical Systems Research Group
+SPDX-FileCopyrightText: Budapest University of Technology and Economics <https://www.ftsrg.mit.bme.hu>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+# BenchExec Extension for Benchmarking via SLURM
+
+This Python script extends BenchExec, a benchmarking framework, to facilitate benchmarking via SLURM, optionally using a Singularity container.
+
+In case of problems, contact / tag in an issue: [Levente Bajczi](https://github.com/leventeBajczi)
```
PhilippWendler, line 16:
```suggestion
In case of problems, please tag in an [issue](https://github.com/sosy-lab/benchexec/issues/new/choose): [Levente Bajczi](https://github.com/leventeBajczi) (@leventeBajczi).
```
Adding the GitHub user name makes it easier to copy and paste. And I would recommend to avoid the term "contact", otherwise people will just send you mails.

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Preliminaries
+
+* *SLURM* is an open-source job scheduling and workload management system used primarily in high-performance computing (HPC) environments.
+* *Singularity* is a containerization platform designed for scientific and high-performance computing (HPC) workloads, providing users with a reproducible and portable environment for running applications and workflows.
```
PhilippWendler, line 21: You already nicely linked SLURM on the top-level readme, maybe do that as well here (and same for Singularity)?

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Requirements
+
+* Singularity (optional), tested with `singularity-ce version 4.0.1`
+* SLURM, tested with `slurm 22.05.7` on `Red Hat Enterprise Linux 8.6 (Ootpa)`, kernel version `4.18.0-372.9.1.el8.x86_64`
```
PhilippWendler, line 26: Do you know anything about likely compatibility? "tested with" is good, but if we can write something like "should work with version 4.0 or newer" in addition, it would be even better..

The versions of RHEL and kernel are likely not so relevant? I would omit them.

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+   ```
+   Options:
+   - `--slurm`: Use SLURM to execute benchmarks. Will revert to regular (local) benchexec if not given. 
+   - `--singularity <path_to_sif>`: Specify the path to the Singularity .sif file to use. See usage later.
+   - `--scratchdir <path>`: Specify the directory for temporary files. The script will use this parameter to create temporary directories for file storage per-run, which get discarded later. By default, this is the CWD, which might result in temporary files being generated by the thousands in the working directory. On some systems, this must be on the same mount, or even under the same hierarchy as the current directory. Must exist, be writable, and be a directory.
+   - `-N <N>`: Specify the factor of parallelism, i.e., how many instances to start at a time. Tested with up to `1000`, probably works with much higher values as well.
+    
+## Overview of the Workflow
+
+The workflow is based on `localexecution.py`, and uses the same general layout. However, instead of delegating to `runexec`, it delegates to `srun`.
```
PhilippWendler, line 41:
```suggestion
This works similarly to BenchExec, however, instead of delegating each run to `runexec`, it delegates to `srun` from SLURM.
```
This is more focused towards users than developers.

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+    
+    Notice the `fuse-overlayfs` package. That is mandatory for the overlay filesystem to work properly.
+    
+    The script parameterizes `singularity exec` with the following params:
+    * `-B $PWD:/lower`: Bind the working directory to `/lower` (could be read-only)
+    * `--no-home`: Do not bind the home directory
+    * `-B {tempdir}:/overlay`: Bind the temporary directory to `/overlay` (must be writeable)
+    * `--fusemount  "container:fuse-overlayfs -o lowerdir=/lower -o upperdir=/overlay/upper -o workdir=/overlay/work $HOME"`: mount an overlay filesystem at $HOME, where modifications go in the temp dir but files can be read from the current dir
+    
+    Modifications to the script are necessary if the user on the host and the container differ (due to using $HOME). Also, files not under the current directory will not be visible.
```
PhilippWendler, line 64: Do you mean that users need to change `slurm-benchmark.py` or `slurmexecutor.py`? Could we improve that? Otherwise just declare it as a limitation?

contrib/slurm/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+2. Currently, the following parameters are passed to `srun` (calculated from the benchmark's parameters):
+   * `-t <hh:mm:ss>` CPU timelimit (generally, SLURM will round up to nearest minute)
+   * `-c <cpus>` number of cpus 
+   * `--threads-per-core=1` only use one thread per core
```
PhilippWendler, line 69: BenchExec typically does not enforce this, only if you pass `--no-hyperthreading`. Can we have the same behavior here? The more that `slurm-benchmark.py` is a drop-in replacement for `benchexec` with the same behavior, the better.

contrib/slurm-benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+# Add ./benchmark/tools to __path__ of benchexec.tools package
+# such that additional tool-wrapper modules can be placed in this directory.
+benchexec.tools.__path__ = [
+    os.path.join(os.path.dirname(__file__), "benchmark", "tools")
+] + benchexec.tools.__path__
+
+
+class Benchmark(benchexec.benchexec.BenchExec):
+    """
+    An extension of BenchExec for use with CPAchecker
```
PhilippWendler, line 32: I guess you want to remove "CPAchecker" here. :-)

contrib/slurm-benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+            dest="singularity",
+            type=str,
+            help="The path to the singularity .sif file to use. Will bind $PWD to $HOME when run.",
+        )
+        slurm_args.add_argument(
+            "--scratchdir",
+            dest="scratchdir",
+            type=str,
+            default="./",
+            help="The path to the singularity .sif file to use. Will bind $PWD to $HOME when run.",
```
PhilippWendler, line 57: The help text here got duplicated from the one above.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def get_system_info():
+    return None
+
+
+def execute_benchmark(benchmark, output_handler):
+    num_of_cores = benchmark.rlimits.cpu_cores
+    mem_limit = benchmark.rlimits.memory
+    run_sets_executed = 0
```
PhilippWendler, line 45: This seems unused.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # wait until workers are finished (all tasks done or STOPPED_BY_INTERRUPT)
+    for worker in WORKER_THREADS:
+        worker.join()
+    assert unfinished_runs == 0 or STOPPED_BY_INTERRUPT
+
+    # get times after runSet
+    walltime_after = time.monotonic()
+    usedWallTime = walltime_after - walltime_before
+    usedCpuTime = 1000  # TODO
```
PhilippWendler, line 113: Please remove the dummy value. It is better to not provide a value instead of a wrong one.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+
+class _Worker(threading.Thread):
+    """
+    A Worker is a deamonic thread, that takes jobs from the working_queue and runs them.
+    """
+
+    working_queue = queue.Queue()
+
+    def __init__(
+        self, benchmark, my_cpus, my_memory_nodes, output_handler, run_finished_callback
```
PhilippWendler, line 137: `my_cpus` is a misleading name, because it seems like it is a list of CPU cores. Please rename. And please remove unused parameters (`my_memory_nodes`).

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+        srun_command = (
+            f"srun "
+            f"-t {srun_timelimit} "
+            f"-c {cpus} "
+            f"-o {log_file} "
+            f"--mem-per-cpu {mem_per_cpu} "
+            f"--threads-per-core=1 "
+            f"--ntasks=1 "
+            f"{singularity_command}"
```
PhilippWendler, line 248: Here are some problems with (missing) quoting/escaping. Any whitespace for example in the name of the scratch dir makes it fail, and whitespace in tool options will also not work as desired. 

We must take care that everything that is passed as a string to other commands is properly treated by escaping or quoting.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+        jobid_command = (
+            f"{srun_command} 2>&1 | grep -o 'job [0-9]* queued' | grep -o '[0-9]*'"
+        )
+        seff_command = f"seff $({jobid_command})"
+        logging.debug("Command to run: %s", seff_command)
+        result = subprocess.run(
+            ["bash", "-c", seff_command], shell=False, stdout=subprocess.PIPE
```
PhilippWendler, line 256: What about `stderr`? Do you want `capture_output=True`?

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+    with open(log_file, "r+") as file:
+        content = file.read()
+        file.seek(0, 0)
+        empty_lines = "\n" * 6
+        file.write(empty_lines + content)
+
+    exit_code, cpu_time, wall_time, memory_usage = parse_seff(str(result.stdout))
+
+    return {
+        "starttime": benchexec.util.read_local_time(),
```
PhilippWendler, line 269: This would produce the current (i.e., end of run) time as start time. If we cannot know the actual start time of the run, the we have to omit it.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+        "walltime": wall_time,
+        "cputime": cpu_time,
+        "memory": memory_usage,
+        "exitcode": ProcessExitCode(raw=exit_code, value=exit_code, signal=None),
```
PhilippWendler, line 273: Please use one of the factory methods of `ProcessExitCode`.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+
+def parse_seff(result):
+    exit_code_pattern = re.compile(r"State: COMPLETED \(exit code (\d+)\)")
+    cpu_time_pattern = re.compile(r"CPU Utilized: (\d+):(\d+):(\d+)")
+    wall_time_pattern = re.compile(r"Job Wall-clock time: (\d+):(\d+):(\d+)")
+    memory_pattern = re.compile(r"Memory Utilized: (\d+\.\d+) MB")
```
PhilippWendler, line 281: Here you use compilation of regexps to speed it up, but you do this for every run. Either the compilation is irrelevant because the `re` module caches them internally anyway, or you want to do this in a module constant such that the regexp instances are reused for all runs.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+    exit_code = int(exit_code_match.group(1)) if exit_code_match else None
+    cpu_time = None
+    if cpu_time_match:
+        hours, minutes, seconds = map(int, cpu_time_match.groups())
+        cpu_time = hours * 3600 + minutes * 60 + seconds
+    wall_time = None
+    if wall_time_match:
+        hours, minutes, seconds = map(int, wall_time_match.groups())
+        wall_time = hours * 3600 + minutes * 60 + seconds
+    memory_usage = float(memory_match.group(1)) * 1000000 if memory_match else None
```
PhilippWendler, line 295: Is the result produced by SLURM really in MB? Unfortunately, many tools wrongly report `MiB` values with the `MB` unit label, so it is better to double check what the value really means.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,297 @@
...
+            f"-t {srun_timelimit} "
+            f"-c {cpus} "
+            f"-o {log_file} "
+            f"--mem-per-cpu {mem_per_cpu} "
+            f"--threads-per-core=1 "
+            f"--ntasks=1 "
+            f"{singularity_command}"
+        )
+        jobid_command = (
+            f"{srun_command} 2>&1 | grep -o 'job [0-9]* queued' | grep -o '[0-9]*'"
```
PhilippWendler, line 251: I would prefer if we could get rid of the pipes and the shell around it, because that would make the quoting easier and also be more robust in general. For example, right now you loose the exit code from `seff`.

PhilippWendler (commented):
Looks mostly good! A few minor wording suggestions, and the only major issue is the quoting/escaping of the commands.

README.md
leventeBajczi: Oh, I did not realize it was ordered! Will fix shortly.

contrib/slurm/README.md
leventeBajczi: Sure, I don't know why I did not do that here

contrib/slurm/README.md
leventeBajczi: I don't have a strong enough understanding of SLURM's/singularity's version history to make claims about compatibility, and unfortunately I cannot change the installed versions. I suppose I could add an extra sentence that within the major version there should not be any compatibility problems.

RHEL / kernel versions are indeed not that relevant. Thanks for pointing this out, I'll remove them.

contrib/slurm/slurmexecutor.py
leventeBajczi: I'll remove it then. For some reason I thought it was mandatory to include a starttime in the result - my mistake!

contrib/slurm/slurmexecutor.py
leventeBajczi: Even though I could not find anything in the documentation, SLURM noticably uses 1000-fold increases in its prefixes `k`/`M`/`G` when parsing inputs, and I'd be very surprised if the measurements were in a different format than the resource allocation inputs.

contrib/slurm/slurmexecutor.py
leventeBajczi: Sure, I'll try to rework this then.

contrib/slurm/README.md
leventeBajczi: You're right, I'll move this to limitations.

contrib/slurm/README.md
leventeBajczi: I'm unsure whether that is a good idea, because unfortunately, I cannot query the nodes about the number of threads per core -- we _can_ assume that it's always 2, but that is not a good idea.
The problem with not knowing this number is that I have to request whole cores from SLURM, and conventionally, benchexec requests threads (correct me if I'm mistaken!). Therefore, I have to divide the requested number of threads by the threads-per-core number to get the desired amount of CPU cores.
Another option is to pass `--threads-per-core=2` and `-c <cpus/2>` by default, and `--threads-per-core=1` and `-c <cpus>` when the `--no-hyperthreading` flag is given. However, this may confuse people who work with non-hyperthreading CPUs in their environment, as their jobs will simply refuse to start, as no nodes will match the desired resource allocation.

contrib/slurm-benchmark.py
leventeBajczi: Thanks, this was left in from `aws-benchmark.py`, which I just copied at first :sweat_smile:

contrib/slurm-benchmark.py
leventeBajczi: Fixed.

contrib/slurm/slurmexecutor.py
leventeBajczi: I reworked argument passing a bit because querying resource limits this early on was not a good idea, as we pass `benchmark` along to the worker, and we can query it there instead.
For the `run_sets_executed`: it was indeed not used, I removed it. However, I noticed that it also present in `localexecution.py`, where it is not used, either - maybe we should remove it there, as well?

contrib/slurm/slurmexecutor.py
leventeBajczi: Done, thanks.

contrib/slurm/slurmexecutor.py
leventeBajczi: Done in the rework of params.

contrib/slurm/slurmexecutor.py
leventeBajczi: Right now, benchexec works for me with Python 3.6. `capture_output=True` would break this, so I'd rather avoid it. However, I'll make sure to capture stderr as well, that is indeed weird I haven't done.

contrib/slurm/slurmexecutor.py
leventeBajczi: Fixed, thanks!

contrib/slurm/README.md
PhilippWendler: I understand. Indeed, in BenchExec "cores" means virtual cores (threads in SLURM terms).  And we should not assume that the number of virtual cores per physical core is always 2.

So how about we require users to explicitly pass `--no-hyperthreading` and keep the current `--threads-per-core=1`? If users do not pass `--no-hyperthreading`, we can print an error message telling them to do so and exit. This avoids silently running benchmarks in a different configuration than the users expect, at the cost of a minor inconvenience. But it would also allow us to improve the situation without compatibility problems if we find a solution in the future.

A note in the limitations would also be good.

contrib/slurm/README.md
```diff
@@ -0,0 +1,85 @@
...
+In case of problems, please tag in an [issue](https://github.com/sosy-lab/benchexec/issues/new/choose): [Levente Bajczi](https://github.com/leventeBajczi) (@leventeBajczi).
+
+## Preliminaries
+
+* [SLURM](https://slurm.schedmd.com/documentation.html) is an open-source job scheduling and workload management system used primarily in high-performance computing (HPC) environments.
+* [Singularity](https://docs.sylabs.io/guides/latest/user-guide/) is a containerization platform designed for scientific and high-performance computing (HPC) workloads, providing users with a reproducible and portable environment for running applications and workflows.
+
+## Requirements
+
+* SLURM, tested with `slurm 22.05.7`, should work withink `22.x.x`
```
PhilippWendler, line 25:
```suggestion
* SLURM, tested with `slurm 22.05.7`, should work within `22.x.x`
```

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+exit_code_pattern = re.compile(r"State: COMPLETED \(exit code (\d+)\)")
+cpu_time_pattern = re.compile(r"CPU Utilized: (\d+):(\d+):(\d+)")
+wall_time_pattern = re.compile(r"Job Wall-clock time: (\d+):(\d+):(\d+)")
+memory_pattern = re.compile(r"Memory Utilized: (\d+\.\d+) MB")
+
+
+def parse_seff(result):
+    logging.debug(f"Got {result} as output from seff")
```
PhilippWendler, line 288: Would this potentially be a lot of output? Then better avoid it. It is just a single line, it can be ok, but I would suggest to move the result it to the end of the log message.

And I just noticed the use of f-strings in log messages. That makes the string formatting eager, i.e., also if the log message is not even printed. Using `%` formatting and separate parameters is lazy:
```suggestion
    logging.debug("Seff output: %s", result)
```
Could you please check other log statements as well?

contrib/slurm/README.md
leventeBajczi: Thanks, fixed

contrib/slurm/slurmexecutor.py
leventeBajczi: This was the only dangerous one, all the others had their potentially large (multiline) outputs at the end.

leventeBajczi:
I think I addressed all of your comments, thanks for the thorough review! I ran some tests just now to see if I've broken anything, and everything seems to work.
I've left two of the threads unresolved where I still need some input from you, but otherwise, I think everything seems to be in order.

contrib/slurm/README.md
leventeBajczi: That's a good idea, I'll do this.

contrib/slurm/slurmexecutor.py
```diff
@@ -40,9 +40,10 @@ def get_system_info():
 
 
 def execute_benchmark(benchmark, output_handler):
-    num_of_cores = benchmark.rlimits.cpu_cores
-    mem_limit = benchmark.rlimits.memory
-    run_sets_executed = 0
+
+    assert (
+        not benchmark.config.use_hyperthreading
+    ), "SLURM can only work properly without hyperthreading enabled. See README.md for details."
```
PhilippWendler, line 46: An `assert` will trigger a crash with an ugly stack trace shown to the user. Can you please use `sys.exit`? And mention that they need to pass `--no-hyperthreading`, like `Pass --no-hyperthreading and see README.md for details.`.

leventeBajczi:
I've added some tweaks, ran some tests, and everything seems to be in order. If you agree, please go ahead with the merge. Thanks!

contrib/slurm/slurmexecutor.py
PhilippWendler: Without the pipe and with the added quoting this is already better. But are still shell invocations and for example more special characters (like quotes) will still break stuff.

My goal would be to get rid of the shell completely. Only if we really need it, it may stay, but then we need to be really really careful with quoting or escaping.

Also, when chaining commands with options like in `foo --abc bar --def` one needs to take care that `foo` handles its own options and the ones it should pass through correctly. Do `srun`, `singularity` and `seff` do this automatically, even if for example the final tool gets a command-line argument that happens to have a name collision with one of the wrapper command's parameters? Or is there something that we can do to improve that?
For example, it is a common convention that such wrapper tools accept `--` as a delimiter between own arguments and the rest. Even `runexec` supports that, you can do `runexec ... (runexec parameters) -- tool ... (tool parameters)`. If the involved tools here also support this, it may be good to add this.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,315 @@
...
+            f"-c {cpus} "
+            f"-o {log_file} "
+            f"--mem-per-cpu {mem_per_cpu} "
+            f"--threads-per-core=1 "  # --use_hyperthreading=False is always given here
+            f"--ntasks=1 "
+            f"{singularity_command}"
+        )
+        srun_result = subprocess.run(
+            ["bash", "-c", srun_command],
+            shell=False,
```
PhilippWendler, line 242: `shell=False` is the default, we do not need to pass it (also below).

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,315 @@
...
+WORKER_THREADS = []
+STOPPED_BY_INTERRUPT = False
+
+
+def init(config, benchmark):
+    tool_locator = tooladapter.create_tool_locator(config)
+    benchmark.executable = benchmark.tool.executable(tool_locator)
+    benchmark.tool_version = benchmark.tool.version(benchmark.executable)
+
+    logging.info("Using %s version %s.", benchmark.tool_name, benchmark.tool_version)
```
PhilippWendler, line 35: I would suggest to remove this to keep the output of BenchExec in this mode and standard mode similar. Either we want this always or never, using SLURM doesn't make a difference here.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,315 @@
...
+            self.run_finished_callback()
+            _Worker.working_queue.task_done()
+
+    def execute(self, run):
+        """
+        This function executes the tool with a sourcefile with options.
+        It also calls functions for output before and after the run.
+        """
+        self.output_handler.output_before_run(run)
+        benchmark = self.benchmark
```
PhilippWendler, line 163: You could inline this variable, it is not really useful.

contrib/slurm/slurmexecutor.py
leventeBajczi: i'll try to research this, I frankly don't know how they handle such params.

leventeBajczi:
I've reworked the commands to not use shells at all, please re-check. Hopefully this will be more robust.
Also, I checked, and singularity, similarly to docker and other containerization tools, will handle all params after the name of the container as params to pass to the container. So singularity will work well. I'm unsure about non-singularity runs, but in my testing, it seems to work just fine, and the command receives the params that overlap with that of srun.

contrib/slurm/slurmexecutor.py
```diff
@@ -215,28 +215,41 @@ def run_slurm(benchmark, args, log_file):
...
+                [
+                    "singularity",
+                    "exec",
+                    "-B",
+                    "./:/lower",
+                    "--no-home",
+                    "-B",
+                    f"{tempdir}:/overlay",
+                    "--fusemount",
+                    f"container:fuse-overlayfs -o lowerdir=/lower -o upperdir=/overlay/upper -o workdir=/overlay/work /home/{os.getlogin()}",
```
PhilippWendler, line 229: Just noticed that this hard-codes `/home/$USER` as the home directory, but this is not guaranteed. We could use `pathlib.Path.home()` instead.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,324 @@
...
+    memory = benchmark.rlimits.memory
+
+    srun_timelimit_h = int(timelimit / 3600)
+    srun_timelimit_m = int((timelimit % 3600) / 60)
+    srun_timelimit_s = int(timelimit % 60)
+    srun_timelimit = f"{srun_timelimit_h}:{srun_timelimit_m}:{srun_timelimit_s}"
+
+    mem_per_cpu = int(memory / cpus / 1000000)
+
+    assert benchmark.config.scratchdir and os.path.exists(benchmark.config.scratchdir)
```
PhilippWendler, line 211: Does this also lead to a crash if the scratch directory does not exist or is not specified? If yes, please avoid the crash by adding proper error handling or create the directory.

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,324 @@
...
+
+    with tempfile.TemporaryDirectory(dir=benchmark.config.scratchdir) as tempdir:
+
+        os.makedirs(os.path.join(tempdir, "upper"))
+        os.makedirs(os.path.join(tempdir, "work"))
+
+        singularity_command = (
+            (
+                [
```
PhilippWendler, line 220: I would suggest to get rid of this deep indentation and at the same time construct the command in the intuitive order (left to right), by something like the following:
```python
command = ['srun', ...]
if benchmark.config.singularity:
    singularity_command = [`singularity`, ...]
    command.extend(singularity_command)
command.extend(args)
```

contrib/slurm/slurmexecutor.py
```diff
@@ -0,0 +1,324 @@
...
+            "-c",
+            str(cpus),
+            "-o",
+            str(log_file),
+            "--mem-per-cpu",
+            str(mem_per_cpu),
+            "--threads-per-core=1",  # --use_hyperthreading=False is always given here
+            "--ntasks=1",
+        ] + singularity_command
+        logging.debug("Command to run: %s", " ".join(srun_command))
```
PhilippWendler, line 250: If we use `" ".join(map(util.escape_string_shell, srun_command))` here and import `benchexec.util`, you should be able to copy-and-paste the command from the log and execute it in your terminal, even in case of spaces and other special characters inside parameters.

PhilippWendler (commented):
Thanks! The argument handling should be safe now, which is a great improvement.

Last round of minor stuff, and then only the question about the desired Python compatibility is left.

contrib/slurm/slurmexecutor.py
leventeBajczi: That should be the path inside the container. I _think_ that is safe to assume it as /home/{user}, because we don't have a better idea -- maybe I could add a new cli parameter, but maybe that is not necessary. What do you think?

contrib/slurm/slurmexecutor.py
leventeBajczi: What is more benchexec-like? an error message, or creating the directory?

contrib/slurm/slurmexecutor.py
leventeBajczi: sure, that's a good idea.

contrib/slurm/slurmexecutor.py
PhilippWendler: Ok, I understand. Then let's keep it as is.

contrib/slurm/slurmexecutor.py
PhilippWendler: For the output directory, we just create it. So that could be somewhat more benchexec-like.

contrib/slurm/slurmexecutor.py
leventeBajczi: Good to know this exists! Will do.

contrib/slurm/slurmexecutor.py
leventeBajczi: Okay, I implemented some error handling regarding this.

leventeBajczi:
Regarding python compatibility: Komondor (which I realize is only relevant for me) has `3.6.8`. Because right now benchexec works for me using this python version, I don't want to break this -- I do understand that other parts of Benchexec may already rely on later versions, and also that this could be modified any time, but I don't want to be the one to break it for myself.

contrib/slurm/slurmexecutor.py
leventeBajczi: Komondor (which I realize is only relevant for me) has 3.6.8. Because right now benchexec works for me using this python version, I don't want to break this -- I do understand that other parts of Benchexec may already rely on later versions, and also that this could be modified any time, but I don't want to be the one to break it for myself.

leventeBajczi: (I moved this comment from the PR's root thread to here, because I did not realize you answered here - github is very bad about loading new answers without a page reload today)

PhilippWendler: I am somewhat surprised that they use 7 year old software on a cluster from last year - I hope for you that not the whole OS is that old, because I would expect that a kernel that old does not deliver optimal performance on an AMD Epyc 7763... But that is just a side note.

If you want to merge `stdout` and `stderr` into one stream, then `capture_output` is irrelevant anyway and there is no need to change anything.

But I still wanted to understand what your expectations and hopes are regarding us dropping support for older Python versions, because I don't want to be expected to keep BenchExec working on Python 3.6. If you are fine with it breaking, then we can move forward.

Our plans are also not set in stone, this is why we ask for feedback in #986. Feel free to comment there, in particular if you have a concrete request or suggestion, such as delaying the version bump for a specific time period or so if that would help you, or for looking at other distributions when determining which Python versions may still be relevant. We just don't want to keep support for old Python versions for a long time if nobody uses them anyway.

PhilippWendler (approved):

PhilippWendler:
Thanks again for the submission, and in particular for your responsiveness and fast integration of suggestions!

contrib/slurm/slurmexecutor.py
leventeBajczi: Unfortunately, the software is quite old in some cases - the kernel is 4.18 (from 2018), which is far from optimal.

I'll copy my answer above to the issue you mentioned, but the worst that could happen for me is that I would need a local copy of a newer python version, which I don't think is that hard to achieve - however, I admit I have not yet tried to go down that route. So I don't want to be the one holding back benchexec with my (admittedly very niche) use-case.

leventeBajczi: I've commented on the issue, I think we should discuss further caveats of python versioning there.

PR #1000, opened 2024-02-24
Fixes of SLURM-related bugs, enhancements

leventeBajczi:
After running a few benchmark sets with SLURM, I discovered some issues with the initial way I implemented some details:
* Sometimes, logs are thrown away by SLURM, when providing an output `-o <file>` parameter to `srun`. I could not get a repeatable example of this happening, but in my testing, it happened a few times (<10) out of a 1000 runs. To fix this, I omit the `-o` parameter, and write the log to file using Python instead. 
* SLURM is particularly bad about reporting its own failures. In my testing, sometimes the host ran out of available ports to open, causing a transport failure, which is reported _the same way as if my command failed_. Thus, I had to make sure that all failures are of SLURM only, and that all benchmarking commands always return 0. I save the exitcode in a file, and read it back afterwards.
* Because of faults (caused by, e.g., transport failures) mentioned above, I implemented a retry-pattern in the benchmarking step. By default this is not enabled, but if a user provides the `--retry-killed <N>` flag, the framework retries any failed benchmark run at most `N` times (or infinitely many times, if a negative number is given).    

Also, some minor tweaks, including:
* Better status handling (connected SLURM statuses to benchexec's 'terminationreason', such as `memory` or `cputime`)
* Better logging in debug mode
* Using `--mem` instead of `--mem-per-cpu`, I don't know why I did not do that before


contrib/slurm-benchmark.py
```diff
@@ -56,6 +56,13 @@ def create_argument_parser(self):
             default="./",
             help="The directory where temporary directories can be created for use within singularity.",
         )
+        slurm_args.add_argument(
+            "--retry-killed",
+            dest="retry",
+            type=int,
+            default="0",
+            help="Retry killed jobs this many times. Use a negative number for unbounded retry attempts.",
```
PhilippWendler, line 64: I would recommend just `-1` for unbounded. This is similar to what BenchExec does for other parameters.

contrib/slurm/slurmexecutor.py
```diff
@@ -164,15 +164,27 @@ def execute(self, run):
...
+                )
+                if (
+                    "terminationreason" not in run_result
+                    or not run_result["terminationreason"] == "killed"
+                    or (attempts >= self.benchmark.config.retry >= 0)
+                    or STOPPED_BY_INTERRUPT
+                ):
+                    break
+                attempts += 1
+                time.sleep(1)  # as to not overcrowd a failing scheduler
```
PhilippWendler, line 182: If this is relevant, maybe a global rate limiting across all threads is needed?

contrib/slurm/slurmexecutor.py
```diff
@@ -196,6 +208,26 @@ def execute(self, run):
...
 
 
+def wait_for(func, timeout_sec=None, poll_interval_sec=1):
+    """
+    Waits until the func() returns non-None
+    :param func: function to call until a value is returned
+    :param timeout_sec: How much time to give up after
+    :param poll_interval_sec: How frequently to check the result
+    """
+    start_time = time.time()
```
PhilippWendler, line 218: This is not monotonic and can have jumps. Use `time.monotonic()`, please.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
                     benchmark.config.singularity,
                 ]
             )
-        srun_command.extend(args)
+        srun_command.extend(
+            [
+                "bash",
+                "-c",
+                f"{' '.join(args)} && echo 0 > exitcode || echo $? > exitcode",
```
PhilippWendler, line 290: This would re-introduce all the escaping problems that we eliminated in the last PR. Is there really no way around a shell invocation? Avoiding shell in these situations is really strongly recommended.

Furthermore, in the past we really had a lot of problems with shells mangling the exit code of the benchmarked process, especially if the process terminates due to a signal.

If we need the shell, we must use proper quoting to safely transport all characters such as whitespaces, quotes, `$`, etc.

I would also question whether we really need `bash` or whether we can be more flexible and use any standard shell available via `sh`?

Why `echo 0 > exitcode || echo $? > exitcode` instead of `echo $? > exitcode`?

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
...
+                    stdout=tmp_log_f,
+                    stderr=subprocess.STDOUT,
+                )
+
+            # we try to read back the log, in the first two lines there should be the jobid
+            with open(tmp_log, "r") as tmp_log_f:
+                first_lines = (
+                    tmp_log_f.readline(100).strip("\r\n")
+                    + "\\n"
+                    + tmp_log_f.readline(100).strip("\r\n")
```
PhilippWendler, line 311: Why the `100` here?

Why the `strip()`? Python's "universal newlines mode" should be active by default and handle all kinds of newline characters.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
...
+                return None
 
-    return {
-        "walltime": wall_time,
-        "cputime": cpu_time,
-        "memory": memory_usage,
-        "exitcode": ProcessExitCode.create(value=exit_code),
-    }
+        # sometimes `seff` needs a few extra seconds to realize the task has ended
+        result = wait_for(get_checked_seff_result, 30, 2)
```
PhilippWendler, line 339: After the timeout, this will make BenchExec crash, and all results are lost. I would recommend either waiting infinitely, or handling this gracefully.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
...
-exit_code_pattern = re.compile(r"exit code (\d+)")
+        if os.path.exists(exitcode_file):
+            with open(exitcode_file, "r") as f:
+                returncode = int(f.read())
+                logging.debug("Exit code in file %s: %d", exitcode_file, returncode)
+        else:
+            if status == "COMPLETED":
+                raise Exception(
+                    "Should never happen: exit code not found, but task was reported COMPLETED."
+                )
```
PhilippWendler, line 353: This indicates a bug? Then you can use `assert`.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
...
+                "TIMEOUT": "cputime",
+                "ERROR": "failed",
+                "FAILED": "killed",
+                "CANCELLED": "killed",
+            }.get(status, status)
+
+            if ret["terminationreason"] == "memory":
+                ret["memory"] = memory
+            if ret["terminationreason"] == "cputime":
+                ret["cputime"] = timelimit
```
PhilippWendler, line 377: Why this? I don't think we should blindly override this. For example, if a run takes longer than the timelimit, we still want to get the real measurement.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,120 @@ def run_slurm(benchmark, args, log_file):
...
+                    f"srun output (retcode: {str(returncode)}), truncated: {str(first_lines)}\n"
+                )
+                file.write(f"seff output: {str(result.stdout)}\n")
+                file.write(f"Parsed data: {str(ret)}\n")
+                empty_lines = "\n" * 3
+                file.write(empty_lines)
+                file.write("\n")
+                file.write(content)
+                if content == "":
+                    file.write("Original log file did not contain anything.")
```
PhilippWendler, line 394: If I understand correctly, this is for writing the log file that will be part of the BenchExec results?

This should look the same as if you run BenchExec locally. In particular, the first line should have the executed command line. This is *extremely* useful because you can just copy the whole line for executing the same thing locally outside of BenchExec.

All the other information here seems like debugging information that is useful for the developers of the SLURM mode. But we don't want debugging information in the tool log. If you really want to store that in a file, then it should be a separate file that is only written with `--debug`.

contrib/slurm/slurmexecutor.py
```diff
@@ -313,10 +408,12 @@ def parse_seff(result):
...
     wall_time_match = wall_time_pattern.search(result)
     memory_match = memory_pattern.search(result)
+    exit_code = None
     if exit_code_match:
-        exit_code = int(exit_code_match.group(1))
+        status = str(exit_code_match.group(1))
+        exit_code = int(exit_code_match.group(2))
     else:
-        raise Exception(f"Exit code not matched in output: {result}")
+        status = "ERROR"
```
PhilippWendler, line 416: In which situation does this happen? `ERROR` would be interpreted as a generic failure of the executed tool. If the problem is with something else, we should set a termination reason instead.

contrib/slurm-benchmark.py
leventeBajczi: Sure, fixed.

contrib/slurm/slurmexecutor.py
leventeBajczi: I don't think I can justify ratelimiting globally - this is very dependent on other jobs in the queue and their workloads, so this code would only get invoked if there is a problem with the scheduler. I can safely spawn 2000 jobs in parallel some times, but fail to spawn even 20 in other cases - and in the latter case, I don't want to overload the system with fast, repeated attempts. However, if you think this needs to be addressed, I'm open to suggestions on how we should achieve that.

contrib/slurm/slurmexecutor.py
leventeBajczi: This will only make the currently running task fail, not Benchexec -- in my testing, any exception I raise in these parts of the code, will just be printed to the terminal (which I admit, is not very elegant), but Benchexec will continue all other tasks without any problem. Is this not the intention? 
I copied the code for the _Worker class from the `localexecution.py` file, which also works the same way, as it catches all `BaseException` exceptions: https://github.com/sosy-lab/benchexec/blob/a75bb4f9745e7a605818f1fa5ffcdef147990772/benchexec/localexecution.py#L291 
So we just log the exception, and carry on.

contrib/slurm/slurmexecutor.py
leventeBajczi: Sure, thanks

contrib/slurm/slurmexecutor.py
leventeBajczi: I think you're right. I will make this behave more like the rest of benchexec. What should I put in the other 5 lines? Unfortunately the first 6 lines have to be padded, otherwise it's cut off before the toolinfo modules get a chance of parsing the output: https://github.com/sosy-lab/benchexec/blob/a75bb4f9745e7a605818f1fa5ffcdef147990772/benchexec/model.py#L1094
(Also, I thought based on this comment, that putting some useful logs here would be OK, since that's what these 6 lines are for -- maybe this comment should be removed/rephrased, as to not confuse further developers?)

contrib/slurm/slurmexecutor.py
leventeBajczi: `ERROR` will be translated to `failed` terminationreason, which I think is correct as a status here.

leventeBajczi: Here, the status is still the status according to SLURM, not to Benchexec. That will be translated later, here: https://github.com/leventeBajczi/benchexec/blob/1895c8e09ff58714408f99f3196f933b5f41fc7a/contrib/slurm/slurmexecutor.py#L365-L372

contrib/slurm/slurmexecutor.py
leventeBajczi: This is to counteract SLURM's weird way of measuring resources. SLURM will report 0 CPU time in the case of a timeout, and often 0 memory usage in the case of an OOM event. I have no idea why. To me, this seems like the next-best thing to a proper measurement: at least in the results we can see that they hit this limit. Otherwise, we'd see 0 time being spent with tasks that hit the time limit.

Example output of a job hititng its timelimit:

```
$ seff xxxxxxx
Job ID: xxxxxxx
Cluster: komondor
User/Group: p_benbl/p_benbl
State: TIMEOUT (exit code 0)
Nodes: 1
Cores per node: 3
CPU Utilized: 00:00:00
CPU Efficiency: 0% of 00:03:09 core-walltime
Job Wall-clock time: 00:01:03
Memory Utilized: 460.39 MB
Memory Efficiency: 9.21% of 4.88 GB
```

contrib/slurm/slurmexecutor.py
leventeBajczi: Unfortunately, because SLURM only reports one exitcode (which is 0 if both `srun` and its subcommand returned 0, non-null if _either_ returned non-zero, but sometimes also 0, together with `status: FAILED`, if `srun` experienced a failure), we have to save the subcommand's exitcode somehow. If we don't use a shell, then the only way to retrieve is through the spawning process (in this case, the scheduler), which does not cooperate with us on this. I think we must use a shell. Please correct me if I'm wrong on this, I'd love to go back to using no shells.

How do you suggest I go about quoting the arguments? Are simple `' '` quotes enough, with all single quotes in the arguments being escaped? Is there a utility for this already in Benchexec, or should I try and implement something here?

>   Why `echo 0 > exitcode || echo $? > exitcode` instead of `echo $? > exitcode`?
I'll use a `{command}; echo $? > exitcode` instead, that is cleaner. In the current state, either side would run based on the failure of the command -- the left side (right after `&&`) in the case of a successful run with 0 exitcode, or the right side (after `||`) in the case of a failure. But using `;` instead is cleaner.

contrib/slurm/slurmexecutor.py
PhilippWendler: Ah, right, because of the worker thread. Hm, not sure how to handle that well.

But right now it will crash with an `AttributeError`, which is a bug. The failure case should at least be handled properly.

contrib/slurm/slurmexecutor.py
PhilippWendler: Hm. In this case, are you really sure it used the CPU time? Does SLURM have any other time limits, like a wall-time limit?

I am really skeptical of "inventing" measurement values, because this makes it really easy to produce wrong results, if some edge case is not considered. And that would be among the worst things a benchmarking framework could do.

contrib/slurm/slurmexecutor.py
leventeBajczi: I'm not sure I undestand - why would it be an `AttributeError`?

PhilippWendler: Won't `result` be `None` and then `result.stdout` is accessed?

contrib/slurm/slurmexecutor.py
leventeBajczi: I think you're right with your point about inventing measurements. I'll remove that part of the code, and just put this in the list of limitations -- afterall, the fix for this in post-processing the results is trivial.

contrib/slurm/slurmexecutor.py
```diff
@@ -251,57 +283,121 @@ def run_slurm(benchmark, args, log_file):
...
+            if ret["terminationreason"] == "memory":
+                ret["memory"] = memory
+            if ret["terminationreason"] == "cputime":
+                ret["cputime"] = timelimit
+
+        # Runexec would populate the first 6 lines with metadata
+        with open(log_file, "w+") as file:
+            with open(tmp_log, "r") as log_source:
+                content = log_source.read()
+                file.write(f"{' '.join(args)}\n")
```
PhilippWendler, line 377: Please also use the escaping here. This is intended (and really useful) for copy-pasting.

contrib/slurm/slurmexecutor.py
leventeBajczi: I think (in my opinion) that this is indeed a bug. If SLURM does not realize that a job has completed for 30 seconds after it is already done, I suspect there are some serious issues in the state of the scheduler. 
However, currently none of this information is conveyed. I'll put this in the argument to the exception.

contrib/slurm/slurmexecutor.py
leventeBajczi: Sure, thanks. Sometimes I feel like I keep repeating my mistakes..

contrib/slurm/slurmexecutor.py
PhilippWendler: After the line with the output, there should be the same as from standard execution:
```python
                "\n\n\n"
                + "-" * 80
                + "\n\n\n"
```
Can we keep the debug output separate? It is already not so nice (but there since forever and really useful) that we have the command line in the same file as the tool output. But for post-processing, it is really better to keep different things separate. For example, debug output could get misinterpreted by the tool-info module. And if you run in debug mode, you might still want to put the results somewhere for other people to see, so let's not confuse them with debug output.

contrib/slurm/slurmexecutor.py
PhilippWendler: Don't feel bad! The lesson from this should only be to remember forever how important it is to have a proper way to handle command lines that does *not* need quoting/escaping, because it is clear that developers will at some point inevitably forget it. This is why I pushed for it. :-) For log output the effect will be limited, but for real execution that would be bad. Same lesson as all security-minded web developers will tell you about including user input in web pages.

contrib/slurm/slurmexecutor.py
PhilippWendler: Would it be a bug in SLURM or in the SLURM module of BenchExec? With "bug" above I meant only the latter, because a stack trace for BenchExec would not be helpful in debugging a SLURM bug anyway (and most users/developers of BenchExec would not be in a position to do so anyway). More info is still good, of course.

contrib/slurm/slurmexecutor.py
leventeBajczi: The tool-info module will never get this result, as the first 6 lines get cut off by the code in `module.py` before the tool-info modules get a chance of looking at it. But if you think it's better to keep it separate, I'm open to suggestions on where to put that (maybe a different file altogether)? If I create a new file next to the logfile, will it be properly handled by later parts of the code, e.g., when zipping the logfiles?

contrib/slurm/slurmexecutor.py
leventeBajczi: I removed the offending code, and the README already included this limitation. I guess I forgot to remove that line when I put this 'solution' together.

contrib/slurm/slurmexecutor.py
leventeBajczi: In SLURM, definitely. Or more precisely, not a bug, but a sign of a very degraded state. My reasoning for using an exception to convey this info (besides being very convenient, instead of a more graceful solution) is that if you see this message, your system is very likely to be affected in other ways as well. So there is no good reason in trying to get this single run to succeed / be reported in a nice way, an exception will do, as your entire benchmark run will likely have to be repeated -- in my testing, waiting for `seff` to realize the job's finished never took more than a second, so 30 seconds is already a way bigger timeout than we'll need in a working system.

contrib/slurm/slurmexecutor.py
PhilippWendler: I am not sure why limiting this here is important? It seems like it also doesn't hurt, but it stands out and confuses the reader as to why it is necessary.

Thinking more about it, I would actually wonder whether it breaks something. What will calling `readline(100)` twice do if the first line is longer than the 100 characters? Will it actually read the second line or just the first (in two parts)? Also note that the limit for the first line cannot be effective if you want the second line, Python would have to read all of the first line anyway.

Bothering with `strip()`, concatenation with `\n`, and `str()` also do not seem necessary at all.

I would recommend something like:
```python
for line in itertools.islice(tmp_log_f, 2):
    jobid_match = jobid_pattern.search(line)
    ...
```

contrib/slurm/slurmexecutor.py
PhilippWendler: Sure, this sounds reasonable. Using an exception internally is also fine, just please make it such that it does not produce a stack trace, just a proper error message - any stack trace that is printed in the output of the tool inevitably indicates a bug in that tool.

contrib/slurm/slurmexecutor.py
PhilippWendler: Yes, a separate file next to the log file with some additional suffix (`run.log_file + ".something"`) is what [similar modules](https://github.com/sosy-lab/benchexec/blob/main/contrib/vcloud/benchmarkclient_executor.py) also use. These files won't be zipped then.

contrib/slurm/slurmexecutor.py
PhilippWendler: I guess you would need some adaptive rate limit to handle this properly - submit jobs in parallel until something goes wrong, then reduce the parallelism. But I don't know how to implement this easily in Python and whether it is worth it. I was just wondering whether slowing down a single thread makes sense when the other parallel worker threads will still "hammer" SLURM as fast as they can.

contrib/slurm/slurmexecutor.py
leventeBajczi: The idea behind concetanation, etc, is that I want that to appear in the log (now only if the debug flag is given). For that, I need it to be single-line and reasonably long. However, that reason does not justify this level of complexity, so I'll rework this part, possibly with your recommendation. Thanks!

contrib/slurm/slurmexecutor.py
leventeBajczi: Only if they haven't experienced the same issue -- during most of the execution of each thread, the worker nodes work on the verification task. During that time, we don't submit new jobs from the same thread, so unless there is only a single thread failing this way (unlikely), all of the threads will be slowed down in this degraded state. While not a proper and fully-fledged solution, I think it's worth keeping in here for cases when multiple threads/processes failed to call `srun` successfully: if a 1000 parallel jobs keep retrying to no avail, I'd rather see them wait a second between each action rather than go as fast as they can.

contrib/slurm/slurmexecutor.py
leventeBajczi: Sure, I changed this to only log instead of raising an exception that gets a stacktrace.

leventeBajczi:
I think I addressed all your reviews, and all problems got resolved besides the shell question. In the meantime, a quite large test run has finished (around 33000 tasks), and I found no problems currently. 
Thanks for the quick responses again, and for the thorough reviews!

contrib/slurm/slurmexecutor.py
leventeBajczi: There is such an interface (`sbatch`), but I am mostly unfamiliar with it. And so far, even with 1000s of parallel jobs, I found no serious resource problems with the current solution.

contrib/slurm/slurmexecutor.py
PhilippWendler: Is there no way to get the tool's exit code out of the run metadata? For example, the same way you get the CPU time etc.?

If not, I also do not see a way anymore to circumvent the shell.

leventeBajczi: Unfortunately, no. A sample output:

```
Cluster: komondor
User/Group: p_benbl/p_benbl
State: TIMEOUT (exit code 0)
Nodes: 1
Cores per node: 3
CPU Utilized: 00:00:00
CPU Efficiency: 0% of 00:03:09 core-walltime
Job Wall-clock time: 00:01:03
Memory Utilized: 460.39 MB
Memory Efficiency: 9.21% of 4.88 GB
```

Here, the exit code is the combined exit code of the `srun` command as well as the wrapped command.

leventeBajczi: In this example as well, the exit code is supposed to be 0, but the verification task wasn't even finished. For 0 exit codes, we can rely on the status ("COMPLETED" with 0 exit code is when everything executed correctly), but for non-zero exit codes, there is no way of distinguishing between an internal `srun` error and an error during tool execution.

PhilippWendler: They should add this - like `runexec` does ;-)

So maybe switch to `/bin/sh`? `$?` should be available in all shells, not just bash.

leventeBajczi: Sure, I'll do this and run a quick test.

And I agree, just to think how much of the headache could have been avoided, only if SLURM relied on something akin to runexec... :)

leventeBajczi:
The re-tests (after moving to `sh`) are done, and I'm happy with their results, it captured all outputs and exit codes seemingly correctly.

Issue #830, opened 2022-05-23
Status drop-down does not correctly display on initial render


PhilippWendler:
When opening the table tab of our HTML tables with a defined status/category filter, e.g., via [this link](https://sosy-lab.github.io/benchexec/example-table/svcomp-simple-cbmc-cpachecker.table.html#/table?filter=0(0*status*(category(in(correct)))))), the drop-down filter does not correctly display the defined filter, it displays `Show all`. This is immediately corrected by doing something with the table, like changing sort order or even just opening the filter sidebar. This is also visible when switching back to the table tab from some othr tab. The text fields for numeric filters have correct behavior.

Git bisect points to 6b5be445f7f5f7fb4d421f7f7253d7ecdebdd85f, which unfortunately is a merge commit, so it is not immediately visible what causes this. Before this commit, the status drop-down displayed the wrong value briefly before re-rendering and showing the correct value immediately after loading. So there is likely some problem that status drop-down does not get the correct values on the first render, and the merge removed the additional re-render. Of course it would be best to have the correct value visible immediately without re-rendering.

JawHawk:
@PhilippWendler I found the source of the issue. It was due to the initial assigning of `filteredColumnValues` to value `{}` at line 103 in [https://github.com/sosy-lab/benchexec/blob/main/benchexec/tablegenerator/react-table/src/components/ReactTable.js](https://github.com/sosy-lab/benchexec/blob/main/benchexec/tablegenerator/react-table/src/components/ReactTable.js) .

To solve it, I defined a function to return the `newFilteredColumnValues` , assigned it to `filteredColumnValues` and made changes in useEffect. The solution works as it is immediately visible, attached is the screen recording of it.

**Can I create a PR to merge the changes from my fork to resolve the issue ?** Also I am interested in contributing to BenchExec in GSOC 2024.
\
![image](https://github.com/sosy-lab/benchexec/assets/86140365/9f2b1998-a368-4dd4-9c92-49e0523d2de3)
![image](https://github.com/sosy-lab/benchexec/assets/86140365/783ca0c0-72f8-43b7-b2c0-4f67da6308ec)

https://github.com/sosy-lab/benchexec/assets/86140365/3c8b446d-b938-4259-bacf-036233aa1cbd

PhilippWendler:
Thanks a lot! Sure, a PR is always welcome.

PR #1002, opened 2024-02-27
Fixed status drop down initial display

JawHawk:
Addresses #830 

Fixed the display on initial render by defining a function named `getNewFileteredColumnValues`. It returns the new Filtered Column Value and is subsequently used in `useEffect` to display changes on changing Filters.


benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -619,8 +621,8 @@ const Table = (props) => {
     }
   }, [columnResizing, columnsResizeValues]);
 
-  // get selected status and category values
-  useEffect(() => {
+  // get new Filtered Column Values
+  function getNewFileteredColumnValues() {
```
PhilippWendler, line 625: Can you please rename to `getNewFilteredColumnValues()`? (typo)

Also, the comment doesn't really describe something, it is just the same as the function name, so it is redundant. Can you make it more descriptive?

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -645,13 +647,20 @@ const Table = (props) => {
...
+  useEffect(() => {
+    let newFilteredColumnValues = getNewFileteredColumnValues();
     if (!deepEqual(newFilteredColumnValues, filteredColumnValues)) {
       setFilteredColumnValues(newFilteredColumnValues);
     }
     // Set current page to new last page if current page is not legit any more after filtering
     if (pageIndex >= pageCount) {
       gotoPage(pageCount - 1);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
PhilippWendler, line 663: Could you please explain why we need to disable this warning? Can we fix the warning instead?

It is always good to leave a comment describing why the warning is not relevant.

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -665,37 +674,40 @@ const Table = (props) => {
...
     <div className="tr headergroup" {...headerGroup.getHeaderGroupProps()}>
-      {headerGroup.headers.map((header) => (
-        <div
-          {...header.getHeaderProps({
-            className: `th header ${header.headers ? "outer " : ""}${
-              header.className
-            }`,
-          })}
-        >
+      {headerGroup.headers.map((header) => {
```
PhilippWendler, line 677: Is there any relevant change in this part of the code, or is it just reformatting? Looks like the latter to me. In this case, please keep only the relevant fix, not the reformatting.

PhilippWendler (commented):
Great! I can confirm that the fix works for me. :tada: 

I have just a few minor suggestions, it would be great if you could update the PR. Also, could you leave out the changes to the `package-lock.json`? We want to update dependencies separately.

benchexec/tablegenerator/react-table/src/components/ReactTable.js
JawHawk: react-hooks/exhaustive-deps shows the warning:  `useEffect` has a missing dependency: `getNewFilteredColumnValues`. 
But `getNewFilteredColumnValues` never changes and `useEffect` has the same working without it as a dependency.

If it needs to be included, then `getNewFilteredColumnValues` needs to be converted to callback function as it may cause unnecessary renders, and then we can't use the `getNewFilteredColumnValues` callback function to assign `filteredColumnValues` as it gives another warning: `getNewFilteredColumnValues` was used before it was defined. 
So ignoring the first warning is more preferable I guess.

JawHawk:
@PhilippWendler I have made the requested changes. Please review

benchexec/tablegenerator/react-table/src/components/ReactTable.js
PhilippWendler: Thanks! This is a really useful explanation.

PhilippWendler (approved):

PhilippWendler:
We have pre-built versions of the React applications in the repository to make it easier for non-JS developers, so when one changes the JS code, these builds also needs to be updated by running `npm run build`. Unfortunately, sometimes slight version changes influence the result somewhat (similar to the changes for `package-lock.json`, I guess), which we want to avoid, so I just ran the build and pushed now. The container with `node:12`, then `npm ci` and then `npm run build` works best for me to create this build in the desired manner.

PhilippWendler:
And merged! Thanks a lot for your fix and for your quick reactions, this was a great first contribution!

PR #1004, opened 2024-02-28
Added case insensitive filtering in HTML tables

### LINKED ISSUE #924: Case-insensitive filtering in HTML tables
In our HTML tables, it is natural to be able to want to filter in a case-insensitive way, for example with regards to task names or other textual columns. We could support this via a checkbox in the filter overlay.

JawHawk:
Fixes #924 .

Converting strings to lowercase first and then comparing for case insensitive filtering. 

Also fixed the Reference error caused by 47cd7772ad460d1d76c30be07736675e67b7d3e8. I apologize for my oversight.


PhilippWendler:
Thanks!

Case insensitive comparisons can be somewhat tricky, because not in all languages there is a 1:1 correspondence between lowercase and uppercase characters. This is why calling `toLowerCase()` might not produce the desired behavior for all locales. See for example the extensive explanation at https://www.b-list.org/weblog/2018/nov/26/case/

Is there a way in JS to do a case insensitive comparison directly instead? Or at least case folding as described in the blog post?

PhilippWendler:
I am not sure, but I think that the current approach only changes the "id" filters, right? We also can have other textual columns, and we should have the same matching behavior there.

And I think it would be really good if we could get some unit tests for this. There are already some in `src/tests/`, also for filters, maybe they just need some extensions? If you could have a look this would be greatly appreciated.

JawHawk:
> Is there a way in JS to do a case insensitive comparison directly instead? Or at least case folding as described in the blog post?

There is a method [toLocaleLowerCase](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase) which converts the provided string to lower case according to any locale-specific case mappings. There's also a way using [normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) as suggested in [blog](https://claritydev.net/blog/diacritic-insensitive-string-comparison-javascript). It may not be exact same as that of case folding, but I think most of the cases would be satisfied. Which one is preferable ?

PhilippWendler:
The problem with `toLocaleLowerCase` would be that it just works well with that locale then. `normalize` could improve things, but it seems it is not the same. Is there maybe a library that implements proper case-insensitive string matching? We are certainly not the only project that wants to have this requirement.

> I found the filters.test.js in src/tests. but I didn't understand exactly what to do. Can you please elaborate.

Right, these are the unit tests for `applyMatcher`. So we can add some test cases there that call `applyMatcher` for the textual matching cases, right?

JawHawk:
Got it, will add the textual matching test cases soon. 

I only found two npm packages which do the case folding - [@ar-nelson/foldcase](https://www.npmjs.com/package/@ar-nelson/foldcase) and [fold-case](https://www.npmjs.com/package/fold-case)  but they are not popular. Will they work ?

PhilippWendler:
Hm, this seems more complex than I thought. The two packages you found also only have relatively old releases.

The one alternative I found would be to use a regexp with case insensitive matching. We just have to make sure that all the special characters in the search string are escaped. And we should probably create an regexp instance and reuse it for matching across all cells instead of compiling the regexp every time. Do you think that would be doable?

JawHawk:
Yes, will try to do it soon.

benchexec/tablegenerator/react-table/src/utils/filters.js
```diff
@@ -228,7 +228,11 @@ const applyMatcher = (matcher) => (data) => {
     const { value: idValue, values: idValues } = matcher.id;
     if (idValue) {
       diffd = diffd.filter(({ id }) =>
-        id.some((idName) => idName === idValue || idName.includes(idValue)),
+        id.some(
+          (idName) =>
+            idName === idValue ||
+            idName.toLowerCase().includes(idValue.toLowerCase()),
```
kfriedberger, line 234: possible optimization: precompute idValue.tolowerCase instead of computing it for every item

JawHawk:
Please review the changes made.

benchexec/tablegenerator/react-table/src/tests/utils.test.js
```diff
@@ -133,8 +133,14 @@ describe("decodeFilter", () => {
     expect(decodeFilter(filter)).toEqual(expected);
   });
 
-  test("should throw errors if there are not exactly two '_' in the filter id", () => {
-    expect(() => decodeFilter("0_cputime")).toThrow();
+  test("should handle empty filters", () => {
```
PhilippWendler, line 136: I guess this test name needs to be updated.

benchexec/tablegenerator/react-table/src/utils/filters.js
```diff
@@ -233,20 +233,34 @@ const applyMatcher = (matcher) => (data) => {
   if (!isNil(matcher.id)) {
     const { value: idValue, values: idValues } = matcher.id;
     if (idValue) {
+      // pre computing RegExp of idValue after excaping the special characters
+      let regexToCompare = new RegExp(
+        idValue.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"),
+        "ui",
+      );
```
PhilippWendler, line 240: Can you please move the regexp construction in a separate utility function? Right now the code exists twice and it would be good to avoid this duplication.

PhilippWendler (commented):
Looks good, just minor suggestions!

I like the test with the special character. How about also adding a test with some characters that are special for regexps, in order to test that the escaping properly works? For example, test that `...` does not match anything or so.

benchexec/tablegenerator/react-table/src/utils/filters.js
JawHawk: I have done the requested changes.

PR #1005, opened 2024-02-29
feat: introduce the `decodeFilter` function to refactor filter parsing

### LINKED ISSUE #731: Refactor filter parsing in JS code
In the HTML tables, filter state is stored (like other state) in URLs and thus serialized in string form. It seems that deserialization and parsing occurs more often than necessary and with duplicated code, which is highly error-prone. For example, the code `.split("_")` for separating a filter id into pieces occurs already four times (in `ReactTable.js`, `utils.js`, `FilterBox.js`, and `filters.js`), and #730 is adding more cases.

This should be refactored such that there is no code duplication and ideally each string is used only once and converted into a proper object-oriented representation, and the latter is used by the rest of the code.

EshaanAgg:
Fixes #731 by creating a `decodeFilter` function, which splits the `filterID` passed to it and returns an object containing all the parts. Refactors all the previous usages of the old `.split("_")` approach as well to make use of this new helper function.


PhilippWendler:
Thank you very much! I like a lot that you also thought of and added documentation, error handling, and unit tests nicely! :+1: This is fine to merge.

I am just wondering whether you maybe see more potential for refactorings in a similar way, i.e., by identifying duplicate (or just complex pieces of) code and moving them out into functions that can be individually documented tested?

EshaanAgg:
Thanks for the review! I'll surely open issues regarding possible refactorings if I come across them in my exploration of the codebase!

PR #1006, opened 2024-03-02
feat: refactor the `setHash` function

### LINKED ISSUE #807: Refactor setHashSearch in JS code
The method [`setHashSearch`](https://github.com/sosy-lab/benchexec/blob/2695205de2376efb3b519b7dc5d1770825c9f0d3/benchexec/tablegenerator/react-table/src/utils/utils.js#L229) is not well designed and should be refactored.

It implements both the creation of the new hash string as well as (optionally) two(!) different side effects of what to do with the string. So the method should be split into the side-effect-free string creation and (if necessary) other methods handling the side effects. In this regard #805 and the avoiding of manual history manipulation could be relevant.

Furthermore, the method is not used well, because for example the method offers the `keepOthers` feature, which only some callers use. The other calls manually merge old and new hash params into a dict before, which could be replaced with setting `keepOthers` to `false`. And if I did not overlook something, there actually does not exist a use case for `keepOthers = false`, so this could be removed.

But first we need more tests for this method.

EshaanAgg:
## Related Issue
Fixes #807 

## Implementation notes
- Refactored the parts of `getHashSearch` to increase it's readability
- Split the `setHashSearch` function into two functions
  - `setHashSearch`: Responsible for actually setting the query parameters correctly in the URL based on the configuration options provided
  - `constructQueryString`: Responsible for taking in an object of `params` and generating a query string from them. This utility offers a layer of abstraction and can be reused in the codebase at many places
  
## Roadblock
The `setHashSeach` function is responsible for creating the correct URL and setting the `document.location.href` to the same. This poses a problem to us, as `JSDom` does not mock routing and thus throws an error when we try to test the same in the unit tests for the function. It throws the [error described in this Stack Overflow thread](https://stackoverflow.com/questions/54090231/how-to-fix-error-not-implemented-navigation-except-hash-changes), but all of the solutions mentioned there, as when the issue for the same on the JSDom repository does not fix it. The same can be verified by pulling this PR, commenting out the test suite for `setHashSearch`, and then running the test suite. 

I could not find the specific version of JSDom we are using in the project in the `package.json`, and I also had trouble understanding the testing setup in the repository. With more context, I can suggest a better approach to tackle this problem.


PhilippWendler:
The splitting is definitively improving things already. Maybe we can refactor even further, like pushing the whole URL construction into the utility method? `baseUrl` and `returnString` could potentially even be removed as a parameter from `setHashSearch`, they seem to be relevant only for testing, but those tests would then test the utility method.

JSDom seems to be a transitive dependency of Jest, and `package-lock.json` refers to version 16.7.0. I would have expected mocking to work, though, because we successfully mock `window.crypto` in the same way for our tests, so mocking `window.location` should also be possible. What did you try in this regard?

EshaanAgg:
> The splitting is definitively improving things already. Maybe we can refactor even further, like pushing the whole URL construction into the utility method? `baseUrl` and `returnString` could potentially even be removed as a parameter from `setHashSearch`, they seem to be relevant only for testing, but those tests would then test the utility method.

Okay, sir. I'll refactor the URL construction into a different utility function altogether. 

> JSDom seems to be a transitive dependency of Jest, and `package-lock.json` refers to version 16.7.0. I would have expected mocking to work because we successfully mock `window.crypto` in the same way for our tests, so mocking `window.location` should also be possible. What did you try in this regard?

I tried a couple of things which were listed in the various resources I consulted online:
- Trying to delete the `window`/`document` object or reassigning the relevant properties on the same (something like the following):
```ts
delete window.location;
window.location = { assign: assignMock };
```
The problem with this approach that I get the error that you can't reassign or delete the properties on the read-only objects `window` and `document`.
The same happens while trying to use the `defineProperty` function:
```ts
const mockResponse = jest.fn();
Object.defineProperty(window, 'location', {
  value: {
    hash: {
      endsWith: mockResponse,
      includes: mockResponse,
    },
    assign: mockResponse,
  },
  writable: true,
});```

- Using third-party libraries such as `jest-location-mock`, but no luck.
- Making use of `window.location,assign` rather than `window.location.href`, but the same throws the following error:
![image](https://github.com/sosy-lab/benchexec/assets/96648934/6eb3b682-18ab-4fd5-8809-2f15a081cb69)

**Comment by PhilippWendler** [comment]
Hm, we mock `window.crypto` via this line:
https://github.com/sosy-lab/benchexec/blob/2217e9f2f6b2800b749bd2cb9bd136d17225691a/benchexec/tablegenerator/react-table/src/tests/setup.js#L13
That seems to work. Maybe try the same? Maybe it also only works in the setup code and not in the test itself? (If you tried out the latter.)

**Comment by EshaanAgg** [comment]
image
This is the error that I face with I try to mock it in the following manner. The reason for the same (as far as I understand it) is because `window.crypto` is not a resevered property, where as `window.location` or `document.location` is a reserved and read-only property that has been setup by JSDOM, but can't be overwritten.

**Comment by EshaanAgg** [comment]
I have updated the PR in the following manner:
- Created a utility `getURLHash` which accepts a URL, a object `params` and `keepOthers` boolean, and has the responsibility of generating the new URLs. This function is 100% unit testable and has been supplemented witht the all the tests.
- The function `setURLHash` now accepts three positional arguments (which is cleaner to read than the earlier approach of specifying the parameters in an `options` object) and performs the desired operation. No tests have been written for this function due to the hiccups above.
- I have also updated the codebase to make use of the new function signature everywhere.

**Review by PhilippWendler — COMMENTED** [review]
Thanks! This is already a nice improvement, and I fully agree with your functionality split and testing strategy. I guess it is the best that we can do.

I have just a few minor comments left, and one question: Could you maybe have a look at the callers of `setHashSearch` and check whether we can simplify them? IIRC some of them call `getHashSearch` and merge values manually, but we could use `keepOthers` instead.

While writing this I noticed that at least in `ReactTable.js`, we have several calls that first check whether the value of a parameter changes compared to `getHashSearch`. We could easily simplify this if we just do nothing in `setHashSearch` if the URL happens to remain the same, right? I don't think this would break anything.

File: benchexec/tablegenerator/react-table/src/utils/utils.js [hunk 1]
```diff
@@ -203,68 +203,78 @@ const EXTENDED_DISCRETE_COLOR_RANGE = [
...
+
+  // Extract the search part of the URL
   const search = urlParts.length > 1 ? urlParts.slice(1).join("?") : undefined;
-  if (search === undefined || search.length === 0) {
-    return {};
-  }
-  const keyValuePairs = search.split("&").map((i) => i.split("="));
 
+  // If there are no search parameters, return an empty object
+  if (search === undefined || search.length === 0) return {};
```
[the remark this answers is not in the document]
EshaanAgg: Sure sir, will update!

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -591,7 +591,7 @@ const Table = (props) => {
     const value = sort.length ? sort : undefined;
     const prevParams = getHashSearch();
     if (prevParams["sort"] !== value) {
-      setHashSearch({ sort: value }, { keepOthers: true });
+      setHashSearch({ sort: value }, true);
```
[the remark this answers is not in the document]
EshaanAgg: Yes sir I do completely agree that boolean parameters are most definitely not the ideal choice, and do add a certain friction when trying to read and understand the code.

I feel that opting for the middle ground is the best in this scenario:
- For the 'setHashSearch' method we can revert back to the options object as it is meant to be used by the various components.
- For the internal 'getHashURL' we can make use of boolean parameter, as it's not meant to be directly called by the various components, and it doesn't make sense to create an options object for just a single parameter.

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -203,68 +203,78 @@ const EXTENDED_DISCRETE_COLOR_RANGE = [
...
+ * Constructs a URL hash from the provided parameters
  *
- * @param {Object} params Object containing the params to be encoded as query params
- * @param {Boolean} [returnString] if true, only returns the url without setting it
+ * @param {string} url - The URL to be processed
+ * @param {Object} params - The parameters to be included in the hash
+ * @param {boolean} [keepOthers=false] - Whether to keep existing parameters in the URL hash or not
+ * @returns {string} - The constructed URL hash
+ */
+export const getHashURL = (url, params = {}, keepOthers = false) => {
```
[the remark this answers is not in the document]
EshaanAgg: Got it sir. Will use the `construct` keyword as it is more descriptive IMO. Thanks for the detailed review!

benchexec/tablegenerator/react-table/src/tests/utils.test.js
```diff
@@ -108,14 +109,64 @@ describe("hashRouting helpers", () => {
...
+    });
+
+    test("should return the same URL with no params if no parameters are provided and keepOthers is false", () => {
+      const baseUrl = "http://example.com?exisitingKey=existingValue";
+      const params = {};
+      const keepOthers = false;
+
+      expect(getHashURL(baseUrl, params, keepOthers)).toEqual(
+        "http://example.com",
+      );
```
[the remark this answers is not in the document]
EshaanAgg: Yes that is indeed an important test case that I missed. Thanks!

EshaanAgg:
I have updated the PR to address all the review comments. Two noteworthy changes:
- To simplify the logic of the functions calling the `setHashSearch` method, I have removed the code where they check if the parameter already exists in the URL and instead replaced them with the `setParam` method, which updates the parameters as needed. 
- In all the places we are calling the `constructHashURL` function with the boolean parameter, I have made sure we explicitly use a named variable to convert the meaning behind the boolean.

PhilippWendler:
Nice idea to use `setParam`. AFAIS there are only two remaining users of `setHashSearch`, and in particular `makeUrlFilterSerializer` could be simplified by making use of the fact that `setHashSearch` deletes params if we pass `undefined`. And then I think all code uses `keepOthers = true`?

Maybe we can even get rid of the `keepOthers` parameter and `setParam` completely? Then we wouldn't even need `options`. Sorry for the back-and-forth, but sometimes refactoring is like this because you gradually end up learning more and more about the code base and seeing more possibilities to improve stuff.

EshaanAgg:
Makes sense! It would be better to remove the `setHashSearch` function instead (as it has a confusing name) and keep using the `setParam` function in the code (with an option to pass in the history object into the same). I would also look into the refactoring the `makeUrlFilterSerializer` function!

EshaanAgg:
I have made the changes as described in the comment above. I have also added a new utility method called `removeParam`, which does the exact opposite of the `setParam` method, as I felt the same was needed for code readability in the `makeUrlFilterSerializer` method. 

PS. I am still a bit unsure if I refactored the `makeUrlFilterSerializer` method correctly, so would highly appreciate any review on the same. Thanks a lot @PhilippWendler sir for all the time and detailed reviews you left along the way

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -589,29 +589,20 @@ const Table = (props) => {
...
 
   // Update the URL page param when the table page changed
   useEffect(() => {
     const value =
       pageIndex && pageIndex !== 0 ? Number(pageIndex) + 1 : undefined;
-    const prevParams = getHashSearch();
-    if (prevParams["page"] !== value) {
-      setHashSearch({ page: value }, { keepOthers: true });
-    }
+    setParam({ page: value }, true);
```
PhilippWendler, line 605: I think here we need to remove the `true`, right?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -661,35 +691,27 @@ const makeFilterDeserializer =
 
 const makeUrlFilterSerializer = (statusValues, categoryValues) => {
   const serializer = makeFilterSerializer({ statusValues, categoryValues });
-  return (filter, options) => {
-    const previousParams = getHashSearch();
+  return (filter, history) => {
     if (!filter) {
-      return setHashSearch(previousParams, options);
+      return setParam({}, history);
```
PhilippWendler, line 696: I think this should also delete the `filter` like the case below.

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -203,68 +203,97 @@ const EXTENDED_DISCRETE_COLOR_RANGE = [
...
+ * Removes the specified search parameter from the URL hash of the current page. Also accepts a history object to update the URL hash without reloading the page.
+ * @param {string} param - The name of the parameter to be removed
+ * @param {Object} [history=null] - The history object to use for updating the URL hash
+ */
+export const removeParam = (param, history = null) => {
+  const params = getHashSearch();
+  delete params[param];
+
+  const keepOthers = false;
+  const newUrl = constructHashURL(document.location.href, params, keepOthers);
```
PhilippWendler, line 291: Isn't this the same as `constructHashURL(document.location.href, {param: undefined}, true)`? `constructHashURL ` first sets the parameter to `undefined` and then `constructQueryString` removes it from the result?

If I am not mistaken, we could remove `removeParam` and document that `setParam` can be used to delete a parameter by setting it to `undefined`. Or alternatively we can replace the content of `removeParam` with `setParam({param: undefined}, history)`, but I don't think it is necessary to keep a separate method. Also we can remove the `keepOthers` parameter from `constructHashURL`.

PhilippWendler (commented):
Yes, you are right, `setParam` is probably the better choice (or maybe setURLParameter`?). I guess we could rename `getHashSearch` then as well, to something like `getURLParameters`?

benchexec/tablegenerator/react-table/src/utils/utils.js
EshaanAgg: Okay!

benchexec/tablegenerator/react-table/src/utils/utils.js
EshaanAgg: Ahh, nice! This trick to remove parameters is super neat! I'll update the implementation and document this behavior in the JSDoc comments. 

Though I don't think we should deprecate the `keepOthers` parameter from the `constructHashURL` function, as it may be needed in the future? We already have a comprehensive test suite for it, and the same is only being utilized once in the codebase, so it might be good to keep support for the same. How do you feel about it?

PhilippWendler: Yes, the tests are good, but I don't really see why we would possibly need it in the future. In the end, it has turned out that all existing users of this feature don't actually need it. In general I think we can prefer less complexity over stuff that is only hypothetically useful in the future. It is also not complex to re-add if we ever need it.

EshaanAgg: Okay sir!

benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -589,29 +589,20 @@ const Table = (props) => {
       )
       .join(";");
     const value = sort.length ? sort : undefined;
-    const prevParams = getHashSearch();
-    if (prevParams["sort"] !== value) {
-      setHashSearch({ sort: value }, { keepOthers: true });
-    }
+    setURLParameter({ sort: value });
```
PhilippWendler, line 592: It is good that we now document the feature of parameter deletion via `undefined` values, because here we can see that it was actually used already!

benchexec/tablegenerator/react-table/src/tests/utils.test.js
```diff
@@ -72,51 +73,128 @@ describe("numericSortMethod", () => {
...
+    test("should remove exisiting parameters if they are updated to null", () => {
+      const baseUrl = "http://example.com?key1=value1&key2=value2";
+      const params = { key2: null };
+
+      expect(constructHashURL(baseUrl, params)).toEqual(
+        "http://example.com?key1=value1",
+      );
+    });
+
+    test("should not remove exisiting parameters if they are updated to falsy values", () => {
```
PhilippWendler, line 187: Very nice thinking!

PhilippWendler (approved):
Really nice now and a significant improvement compared to the old code. Exactly what I had hoped for in #807. Thank you very much!

EshaanAgg:
Thanks a lot for the continued support and reviews!

Issue #1007, opened 2024-03-02
Feature Request: Replication of CI/CD pipelines via GitHub workflows


EshaanAgg:
Currently, we are using `GitLab Pipelines` to run some of the essential tests on the PRs, like linting, tests, and builds. This is not ideal, as this feature is only available to branches on the central repository; thus, all the external contributors can't benefit from it. Instead, we should add these pipelines to GitHub workflows so they can be run on every PR. 

As a proof of concept of the same, I would first like to add a new workflow (primarily targeting JS codebase) to the repository that does the following:
- Installs the dependencies via `npm` and caches the same as an artifact
- Runs the linting and formatting checks on the code
- If the previous passes, then run the test suite on the code
- If the previous passes, then does either of the following (which would be more beneficial is something I would like input on):
  - Builds the application with `node:12`, and if the PR does not have the updated build, then the check fails
  - Builds the app with `node:12`, and then automatically commits the build to the PR so that the contributors or the maintainers don't have to do the same manually

PhilippWendler:
Yes, I guess we should do this. Having all on GitLab and only need to worry with one CI system is nice, but the advantage for outside contributors is larger.

We can run all the checks (linter, tests, build) in separate jobs such that we always execute all of them. This improves feedback time for contributors (you don't have to first fix your formatting to see that also some tests fail).

EshaanAgg:
Okay, sir. I would try to add these CI/CD checks and raise a PR for the same soon! Thanks.

PhilippWendler:
Implemented by #1009

PR #1009, opened 2024-03-04
feat: add javascript workflows via GitHub Workflows

EshaanAgg:
Fixes part of #1007 
Adds the JavaScript workflows for
- Linting
- Running the tests
- Building the application

It also makes sure that the `node_modules` folder is cached between runs so that the workflows run in parallel and quickly.

### DIFF
--- a/.github/workflows/javascript.yml
+++ b/.github/workflows/javascript.yml
@@ -1,0 +1,97 @@
+name: JavaScript Pipeline
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  lint:
+    name: Linting JavaScript code
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12"
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: benchexec/tablegenerator/react-table/node_modules
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: benchexec/tablegenerator/react-table
+
+      - name: Lint
+        run: npm run lint
+        working-directory: benchexec/tablegenerator/react-table
+
+  tests:
+    name: Run JavaScript tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12"
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: benchexec/tablegenerator/react-table/node_modules
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: benchexec/tablegenerator/react-table
+
+      - name: Run tests
+        run: npm run test
+        working-directory: benchexec/tablegenerator/react-table
+
+  javascript-build:
+    name: Build JavaScript
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12"
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: benchexec/tablegenerator/react-table/node_modules
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: benchexec/tablegenerator/react-table
+
+      - name: Build
+        run: |
+          npm run build
+          git diff --stat --exit-code
+        working-directory: benchexec/tablegenerator/react-table
--- a/.github/workflows/javascript.yml
+++ b/.github/workflows/javascript.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.npm
-          key: npm-${{ hashFiles('package-lock.json') }}
+          key: npm-${{ hashFiles('benchexec/tablegenerator/react-table/package-lock.json') }}
           restore-keys: npm-
 
       - name: Install dependencies
@@ -53,7 +53,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.npm
-          key: npm-${{ hashFiles('package-lock.json') }}
+          key: npm-${{ hashFiles('benchexec/tablegenerator/react-table/package-lock.json') }}
           restore-keys: npm-
 
       - name: Install dependencies
@@ -80,7 +80,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.npm
-          key: npm-${{ hashFiles('package-lock.json') }}
+          key: npm-${{ hashFiles('benchexec/tablegenerator/react-table/package-lock.json') }}
           restore-keys: npm-
 
       - name: Install dependencies
--- a/.github/workflows/javascript.yml
+++ b/.github/workflows/javascript.yml
@@ -26,8 +26,6 @@ jobs:
 
       - name: Setup Node.js
         uses: actions/setup-node@v2
-        with:
-          node-version: "12"
 
       - name: Cache node modules
         uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/javascript.yml
```diff
@@ -0,0 +1,98 @@
...
+name: JavaScript Pipeline
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
```
PhilippWendler, line 17: Is there a reason it shouldn't run on other branches?

For the other CI checks that we already have on GitHub Actions we use `on: [push, pull_request]` and I am wondering what is better.

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,98 @@
...
+jobs:
+  lint:
+    name: Perform linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
```
PhilippWendler, line 28: Can we pin Node to the latest LTS version here and for the other checks? This would make it consistent with the version we use in GitLab CI?

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: benchexec/tablegenerator/react-table/node_modules
```
PhilippWendler, line 33: Thanks for thinking of caching, this is good!

However, as far as I understand [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci/#description), it always deletes `node_modules`. On GitLab CI we instead cache the cache directory of npm, such that at least the downloading of packages is skipped:
https://github.com/sosy-lab/benchexec/blob/2217e9f2f6b2800b749bd2cb9bd136d17225691a/.gitlab-ci.yml#L15-L28

Can we do the same here? And can we also set `prefer_offline` such that `npm` doesn't attempt to look up stuff online as long as it is not necessary?

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,98 @@
...
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-
```
PhilippWendler, line 36: I am wondering about `restore-keys`: As far as I understand the [documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action), the current definition does not match anything because we never create a cache with the key `${{ runner.os }}-node-modules-` anywhere, so it would never successfully look it up under this name, right?

It is actually a nice trick to use the hash of `package-lock.json` as a cache key, I like that. But if we cache only the cache directory of npm, we might not even need that and can just reuse the cache always?

PhilippWendler (commented):
Thanks! This looks good, I just have 3 questions.

And is there maybe a way to refactor out the common steps such that we don't have to write down everything 3 times?

.github/workflows/javascript.yml
EshaanAgg: There is no specific reason for the same, but just the fact that being explicit above the branches that you want to run the workflow on usually helps limit the runs. It'd probably be better to standardize this workflow following others and run it on all PRs.

.github/workflows/javascript.yml
EshaanAgg: `setup-node` uses the LTS version of Node by default. I'll be explicit above it by specifying the `node_version`.

.github/workflows/javascript.yml
EshaanAgg: > I am wondering about `restore-keys`: As far as I understand the [documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action), the current definition does not match anything because we never create a cache with the key `${{ runner.os }}-node-modules-` anywhere, so it would never successfully look it up under this name, right?

No. The key with the hashed `package.json` is used to build the artifact for caching, but it can be restored by any of the keys mentioned (that is without the hashed `package-lock.json`), so we do indeed make use of the cache. I think the best would be to cache the `.npm` directory with the `package-lock.json` as key, so we do update the `.npm` content as and when the contents of the lockfile change.

.github/workflows/javascript.yml
EshaanAgg: Got it! I would look into the ways of doing the same. Using `npm ci` is preferred as it would respect the `package-lock.json` and would not fetch the lastest updates.

EshaanAgg: I couldn't find any resource for setting `prefer_offline` setting for this workflow, but I found this [article](https://www.voorhoede.nl/en/blog/super-fast-npm-install-on-github-actions/), which in my opinion does a great job of explaining the two different caching strategies and what the ideal idiomatic think should be. Using `npm ci` along with caching `.npm` seems to our best bet!

.github/workflows/javascript.yml
EshaanAgg: Yes, my bad. You are correct here. I don't think the `restore_keys` actually serve a purpose for us here (as we have a fixed runner, and the only thing changing is the `package-lock.json`)! I'd prefer to remove the same to simplify the workflow as much as possible. I also tried to find a way to research a way so that we don't have to duplicate the setting up of node and checking out the repository part of the workflows, but I wasn't able to find something on it (other than using a build matrix with different commands like `npm run lint`, `npm run test` etc., which IMO profoundly decreases the legibality of the codebase.

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,91 @@
...
+    name: Perform linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18.12.0"
```
PhilippWendler, line 23: This is quite specific and would need constant updating. So far we it seems to work fine in GitLab CI to always use the latest LTS version, so I guess we can do so here as well. Is `lts` supported as a version specification?

EshaanAgg: It supports versions in NVM LTS syntax: `lts/erbium`, `lts/fermium`, `lts/*` etc, but not just plain `lts`.

.github/workflows/javascript.yml
EshaanAgg: If I understand if correctly, then we do need the hash. Let's say I added a new dependency `XYZ` to the project. Then, in that process, I would update my `package-lock.json` and would re-cache the `.npm` directory. This is essential cause now the newly cached `.npm` directory would also have the compiled files for `XYZ`, and thus during `npm ci` would fetch the same from local storage and not the internet. If we omit the hash, then we are essentially fixing the content of the `.npm` directory to the current state. 

Or am I misinterpreting something here?

.github/workflows/javascript.yml
PhilippWendler: Ok, what does that mean? I haven't heard of `erbium` etc. For GitLab CI we just use the [official node Docker image](https://hub.docker.com/_/node) and that has a version named `node`. Do you know which of the above would correspond to this image?

.github/workflows/javascript.yml
PhilippWendler: Hm, maybe I am also wrong.

I would think that a cache works such that in your example, after you add the dependency and run CI once, the cache gets updated such that it now includes the new dependency. Every other CI job after that would use the new cache with the added dependency, but it doesn't hurt if there is more in `.npm` than what the job needs.

This is at least what GitLab's cache does and how I understand the term "cache" in general. Is GitHub different?

...

Ok, [found it](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache). Indeed, GitHub's cache is a one-time thing. And the next section on that page even says it doesn't use the same cache across branches? This is really weird.

Ok, so we really need the hash, and we need the fallback key as well. Note our `package-lock.json` is in `benchexec/tablegenerator/react-table/`, I think we need to specify this path.

If we run `npm ci` with the correct working directory, does it put the `.npm` in the working directory or in the root of the project? This would also change what we need to cache.

EshaanAgg: Aight. I will restore the fallback keys into the workflow so that we should get the caching benefits. The `.npm` is always created with respect to the root (i.e. at `~./npm`), so once we setup the cache in the next step, no matter the directory we run `npm ci` from, we should use the cached dependencies.

EshaanAgg:
image
The cache seems to be working as intended 🎉

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "lts/iron"
+
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.npm
+          key: npm-${{ hashFiles('package-lock.json') }}
```
PhilippWendler, line 29: Last question: I guess we need the path `benchexec/tablegenerator/react-table/package-lock.json` here?

EshaanAgg: Yes totally. I'm so sorry for missing it and not being careful!

PhilippWendler: No problem! Now it works perfectly:
image
It shows that the primary key has the hash (was missing previously), but that it successfully used the cache named `npm-`.

PhilippWendler (approved):

PR #1016, opened 2024-03-14
Handle special characters in filters

### LINKED ISSUE #806: Handle special characters in filters of HTML tables
Not all special characters are handled correctly if a user enters them as a filter for task ids or textual columns because of missing escaping. Some of them just do not work, some actually break the parsing of the URL.

Some characters are actually handled differently depending on whether they are entered on the table tab or in the filter overlay, but not sure whether either of them is working.

We should check and handle at least the following characters, and also add tests for them:
- `?`
- `#`
- `&`
- `=`
- `(`
- `)`
- `,`
- `*`

### LINKED ISSUE #1015: Id filter is removed when other filters are used
Due to some very recent commits, id filter gets removed when other filters are applied. Also there is a very abrupt reload and search bar gets out of focused. Attached is the video of the same.

https://github.com/sosy-lab/benchexec/assets/86140365/ca83eb9a-7f29-4ea2-a78e-0c98392fbfdc

I am interested in fixing this along with #806 . Thorough testing is required before commits to avoid such unnecessary breaks.

JawHawk:
Fixes #1015
Fixes #806

Solved the Id filter being removed when other filters are applied by fixing the `setURLParameter` function. 

Added `encodeURIComponent` in `makeFilterSerializer` function to escape special characters and `decodeURIComponent` in `tokenizePart` function to get the filter value back.
Also fixed `decodeFilter` for column names having `_`


benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -610,6 +623,7 @@ const makeFilterDeserializer =
         out.push({
           id: "id",
           ...tokenizePart(filter),
+          isTableTabFilter: true,
```
PhilippWendler, line 626: What does this do? It was added in 2f03808b016150b2c954d27ae86b263eeb39096d, but the commit message only mentions changes to tests.

PhilippWendler (commented):
Thanks a lot!

Does the handling of `_` and other special characters also work for filters that are set for other textual columns, i.e., not the id column?

I am also a little bit skeptical about the existing filter-parsing logic. Could you please add tests that have just one an opening parentheses? I would like to see that this works as well.

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -269,10 +272,14 @@ export const constructHashURL = (url, params = {}) => {
  * @returns {void}
  */
 const setURLParameter = (params = {}, history = null) => {
-  const newUrl = constructHashURL(document.location.href, params);
-
+  const { newUrl, queryString } = constructHashURL(
+    document.location.href,
+    params,
+  );
+  console.log(history, "s");
```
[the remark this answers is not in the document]
JawHawk: I am sorry for the oversight but have already removed it in 2f03808b016150b2c954d27ae86b263eeb39096d

benchexec/tablegenerator/react-table/src/utils/utils.js
JawHawk:
```
const makeFilterSerializer =
  ({ statusValues: allStatusValues, categoryValues: allCategoryValues }) =>
  (filter) => {
    const groupedFilters = {};
    const tableTabIdFilters = [];
    for (const { id, value, values, isTableTabFilter } of filter) {
      if (id === "id") {
        if (values && values.length > 0) {
          groupedFilters.ids = {
            values: values.map((val) => (val ? val : "")),
          };
        } else if (isTableTabFilter) {
          tableTabIdFilters.push({ id, value });
        }
        continue;
      } ...
```
```
if (tableTabIdFilters) {
      tableTabIdFilters.forEach((filter) => {
        runsetFilters.push(
          `id_any(value(${encodeURIComponent(filter.value)}))`,
        );
      });
    }
```
In the `makeFilterSerializer` function, we are pushing to `tableTabIdFilters` array by checking the `isTableTabFilter` property in filter element. Same property should be added while deserializing it but wasn't done. This resulted in id filter getting null even before the recent commits on a special case , i.e if reloaded the page with only id filter applied and then added any other filter.

benchexec/tablegenerator/react-table/src/utils/utils.js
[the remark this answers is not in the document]
PhilippWendler: Hm, I must have looked at an intermediate state of the PR. Sorry.

JawHawk:
> I am also a little bit skeptical about the existing filter-parsing logic. Could you please add tests that have just one an opening parentheses? I would like to see that this works as well.

Yes, even the online example `https://sosy-lab.github.io/benchexec/example-table/svcomp-simple-cbmc-cpachecker.table.html#/table?filter=id_any(value()))` breaks when only one closing parentheses is used. This is due to the `tokenizePart` function logic of considering '(' and ')'. For this, parantheses in the string values should also be escaped but `encodeURIComponent` doesn't do so. String values need to be `replacedAll` with `%28` and `%29` respectively. That will make it work.

PhilippWendler:
Thanks for investigating! Would you like to add this to this PR?

JawHawk:
> Thanks for investigating! Would you like to add this to this PR?

I have done the necessary changes. Please review them.

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -210,7 +210,7 @@ const EXTENDED_DISCRETE_COLOR_RANGE = [
  */
 const getURLParameters = (str) => {
   // Split the URL string into parts using "?" as a delimiter
-  const urlParts = (str || decodeURI(document.location.href)).split("?");
+  const urlParts = (str || document.location.href).split("?");
```
PhilippWendler, line 213: I guess this function should get some comment explaining that it handles decoding of the parameters, except for `filter` where it does not do that on purpose to allow the filter handling do it on its own?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -384,13 +398,14 @@ export const decodeFilter = (filterID) => {
...
+  if (splitedArray.length === 2) {
     throw new Error("Invalid filter ID");
   }
   return {
     tool: splitedArray[0],
-    name: splitedArray[1],
-    column: splitedArray[2],
+    name:
+      splitedArray.length > 2 ? splitedArray.slice(1, -1).join("_") : undefined,
+    column: splitedArray.length > 2 ? splitedArray.at(-1) : undefined,
```
PhilippWendler, line 408: `splitedArray.length > 2` should always be true, right (because of the check above)?

And it would be nice to have a comment that explains that `tool` is the first value, `column` the last, and `name` all values in between.

PhilippWendler (commented):
Thanks a lot! Seems to work fine in my tests. I also like that many unit tests are added, this is always good.

I have just some minimal suggestions left, but otherwise it is ready to merge from my point of view. If you agree, could you please also update the pre-built files?

benchexec/tablegenerator/react-table/src/utils/utils.js
JawHawk: The check above is just for `splitedArray.length === 2` . `filterID` has value `"id"` when id filters are applied, so `splitedArray` has length 1 and in this case, name & column should be `undefined` .  `splitedArray` will always have length equal to 1 or >= 3 .

PhilippWendler (approved):

PR #1021, opened 2024-03-22
minor changes to coccinelle tool

justfeelix:
modified coccinelle tool - do not pass the spatch template as a property file anymore but rather in options.


benchexec/tools/coccinelle.py
```diff
@@ -22,20 +22,15 @@ def name(self):
         return "Coccinelle"
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("spatch")
+        return tool_locator.find_executable("_spatch")
```
PhilippWendler, line 25: Is this a typo?

PhilippWendler:
This seems like a backwards-incompatible change - benchmark definitions written for the old tool-info module would not work anymore. Luckily, the coccinelle tool-info module was not part of a BenchExec release anymore, so this is not a problem, we can still change it in any way. But note that in the future (after the next release) such changes would be more difficult. Is there anything else you need to change before we make a release and thus backwards compatibility becomes important?

justfeelix:
fixed the typo, this should be the final version of coccinelle spatch tool definition.

PR #1029, opened 2024-04-02
Documentation: Part I

incaseoftrouble:
* Improve README with usage notes
 * Consistent 80-col + sentence line wrap

this only touches the readme and should be self-contained. I made all parts of the document adhere to the 80 col rule but I can't figure out the exact formatting style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 # BenchExec
+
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Sorry, it seems there was a misunderstanding. I would not have dared to ask you to improve the formatting of the existing documentation when improving the documentation content, and in fact this makes reviewing much harder. Could you please provide the changes without irrelevant reformattings?

Issue #162, opened 2016-11-25
Update test framework


PhilippWendler:
Currently the tests are executed with `nose`, but this framework is dead. We should switch to a new framework, for example `nose2` or `py.test`.

It seems `py.test` has quite useful support for fixtures and parameterized tests, not sure whether `nose2` also supports this.

PhilippWendler:
Furthermore, `setup.py test` is [deprecated](https://setuptools.readthedocs.io/en/latest/userguide/commands.html#test-build-package-and-run-a-unittest-suite) and will be removed in the future. We might just tell people to use the command of the test framework that we decide on, or use [tox](https://tox.readthedocs.io/en/latest/).

PhilippWendler:
Problems I found while investigating:
- We have tests directly in packages (in `__init__.py`), but nose2 and other testing frameworks do not support this.
- We have test modules with names that are not globally unique (e.g., `test_integration`), this also seems problematic for many testing frameworks.
- We have tests below `contrib/`, but `contrib/` is not a proper package (and shouldn't be).

Mihan786Chistie:
Hi, can you explain what we need to do to fix the issue, I am a beginner so I am having trouble understanding.

PhilippWendler:
As a first step, an appropriate framework needs to be chosen. This should be done systematically based on criteria like ease of use, how common the framework is, whether it is well maintained etc. Next, the framework needs to be integrated in BenchExec's project structure and the existing tests need to adjusted where necessary. This might also include restructuring the tests and in general improving their state because as mentioned above, the current test infrastructure might be problematic for some test frameworks.

samadpls:
hey @PhilippWendler 
I'm interested in contributing to help address these problems. I have experience with Python and I'm excited to use my skills to improve the testing framework of the project.

I understand that the current testing framework uses nose, which is no longer maintained. I agree that it would be a good idea to evaluate alternative testing frameworks like nose2 or py.test, and I'd be happy to help with this evaluation process. Additionally, I'd like to contribute to improving the current test infrastructure, such as restructuring tests that are problematic for the new framework and adjusting existing tests where necessary.

Please let me know if there are any specific tasks related to the testing infrastructure that I can work on. I'm excited to be part of the project and contribute to its success.

Also, could you provide me with more details about the project and how I can contact you or the team to get started?

Thank you for considering my interest in contributing.

Naman-Priyadarshi:
[...]

PR #1037, opened 2024-05-06
Draft: Add a Pytest-based Testsuite

schroeding:
This branch introduces a second testsuite, using pytest, in addition to the existing nose-based testsuite. The goal is to migrate the current testsuite to pytest, making it obsolete (see [Issue 162](https://github.com/sosy-lab/benchexec/issues/162)).

Note that, if we want to use pytest features like test function parameterization (i.e. using a pytest decorator instead of having helper methods, which call another test method multiple times with different parameters),
the new testsuite will have a larger number of tests - nose sees e.g. a helper method which calls the actual test with 4 different parameter combinations and assert statements as a single test, pytest will report it as 4 tests.

Short feedback if we want to use pytest parameterization or just translate the current tests as they are with as few changes as possible would be appreciated. For an example how the tests would be rewritten, see e.g. commit [dad90346f320a2447b2dd1d91a15c3a3e3342a18](https://github.com/sosy-lab/benchexec/commit/dad90346f320a2447b2dd1d91a15c3a3e3342a18).

Upside would be that any failing call to a test method, which currently is hidden inside helper methods, would be directly reported by pytest without having to check manually which assertion failed, i.e. tests would be finer. Would also clearly separate the test data from the actual tests. Downside would be the need to rewrite significant parts of the testsuite and that there wouldn't be a 1:1 relation between the current and new testsuite, but a 1:n relation.

### DIFF
--- a/benchexec/pytest_result.py
+++ b/benchexec/pytest_result.py
@@ -1,0 +1,952 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import sys
+import tempfile
+import unittest
+
+from benchexec.result import *  # noqa: F403 @UnusedWildImport everything is tested
+from benchexec.result import (
+    _SCORE_CORRECT_FALSE,
+    _SCORE_CORRECT_TRUE,
+    _SCORE_WRONG_TRUE,
+    _SCORE_WRONG_FALSE,
+)
+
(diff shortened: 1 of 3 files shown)


PhilippWendler:
> Short feedback if we want to use pytest parameterization or just translate the current tests as they are with as few changes as possible would be appreciated.

Your discussion of this question is fully correct. If parameterized tests are now easy, we would like to have their advantages.

I would, however, tend to say this should be a two-step process: First migrate to pytest with the existing pattern of tests, and afterwards improve the tests to be parameterized. This would have the advantage of making the first step less complex and make it easier to compare the number of tests during the migration and to review the changes.

PhilippWendler:
Can you please use a branch in this repository and not a fork? Only this way you get CI that executes the tests.

Issue #880, opened 2022-11-18
Deb package from PPA not installable on Debian 


quentinhaenn:
The debian package is not installing because of a compression tool not supported by dpkg-deb :`.tar.zst` 

This has already been treated for a previous version here : #818 

![image](https://user-images.githubusercontent.com/83473151/202686336-e27cf464-1e6b-4ccd-864b-471618265e79.png)

quentinhaenn:
Yes I do use your PPA, in fact I followed the instruction mentioned [here](https://github.com/sosy-lab/benchexec/blob/main/doc/INSTALL.md). 
This lead me to use those commands :
```bash 
sudo add-apt-repository ppa:sosy-lab/benchmarking
sudo apt install benchexec
̀̀```
And then I got the error mentioned. I also tried with downloading the `.deb` archive as its also mentioned in your install process, but got the same exact error. 

Unfortunately I cannot say if it worked for 3.11, but I managed to install 3.13 correctly somehow.


I cannot help you either to make Launchpad build your package so it works on Debian, sorry.
I

PhilippWendler:
You downloaded [this file](https://github.com/sosy-lab/benchexec/releases/download/3.14/benchexec_3.14-1_all.deb) and it produced a compression error? That would be highly surprising, as I have confirmed manually that it uses the XZ compression, which is supported by Debian, and not ZST. And in fact I can install it just fine on Debian (just tried it out on Debian 11).

On the other hand BenchExec 3.13 from the PPA did use the unsupported ZST. Maybe you have mixed something up?

A final resolution for this will be if [this Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892664) is fixed, or if we find a way to publish PPAs specifically for Debian. Until then, I guess Debian users have to install the BenchExec package manually.

quentinhaenn:
I tried again with your file linked, and it worked, I can't really explain how I have mixed it up, but I definitely downloaded the exact same file and then produced the error,  as I saved the terminal process and the error to check each step.  

Although it produced the error using the PPA for 3.14 and 3.13, I managed to install it manually too.

Thanks for sharing this Debian bug and for answering me, I guess this should be closed now. 
Have a nice day !

PhilippWendler:
Debian Bookworm (current stable) is supposed to support ZST for packages. @quentinhaenn Would you mind testing with the packages from our PPA again, if you run this Debian version?

PhilippWendler:
[...]

PR #1038, opened 2024-05-18
Revert "Update installation instructions for Debian due to #880"

lorenzleutgeb:
According to https://github.com/sosy-lab/benchexec/issues/880#issuecomment-1774639931 Debian supports zstd now, so let's revert be7ee3535689636de49cd5c939b18e4062dfe6f0.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -29,9 +29,9 @@ The following packages are optional but recommended dependencies:
 
 Note that the `table-generator` utility requires only Python and works on all platforms.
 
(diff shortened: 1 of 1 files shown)


PhilippWendler (approved):
There is still a supported Debian version that is incompatible with our PPA, but I guess most users installing BenchExec for the first time on a Debian machine will already have the latest Debian version (released almost a year ago). So I guess we can skip mentioning the old Debian version in our docs.

Thanks for remembering this and submitting the PR!

PR #1040, opened 2024-05-21
Fixed bug on task filter clearance

### LINKED ISSUE #1032: Column filters disabled after clearing Task Filter
Column filters are disabled when Task filter is applied. But even after clearing Task filter, column filters are disabled and shows message `To edit, please clear task filter in the sidebar`. The page needs to be refreshed to enable to column filters again. I think, the expected behavior should be to enable column filters whenever Task Filter is removed. 
Following is the recording of the same.

https://github.com/sosy-lab/benchexec/assets/86140365/0b5eb0d2-fc99-4bbf-8e5e-e860ce32a708

JawHawk:
Fixes #1032 .
- Enables task text when task filter is not applied


benchexec/tablegenerator/react-table/src/components/ReactTable.js
```diff
@@ -641,12 +641,22 @@ const Table = (props) => {
...
+      if (id === "id" && !isNil(values)) {
+        taskFilterExists = true;
+        setDisableTaskText(true);
       }
     }
+
+    // Enable task text when task filter is not applied
+    if (!taskFilterExists) {
+      setDisableTaskText(false);
+    }
```
PhilippWendler, line 658: Wouldn't it work if we just call `setDisableTaskText` once in the end (after the loop) with the result of whether a task-filter exists? I guess this would be easier to understand when reading the code.

We could also maybe use functional style with `some()` instead of a manual loop?

PhilippWendler (commented):
Thanks, this looks like  a plausible fix.

I have just some code-style question.

And could you please also commit the rebuilt JS file?

benchexec/tablegenerator/react-table/src/components/ReactTable.js
JawHawk: Yes, I have replaced the loop with `some()`.

PhilippWendler:
Thanks! It seems something in your environment creates a different `vendors.min.js` file than what CI produces. Maybe try running `npm ci` to install the dependencies before building?

PhilippWendler:
For this PR, I have recreated the file and will merge it, no need to do anything.

PhilippWendler (approved):

Issue #1041, opened 2024-05-23
BenchExec not working after upgrading from Ubuntu 23.10 to 24.04


younghojan:
I recently upgraded my system from Ubuntu 23.10 to 24.04, and BenchExec stopped working. I installed BenchExec from the PPA following the instructions provided [here](https://github.com/sosy-lab/benchexec/blob/main/doc/INSTALL.md). I believe I have configured everything correctly, as `python3 -m benchexec.check_cgroups` did not produce any output.

However, when I run `containerexec --debug /usr/bin/ls`, I encounter the following error:

```
haoranyang@haoranyang-fudan:~/benchexec-younghojan$ containerexec --debug /usr/bin/ls
2024-05-23 20:02:47 - DEBUG - This is containerexec 3.21.
2024-05-23 20:02:47 - INFO - Starting command ls
2024-05-23 20:02:47 - DEBUG - Available Cgroups: {}
2024-05-23 20:02:47 - DEBUG - Starting process.
2024-05-23 20:02:47 - DEBUG - Parent: child process of RunExecutor with PID 12007 started.
2024-05-23 20:02:47 - DEBUG - Child: child process of RunExecutor with PID 12007 started
2024-05-23 20:02:47 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
2024-05-23 20:02:47 - CRITICAL - Failed to configure container with operation 'fcntl.ioctl(sock, _SIOCSIFFLAGS, ifreq)': [Errno 1] Operation not permitted
```

BenchExec then exits with an error. This seems to be caused by Ubuntu 24.04 now restricting user namespaces by default.

Referencing the [Ubuntu 24.04 LTS Release Notes](https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#heading--known-issues:~:text=Unprivileged%20user%20namespace%20restrictions), it states:

> In combination with the apparmor package, the Ubuntu kernel now restricts the use of unprivileged user namespaces.

I disabled this restriction by executing:

```
echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
```

However, this solution is temporary and will be reset upon reboot. A persistent setting can be applied by creating a new file `/etc/sysctl.d/60-apparmor-namespace.conf` with the following contents:

```
kernel.apparmor_restrict_unprivileged_userns=0
```

[...]

PR #1042, opened 2024-05-24
fix: Add error message output to respond to the impact of Ubuntu 24.04

younghojan:
- Updated `container.py` and `containerexecutor.py` to detect whether BenchExec is being used with container mode enabled on Ubuntu 24.04 and the use of unprivileged user namespaces is restricted in system configuration, output the corresponding error message if so.
- Updated the relevant documentation.

Part of #1041

### DIFF
--- a/benchexec/container.py
+++ b/benchexec/container.py
@@ -115,6 +115,14 @@ NATIVE_CLONE_CALLBACK_SUPPORTED = (
 )
 """Whether we use generated native code for clone or an unsafe Python fallback"""
 
+_ERROR_MSG_USER_NS_RESTRICTION = (
+    "Unprivileged user namespaces forbidden on this system, please "
+    "enable them with 'sysctl -w kernel.apparmor_restrict_unprivileged_userns=0'. "
+    "Ubuntu disables them by default since 24.04, refer to "
+    "https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces "
+    "for more information."
+)
+
 
 @contextlib.contextmanager
 def allocate_stack(size=DEFAULT_STACK_SIZE):
--- a/benchexec/container.py
+++ b/benchexec/container.py
@@ -44,6 +44,7 @@ __all__ = [
     "CONTAINER_GID",
     "CONTAINER_HOME",
     "CONTAINER_HOSTNAME",
+    "check_apparmor_userns_restriction",
 ]
 
 
@@ -122,6 +123,20 @@ _ERROR_MSG_USER_NS_RESTRICTION = (
     "https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces "
     "for more information."
 )
+
+
+def check_apparmor_userns_restriction(error: OSError):
+    """Check whether the passed OSError was likely caused by Ubuntu's AppArmor-based
+    restriction of user namespaces."""
+    return (
+        error.errno
+        in [
+            errno.EPERM,
+            errno.EACCES,
+        ]
+        and util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns")
+        == "1"
+    )
 
 
 @contextlib.contextmanager
--- a/benchexec/containerexecutor.py
+++ b/benchexec/containerexecutor.py
@@ -754,12 +754,7 @@ class ContainerExecutor(baseexecutor.BaseExecutor):
                         traceback.extract_tb(e.__traceback__, limit=-1)[0].line,
                         e,
                     )
-                    if util.try_read_file(
-                        "/proc/sys/kernel/apparmor_restrict_unprivileged_userns"
-                    ) == "1" and e.errno in [
-                        errno.EPERM,
-                        errno.EACCES,
-                    ]:
+                    if container.check_apparmor_userns_restriction(e):
                         logging.critical(container._ERROR_MSG_USER_NS_RESTRICTION)
                     return CHILD_OSERROR
 
--- a/benchexec/containerexecutor.py
+++ b/benchexec/containerexecutor.py
@@ -754,6 +754,13 @@ class ContainerExecutor(baseexecutor.BaseExecutor):
                         traceback.extract_tb(e.__traceback__, limit=-1)[0].line,
                         e,
                     )
+                    if util.try_read_file(
+                        "/proc/sys/kernel/apparmor_restrict_unprivileged_userns"
+                    ) == "1" and e.errno in [
+                        errno.EPERM,
+                        errno.EACCES,
+                    ]:
+                        logging.critical(container._ERROR_MSG_USER_NS_RESTRICTION)
                     return CHILD_OSERROR
 
                 try:
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -124,6 +124,14 @@ def _init_container_and_load_tool(tool_module, *args, **kwargs):
     try:
         _init_container(*args, **kwargs)
     except OSError as e:
+        if (
+            util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns")
+            == "1"
+        ) and e.errno in [
+            errno.EPERM,
+            errno.EACCES,
+        ]:
+            raise BenchExecException(container._ERROR_MSG_USER_NS_RESTRICTION)
         raise BenchExecException(f"Failed to configure container: {e}")
     return _load_tool(tool_module)
 
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -124,13 +124,7 @@ def _init_container_and_load_tool(tool_module, *args, **kwargs):
     try:
         _init_container(*args, **kwargs)
     except OSError as e:
-        if (
-            util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns")
-            == "1"
-        ) and e.errno in [
-            errno.EPERM,
-            errno.EACCES,
-        ]:
+        if container.check_apparmor_userns_restriction(e):
             raise BenchExecException(container._ERROR_MSG_USER_NS_RESTRICTION)
         raise BenchExecException(f"Failed to configure container: {e}")
     return _load_tool(tool_module)
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -169,6 +169,10 @@ that are not usable on all distributions by default:
   On CentOS it can be necessary to enable this feature with
   `sudo sysctl -w user.max_user_namespaces=10000` or a respective entry
   in `/etc/sysctl.conf` (the exact value is not important).
+  On Ubuntu 24.04 (or newer versions) it can be necessary to enable this feature with
+  `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0` or a respective entry
+  in `/etc/sysctl.conf`.
+  
 
 - **Unprivileged Overlay Filesystem**: This is only available since Linux 5.11
   (kernel option `CONFIG_OVERLAY_FS`),
--- a/doc/container.md
+++ b/doc/container.md
@@ -211,7 +211,7 @@ in a container with `containerexec` than using `benchexec` or `runexec`.
 
 #### `Cannot execute ...: Unprivileged user namespaces forbidden on this system...`
 Unprivileged user namespaces are forbidden on your system
-(this is the default on some distributions like Debian, Arch Linux, and CentOS).
+(this is the default on some distributions like Debian, Arch Linux, CentOS, and Ubuntu since 24.04).
 Please check the [system requirements](INSTALL.md#kernel-requirements)
 how to enable them.
 


benchexec/container.py
```diff
@@ -335,13 +337,30 @@ def setup_user_mapping(
...
         gid_map = f"{gid} {parent_gid} 1"
         util.write_file(gid_map, proc_child, "gid_map")
     except OSError as e:
         logging.warning("Creating GID mapping into container failed: %s", e)
+        exception_occurred = True
+
+    # Ubuntu 24.04 restricts user namespaces, output error message here
+    if (
+        exception_occurred
+        and util.check_ubuntu_version() == "24.04"
```
PhilippWendler, line 353: I think we can skip this check. Future Ubuntu versions will likely behave the same, and other distributions might also adopt the behavior from Ubuntu. So just check whether `apparmor_restrict_unprivileged_userns` is enabled, and the error message can then tell people do disable it, no matter what the distribution is.

younghojan: Agree with that. btw, is there a way to locally deploy these CI checks? I want to be able to run them before pushing to avoid excessive commits :)

PhilippWendler: Yes, you can basically run everything locally if you install the tools like Black etc., have a look in `.gitlab-ci.yml` for what exactly gets executed.

Note that running the tests in all combinations is more tricky. The CI tests with cgroups v1, if you test locally you will probably have cgroups v2. But we want to have it work with all versions anyway. :-)

younghojan: Please review if there are any other modifications needed. I have already skipped the Ubuntu version check.

benchexec/container.py
```diff
@@ -335,13 +337,30 @@ def setup_user_mapping(
...
+
+    # Ubuntu 24.04 (and possibly later versions) restricts user namespaces,
+    # output error message here
+    if (
+        exception_occurred
+        and util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns")
+        == "1"
+    ):
+        logging.warning(
+            "Ubuntu 24.04 restircts user namespaces, "
```
PhilippWendler, line 358:
```suggestion
            "Ubuntu 24.04 restricts user namespaces, "
```
(also in other file)

benchexec/containerexecutor.py
```diff
@@ -713,11 +713,28 @@ def child():
...
+                                )
+                                == "1"
+                            ):
+                                logging.warning(
+                                    "Ubuntu 24.04 restircts user namespaces, "
+                                    "preventing changing hostname in container, "
+                                    "real hostname will leak into the container. "
+                                    "Please try 'echo 0 | sudo tee "
+                                    "/proc/sys/kernel/apparmor_restrict_unprivileged_userns' "
+                                    "as a temporary workaround, or disable container mode."
```
PhilippWendler, line 730: In this case, we do not want to recommend to disable container mode. If we are in a container where we cannot change the hostname, it is not ideal because of the visible real hostname. But disabling container mode would mean that there is no isolation at all, which is clearly worse than a container with just the unwanted hostname. So this recommendation is like [throwing out the child with the bathwater](https://en.wikipedia.org/wiki/Don%27t_throw_the_baby_out_with_the_bathwater).

benchexec/containerexecutor.py
```diff
@@ -713,11 +713,28 @@ def child():
                         try:
                             socket.sethostname(container.CONTAINER_HOSTNAME)
                         except PermissionError:
-                            logging.warning(
-                                "Changing hostname in container prevented "
-                                "by system configuration, "
-                                "real hostname will leak into the container."
-                            )
+                            # Ubuntu 24.04 (and possibly later versions) restricts user namespaces,
+                            # output error message here
```
PhilippWendler, line 717:
```suggestion
```
This comment doesn't really add anything non-obvious.

benchexec/containerexecutor.py
PhilippWendler, line 730: For such long and deeply indented log messages it can be useful to define them in a module-level constant, where the usable parts of lines are longer and the message is thus easier to read.

PhilippWendler (commented):
A general question:

Does the container mode work at all in Ubuntu 24.04 with the default settings?
In #1041 it fails due to network configuration, does it work with `--network-access`?

If it does not work at all due to further problems, we don't want to print several warnings just to fail with a different error in the end. Instead, it would be easier for the user to understand if we just detect the situation once the first problem occurs and abort immediately with an appropriate error message that includes the explanation.

If container mode works in principle, but several features are missing, then it would probably be appropriate to print a warning with the explanation and continue. But then we should also handle the problem with the networking.

younghojan:
> A general question:
> 
> Does the container mode work at all in Ubuntu 24.04 with the default settings? In #1041 it fails due to network configuration, does it work with `--network-access`?
> 
> If it does not work at all due to further problems, we don't want to print several warnings just to fail with a different error in the end. Instead, it would be easier for the user to understand if we just detect the situation once the first problem occurs and abort immediately with an appropriate error message that includes the explanation.
> 
> If container mode works in principle, but several features are missing, then it would probably be appropriate to print a warning with the explanation and continue. But then we should also handle the problem with the networking.



After adding `--network-access`, the function `activate_network_interface()` will no longer encounter an error. An error will subsequently occur in `child()`->`_setup_container_filesystem()`->`libc.mount(None, temp_dir, b"tmpfs", 0, tmpfs_opts)`.
```
(benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ containerexec --network-access --debug /usr/bin/ls2024-05-27 15:36:22 - DEBUG - This is containerexec 3.22-dev.
2024-05-27 15:36:22 - WARNING - The container configuration disables DNS, host lookups will fail despite --network-access. Consider using --keep-system-config.
2024-05-27 15:36:22 - INFO - Starting command /usr/bin/ls
2024-05-27 15:36:22 - DEBUG - Available Cgroups: {}
2024-05-27 15:36:22 - DEBUG - Starting process.
2024-05-27 15:36:22 - DEBUG - Parent: child process of RunExecutor with PID 355923 started.
2024-05-27 15:36:22 - DEBUG - Child: child process of RunExecutor with PID 355923 started
2024-05-27 15:36:22 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
2024-05-27 15:36:22 - CRITICAL - Failed to configure container with operation 'raise OSError(errno, msg)': [Errno 13] mount(None, b'/tmp/BenchExec_run_kzzs3je7', b'tmpfs', 0, b'size=100%') failed: Permission denied
2024-05-27 15:36:22 - DEBUG - Waiting for process /usr/bin/ls with pid 355923
2024-05-27 15:36:22 - DEBUG - Parent: child process of RunExecutor with PID 355923 terminated with return value 128.
2024-05-27 15:36:22 - DEBUG - Process terminated, exit code 0.
2024-05-27 15:36:22 - DEBUG - Cleaning up temporary directory.
2024-05-27 15:36:22 - ERROR - execution in container failed, check log for details
  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 952, in _start_execution_in_container
    grandchild_pid = int(os.read(from_grandchild, 10))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: b''

During handling of the above exception, another exception occurred:

  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 297, in main
    result = executor.execute_run(
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 466, in execute_run
    pid, result_fn = self._start_execution(
                     ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 535, in _start_execution
    result = self._start_execution_in_container(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 956, in _start_execution_in_container
    check_child_exit_code()
  File "/home/haoranyang/benchexec/benchexec/containerexecutor.py", line 909, in check_child_exit_code
    raise BenchExecException(
benchexec.BenchExecException: execution in container failed, check log for details
Cannot execute /usr/bin/ls: execution in container failed, check log for details.
```
So, should we let child `return CHILD_OSERROR` when `child()`->`socket.sethostname()` is called and failed?

benchexec/container.py
younghojan: I will fix this typo in a later commit. Thanks

benchexec/containerexecutor.py
younghojan: That's an interesting analogy, I understand.: ) I will delete this recommendation.

PhilippWendler:
> > Does the container mode work at all in Ubuntu 24.04 with the default settings? In #1041 it fails due to network configuration, does it work with `--network-access`?
> 
> After adding `--network-access`, the function `activate_network_interface()` will no longer encounter an error. An error will subsequently occur in `child()`->`_setup_container_filesystem()`->`libc.mount(None, temp_dir, b"tmpfs", 0, tmpfs_opts)`.

Thanks for trying out! So this means that we cannot mount anything in the container, but mounting stuff is a hard requirement for our container mode. So there is no way to make container mode work on Ubuntu 24.04 except by letting the user change the system config.

> So, should we let child `return CHILD_OSERROR` when `child()`->`socket.sethostname()` is called and failed?

I am not 100% sure how to best implement this. Let's think about this together. I see the following goals:
- Detection must work no matter which options the user has, so for example also with `--keep-system-config`.
- On other systems (non-Ubuntu) BenchExec should not abort if changing the hostname fails, because it might be able to continue. We can distinguish such cases with the check for the current value of the sysctl setting (what you implemented already).
- We only want to do something special once we encounter an actual error. The alternative would be to check for the sysctl in the very beginning and simply give up if it has the wrong value, but that is not so robust because it could trigger in unwanted situations, for example if Ubuntu changes something in the future. Thus it is better to always try setting up the container.
- It is better if there is only one error message for the user with an explanation and the recommendation, not a bunch of technical warnings. But if it is complicated to avoid this, it is not a hard requirement.
- We can to keep the implementation maintainable without lots of special cases.

Could you help thinking about how to implement this?
I would like if in the end we would raise a `BenchExecException` with the error message in the main process, like we do for similar cases. However, I see that we first have to do something in the child. Maybe add some other special return code for the child?

younghojan:
I have pushed a new version that meets a part of our vision.
>Detection must work no matter which options the user has

Having a global variable in `container.py` might satisfy this:
```
USER_NS_RESTRICTION = (
    util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns") == "1"
)
```

>We only want to do something special once we encounter an actual error

After catching `OSError` in `child()`, check `container.USER_NS_RESTRICTION` and print a warning before `return CHILD_OSERROR`. This is quite rough, as I'm not sure if the `OSError` is necessarily caused by the restriction of unprivileged user namespaces in this situation.

>It is better if there is only one error message for the user with an explanation and the recommendation

Now, if `sethostname` fails, no information regarding the restriction of user-ns will be printed. If we output error messages when `sethostname` fails without terminating BenchExec, the creation of container goes on. Later when the network setup or the mount tmpfs error happens, the same error messages will be printed again. Therefore, I decided to discard the handling of `sethostname` failure and only print special info in the event of a critical error.

PhilippWendler:
> > Detection must work no matter which options the user has
> 
> Having a global variable in `container.py` might satisfy this:
> 
> ```
> USER_NS_RESTRICTION = (
>     util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns") == "1"
> )
> ```

Ah sorry, this is a misunderstanding because I didn't write it clearly.
Whether the check is implemented on module-import time or later on doesn't really matter here, we can keep the standard way. What I wanted to say is that for example we shouldn't implement the check *only* in the `sethostname` part of the code, because that is optional.

> > We only want to do something special once we encounter an actual error
> 
> After catching `OSError` in `child()`, check `container.USER_NS_RESTRICTION` and print a warning before `return CHILD_OSERROR`. This is quite rough, as I'm not sure if the `OSError` is necessarily caused by the restriction of üunprivileged user namespaces in this situation.

Yes, `OSError` in general can have many causes. That's why we need to have this additional `try_read_file` check. We can also check the errno value of the `OSError` whether it matches our expectation.

> > It is better if there is only one error message for the user with an explanation and the recommendation
> 
> Now, if `sethostname` fails, no information regarding the restriction of user-ns will be printed. If we output error messages when `sethostname` fails without terminating BenchExec, the creation of container goes on. Later when the network setup or the mount tmpfs error happens, the same error messages will be printed again. Therefore, I decided to discard the handling of `sethostname` failure and only print special info in the event of a critical error.

I see, so we get all of these warnings. Could you post an example of a such a failing run of `runexec`? I wonder whether it is worth the effort to complicate the code by adding more checks to cases like `sethostname`.

younghojan:
> Whether the check is implemented on module-import time or later on doesn't really matter here, we can keep the standard way.

Oh I see. However, I think it would be cooler to place this check at the import time, similar to how `NATIVE_CLONE_CALLBACK_SUPPORTED` is checked. Since checking unprivileged user-NS is necessary, doing it this way would not incur any additional overhead.

> We can also check the errno value of the OSError whether it matches our expectation.

I think maybe we can do it this way:
```
                    if container.USER_NS_RESTRICTION and e.errno in [
                        errno.EPERM,  # container.activate_network_interface() failed
                        errno.EACCES,  # _setup_container_filesystem failed
                    ]:
                        logging.warning(
                            "Ubuntu 24.04 restircts unprivileged user namespaces,"
                            " please try 'echo 0 | sudo tee /proc/sys/kernel/"
                            "apparmor_restrict_unprivileged_userns' as a temporary workaround."
                        )
```

> Could you post an example of a such a failing run of `runexec`?

Yes, I tested `runexec` with and without network access. This looks all "right", and from the results, it appears to be what we want.
```
(benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ runexec /usr/bin/ls
2024-05-28 09:10:54 - INFO - Starting command /usr/bin/ls
2024-05-28 09:10:54 - INFO - Writing output to output.log
2024-05-28 09:10:54 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
2024-05-28 09:10:54 - CRITICAL - Failed to configure container with operation 'fcntl.ioctl(sock, _SIOCSIFFLAGS, ifreq)': [Errno 1] Operation not permitted
2024-05-28 09:10:54 - WARNING - Ubuntu 24.04 restircts unprivileged user namespaces, please try 'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' as a temporary workaround.
2024-05-28 09:10:54 - CRITICAL - Cannot execute '/usr/bin/ls': execution in container failed, check log for details.
terminationreason=failed
```
```
(benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ runexec --network-access /usr/bin/ls
2024-05-28 09:12:20 - WARNING - The container configuration disables DNS, host lookups will fail despite --network-access. Consider using --keep-system-config.
2024-05-28 09:12:20 - INFO - Starting command /usr/bin/ls
2024-05-28 09:12:20 - INFO - Writing output to output.log
2024-05-28 09:12:20 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
2024-05-28 09:12:20 - CRITICAL - Failed to configure container with operation 'raise OSError(errno, msg)': [Errno 13] mount(None, b'/tmp/BenchExec_run_5hgyaihz', b'tmpfs', 0, b'size=100%') failed: Permission denied
2024-05-28 09:12:20 - WARNING - Ubuntu 24.04 restircts unprivileged user namespaces, please try 'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' as a temporary workaround.
2024-05-28 09:12:20 - CRITICAL - Cannot execute '/usr/bin/ls': execution in container failed, check log for details.
terminationreason=failed
```

PhilippWendler:
> > Whether the check is implemented on module-import time or later on doesn't really matter here, we can keep the standard way.
> 
> Oh I see. However, I think it would be cooler to place this check at the import time, similar to how `NATIVE_CLONE_CALLBACK_SUPPORTED` is checked. Since checking unprivileged user-NS is necessary, doing it this way would not incur any additional overhead.

Well, we do not need to check for it if everythings works. But overhead is not my concern, I am mostly wondering whether this has the potential to break something. For example, we had problems in the past where modules did stuff during import time, and this caused problems for situations where the module was not actually used but still imported, e.g., when using other parts of BenchExec on Windows, or in tests, or by type checkers. The use of `try_read_file` should make it robust, but as long as there is no real reason to use the global variable, I prefer to not risk it.

> > We can also check the errno value of the OSError whether it matches our expectation.
> 
> I think maybe we can do it this way:
> 
> ```
>                     if container.USER_NS_RESTRICTION and e.errno in [
>                         errno.EPERM,  # container.activate_network_interface() failed
>                         errno.EACCES,  # _setup_container_filesystem failed
>                     ]:
>                         logging.warning(
>                             "Ubuntu 24.04 restircts unprivileged user namespaces,"
>                             " please try 'echo 0 | sudo tee /proc/sys/kernel/"
>                             "apparmor_restrict_unprivileged_userns' as a temporary workaround."
>                         )
> ```

Like this, yes. But the comments are not correct, because it is not guaranteed that `activate_network_interface` always produces `EPERM` and `_setup_container_filesystem` always produces `EACCES`. So this can be misleading and it is better to remove the comments.

> > Could you post an example of a such a failing run of `runexec`?
> 
> Yes, I tested `runexec` with and without network access. This looks all "right", and from the results, it appears to be what we want.
> 
> ```
> (benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ runexec /usr/bin/ls
> 2024-05-28 09:10:54 - INFO - Starting command /usr/bin/ls
> 2024-05-28 09:10:54 - INFO - Writing output to output.log
> 2024-05-28 09:10:54 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
> 2024-05-28 09:10:54 - CRITICAL - Failed to configure container with operation 'fcntl.ioctl(sock, _SIOCSIFFLAGS, ifreq)': [Errno 1] Operation not permitted
> 2024-05-28 09:10:54 - WARNING - Ubuntu 24.04 restircts unprivileged user namespaces, please try 'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' as a temporary workaround.
> 2024-05-28 09:10:54 - CRITICAL - Cannot execute '/usr/bin/ls': execution in container failed, check log for details.
> terminationreason=failed
> ```

Hm, I find the explanation quite hidden. It is in between other log messages, and it is even a warning where the rest is marked critical. Ideal would be a separate error message, and I thought we have that in similar cases. But I realize now that maybe this was only done for `benchexec` and not for `runexec`. Could you please also try with `benchexec`?

For comparison, consider for example the case if BenchExec cannot create a cgroup because pystemd is missing or so.

younghojan:
>  > Oh I see. However, I think it would be cooler to place this check at the import time, similar to how `NATIVE_CLONE_CALLBACK_SUPPORTED` is checked. Since checking unprivileged user-NS is necessary, doing it this way would not incur any additional overhead.
>
> Well, we do not need to check for it if everythings works. But overhead is not my concern, I am mostly wondering whether this has the potential to break something. For example, we had problems in the past where modules did stuff during import time, and this caused problems for situations where the module was not actually used but still imported, e.g., when using other parts of BenchExec on Windows, or in tests, or by type checkers. The use of `try_read_file should` make it robust, but as long as there is no real reason to use the global variable, I prefer to not risk it.

That's correct. Using such a global variable does indeed carry unpredictable risks. But should we wrap this check into a function and perhaps place it in `util.py` and place the module-constant error message in `container.py`?
According to BenchExec's execution process, when using container mode, due to the kernel's restrictions on unprivileged user namespaces, BenchExec will either abort at `activate_network_interface()` or at `_setup_container_filesystem()`. Can we temporarily add checks only at these two points? Because even if the restriction on unprivileged user namespaces affects other steps in the container setup, BenchExec's execution flow will not reach those steps and will terminate early.

As I mentioned before, if an error message is printed at `sethostname` without terminating BenchExec, the same error message will be printed again when `activate_network_interface` is executed, which is redundant. Perhaps we can adjust the execution flow. Currently, it is `sethostname`->`activate_network_interface`->`_setup_container_filesystem`, which results in two repetitive warnings being printed (one at `sethostname` and one at `activate_network_interface`). Can we change it to `activate_network_interface`->`_setup_container_filesystem`->`sethostname`? This way, BenchExec would print the error message and abort at `activate_network_interface`, preventing duplicate warnings.

Additionally, I noticed that in `containerized_tool._init_container`, the execution order is `_setup_container_filesystem`->`sethostname`->`activate_network_interface`, which is different from that in `ContainerExecutor`. The execution order of these three operations should not affect the container setup.

> Like this, yes. But the comments are not correct, because it is not guaranteed that` activate_network_interface` always produces `EPERM` and `_setup_container_filesystem` always produces `EACCES`. So this can be misleading and it is better to remove the comments.

I'll remove them in later commits.

> Hm, I find the explanation quite hidden. It is in between other log messages, and it is even a warning where the rest is marked critical. Ideal would be a separate error message, and I thought we have that in similar cases. But I realize now that maybe this was only done for `benchexec` and not for `runexec`. Could you please also try with `benchexec`?

In fact, after installing `coloredlogs`, the `warning`/`critical` messages are color-coded and not that difficult to distinguish. However, these fatal errors (which prevent the container from being created) are actually caused by Ubuntu's restrictions on user namespaces. We could also change the warnings here to critical.

I've tested `benchexec`, it failed when mounting tmpfs:
```
(benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ benchexec doc/benchmark-example-rand.xml --tasks "XML files" --limitCores 1 --timelimit 10s --numOfThreads 4
2024-05-29 13:51:21 - WARNING - Creating UID mapping into container failed: [Errno 1] Operation not permitted
2024-05-29 13:51:21 - WARNING - Could not write to setgroups file in /proc: [Errno 13] Permission denied: '/proc/4966/setgroups'
2024-05-29 13:51:21 - WARNING - Creating GID mapping into container failed: [Errno 1] Operation not permitted
2024-05-29 13:51:21 - WARNING - Ubuntu 24.04 restircts unprivileged user namespaces, please try 'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' as a temporary workaround.
Error: Failed to configure container: [Errno 13] mount(None, b'/tmp/Benchexec_tool_info_container_obb1i6y4', b'tmpfs', 0, b'size=100%') failed: Permission denied
```

> For comparison, consider for example the case if BenchExec cannot create a cgroup because pystemd is missing or so.

I tested `runexec` without `pystemd`:
```
(benchexec-venv) haoranyang@haoranyang-fudan:~/benchexec$ runexec /usr/bin/ls
2024-05-29 14:01:10 - WARNING - Cgroup subsystem cpu is not available. Please make sure it is supported by your kernel and available.
2024-05-29 14:01:10 - WARNING - Cannot measure CPU time without cpuacct cgroup.
2024-05-29 14:01:10 - WARNING - Cgroup subsystem freeze is not available. Please make sure it is supported by your kernel and available.
2024-05-29 14:01:10 - WARNING - Cgroup subsystem memory is not available. Please make sure it is supported by your kernel and available.
2024-05-29 14:01:10 - WARNING - Cannot measure memory consumption without memory cgroup.
2024-05-29 14:01:10 - INFO - Starting command /usr/bin/ls
2024-05-29 14:01:10 - INFO - Writing output to output.log
2024-05-29 14:01:10 - WARNING - Changing hostname in container prevented by system configuration, real hostname will leak into the container.
2024-05-29 14:01:10 - CRITICAL - Failed to configure container with operation 'fcntl.ioctl(sock, _SIOCSIFFLAGS, ifreq)': [Errno 1] Operation not permitted
2024-05-29 14:01:10 - WARNING - Ubuntu 24.04 restircts unprivileged user namespaces, please try 'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' as a temporary workaround.
2024-05-29 14:01:10 - CRITICAL - Cannot execute '/usr/bin/ls': execution in container failed, check log for details.
terminationreason=failed
```

When handling the missing pystemd case, the warning messages are somewhat repetitive, such as `Please make sure it is supported by your kernel and available`.

younghojan:
> But why can't we handle this more generally, such that if there is some permission error and the problematic sysctl is set, then we terminate BenchExec and show a final error message. 

In previous discussions, you have also mentioned this idea. However, implementing it seems to require quite a few changes, including adding other special return codes for the child. 

Errors caused by the restriction of user-ns are caught as exceptions within the `child()` function, followed by printing the error message, and the `child()` function returns the error code `CHILD_OSERROR`. In the parent function, this error code is parsed by the `check_child_exit_code()` function, which throws a `BenchExecException` based on the type of error code. This `BenchExecException` is caught in `containerexecutor.main`, which then calls `sys.exit()` to terminate the entire containerexec program.

The problem here is evident: we cannot delay the output of the error message until `containerexecutor.main` because the errno of `OSError` has already been lost. Furthermore, in `parent->check_child_exit_code()`, it is assumed that all error messages have been recorded within the child, making it impossible to determine outside of the child whether the containerexec error was caused by the problematic sysctl set.

```python
# containerexecutor.py > ContainerExecutor > _start_execution_in_container > check_chile_exit_code
# Line 914 - 919

                        if child_exitcode.value == CHILD_OSERROR:
                            # This was an OSError in the child,
                            # details were already logged
                            raise BenchExecException(
                                "execution in container failed, check log for details"
                            )
```

I am now feeling a bit confused. Should we continue to implement the same way as in the previous commits, where we only check if the problematic sysctl is set and print an error message when errors occur in `activate_network_interface` and `_setup_container_filesystem`? Or should we consider other implementation methods to achieve the desired effect with minimal changes to the code structure?

PhilippWendler:
I see. Thanks for digging through this and the summary! Might be somewhat awkward in the code to handle this specific case so differently. So now I tend to say we can proceed as in the previous commits.

benchexec/containerexecutor.py
```diff
@@ -754,6 +754,13 @@ def child():
                         traceback.extract_tb(e.__traceback__, limit=-1)[0].line,
                         e,
                     )
+                    if util.try_read_file(
+                        "/proc/sys/kernel/apparmor_restrict_unprivileged_userns"
+                    ) == "1" and e.errno in [
+                        errno.EPERM,
+                        errno.EACCES,
+                    ]:
+                        logging.critical(container._ERROR_MSG_USER_NS_RESTRICTION)
```
PhilippWendler, line 763: We could leave a comment with a link to https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces here.

doc/container.md
```diff
@@ -266,6 +266,26 @@ with several versions of the Linux kernel, including at least kernel versions up
...
+- `Failed to configure container with operation 'fcntl.ioctl(sock, _SIOCSIFFLAGS, ifreq)': [Errno 1] Operation not permitted`
+-  `Failed to configure container with operation 'raise OSError(errno, msg)': [Errno 13] mount(None, b'/tmp/BenchExec_run_aiwscwd2', b'tmpfs', 0, b'size=100%') failed: Permission denied`
+- `Creating UID mapping into container failed: [Errno 1] Operation not permitted`
+- `Could not write to setgroups file in /proc: [Errno 13] Permission denied`
+- `Could not write to setgroups file in /proc: [Errno 13] Permission denied`
+- `Changing hostname in container prevented by system configuration, real hostname will leak into the container.`
+
+It seems that you are using BenchExec on Ubuntu 24.04 (or newer versions), which restricts the use of unprivileged user namespaces, making it impossible to change the hostname, create GID/UID mappings, or do some other operations to set up a container. You can temporarily disable AppArmor's restriction on unprivileged user namespaces by executing 
+```
+echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
```
PhilippWendler, line 282: Please apply the same changes as above (version number, link to Ubuntu announcement).

I would also be great if you could adopt to our formatting style for continuous text. We hadn't documented this so far, but I just described it in https://github.com/sosy-lab/benchexec/issues/999#issuecomment-2160508898.

doc/container.md
```diff
@@ -266,6 +266,26 @@ with several versions of the Linux kernel, including at least kernel versions up
...
+```
+This setting will be reset upon reboot. To apply a persistent setting, create a new file at `/etc/sysctl.d/60-apparmor-namespace.conf` with the following contents:
+```
+kernel.apparmor_restrict_unprivileged_userns=0
+```
```
PhilippWendler, line 287: We can move this to `INSTALL.md`, where we already explain how to do this for similar cases. And then just link to that part from here, like for the other problems.

benchexec/container.py
```diff
@@ -115,6 +115,12 @@
 )
 """Whether we use generated native code for clone or an unsafe Python fallback"""
 
+_ERROR_MSG_USER_NS_RESTRICTION = (
+    "Ubuntu 24.04 restircts unprivileged user namespaces, please try "
+    "'echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns' "
```
PhilippWendler, line 120: The standard way of changing sysctls is with `sysctl -w ...`.
This would be `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0` here.

We could also use the same style of error message as we have for similar cases, for example we already have this:
> Unprivileged user namespaces forbidden on this system, please enable them with 'sysctl -w kernel.unprivileged_userns_clone=1' or disable container mode

doc/container.md
```diff
@@ -266,6 +266,26 @@ with several versions of the Linux kernel, including at least kernel versions up
 If a kernel upgrade does not help, please use a different access mode for NFS-mounted directories,
 such as `--hidden-dir` or `--read-only-dir`.
 
+#### Problems due to default user namespaces restrictions on Ubuntu 24.04 (and possibly later versions)
```
PhilippWendler, line 269: In this section, we try to show the error message in the heading, such that it is easy to find. Could you please also do this?

benchexec/containerized_tool.py
```diff
@@ -124,6 +124,14 @@ def _init_container_and_load_tool(tool_module, *args, **kwargs):
...
         _init_container(*args, **kwargs)
     except OSError as e:
+        if (
+            util.try_read_file("/proc/sys/kernel/apparmor_restrict_unprivileged_userns")
+            == "1"
+        ) and e.errno in [
+            errno.EPERM,
+            errno.EACCES,
+        ]:
+            logging.critical(container._ERROR_MSG_USER_NS_RESTRICTION)
```
PhilippWendler, line 134: Here we could instead easily raise a `BenchExecException` with the error message, right? Then at least for users of `benchexec` we would get the nice way of showing the error message, only for `runexec` and `containerexec` not.

PhilippWendler (commented):
Thanks! Would you be willing to do some polishing?

benchexec/container.py
```diff
@@ -115,6 +115,12 @@
 )
 """Whether we use generated native code for clone or an unsafe Python fallback"""
 
+_ERROR_MSG_USER_NS_RESTRICTION = (
+    "Ubuntu 24.04 restircts unprivileged user namespaces, please try "
```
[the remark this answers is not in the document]
younghojan: I reviewed the release notes for Ubuntu 23.10 and found that this version introduced the restriction, but it was not enabled by default. It became the default in version 24.04. Before updating my machine to 24.04, I was using 23.10 and did not encounter any errors caused by the restriction on user namespaces. To confirm when the restriction was introduced as a default option in the Ubuntu kernel, the best approach would be to check on a machine running Ubuntu 23.10, but this is somewhat inconvenient. However, the statement "Ubuntu restricts unprivileged user namespaces since version 23.10" is not incorrect, I will amend this error message.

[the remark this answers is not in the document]
PhilippWendler: Ah, thanks for checking! I hadn't noticed. Then we can just write "since 24.04".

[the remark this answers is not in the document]
younghojan: I just forgot to include the link to the release notes. Please check this [link](https://discourse.ubuntu.com/t/mantic-minotaur-release-notes/35534?_gl=1*178yt1l*_gcl_au*MTM0MTg5ODM0MC4xNzE4MTU4NDAz&_ga=2.40433895.455817366.1718158403-1173811805.1718158401#security-improvements-12). :-)

benchexec/containerexecutor.py
younghojan: Should we replace this link with the corresponding blog for 24.04, https://ubuntu.com/blog/whats-new-in-security-for-ubuntu-24-04-lts , since this restriction became the default in 24.04?

benchexec/container.py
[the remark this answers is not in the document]
PhilippWendler: Sure, it confirms what you said. But I would prefer the link to the blog post to be used as documentation.

benchexec/containerexecutor.py
PhilippWendler: Hm, the newer blog post doesn't describe much, and it does not even mention the `sysctl` flag. And they do not even link to more information. So I guess the best link that we have remains the 23.10 blog post. Do you agree?

benchexec/container.py
younghojan: I have changed the error message to this:
```Python
_ERROR_MSG_USER_NS_RESTRICTION = (
    "Unprivileged user namespaces forbidden on this system, please "
    "enable them with 'sysctl -w kernel.apparmor_restrict_unprivileged_userns=0', "
    "Ubuntu disable it by default since 24.04, refer to "
    "https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces "
    "for more information."
)
```
 I noticed that `container.md` already has an entry for `Cannot execute ...: Creating namespace for container mode failed` and a link to `INSTALL.md#kernel-requirements` followed. We could just add the solution for Ubuntu 24.04 in `INSTALL.md#kernel-requirements`, and leave `container.md` unchanged. Users should be able to find the solution by following the error message.

benchexec/containerized_tool.py
younghojan: That's true, I'll raise a `BenchExecException` here. The output would be this:

```
(benchexec-venv) benchexec-venvhaoranyang@haoranyang-fudan:~/benchexec$ benchexec doc/benchmark-example-rand.xml --tasks "XML files" --limitCores 1 --timelimit 10s --numOfThreads 4
2024-06-12 13:28:05 - WARNING - Creating UID mapping into container failed: [Errno 1] Operation not permitted
2024-06-12 13:28:05 - WARNING - Could not write to setgroups file in /proc: [Errno 13] Permission denied: '/proc/71549/setgroups'
2024-06-12 13:28:05 - WARNING - Creating GID mapping into container failed: [Errno 1] Operation not permitted
Error: Unprivileged user namespaces forbidden on this system, please enable them with 'sysctl -w kernel.apparmor_restrict_unprivileged_userns=0'. Ubuntu disable it by default since 24.04, refer to https://ubuntu.com/blog/whats-new-in-security-for-ubuntu-24-04-lts for more information.
```

benchexec/containerexecutor.py
younghojan: Sure I agree with that 😃

doc/INSTALL.md
```diff
@@ -169,6 +169,9 @@ that are not usable on all distributions by default:
   On CentOS it can be necessary to enable this feature with
   `sudo sysctl -w user.max_user_namespaces=10000` or a respective entry
   in `/etc/sysctl.conf` (the exact value is not important).
+  On Ubuntu 24.04 (or newer versions), execute `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0`
+  to enable it. This setting will be reset upon reboot. To apply a persistent setting, create a new file
+  at `/etc/sysctl.d/60-apparmor-namespace.conf` with `kernel.apparmor_restrict_unprivileged_userns=0`.
```
PhilippWendler, line 174: Just point to `/etc/sysctl.conf` as above?

For changing `sysctl` values automatically, i.e., via some tool for configuration management, `/etc/sysctl.d` is great. For manual editing by a human it provides little benefit.

benchexec/container.py
```diff
@@ -115,6 +115,14 @@
 )
 """Whether we use generated native code for clone or an unsafe Python fallback"""
 
+_ERROR_MSG_USER_NS_RESTRICTION = (
+    "Unprivileged user namespaces forbidden on this system, please "
+    "enable them with 'sysctl -w kernel.apparmor_restrict_unprivileged_userns=0'. "
+    "Ubuntu disable it by default since 24.04, refer to "
```
PhilippWendler, line 121:
```suggestion
    "Ubuntu disables them by default since 24.04, refer to "
```

doc/INSTALL.md
younghojan: Like this, or use the same sentence structure as "On CentOS..."?

```markdown
  On Ubuntu 24.04 (or newer versions), execute `sysctl -w kernel.apparmor_restrict_unprivileged_userns=0`
  or add a respective entry in `/etc/sysctl.conf` to enable it.
```

PhilippWendler: Sure, why not the same sentence structure?

younghojan: Maybe I like to be grammatically flexible. Just kidding 😆, I've used the same sentence structure in the next commit.

PhilippWendler:
I squashed the commits of this PR. Usually we do not do this, but in this case the back-and-forth just clutters the history and does not provide a benefit. I also added some slight refactoring and will now merge this PR. Thanks!

younghojan:
Very honored that these codes has been merged into the main branch!

Issue #991, opened 2024-02-14
Better handling of rounding-mode context for decimal calculations in table-generator


PhilippWendler:
In table-generator, we use the rounding mode `ROUND_HALF_UP`. We set this globally in `columns.py`, which creates problems if someone imports table-generator, and for #917.
We also do not want to specify the rounding mode for every single operation because it can easily be forgotten. But we can set the rounding mode using a local context by wrapping `with decimal.localcontext() as context:` around all entry points. That would avoid the change of the global state and still be less risk. We just have to check whether this automatically works for computations done in worker processes with the multiprocessing module.

PhilippWendler:
Implemented by #1043.

PR #1043, opened 2024-05-26
Use Pytest for test execution

### LINKED ISSUE #162: Update test framework

schroeding:
This branch changes the test runner to pytest. Fixes #162.


benchexec/pytest_analyze_run_result.py
```diff
@@ -0,0 +1,197 @@
...
+from benchexec.model import Run
+from benchexec.result import (
+    RESULT_FALSE_REACH,
+    RESULT_ERROR,
+    RESULT_UNKNOWN,
+    RESULT_TRUE_PROP,
+)
+from benchexec.tools.template import BaseTool
+
+sys.dont_write_bytecode = True  # prevent creation of .pyc files
```
PhilippWendler, line 23: I don't think we need this in test code. It was intended to prevent Python writing all the `.pyc` files out for regular users.

benchexec/pytest_analyze_run_result.py
```diff
@@ -0,0 +1,197 @@
...
+
+normal_result = ProcessExitCode(raw=0, value=0, signal=None)
+
+
+@pytest.fixture(scope="class")
+def disable_non_critical_logging():
+    logging.disable(logging.CRITICAL)
```
PhilippWendler, line 30: There was some discussion around this in #908, please have a look there. Ideal is a solution where
- we do not have to add something in every test file,
- log output does not clutter standard test execution (if everything works), and
- as much information as possible (including log messages) is shown for every failed test case.

IIRC we found some good way to do this in #908. If you see further possible improvements, please don't hesitate to add them.

PhilippWendler (commented):
It would be great if the final diff of the PR (before merging) is somewhat useful and readable (to allow reviewing). I know that there will inevitably be quite some changes due to the test structure and the assertions, but can you maybe aim at making these changes as localized as possible? Such that for example comparing `pytest_foo.py` against `test_foo.py` is doable.

We also want to compare the old tests against the new tests, in particular with regards to the question how many tests are executed, in order to find any overlooked things.

For this I wonder what the best thing to do is?
- Copy the test suite in a branch (as you do) such that we can always execute both? And only at the end of the project delete the old test suite and rename the new one?
- Simply replace the test suite in a branch and then compare it with an execution of the old test suite on `main`? This allows to continually look at the diff between the test suites using GitHub.
- Replace the test suite incrementally, module by module, after checking that the module is rewritten correctly. Makes everything smaller and easier to follow, but with the disadvantage of having two independent test suites in CI for a while.

PhilippWendler:
> In this case, this may be fine though, as the only reason for duplicating the testsuite in the branch would be to make sure we don't miss any changes to the testsuite on main in the meantime (e.g. added tests which may not cause a merge conflict)

I would expect that almost all changes to the test suite on main in the meantime would cause a merge conflict. And to double check it is easy to get a list of all commits touching test files.

> and to make comparing the results between the two testsuites (especially coverage) more comfortable, right?
> 
> Without copying the testsuite (or simply not checking it into git), this would be equivalent to option 2.

Then we can simply do option 2, if you like that best.

benchexec/tablegenerator/columns.py
```diff
@@ -18,12 +18,15 @@
 
 __all__ = ["Column", "ColumnType", "ColumnMeasureType"]
 
-# This sets the rounding mode for all Decimal operations in the process.
-# It is actually used only as default context for new contexts, but because we set this
-# at import time and before any threads are started, it should work according to its
-# documentation. We double check with the context of the current thread.
+# This sets the rounding mode for all Decimal operations in the process and explicitly for the current context
+# As asserting the rounding mode for the current context is almost a tautology, we also assert in the methods
```
PhilippWendler, line 22: True, but for those who read it and do not know the history of this line, it would be quite confusing to state this and then still do the tautological assertion. I think that we can just get rid of the assert and this explanation then, but we would need to explain why checks in the methods are needed, and ideally a reference to #991.

PhilippWendler (commented):
Can you please also check the amount of executed tests before and after this change, to make sure that pytest picks up the same tests and none is forgotten? Ideal would even be to compare a list of executed tests.

benchexec/tablegenerator/columns.py
```diff
@@ -18,12 +18,10 @@
...
 __all__ = ["Column", "ColumnType", "ColumnMeasureType"]
 
-# This sets the rounding mode for all Decimal operations in the process.
-# It is actually used only as default context for new contexts, but because we set this
-# at import time and before any threads are started, it should work according to its
-# documentation. We double check with the context of the current thread.
-decimal.DefaultContext.rounding = decimal.ROUND_HALF_UP
-assert decimal.getcontext().rounding == decimal.ROUND_HALF_UP
+# Important: It's important to make sure on *all* entry points / methods which perform arithmetics that the correct
+#            rounding / context is used.
```
PhilippWendler, line 22: Weird formatting.

benchexec/tablegenerator/columns.py
```diff
@@ -18,12 +18,10 @@
...
+local_context = decimal.getcontext()
```
PhilippWendler, line 23: This feels more like a module-level constant, so, I would use CONSTANT_CASE. And name would be better if it "decimal" ("rounding" would also be good at first thought, but the context defines more than that). "local context" could be anything. Just `DECIMAL_CONTEXT` would already be better.

Furthermore, this is a weird strategy: Setting the rounding mode on the current thread's context, but then *also* using `with decimal.localcontext`. If we already bite the bullet of having to use `localcontext`, then we do not need to and should not change the current thread's context and affect other code. So we should create a new context.

benchexec/tablegenerator/columns.py
```diff
@@ -132,32 +130,37 @@ def __init__(
...
     ):
-        # If scaling on the variables is performed, a display unit must be defined, explicitly
-        if scale_factor is not None and scale_factor != 1 and unit is None:
-            raise util.TableDefinitionError(
-                f"Scale factor is defined, but display unit is not (in column {title})"
-            )
+        with decimal.localcontext(local_context):
+            assert (
+                decimal.getcontext().rounding == decimal.ROUND_HALF_UP
+            ), f"rounding of context is {decimal.getcontext().rounding}, expected ROUND_HALF_UP"
```
PhilippWendler, line 136: These assertions are not needed anymore once we properly use a local context. We do not fear anyone accidentally changing our context, we fear that people forget to use `localcontext` where necessary, but the assertions don't help with that.

pyproject.toml
```diff
@@ -70,3 +70,7 @@ ignore = [
     # wildcard imports significantly shorten test code,
     'F405',
 ]
+
+[tool.pytest.ini_options]
+python_files = ["test_*.py", "test_integration/__init__.py", "test.py"]
+norecursedirs = ["docker_files"] # otherwise we execute files which are intended to be executed inside a docker container
```
PhilippWendler, line 76:
```suggestion
norecursedirs = ["contrib/p4/docker_files"]
```
I did not even know that this directory exists, so I was confused what this is referring to even with the comment.

setup.cfg
```diff
@@ -82,9 +81,4 @@ console_scripts =
 benchexec.tablegenerator =
   react-table/build/*.min.js
   react-table/build/*.min.css
-
-[nosetests]
-# Necessary for nose since Python 3.8 to find the tests on Windows
-traverse-namespace=1
-# Necessary to find tests in non-package
-include=contrib
+
```
PhilippWendler, line 84: Please add trailing space.

PhilippWendler (commented):
Great, thanks for the work and comparing the test results!

Just some minor cleanup now.

release.sh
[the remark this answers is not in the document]
PhilippWendler: Of course, don't worry. The particular part of the script that is changed can also be relatively easily tested separately by just setting `DIR` and copying the commands into a shell. But in this case I have just worked on this part recently and thus happen to know that your change is correct, so it is fine.

PhilippWendler (approved):
Thanks a lot! I am happy that this is implemented and one of our oldest open issues is resolved.

PR #1044, opened 2024-05-26
Quickstart tutorial

incaseoftrouble:
First draft for the quickstart guide. It should include a short point how to add it for docker / what to consider there and maybe some basic troubleshooting without getting too large (I think this should remain at about 1-2 pages so it doesn't look more daunting than necessary).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 # BenchExec
+
 ## A Framework for Reliable Benchmarking and Resource Measurement
 
 [![Build Status](https://gitlab.com/sosy-lab/software/benchexec/badges/main/pipeline.svg)](https://gitlab.com/sosy-lab/software/benchexec/pipelines)
 [![Apache 2.0 License](https://img.shields.io/badge/license-Apache--2-brightgreen.svg)](https://www.apache.org/licenses/LICENSE-2.0)
 [![PyPI version](https://img.shields.io/pypi/v/BenchExec.svg)](https://pypi.python.org/pypi/BenchExec)
 [![DOI](https://zenodo.org/badge/30758422.svg)](https://zenodo.org/badge/latestdoi/30758422)
-
 
 **News and Updates**:
 - Two projects accepted for BenchExec as part of [Google Summer of Code](https://summerofcode.withgoogle.com/)!
@@ -28,6 +28,11 @@ SPDX-License-Identifier: Apache-2.0
 - An extended version of our paper on BenchExec and its background was published as open access in the journal STTT,
   you can read [Reliable Benchmarking: Requirements and Solutions](https://doi.org/10.1007/s10009-017-0469-y) online.
   We also provide a set of [overview slides](https://www.sosy-lab.org/research/prs/Latest_ReliableBenchmarking.pdf).
+
+> To help new or inexperienced users get started with reliable benchmarking
+> right away, we offer a [quickstart guide](doc/quickstart.md) that contains
+> a brief explanation of the issues of common setups as well as the (few)
+> steps necessary to setup and use BenchExec instead.
 
 BenchExec is a framework for reliable benchmarking and resource measurement
 and provides a standalone solution for benchmarking
--- a/doc/benchexec-in-container.md
+++ b/doc/benchexec-in-container.md
@@ -239,7 +239,6 @@ There are many ways to achieve the required setup and users familiar with
 Docker may choose to adapt the above procedure. There are a few peculiarities
 to be aware of.
 
-
 As explained above, the goal is to give BenchExec its own cgroup to use. So,
 you need to start BenchExec processes inside an otherwise empty cgroup (with
 appropriate controllers available). If that is not the case, BenchExec will
@@ -251,7 +250,7 @@ inside the container, i.e., one where BenchExec has its own separate cgroup.
 In any case, the cgroup which BenchExec then uses should have as many
 controllers enabled and delegated to sub-cgroups as possible, for example like
 this:
-```
+```bash
 mkdir -p /sys/fs/cgroup/benchexec
 for controller in $(cat /sys/fs/cgroup/cgroup.controllers); do
   echo "+$controller" > /sys/fs/cgroup/cgroup.subtree_control
--- a/doc/benchexec-in-container.md
+++ b/doc/benchexec-in-container.md
@@ -1,0 +1,261 @@
+<!--
+This file is part of BenchExec, a framework for reliable benchmarking:
+https://github.com/sosy-lab/benchexec
+
+SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# BenchExec: Containerized Environments
+
+This document explains the complications of using BenchExec inside
+containerized environments (such as Docker or Podman) and shows how you can
+create your own interactive Docker image (or adapt existing ones) to use
+BenchExec within it. We focus mainly on the (nowadays standard) cgroups v2,
+a brief guideline for the (outdated) cgroups v1 is provided
+[below](#cgroups-v1).
+
+There is a difference between executing only BenchExec inside a
+(non-interactive) container and a fully interactive one which contains
+multiple processes, with the former being significantly simpler. Both are
+discussed here.
+
+Note: It is possible to use BenchExec inside other container environments, e.g.
+Docker, but we strongly recommend to use [Podman](https://podman.io/)
+(compatible with Docker), as it provides "rootless" containers, i.e. its
+containers are started as a regular user without sudo (just like BenchExec
+containers). However, as the setup is largely the same, we provide it for
+either case. In case you do want to use Docker, please also consider following
+the (easy to implement) [security recommendations](#securing-docker-execution).
+
+
+## Non-interactive Setups
+
+In case you intend to only execute one BenchExec process and nothing else, the
+setup is quite simple. It is sufficient to execute the BenchExec process as
+entry point in a container with all required dependencies installed. For
+example, if BenchExec is installed in a container with tag `my-container`, you
+can simply execute
+```
+podman run --security-opt unmask=/sys/fs/cgroup --cgroups=split \
+  --security-opt unmask=/proc/* \
+  --security-opt seccomp=unconfined \
+  -t my-container <arguments>
+```
+or
+```
+docker run --privileged --cap-drop=all -t my-container benchexec <arguments>
+```
+
+
+## BenchExec in Interactive Containers
+
+Next follows a step-by-step guide to create a Docker / Podman image with
+BenchExec (assuming cgroups v2). Some further background and reasoning is
+provided later. Summarized, the main reason why BenchExec needs a "custom"
+setup for containers is due to how cgroups work in combination with containers;
+we need to "manually" set up a separate cgroup for BenchExec.
+
+While this setup should work on most recent system, we cannot guarantee this,
+since there simply are too many variables. In some cases, you may need to
+slightly adapt some parts.
+
+### Creating the Image
+
+First, create the file `init.sh` with the following content:
+```sh
+#!/bin/sh
+set -eu
+
+# Create new sub-cgroups
+# Note: While "init" can be renamed, the name "benchexec" is important
+mkdir -p /sys/fs/cgroup/init /sys/fs/cgroup/benchexec
+# Move the init process to that cgroup
+echo $$ > /sys/fs/cgroup/init/cgroup.procs
+
+# Enable controllers in subtrees for benchexec to use
+for controller in $(cat /sys/fs/cgroup/cgroup.controllers); do
+  echo "+$controller" > /sys/fs/cgroup/cgroup.subtree_control
+  echo "+$controller" > /sys/fs/cgroup/benchexec/cgroup.subtree_control
+done
+
+# ... or whatever your init process should be
+exec "$@"
+```
+and set it executable (`chmod +x init.sh`).
+
+Now, pack this script into your Docker image and set it as entry point. If you
+are working with a standard Docker image, create a file `Dockerfile`
+```dockerfile
+FROM debian:bookworm-slim
+# Or whichever base image you are using
+
+# Install all dependencies of your tool and BenchExec
+# Example:
+# RUN apt-get update && apt-get -y install \
+#     python3-minimal \
+#   && rm -rf /var/lib/apt/lists/*
+
+# Install BenchExec with any method (apt install, pip install, or just copy the .whl)
+# Examples:
+# RUN pip install benchexec
+# RUN wget https://github.com/sosy-lab/benchexec/releases/download/<current release>.whl -O /opt/benchexec.whl
+
+# Copy the created script
+COPY init.sh /init.sh
+
+# Set init.sh as the entrypoint -- It is important to use brackets here
+ENTRYPOINT [ "/init.sh" ]
+# Set the default process to run
+CMD [ "bash" ]
+```
+If you already have a Dockerfile, you only need to install BenchExec into it
+and add the last few commands (i.e. copy `init.sh` and set the entrypoint).
+
+With this finished, execute `podman build -t <tag> .` (or
+`docker build -t <tag> .` when using Docker) in the directory where the
+Dockerfile is located to build the container.
+
+### Executing BenchExec in the Container
+
+Start the image with
+```
+podman run --cgroups=split \
+  --security-opt unmask=/proc/* \
+  --security-opt seccomp=unconfined \
+  -it <tag>
+```
+or
+```
+docker run --privileged --cap-drop=all -it <tag>
+```
+<!--
+In principle, `--security-opt systempaths=unconfined --security-opt seccomp=unconfined`
+should also be sufficient as Docker arguments,
+but then mounting within the container still fails.
+-->
+
+> **IMPORTANT**: The `--privileged` argument gives the Docker container *full
+> root access* to the host, so make sure to include the `--cap-drop=all` flag,
+> use this command only with trusted images, and configure your Docker
+> container such that everything in it is executed under a different user
+> account, not as root (more details [below](#securing-docker-execution)).
+> BenchExec is not designed to run as root and does not provide any safety
+> guarantees regarding its container under these circumstances.
+
+With this, you should be able to execute, for example, `runexec echo`
+inside the Docker container. (In case you opted for the `.whl` install, you
+need to execute
+`PYTHONPATH=/opt/benchexec.whl python3 -m benchexec.runexecutor <program>`
+instead.)
+
+In case you want to modify this setup, please consider the
+[background information below](#background-and-technical-details).
+
+### Securing Docker Execution
+
+When running under Docker, the user which executes the commands by default has
+`UID` 0, i.e. has the permissions of `root` on the host system. For example, if
+you mount a folder into the container, the processes started in the container
+have *full write access* to these files, no matter what their permissions are.
+As such, a small mistake in, e.g., an evaluation script that deletes temporary
+files, could easily wreak havoc on your system. Using Podman directly mitigates
+this issue, as the Podman container runs with the permissions of the current
+user, not `root`, and thus cannot mess with system files (in particular, the
+`root` user in a Podman container is mapped to your current user).
+
+To significantly increase security of your Docker execution, add
+```dockerfile
+# Create non-root user and set it as default
+RUN useradd -ms /bin/bash user
+WORKDIR /home/user
+USER user
+```
+at the end of your Dockerfile. Note: This is still worse than using Podman
+(e.g. privilege escalation within the container leads to obtaining root
+permissions on the host) but drastically reduces the chance for accidental
+problems.
+
+
+## Cgroups v1
+
+With cgroups v1, use the following command line to start your container
+```
+podman run -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
+  --security-opt unmask=/sys/fs/cgroup --cgroups=split \
+  --security-opt unmask=/proc/* \
+  --security-opt seccomp=unconfined \
+  -it <tag>
+```
+or
+```
+docker run -v /sys/fs/cgroup:/sys/fs/cgroup:rw --privileged --cap-drop=all -it <tag>
+```
+Note that this will not work with cgroups v2.
+
+The special setup for interactive containers is not necessary in this case,
+since cgroup v1 is designed differently. However, be advised that in either
+case, the container unavoidably has access to the root cgroup of your entire
+system, and, in case of Docker, full write access to this cgroup. Thus,
+following the [security advice](#securing-docker-execution) is especially
+recommended here.
+
+
+## Background and Technical Details
+
+Just like BenchExec, tools like Docker and Podman make heavy use of cgroups to
+isolate processes. Thus, in order to get them working nicely together, some
+additional tricks are required, especially in an interactive container (e.g.
+one with a shell, where multiple commands can be executed).
+
+BenchExec needs a dedicated cgroup together with enabled controllers to cleanly
+isolate and precisely measure each tool execution. On a standard system,
+cgroups are usually managed by systemd, from which BenchExec can automatically
+obtain such a cgroup. In contrast, in an interactive container, the init
+process usually is just a shell, which is placed in the "root" cgroup of
+the container, without dedicated cgroup management. Thus, the cgroup needs to
+be managed manually.
+
+One might assume that BenchExec should be able to do this on its own.
+Unfortunately, due to the "no internal processes" rule (see the
+[cgroup documentation](https://www.man7.org/linux/man-pages/man7/cgroups.7.html)
+for more information) any (non-root) cgroup can *either* contain processes *or*
+delegate controllers to a child group. Thus, the shell in the root cgroup of
+the container (which is not the root cgroup of the overall system) prevents
+child cgroups with controllers being created. So, we need to move the init
+process (and thus all subsequent ones) into a separate cgroup, which is the
+purpose of the `init.sh` script. Only then can we create a separate, empty
+cgroup dedicated to BenchExec.
+
+Note that this also is the reason why non-interactive setups do not need this.
+There, BenchExec is the sole root process, so it can use the root cgroup of the
+container.
+
+
+### Adaptation
+
+There are many ways to achieve the required setup and users familiar with
+Docker may choose to adapt the above procedure. There are a few peculiarities
+to be aware of.
+
+The goal is to create a cgroup called `/benchexec` with relevant controllers
+enabled. Note that it is important that the cgroup has exactly this path, as
+BenchExec uses this as a "fallback" on non-systemd setups.
+
+Additionally, you should be aware of the implications of the "no internal
+process"-rule. Effectively, this means that there cannot be any processes in
+the root cgroup of the container (which, notably, is *not* the root cgroup of
+the system). So, for the provided `init.sh` to work, it needs to be directly
+executed (with `ENTRYPOINT [ "/init.sh" ]`). In particular, using, for example,
+`ENTRYPOINT "/init.sh"` would not work, since this creates a shell that then
+runs `init.sh` as a sub-process, meaning that the root cgroup would not be
+empty when we try to enable subtree control.
+
+In case you run into `Device or resource busy` errors, the problem likely is
+that you want to enable subtree control in a non-empty cgroup or, vice versa,
+move a process to cgroup with enabled subtree control, both of which is
+prohibited for cgroups by design. For debugging, it is useful to inspect the
+`cgroup.procs` and `cgroup.subtree_control` of the cgroup in question. Here,
+note that calls like `cat` do also create a separate process.
+
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -1,0 +1,145 @@
+<!--
+This file is part of BenchExec, a framework for reliable benchmarking:
+https://github.com/sosy-lab/benchexec
+
+SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# A Beginner's Guide to Reliable Benchmarking
+
+> If your current setup looks similar to the below example (or you are thinking
+> about such a setup), we strongly recommend following this guide for a much
+> more reliable process.
+
+## Audience
+
+This guide provides a brief summary of instructions to set up reliable
+benchmark measurements using BenchExec and important points to consider. It is
+meant for users who either want to set up benchmarking for a small number of
+executions from scratch or already have a simple setup using, e.g., 
+[`time`](https://linux.die.net/man/1/time),
+[`timeout`](https://linux.die.net/man/1/timeout),
+[`taskset`](https://linux.die.net/man/1/taskset),
+[`ulimit`](https://linux.die.net/man/3/ulimit), etc. Concretely, this guide
+will show you how to use [`runexec`](runexec.md) as a simple but much more
+reliable "drop-in" replacement for these tools. If you want to benchmark large
+number of executions or get the most out of what BenchExec provides as a
+benchmarking framework, consider using the tool [`benchexec`](benchexec.md)
+instead (further details below).
+
+## Why Should I use BenchExec?
+
+As a simple example, suppose that you want to measure the performance of your
+newly implemented tool `program` with arguments `--foo` and `--bar` on the
+input files `input_1.in` to `input_9.in`. To measure the runtime of the tool,
+you might execute `/usr/bin/time program --foo input_1.in` etc. and note the
+results. In case resource limitations are desired (e.g. limiting to 1 CPU and
+60 seconds of wallclock time), the calls might be
+`taskset -c 0 timeout 60s /usr/bin/time program ...` or similar.
+
+While useful, these utilities (i.e. `time`, `ulimit`, etc.) unfortunately are
+not suitable for reliable benchmarking, especially when parallelism or
+sub-processes are involved, and may give you *completely wrong* results.
+For further details and insights into peculiarities and pitfalls of reliable
+benchmarking (as well as how BenchExec is mitigating them where possible), we
+recommend the
+[overview slides](https://www.sosy-lab.org/research/prs/Latest_ReliableBenchmarking.pdf)
+and [the corresponding paper](https://doi.org/10.1007/s10009-017-0469-y).
+
+BenchExec takes care of most of the underlying problems for you, which is why
+we recommend using it instead. Concretely, by following this guide, you can
+significantly increase the reliability of your results without much effort.
+
+## Reliable Benchmarking with BenchExec
+
+The following steps show you how to increase the reliability and quality of
+measurements by using BenchExec instead of the standard system utilities.
+
+### Step 1. Install BenchExec
+
+Depending on your distribution, the setup steps vary slightly.
+On Debian/Ubuntu and similar, you can
+```
+sudo add-apt-repository ppa:sosy-lab/benchmarking
+sudo apt update && sudo apt install benchexec
+```
+Otherwise, try the pip-based setup
+```
+pip3 install --user benchexec[systemd] coloredlogs
+```
+Advanced users can also directly download and run the packaged python wheel
+```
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -43,15 +47,15 @@ In particular, BenchExec provides three major features:
 
 - execution of arbitrary commands with precise and reliable measurement
   and limitation of resource usage (e.g., CPU time and memory),
-  and isolation against other running processes  
+  and isolation against other running processes
```
PhilippWendler, line 50: Be careful with your editor config. Trailing whitespace is (unfortunately) significant in Markdown and should not be removed (although we do want this in other file types).

README.md
```diff
@@ -8,13 +8,17 @@ SPDX-License-Identifier: Apache-2.0
...
 # BenchExec
+
 ## A Framework for Reliable Benchmarking and Resource Measurement
 
 [![Build Status](https://gitlab.com/sosy-lab/software/benchexec/badges/main/pipeline.svg)](https://gitlab.com/sosy-lab/software/benchexec/pipelines)
 [![Apache 2.0 License](https://img.shields.io/badge/license-Apache--2-brightgreen.svg)](https://www.apache.org/licenses/LICENSE-2.0)
 [![PyPI version](https://img.shields.io/pypi/v/BenchExec.svg)](https://pypi.python.org/pypi/BenchExec)
 [![DOI](https://zenodo.org/badge/30758422.svg)](https://zenodo.org/badge/latestdoi/30758422)
 
+> [!NOTE]
```
PhilippWendler, line 19: Interesting, I wasn't aware of this Markdown feature. Do you happen to know how supported this is outside of GitHub? We copy our Readme for example into the project description on PyPI, and I wonder whether this would look nice there or break rendering.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+```
+$ /usr/bin/time program --foo input_1.in
+$ /usr/bin/time program --bar input_1.in
+$ /usr/bin/time program --foo input_2.in
+...
+```
+etc. and note the results. In case resource limitations are desired (e.g. 
+limiting to 1 CPU and 60 sec of wallclock time), the calls might be
+```
+$ taskset -c 0 timeout 60s /usr/bin/time program ...
```
PhilippWendler, line 29: I wonder whether we should show negative examples, and in particular this prominently (as highly visible listings on the top of the file, with copy-and-paste buttons from GitHub, etc.). I fear that some users will stop reading here...

If you think these add value, how about having just 1-2 command lines, and inline with single backticks instead of as blocks, such that they are less strongly highlighted?

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+```
+or similar.
+
+## Benchmarking with BenchExec
+
+The following steps guide you to increase the reliability and quality of
+measurements drastically by using BenchExec instead of these small
+utilities.
+
+### Step 1. Install BenchExec
```
PhilippWendler, line 39: I wonder whether it is good to duplicate a part of the instructions here, where they could become more easily outdated and where users might give up more easily if the full help is hidden behind a link. On the other hand, from the tutorial point-of-view it makes sense and it might highlight that is in fact easy to install BenchExec for a large share of users (those on common modern distributions).

My current tendency is to say that the benefits outweigh the drawbacks.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Step 2. Consider the Setup of Your Benchmark
+
+Consider and document the benchmarking process *beforehand*. In particular,
+think about which executions you want to measure, what resource limits should
+be placed on the benchmarked tool(s), such as CPU time, CPU count, memory, etc.
+Also consider how timeouts should be treated.
+
+For more complicated setups, please also refer to the
+[benchmarking setup guide](benchmarking.md). For example, in case you want to
```
PhilippWendler, line 64: I would say we should make the recommendation to look at `benchmarking.md` (btw., it is not a "setup guide") stronger and without conditions, not just for complicated setups. I would also like to have a reference to the paper (and slides?) here.

But we can change the wording such that the tutorial still keeps its flow and users are not discouraged. For example, we could add an "Independently from using BenchExec, ..." or so in the beginning and something with the idea of "to continue with trying out BenchExec, simply follow the rest of the tutorial" at the end.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+resources (e.g. the memory bus and CPU cache). For such cases, we recommend
+using `benchexec` instead, which takes care of managing parallel invocations.
+
+
+### Step 3. Gather Measurements using runexec
+
+Using the example from above, suppose that we want to limit the process to 60s
+wall time, 1 GB of memory, and cpu core 0. Then, simply run
+```
+$ runexec --quiet --walltimelimit 60s --memlimit 1GB --cores 0 --output output_1_foo.log -- \
```
PhilippWendler, line 75: I would remove `--quiet` and `--output` here. The less options, the better.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+[benchmarking setup guide](benchmarking.md). For example, in case you want to
+execute multiple benchmarks in parallel, think about how to deal with shared
+resources (e.g. the memory bus and CPU cache). For such cases, we recommend
+using `benchexec` instead, which takes care of managing parallel invocations.
+
+
+### Step 3. Gather Measurements using runexec
+
+Using the example from above, suppose that we want to limit the process to 60s
+wall time, 1 GB of memory, and cpu core 0. Then, simply run
```
PhilippWendler, line 73: We already had users confusing the `--cores` setting of `runexec` with a limit on the number of cores. I would take the chance here to try and clarify this with something like "... and one CPU core (pinning the process to CPU core 0 of the system)".

README.md
```diff
@@ -8,13 +8,17 @@ SPDX-License-Identifier: Apache-2.0
...
+> To get started with reliably benchmarking right away, follow the
+> [quickstart guide](doc/quickstart.md).
```
PhilippWendler, line 21: I would move this below the "News and Updates".

Also, we have to restrict this statement somehow. Right now it reads as if everyone could (and should?) just follow this link, but the tutorial targets only a part of potential BenchExec users. So we send some people the wrong way.

Besides that, using the phrase "we offer" (a quickstart guide) makes it sound more like a nice option instead of telling people to not bother with the rest of the readme.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
+# A Quickstart Guide to Proper Benchmarking with BenchExec
+
+This guide provides a brief summary of instructions to set up reliable
+benchmark measurements using BenchExec and important points to consider. It is
+meant for users who either want to set up benchmarking from scratch or already
+have a simple setup using, e.g., `time`, `timeout`, `taskset`, `ulimit`, etc.
```
PhilippWendler, line 6: I would prefer to rephrase this slightly. While I agree with the second half, I would say that users who do not have anything already should be encouraged to try out `benchexec` in many cases.

Something like the following maybe?

> It is meant for users who already have a basic setup using, e.g., `time`, `timeout`, `taskset`, `ulimit`, etc., or who want to try out BenchExec for small numbers of benchmarking runs. If you want to benchmark large number of runs or get the most out of what BenchExec provides as a benchmarking framework, consider using benchexec.

doc/quickstart.md
```diff
@@ -0,0 +1,98 @@
...
+
+Using the example from above, suppose that we want to limit the process to 60s
+wall time, 1 GB of memory, and cpu core 0. Then, simply run
+```
+$ runexec --quiet --walltimelimit 60s --memlimit 1GB --cores 0 --output output_1_foo.log -- \
+  program --foo input_1.in
+```
+This executes `program --foo input_1.in` and prints measurements to standard
+output, such as walltime, cputime, memory, I/O, etc. in a simple to read and
+parse format. The output of program is redirected to `output_1_foo.log`.
```
PhilippWendler, line 80: We could add an example of the output here? Like in `runexec.md`. And maybe a link to `run-results.md`?

PhilippWendler (commented):
Thanks a lot!

Besides the idea of this tutorial, I also like the overall structure, its scope, length, and level of detail.

I mostly have suggestions and questions about tweaking the phrasing a little bit.

I guess we should add more cross references from the Documentation section of the readme, the doc INDEX, and `runexec.md`?

README.md
incaseoftrouble: but these are two trailing whitespace (reduced to one). this shouldn't make a difference I think.

README.md
incaseoftrouble: I found it while searching for "nice" markdown boxes, I think this is a GitHub-specific feature, can remove it

doc/quickstart.md
incaseoftrouble: Fair point, will change

doc/quickstart.md
incaseoftrouble: Yes, agree.

doc/quickstart.md
incaseoftrouble: Agree with quiet but without --output the run output is put into `output.log` which can be unexpected (overwriting the previous outputs)

doc/quickstart.md
incaseoftrouble: I agree in spirit, but I would rather reformulate the subsequent text to indicate that there is more than `runexec`. So that the tutorial reads like "if you have something like this, use BenchExec. For a quick&dirty fix, simply replace these commands with `runexec`, for a proper solution, use `benchexec`"?

doc/quickstart.md
incaseoftrouble: I wasn't sure how much value it adds, but yes, agree.

README.md
incaseoftrouble: pandoc isn't aware of it, so I'll remove it

README.md
PhilippWendler: Two trailing whitespaces define a line break in Markdown, one trailing whitespace is ignored. So this makes the difference.

(As a side note, please don't mark discussion as resolved that others created. This makes it hard to follow the discussion and making sure that everything is actually resolved.)

doc/quickstart.md
incaseoftrouble: On this note: Would it be useful to maybe have `output-%Y%M%D-%h%m%s.log` (or similar)  as default?

PhilippWendler: Good question. Then we would have to do the same for `output.files`, of course. I am not sure how useful it is, because if you want to do real benchmarking, just knowing the time stamp will not be enough, you need more metadata anyway, but it likely wouldn't hurt. But we could change this only for the next major version of BenchExec.

incaseoftrouble: Arguably, if you do "real benchmarking" you probably set the output path manually too ;)

On that note, there might also be a point in actually printing that meta information too. Concretely, this would probably mean generating a unique identifier for each execution (e.g. id = time_random number, or even just time up to milliseconds) and then writing out `run_<id>.out` `run_<id>.err` and `run_<id>.meta`, where the "meta" file contains everything relevant (e.g. things like version of benchexec, the complete command line invocation, cwd, env, uname, and maybe the run results?)

(this is not the right place to discuss this, if this sounds interesting I can create an issue for this)

incaseoftrouble:
I think I found something weird with cgroup v2 in docker :) Not sure if this is related to benchexec, but maybe you know more:

I made the following script:
```
#!/bin/sh

echo $$ > /sys/fs/cgroup/benchexec/cgroup.procs
cat /proc/$$/cgroup

export PYTHONPATH=/opt/benchexec.whl
exec python3 -m benchexec.check_cgroups
```
Output: `0::/../../benchexec` -- so the path of the cgroup is somehow reported wrongly?!

PhilippWendler:
> EDIT: I completely removed my writing-while-thinking comments, the results are now moved to separate issues.

Thanks. Continuing separately makes a lot of sense.

Regarding the removed comments, I want to add that it should have worked with Podman already, it is just that after my last comment you seem to have switched to Docker only, and there got some arguments wrong. I also tried to improve the documentation in this regard in 901b50c303590e0fd12543f49a804e370f30b777. If you find something that does not work as documented now, this would be another issue.

> It seems that `docker run --privileged --cap-drop=all --cgroupns=host` is needed to work with cgroups-v2, but there are some other issues related to benchexec.

`--cgroupns=host` is not needed for me and would actually be contrary to what one wants. With cgroups v2, nesting of containers finally works properly, but with `--cgroupns=host` one disables this.

> I think it would be really beneficial to have a tutorial of how to actually manage the cgroups in an interactive container. A lot of artifact evaluations want everything to be self-contained inside the image (i.e. include data processing etc.). But since I believe this requires more work, it is probably best if we delay this.

It is tricky, because use cases and systems are so diverse. We can probably not hope to give a tutorial "this is how it works", but we can likely give an example "this is how you can do it in this particular case". But what needs to be done for example already differs between the case "interactive container, configure cgroups, start BenchExec" and "non-interactive container with script wrapping BenchExec".

incaseoftrouble:
> --cgroupns=host is not needed for me and would actually be contrary to what one wants. With cgroups v2, nesting of containers finally works properly, but with --cgroupns=host one disables this.

I needed that to be able to modify subtree_control easily. With --cgroupns=private, the docker container is put into a cgroup with empty subtree_control and this cannot be modified if there is a process in there. So, another solution is to create a sub-cgroup of the container's group, move the interactive shell there, set the subtree_control of the container's cgroup, create another group for benchexec, and then start benchexec inside that one. This seems kinda brittle.

But being able to run easily in an interactive docker setting is (I think) really important.

PhilippWendler:
> > --cgroupns=host is not needed for me and would actually be contrary to what one wants. With cgroups v2, nesting of containers finally works properly, but with --cgroupns=host one disables this.
> 
> I needed that to be able to modify subtree_control easily.

Oh, I see now also from the explanation in the other issue. But this means that it works only because you escape the cgroups of the (outer) container completely. So for example any resource limits configured on the Docker container won't work. I won't recommend this.

> So, another solution is to create a sub-cgroup of the container's group, move the interactive shell there, set the subtree_control of the container's cgroup, create another group for benchexec, and then start benchexec inside that one. This seems kinda brittle.

Well, this is exactly how cgroups v2 are supposed to be used (and what BenchExec does internally for itself), so this is not brittle per se. It starts getting brittle once you think about "moving the interactive shell there", because if you have more than that shell process (e.g., background processes), then you also need to move them. And if someone starts `runexec` twice in parallel in a script, you will have problems with race conditions in this step. Reading the existing list of processes and moving them can't be done atomically. And there is also no atomic "move process P from A to B" operation.

> But being able to run easily in an interactive docker setting is (I think) really important.

Not arguing against this.

> EDIT: I found a somewhat reliable way for docker, I will add it to the quickstart

Maybe discuss in a separate issue first? If we find something like this, I don't want to have it only inside the tutorial, but also in the main documentation. And if the final solution needs several iterations, it is nicer to have the discussion recorded in a specific issue.

incaseoftrouble:
>  It starts getting brittle once you think about "moving the interactive shell there"

Yes, this is what I meant. But now that I understood what is the issue I think it is fine to just put this into the entrypoint of the docker container (which is what I did)

> Maybe discuss in a separate issue first?

Done in #1048

incaseoftrouble:
Ok, I merged the two things and discovered an important bit: Even for non-interactive cases, we need the "complicated" setup due to permissions!

If we only run `podman run ... tag runexec ...` this process is started as `root` inside the podman container -- which we explicitly advise against. However, if you directly start the container as `user`, that user does not have permissions to modify cgroups, so runexec will fail. So, in any case, we need to chown the root cgroup to the non-root user and then switch users only afterwards!

PhilippWendler:
> If we only run `podman run ... tag runexec ...` this process is started as `root` inside the podman container -- which we explicitly advise against. However, if you directly start the container as `user`, that user does not have permissions to modify cgroups, so runexec will fail. So, in any case, we need to chown the root cgroup to the non-root user and then switch users only afterwards!

Out of interest: How do you directly start a container as `user`?

And are you talking about Docker here or about rootless Podman? I wasn't aware that there is support for this, in particular that you can use something like `su` in a rootless Podman container (you cannot in a BenchExec container).

> So either we remove that part ("benchexec should not be run as root" -> "its fine inside a container for non-interactive setups") or simply say "this is the way to do it"

Ok, so the following is true:
- Outside of a container, BenchExec should not run as root.
- In a Docker container (with `--privileged` or not), BenchExec should not run as root (because container root is host root). (There is AFAIK an optional mode in Docker to make use of user namespaces like Podman does, but I don't know exactly how to use it and I don't think anybody uses it, so I ignore this.)

To be honest, for the case of a rootless Podman I hadn't thought much about whether root is acceptable or not. My first intuition would be that it is probably acceptable.

---

After thinking more about this:

If you managed to use `su` in a rootless Podman container, then the container was not configured fully rootlessly, because this would not be possible. Instead what must have happened is that you have the suid tool `newuidmap` installed and configured (`/etc/subuid`) such that your user account is allowed to use a certain range of other uids as well. Podman optionally makes use of this to create containers with more than one user inside. You can probably verify this by looking under which uids the processes appear outside of all containers.

In a setup with Podman and usable `newuidmap`, using `su` provides a slight additional security boundary compared to not using it (for example, a process escaping both containers would not have access to the user's files).

However, I don't want to say "you have to have `newuidmap`". In a setup with Podman but without `newuidmap`, `su` is not usable and I think we can live with that.

What would be middle ground is to have a Podman container with just one user inside (no `newuidmap` required), but that user not having the uid 0 but some other uid. This is what BenchExec does by default, and Podman's default is what `containerexec --root` does. But I am not sure whether Podman supports this and if cgroups would be usable.

So, we would want to recommend `su` for Docker containers, but for rootless Podman is not necessary and restricts the potential use cases slightly. I would be fine with explaining the case for rootless Podman in detail (with Dockerfile and script) and telling the users what they additionally need to do for Docker in prose.

--- 
If you are interested in even more details about this (feel free to ignore, but for cgroups you were quite interested, so I am leaving this here):

There are several layers of isolation with regards to containers:
0. `docker run --privileged` provides almost no isolation and is trivial to escape, probably even accidentally.
1. `docker run` is what AFAIK kernel developers call a privileged and insecure container, but Docker seems to get away with it in practice, so it is maybe not trivially escapable. This has still no separation between a user inside a container and the same uid outside the container.
2. A container with a user namespace, but started as root. This is AFAIK the first level for which kernel developers would agree that some security boundary exists, but I don't think anyone wants to make a guarantee. If you start BenchExec as root, this is what you get.
3. A container with a user namespace, not started as root, but with some suid helpers for setup. This is "rootless Podman" if you have `newuidmap` installed. The container itself is fully rootless and as secure as the next level, just with more usable users inside.
4. A container with a user namespace, not started as root. This is the only way to start a container on a standard Linux system without the need for cooperation of the system administrator at any point. (At least for vanilla Linux, many distributions restrict even this.) This is what BenchExec provides if you start it as a regular user.

The main reason why I do not recommend 2. is that it is easy to overlook something during container setup, and then there is immediately an escape to being root at the host. I *think* there are also more things that a container tool needs to take care of if 2. is the target instead of 4., and BenchExec was written only for 4.

And 2. inside 0. or 1. (BenchExec in Docker) is just the same as 2. so we recommend 4. in 0. (Docker+`su`). But 2. inside 3. or 4. (BenchExec in rootless Podman) should not be worse than just 4. (BenchExec outside of container). The combination 4. in 3. is the solution with Podman+`su`. 4. in 4. is technically possible (you can nest BenchExec containers), but I don't know whether it is possible if Podman is the outer layer.

incaseoftrouble:
> Out of interest: How do you directly start a container as user?

Add `USER` in the Dockerfile or `--user=xy` to the `run` command.

> If you managed to use su in a rootless Podman container, then the container was not configured fully rootlessly, because this would not be possible. 

IDK :D `podman run -t debian:latest id` gives me `uid=0(root) gid=0(root) groups=0(root)`

My `/etc/subuid` says `<user>:100000:65536`. I think this `newuidmap` was installed as a dependency of docker / podman.

> You can probably verify this by looking under which uids the processes appear outside of all containers.

Running as `0` inside the container: My account; creating a new user and `su` into that: Some high id (`100999`)

So I guess that confirms the suspicion + I think I understand the issue.

>  telling the users what they additionally need to do for Docker in prose.

What will happen then is that people copy-paste the "unsafe" Dockerfile and use it with docker I think (we can be fine with that and explicitly say "this is at your own risk"; I'm just saying that this is probably what would happen).

I would be happy with this variant. Especially because the su variant (a) looks confusing and (b) (as I understand it) breaks on a "recommended" setup (podman without newuidmap); and going into detail what needs to be done then just makes this more daunting. I think especially for a "basic guide" it is better if the approach is universal than being perfectly secure.

> but Docker seems to get away with it in practice

I think the working assumption is that docker images are trusted; but yeah.

PhilippWendler:
Hm, all this container handling by Podman seems to be even more complex. It seems that differently from what I thought and what I claimed above it does not use a user namespace per container, you need to request that explicitly with `--userns=`. It does seem to put all containers that you run into a shared user namespace, though ([source](https://github.com/containers/podman/discussions/13728)). And if you want to have a separate user namespace without `newuidmap` (i.e., case 5 that BenchExec does by default) then you even need to manually specify a single-uid-mapping with `--uidmap`.

So although security-wise Podman containers are worse than I thought the complexity to explain all what is necessary for a proper solution is way beyond what we can and want to manage here.

This is just to leave my latest knowledge here, I will now review the current state of the PR.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+
+With this finished, execute `podman build -t <tag> .` (or
+`docker build -t <tag> .` when using Docker) in the directory where the
+Dockerfile is located to build the container.
+
+### Executing BenchExec in the Container
+
+Start the image with
+```
+  podman run --security-opt unmask=/sys/fs/cgroup --cgroups=split --security-opt unmask=/proc/* --security-opt seccomp=unconfined -it <tag>
```
PhilippWendler, line 100:
```suggestion
podman run --security-opt unmask=/sys/fs/cgroup --cgroups=split --security-opt unmask=/proc/* --security-opt seccomp=unconfined -it <tag>
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+```
+or
+```
+  docker run --privileged --cap-drop=all -it <tag>
```
PhilippWendler, line 104:
```suggestion
docker run --privileged --cap-drop=all -it <tag>
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+you mount a folder into the container, the processes started in the container
+have *full write access* to these files, no matter what their permissions are.
+As such, a small mistake in, e.g., an evaluation script that deletes temporary
+files, could easily wreak havoc on your system. Using Podman directly mitigates
+this issue, as the Podman container runs with the permissions of the current
+user, not `root`, and thus cannot mess with system files (in particular, the
+"`root`" user in a Podman container is "mapped" to your current user).
+
+To significantly increase security of your Docker execution, add
+```docker
```
PhilippWendler, line 142:
```suggestion
```dockerfile
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+```
+  docker run --privileged --cap-drop=all -it <tag>
+```
+<!--
+In principle, `--security-opt systempaths=unconfined --security-opt seccomp=unconfined`
+should also be sufficient as Docker arguments,
+but then mounting within the container still fails.
+-->
+
+> **IMPORTANT**: The `--privileged` argument, gives the Docker container *full
```
PhilippWendler, line 112:
```suggestion
> **IMPORTANT**: The `--privileged` argument gives the Docker container *full
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+> guarantees regarding its container under these circumstances.
+
+With this, you should be able to execute, for example, `runexec echo`
+inside the Docker container. (In case you opted for the `.whl` install, you
+need to execute
+`PYTHONPATH=/opt/benchexec.whl python3 -m benchexec.runexecutor <program>`
+instead.)
+
+In case you want to modify this setup, please consider the notes mentioned
+[below](#background-and-technical-details).
```
PhilippWendler, line 127:
```suggestion
In case you want to modify this setup, please consider the
[background information below](#background-and-technical-details).
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Securing Docker Execution
+
+When running under Docker, the user which executes the commands by default has
+`UID` 0, i.e. has the permissions of `root` on the host system. For example, if
+you mount a folder into the container, the processes started in the container
+have *full write access* to these files, no matter what their permissions are.
+As such, a small mistake in, e.g., an evaluation script that deletes temporary
+files, could easily wreak havoc on your system. Using Podman directly mitigates
```
PhilippWendler, line 136: I like that you mentioned a robustness problem such that people do not think "security is not important for me, so I do not need this"!

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+this issue, as the Podman container runs with the permissions of the current
+user, not `root`, and thus cannot mess with system files (in particular, the
+"`root`" user in a Podman container is "mapped" to your current user).
```
PhilippWendler, line 139:
```suggestion
`root` user in a Podman container is mapped to your current user).
```
"mapped" is actually the correct term, so no scare quotes necessary. Same for `root`.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+problems.
+
+### Non-interactive Setups
+
+In case you intend to only execute one BenchExec process inside the container,
+the setup simplifies notably. In particular, then, it is sufficient to execute
+the BenchExec process as entry point, without requiring the `init.sh` script.
+For example, if BenchExec is installed in a container with tag `my-benchexec`,
+you can simply execute
+`podman run <same flags as above> -t my-benchexec benchexec <arguments>`
```
PhilippWendler, line 160: I would like to have this as a ready-to-use command line like in the current docs, so let's copy the flags here and make it a separate code section with three back-ticks.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+at the end of your Dockerfile. Note: This is still worse than using Podman
+(e.g. privilege escalation within the container leads to obtaining root
+permissions on the host) but drastically reduces the chance for accidental
+problems.
+
+### Non-interactive Setups
+
+In case you intend to only execute one BenchExec process inside the container,
+the setup simplifies notably. In particular, then, it is sufficient to execute
+the BenchExec process as entry point, without requiring the `init.sh` script.
```
PhilippWendler, line 157:
```suggestion
the BenchExec process as entry point in a standard container image with all required dependencies installed.
No `init.sh` script and no special steps in the Dockerfile are required.
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+Docker, but we strongly recommend to use [Podman](https://podman.io/)
+(compatible with Docker), as it provides "rootless" containers, i.e. its
+containers are started as a regular user without sudo (just like BenchExec
+containers). However, as the setup is largely the same, we provide it for
+either case. In case you do want to use Docker, please also consider following
+the (easy to implement) [security recommendations](#securing-docker-execution).
+
+If you only require a non-interactive setup, i.e. only run a single BenchExec
+process inside a container, you can avoid most of the complications. See
+[below](#non-interactive-setups) for more details.
```
PhilippWendler, line 27: How about handling the easy case first? Basically just move that section up here and adjust the glue text. It would make it more clear that it is possible to run BenchExec inside a container without lots of configuration at least in some cases. Otherwise people might see the length of the instructions and just give up.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+results separately.
+
+## Background and Technical Details
+
+Just like BenchExec, tools like Docker and Podman make heavy use of cgroups to
+isolate processes. Thus, in order to get them working nicely together, some
+additional tricks are required, especially in an interactive container (e.g.
+one with a shell, where multiple commands can be executed).
+
+The main complication of using BenchExec with Docker is that benchexec needs a
```
PhilippWendler, line 174:
```suggestion
The main complication of using BenchExec with Docker is that due to how cgroups work, BenchExec needs a
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+"clean" cgroup together with enabled controllers. On a standard system, cgroups
+are usually managed by, for example, systemd, from which BenchExec can
+transparently obtain its cgroup. In contrast, in an interactive container, the
```
PhilippWendler, line 177:
```suggestion
automatically obtain a usable cgroup. In contrast, in an interactive container, the
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+root process usually is just a shell, which is placed in the "root" cgroup of
+the container. Due to the "no internal processes" rule (see the
+[cgroup documentation](https://man.archlinux.org/man/cgroups.7) for more
+information) any (non-root) cgroup can *either* contain processes *or* delegate
+controllers to a child group. Thus, the shell in the root cgroup of the
```
PhilippWendler, line 182:
```suggestion
the container. Due to the "no internal processes" rule (see the
[cgroup documentation](https://www.man7.org/linux/man-pages/man7/cgroups.7.html)
for more information)
a cgroup can *either* contain processes *or* delegate controllers to a child group
(the exception for the root cgroup only applies for the root cgroup of the host).
Thus, the shell in the root cgroup of the
```
We use the official links to the man-page project. And I think we can shorten the discussion of the special root case here.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+container (which is not the root cgroup of the overall system) prevents child
+cgroups with controllers being created. So, we need to move the init process
+(and thus all subsequent ones) into a separte cgroup, which is the purpose of
```
PhilippWendler, line 185:
```suggestion
(and thus all subsequent ones) into a separate cgroup, which is the purpose of
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Adaptation
+
+There are many ways to achieve the required setup and users familiar with
+Docker may choose to adapt the above procedure. There are a few peculiarities
+to be aware of.
+
+The goal is to create a cgroup called `/benchexec` with relevant controllers
+enabled. here, note that it is important that the cgroup has exactly this path,
+as benchexec uses this as a "fallback" on non-systemd setups.
```
PhilippWendler, line 196:
```suggestion
enabled. Note that it is important that the cgroup has exactly this path,
as BenchExec uses this as a fallback on non-systemd setups.
```

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+`cgroup.procs` and `cgroup.subtree_control` of the cgroup in question. Here,
+note that calls like `cat` do also create a separate process.
+
+Chowning the cgroup and only then switching to the created user is only
+required for Podman. In Docker, it is sufficient to add `USER user` to the
+Dockerfile and at the end of `init.sh` simply add `exec "$@"`. This is due to
+the docker image being executed with `--privileged`, which essentially
+disables proper permissions handling on, e.g., the cgroups, and any user in the
+image can modify the cgroups.
+
```
PhilippWendler, line 220:
```suggestion
```
This seems to be a left-over?

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+<!--
+## Cgroups v1
```
PhilippWendler, line 222: I think we should have this visible in the docs. Maybe above the background part?

doc/quickstart.md
```diff
@@ -0,0 +1,144 @@
...
+PYTHONPATH=/path/to/benchexec.whl python3 -m benchexec.runexecutor <the program>
+```
+
+You should be able to execute the command
+`python3 -m benchexec.check_cgroups` without issues. See
+[the installation guide](INSTALL.md) for further details and troubleshooting.
+
+> Note: To run inside a docker (or similar) container, unfortunately some
+> more tinkering might be required due to how process isolation with cgroups
+> works. A step-by-step guide for docker follows at the end of the document.
```
PhilippWendler, line 83: Just that we do not forget it: This part needs to be replaced with a link.

PhilippWendler (commented):
Thanks a lot, looks great already! I have just minor comments.

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+Note: It is possible to use BenchExec inside other container environments, e.g.
+Docker, but we strongly recommend to use [Podman](https://podman.io/)
+(compatible with Docker), as it provides "rootless" containers, i.e. its
+containers are started as a regular user without sudo (just like BenchExec
+containers). However, as the setup is largely the same, we provide it for
+either case. In case you do want to use Docker, please also consider following
+the (easy to implement) [security recommendations](#securing-docker-execution).
+
+If you only require a non-interactive setup, i.e. only run a single BenchExec
+process inside a container, you can avoid most of the complications. See
```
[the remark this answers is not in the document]
incaseoftrouble: ... unless you `exec` benchexec :) but yes, agree, will change

doc/benchexec-in-container.md
```diff
@@ -0,0 +1,233 @@
...
+Now, pack this script into your Docker image and set it as entry point. If you
+are working with a standard Docker image, create a file `Dockerfile`
+```dockerfile
+FROM debian:bookworm-slim
+# Or whichever base image you are using
+
+# Install python -- add further dependencies of your application here
+RUN apt-get update && apt-get -y install \
+    python3-minimal \
+  && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
incaseoftrouble: I see your point, but there also would be merit in having something as close to being copy-pasteable as possible. What about putting this into a comment? (Similar to the "install benchexec" part)

doc/benchexec-in-container.md
incaseoftrouble: Hmmm, yes, lets have a look.

doc/benchexec-in-container.md
incaseoftrouble: ah, I wasn't sure which page is the most "official" one :D

doc/benchexec-in-container.md
incaseoftrouble: I just don't know how to do it with cgroups v1

doc/benchexec-in-container.md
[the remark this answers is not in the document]
PhilippWendler: Ok.

doc/benchexec-in-container.md
PhilippWendler: I think the current text is fine for this case.

incaseoftrouble:
All comments should be addressed

PhilippWendler (approved):
Great! Two small typos, otherwise ready to merge for me. Do you agree?

PhilippWendler:
Oh, I overlooked the conflicts. The changes from 901b50c303590e0fd12543f49a804e370f30b777 and 29480b32914356e182b598bf174e4c5ed2e63acc need to be merged here.

One is the removal of `-security-opt unmask=/sys/fs/cgroup` and the other is [this text](https://github.com/sosy-lab/benchexec/compare/6467aa5e1117155586f50928c17097fa87f341de..5d7a519db86f8ba2d70237e0805ce3b612ee53d3#diff-c60025a17472627516fb897916bc73880707de9aecab6b2f381c437fec7e8a88) that was added to `INSTALL.md`. I think most of it is covered already by the new file, but can you double check? Maybe what I wrote about manual creation of the cgroup makes sense to have in the background section?

PhilippWendler:
> > One is the removal of -security-opt unmask=/sys/fs/cgroup
> 
> With this (`podman run --cgroups=split --security-opt unmask=/proc/* --security-opt seccomp=unconfined -it ...`) I get `mkdir: cannot create directory '/sys/fs/cgroup/init': Read-only file system`. I kept this in for now; maybe we can find a solution?

Hm, maybe a version difference? I guess with the `unmask` it should work all the time.

> I merged the rest I think.

Thank you very much for your extensive help with all this! It is a great improvement of our documentation.

I will merge this now and of course also make sure to list you as a contributor

incaseoftrouble:
> Hm, maybe a version difference? I guess with the unmask it should work all the time.

`podman version 3.4.4`

> Thank you very much for your extensive help with all this! It is a great improvement of our documentation.

No problem, its helping me twofold, too. Esp. having runexec for my evaluation makes my life easier :)

> list you as a contributor for the next release.

thanks!

PR #1045, opened 2024-05-27
Create a tool-info module for BtorMC

Po-Chun-Chien:


benchexec/tools/template.py
```diff
@@ -137,6 +137,7 @@ def version(self, executable):
     def _version_from_tool(
         executable,
         arg="--version",
+        stdin=None,
```
Po-Chun-Chien, line 140: @PhilippWendler When getting the version,
the tool somehow requires passing EOF (ctrl-D) to its stdin.
(Otherwise, it won't terminate.)

So I added a new argument to `BaseTool2._version_from_tool`
for specifying stdin.
Or would you prefer not touching the template?
In that case, I will move everything into BtorMC's `version`,
instead of calling `_version_from_tool`.

PhilippWendler: Hm, this tool behavior is weird. If it requires EOF on stdin, then this should also mean that the tool similarly hangs when executed manually. Can you confirm this? If yes, I would suggest asking the tool developers to change this.

Independently, I wonder if we should change the behavior of BenchExec here. Without specifying stdin, the process inherits its stdin from the BenchExec process, i.e., it could end up receiving user input. This seems unwanted, confusing, and kind of breaks the isolation. Maybe we should just hardcode `subprocess.DEVNULL`? (Not in this PR, though.)

benchexec/tools/btormc.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable] + options + [task.single_input_file]
+
+    def determine_result(self, run):
+        """
+        @return: status of BtorMC after executing a run
+        """
+        if run.was_timeout:
+            return result.RESULT_TIMEOUT
```
PhilippWendler, line 39: Note that it is only necessary to do this if you want to make sure that the below code is not executed in timeout cases. This could be the case if the tool always prints out lines matching the patterns below, but they are meaningless in timeouts (though then you probably want to broaden this check to `was_terminated`).

It is not the responsibility of the tool-info module to check for timeouts.

benchexec/tools/template.py
Po-Chun-Chien: > Hm, this tool behavior is weird. If it requires EOF on stdin, then this should also mean that the tool similarly hangs when executed manually.

The tool takes as input a Btor2 circuit via a file or stdin.
If a file is given, then it won't wait for stdin, i.e., won't hang.

benchexec/tools/btormc.py
```diff
@@ -0,0 +1,47 @@
...
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for BtorMC -- A Bounded Model Checker for Btor2
+    """
+
+    REQUIRED_PATHS = ["build/"]
```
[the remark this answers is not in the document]
Po-Chun-Chien: I followed the instructions in Boolector's [README](https://github.com/Boolector/boolector?tab=readme-ov-file#linux-and-unix-like-os).
The compiled statically-linked binary of BtorMC is `build/bin/btormc`.
We could of course specify other directory to place the build artifacts.

> everything outside of it being irrelevant for execution

This is indeed the case for BtorMC.
Since the executable is statically-linked, only the binary itself is required.
So I think we could remove the `REQUIRED_PATHS` since no other file is needed.

benchexec/tools/btormc.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def determine_result(self, run):
+        """
+        @return: status of BtorMC after executing a run
+        """
+        if run.was_timeout:
+            return result.RESULT_TIMEOUT
+        for line in run.output[::-1]:
+            # BtorMC's option `--trace-gen` must be set to 1
+            # such that "sat/unsat" is printed to the output
```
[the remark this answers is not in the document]
Po-Chun-Chien: This option is set to 1 by BtorMC in its default mode,
but I could add it explicitly to `cmdline` as well.

benchexec/tools/template.py
Po-Chun-Chien: But it does seem like a mistake of the tool, because when specifying the other similar option `--copyright`,
BtorMC prints the message and terminates.
I'll ask the developer for clarification.

benchexec/tools/btormc.py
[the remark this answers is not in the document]
PhilippWendler: Ok, then I would tend to not enforce it. One could only consider checking it in `cmdline`, but even then I would say it is only a warning, not a failure, if the option is set to a different value. But likely other tools also have options that mangle the output in a way that the tool-info module would fail and we don't check for it.

Clarifying the comment that this is the default would be useful, though. Otherwise people like me think they have to do extra steps.

benchexec/tools/btormc.py
Po-Chun-Chien: Addressed in d90d0fe

benchexec/tools/btormc.py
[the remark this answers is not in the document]
Po-Chun-Chien: Addressed in 7427ddf

benchexec/tools/btormc.py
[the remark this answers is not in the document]
Po-Chun-Chien: Addressed in 78d4373

benchexec/tools/template.py
Po-Chun-Chien: I've implemented a fix myself and posed a [PR](https://github.com/Boolector/boolector/pull/225), but there is no response yet.

benchexec/tools/btormc.py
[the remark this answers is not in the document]
PhilippWendler: I think following this we should also remove `build/` from the executable search path. And maybe if there is really just the executable and nothing else, also remove `bin/`? If I take the tool and copy it somewhere, I would not expect to be forced to create a `bin/` subdirectory just in order to satisfy the tool-info module.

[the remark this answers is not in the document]
Po-Chun-Chien: If I understand `ToolLocator.find_executable` correctly,
it searches for executable under `./` and `./build/bin/` ([lines](https://github.com/sosy-lab/benchexec/blob/e22ca4443024a58dd7ea5828755acf833b028c53/benchexec/tools/template.py#L371-L374)).

If a user (like me), follows the build instructions to compile the tool,
he/she does not need to specify additional parameters to locate the executable when using BenchExec.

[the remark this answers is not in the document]
Po-Chun-Chien: Ok, now I just notice that when `--tool-directory` is specified,
`ToolLocator` only looks into `tool_dir/subdir/`.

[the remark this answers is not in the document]
Po-Chun-Chien: `subdir` removed in baadd81

benchexec/tools/template.py
Po-Chun-Chien: I would like to evaluate this tool in our paper.
It would be nice if this PR could be merged such that I can use the main repository of BenchExec instead of my own fork,

Should I drop the `version` function from BtorMC's tool-info module for now?

PhilippWendler: Given that it seems temporary, I wouldn't merge the current state of the PR with the changes to `template.py`. Because then I have to keep this API forever, and I am not sure I want that. But I could merge a PR with a `version` method if you do everything necessary there.

Po-Chun-Chien: > But I could merge a PR with a version method if you do everything necessary there.

Done

I also just noticed that my fix (https://github.com/Boolector/boolector/pull/225) has been approved.
I'll create another PR here after that has been merged.

PhilippWendler: I now thought more about https://github.com/sosy-lab/benchexec/pull/1045#discussion_r1617030736 and decided that there seems to be no risk in using `/dev/null` as input for all tools when getting the version number. So I just did that in 9997760a495853458f8cbcc9b56ac8a8b3f9cf05. This should work for all versions of BtorMC if you just call the standard utility method. Would you mind trying that out and updating the PR?

Sorry for the back-and-forth, but this is the best overall solution I think.

Po-Chun-Chien: No problem at all!

I've tested 773f3d5 and it works.

PR #1049, opened 2024-05-29
Update Btor2C's tool-info module

Po-Chun-Chien:
The existing tool-info module for Btor2C is pretty outdated.
The main changes of this PR include:

- Fix executable name
- Update letter cases of tool name
- Support version extraction

### DIFF
--- a/benchexec/tools/btor2c.py
+++ b/benchexec/tools/btor2c.py
@@ -12,14 +12,14 @@ import benchexec.tools.template
 
 class Tool(benchexec.tools.template.BaseTool2):
     """
-    Tool info for BTOR2C: A Converter from BTOR2 models to C programs
+    Tool info for Btor2C: A translator from Btor2 circuits to C programs
     """
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("btor2code", subdir="build")
+        return tool_locator.find_executable("btor2c", subdir="build")
 
     def name(self):
-        return "BTOR2C"
+        return "Btor2C"
 
     def project_url(self):
         return "https://gitlab.com/sosy-lab/software/btor2c"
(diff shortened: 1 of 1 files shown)


benchexec/tools/btor2c.py
```diff
@@ -12,25 +12,28 @@
 
 class Tool(benchexec.tools.template.BaseTool2):
     """
-    Tool info for BTOR2C: A Converter from BTOR2 models to C programs
+    Tool info for Btor2C: A translator from Btor2 circuits to C programs
     """
 
     def executable(self, tool_locator):
-        return tool_locator.find_executable("btor2code", subdir="build")
+        return tool_locator.find_executable("btor2c", subdir="build")
```
PhilippWendler, line 19: Rest of the changes look fine, but this is not backwards compatible. Was the old version only some preliminary beta, or was it publicly distributed and used? (Why not keep btor2c backwards compatible?)

Po-Chun-Chien: > Was the old version only some preliminary beta?

Yes, `btor2code` was just a preliminary version.
It was not distributed publicly.

benchexec/tools/btor2c.py
```diff
@@ -12,25 +12,28 @@
...
 
     def name(self):
-        return "BTOR2C"
+        return "Btor2C"
 
     def project_url(self):
         return "https://gitlab.com/sosy-lab/software/btor2c"
 
+    def version(self, executable):
+        return self._version_from_tool(executable, line_prefix="Btor2C version")
```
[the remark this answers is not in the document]
Po-Chun-Chien: The old version (when the binary was still `btor2code`) does not support `--version`.
The newer versions we used in TACAS papers support `--version`.
And ever since `--version` is introduced,
such prefix is printed.

PR #1054, opened 2024-06-12
Upgrade React Router to V6

### LINKED ISSUE #805: Upgrade to react-router 6
We are using react-router 5 in our HTML table, and [version 6 has been released](https://remix.run/blog/react-router-v6). Upgrading needs some work on our side ([guide](https://reactrouter.com/docs/en/v6/upgrading/v5)).

We have to replace `<Switch>` with `<Routes>` and change a few things inside it, but this seems easy.

More complicated is that we use the `useHistory` hook to `listen` on history changes in `ReactTable.js`, and `useHistory` is gone now. We might be able to use [`useLocation`](https://reactrouter.com/docs/en/v6/api#uselocation) or the [`history` package](https://www.npmjs.com/package/history) that is used by `react-router`.

However, we also directly access the browser history in some other components and do `window.history.pushState()` and `window.addEventListener("popstate", ...)`, and sometimes we also access the `history` property from `react-router`, which also seems to be gone. So this also needs changes and we should probably rewrite all this history handling to use the same kind of code based on react-router.

EshaanAgg:
Upgrades from React Router V5 to React Router V6.

The primary changes included in this PR:
- Converted the `App` component from a class component to a functional component:
I did it while experimenting with different ways to use hooks like `useNavigate` and `useLocation` in the `Overview` component. In the end, I did not need it, but it is a good change to keep in my opinion. 
- Changed the `Switch` components to `Route` components
- Changed all the existing instances of using the `history` object to other means

Initially, I tried converting the `Overview` component to a functional one. Still, the same resulted in many breaking changes as I am unfamiliar with what and how each function works. I think so I have found a way that allows us to continue using class components, which integrating it with hooks, that doesn't involve too much of a rewrite and refactoring. 

Fixes #805.


EshaanAgg:
The immediate issue I am facing is that the tests for `QuantilePlot` seem to fail now and I have no clue what is going wrong with it. Would really appreciate any help with the same.

PhilippWendler:
Seems to work seamlessly in my first tests. Very nice! And a much smaller PR than I had feared. :-)

> The immediate issue I am facing is that the tests for `QuantilePlot` seem to fail now and I have no clue what is going wrong with it. Would really appreciate any help with the same.

There are several errors with the message "Error: Not implemented: navigation (except hash changes)" shown in the log. My guess would be that something attempts to change the URL. There are also stacktraces available for these errors that should help.

EshaanAgg:
Yes sir. I have been trying to figure out this error and tried again all the solutions that I had researched when I [faced this problem earlier while working on different PR](https://github.com/sosy-lab/benchexec/pull/1006#issuecomment-1976388147), but to no avail. I have some other ideas and hope to complete my experimentation with them by this weekend. I will keep adding updates on my progress here.

EshaanAgg:
Update: I got the URL navigation working! We are on a particular version of Jest, which doesn't allow overwriting the `window.location` object. Still, we can use the `window.history.pushState` function to set the value of the `window.location.href` as needed. All the tests now run successfully :)

There is another minor hiccup, though. Some snapshots for the `Quantile.js` component seem to have changed, but I have not changed the component's markup. I tried manually using the various interactions mocked in the test in the browser, and they seem correct. I have the following questions now:
- Is the change in markup possible because we introduced the `NavSync` component in the Overview component?
- Is there a way to visually check the Quantile snapshots and the current renders so we can see the difference between them and compare them?
- Can anybody else set up this branch locally, test out the `Quantile` tab, and let me know if everything behaves as expected? If it does, then I feel we can update the snapshots with the current render.

PhilippWendler:
> Update: I got the URL navigation working! We are on a particular version of Jest, which doesn't allow overwriting the `window.location` object. Still, we can use the `window.history.pushState` function to set the value of the `window.location.href` as needed. All the tests now run successfully :)

:tada: 

> There is another minor hiccup, though. Some snapshots for the `Quantile.js` component seem to have changed, but I have not changed the component's markup. I tried manually using the various interactions mocked in the test in the browser, and they seem correct. I have the following questions now:
> 
> * Is the change in markup possible because we introduced the `NavSync` component in the Overview component?

Hm. Could be, but if I look at the [start of the changes](https://github.com/sosy-lab/benchexec/actions/runs/9541855559/job/26295677656?pr=1054#step:6:87) I see that the test now renders a different column in the plot. This might look more like if some state change is not picked up? Note that the test is named "Quantile Plot should match HTML snapshot with selection of the type status" and we previously had the status column, but now there is a different column.

> * Is there a way to visually check the Quantile snapshots and the current renders so we can see the difference between them and compare them?

Not that I am aware of. Maybe google for this?

> * Can anybody else set up this branch locally, test out the `Quantile` tab, and let me know if everything behaves as expected? If it does, then I feel we can update the snapshots with the current render.

I think it is more likely that the test is not behaving correctly right now due to the mismatch between test name and test result.

EshaanAgg:
Got it sir. Will try to fix this as soon as possible.

EshaanAgg:
I have fixed the errors in the test and the application. This PR is now ready for review!

benchexec/tablegenerator/react-table/src/tests/QuantilePlot.test.js
```diff
@@ -65,25 +78,29 @@ files
...
             expect(plot).toMatchSnapshot();
           },
         );
       });
 
       describe("Direct Plot should match HTML snapshot", () => {
-        setURLParameter({ plot: plotInstance.plotOptions.direct });
+        console.log(window.location.href);
+        updateURLParams({ plot: plotInstance.plotOptions.direct });
+        console.log(window.location.href);
```
PhilippWendler, line 97:
```suggestion
        updateURLParams({ plot: plotInstance.plotOptions.direct });
```
Let's remove this if everything is working now in order to not clutter the output?

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -253,7 +247,7 @@ export default class Overview extends React.Component {
       clearImmediate(this.lastImmediate);
     }
     this.lastImmediate = setImmediate(() => {
-      this.filterUrlSetter(filter, this.routerRef.current.history);
+      this.filterUrlSetter(filter, window.history);
```
PhilippWendler, line 250: I see that here we switch to `window.history`, and in the tests `document.history` was changed to `window.history` as well. I don't know that much about such details, would you mind explaining what is the background of this such that I can learn something?

PhilippWendler:
Thanks a lot!

I found two related problems with state updates. If I configure a filter in the filter sidebar and keep the sidebar open while changing tabs, the filter is lost. Similarly, if I change a filter via the drop down and then change tabs, the filter is lost. Here is a video of the former case:
[Bildschirmaufzeichnung vom 18.06.2024, 16:30:33.webm](https://github.com/sosy-lab/benchexec/assets/2545335/eb6f8964-2e5e-478e-ba78-7ed407ac8481)

Could you please have a look?

Aside from this, everything seems to work fine. :+1:

benchexec/tablegenerator/react-table/src/tests/QuantilePlot.test.js
EshaanAgg: Sure sir! I forgot to remove it post-debugging.

benchexec/tablegenerator/react-table/src/components/Overview.js
EshaanAgg: Sure sir. In the context of web development, `document` and `window` are two distinct objects provided by the browser's JavaScript environment. 

The `window` object represents the browser window or the frame that contains the document. It is the global object in the browser's JavaScript environment, meaning all global variables and functions are properties of `window`. Thus it is able to expose properties like `window.history` and `window.location` that we rely on to navigate (ideally these should be done via `React Router V6`, but due to the hooks and class component problem, it makes most sense to form the URLs ourselves and navigate using the URLs directly).

The `document` object represents the HTML or XML document that is loaded in the browser window. It is a property of the `window` object (`window.document`) itself and provides access to the DOM (Document Object Model) of the page. It is meant to access the content of the page (via query selectors) and not history or navigation. There is no property like `document.history`

EshaanAgg: Mostly the `document` and `window` point to the same URL, so we can access either `document.location.href` or `window.location.href`, but they may be different in some cases (like modals). So it is better to ensure consistency, and preferably use `window` for navigation, till you have a really stong use for `document`.

Also the `this.routerRef.current.history` is a `React Router V5` specific property which maps to the `window.history`, but now stands deprecated. So we are now accessing it directly.

EshaanAgg:
Okay, sir! Will try to patch the same and will update here.

benchexec/tablegenerator/react-table/src/components/Overview.js
PhilippWendler: Ah, I had mixed up `document.location` and `window.history`. Thanks!

PhilippWendler:
I still get some weird behavior when clicking around with filters. For example, the drop down filter first doesn't update and then the UI seems to lag one filter behind what is actually being applied as a filter. Similar with the sidebar. Can you reproduce or should I make a video?

benchexec/tablegenerator/react-table/src/utils/utils.js
```diff
@@ -267,23 +267,16 @@ export const constructHashURL = (url, params = {}) => {
...
-  const { newUrl, queryString } = constructHashURL(
-    document.location.href,
-    params,
-  );
-  if (history && history.push) {
-    history.push(queryString);
-    return;
-  }
-  document.location.href = newUrl;
+const setURLParameter = (params = {}, history = undefined) => {
```
PhilippWendler, line 276: Remove `history`?

EshaanAgg:
> I still get some weird behavior when clicking around with filters. For example, the drop down filter first doesn't update and then the UI seems to lag one filter behind what is actually being applied as a filter. Similar with the sidebar. Can you reproduce or should I make a video?

I'm aware of this bug and faced it yesterday, too. But with the current changes, the same seems resolved on my end. Maybe there is some difference in the version of the code changes or cache?

https://github.com/sosy-lab/benchexec/assets/96648934/cc6a2853-2de3-45d1-8682-f7db56411fbe

benchexec/tablegenerator/react-table/src/utils/utils.js
EshaanAgg: Yes! It is not used in the codebase now.

PhilippWendler:
I can reproduce this with 224c784e, both with `npm run start` and with the build files that are checked in into the repository. Maybe it is a browser issue? I am using Firefox.

EshaanAgg:
It sure seems like it! It works well on Chromium-based browsers like Chrome and Brave, but Firefox seems different. If probably is due to how these browsers handle the `pushState` actions and syncing of state based on it. I will install Firefox and try to debug this there.

EshaanAgg:
@PhilippWendler, I have pushed the changes related to Firefox. 

I also encountered a small bug that occurred if the root URL did not include a `#` (occurred when you visited `localhost:3000/` instead of `localhost:3000/#/`): 
1. Visit the homepage
2. Change the columns to show using `Select Columns` 
3. Navigate to some other page 
This would cause the URL to become malformed, with the filters being added twice to the URL. Have pushed a small fix for that as well.

benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -169,18 +170,16 @@ export default class Overview extends React.Component {
...
+      mounted: true,
+    });
+    this.updateState();
 
-  componentWillUnmount() {
-    this.removeHistoryListener();
+    // window.history.onpushstate = () => {
+    //   this.updateState();
+    //   this.updateFiltersFromUrl();
+    // };
```
PhilippWendler, line 182:
```suggestion
```

PhilippWendler:
Thanks! I can confirm that this problem is fixed.

I have one little request though: When I configure a status filter and then change it such that all checkboxes are clicked in the sidebar, the filter box immediately vanishes. To further configure a status filter, I have to re-add the filter. Here you can see this at ca. 50% of the video:

https://github.com/sosy-lab/benchexec/assets/2545335/4376e708-ef57-424c-b432-e1eaed063c27

Previously this was different, and more convenient. Can we restore the old behavior?

Also, I noticed that the statistics render twice when I change a filter. You can see a short flicker in the video. I do not know if this is actually caused by your changes or whether this was already the case previously, but maybe we can investigate and fix this anyway? (potentially as a separate PR if not caused by the current set of changes)

I played the video on slow motion and managed to get a screenshot of the intermediate state:

image

This is actually even slightly wrong, the row "local summary" should never be shown if filters apply. "local summary of selected rows" is wrong, we do not have that.

EshaanAgg:
Update: I had been working on this over the weekend and have been able to fix the screen loading twice bug. The other behavior involving the vanishing checkbox on all filters being selected is tricker to handle, as:
- Having all filters set is equivalent to having no filters, and thus we don't push anything in the URL for the same
- The state of the component is automatically synced from the URL
These two opposite interactions are posing somewhat of a hiccup. I am still trying out to figure out a solution to the same.

PhilippWendler:
I understand. Maybe some temporary state (not in the URL) that is only kept while the sidebar is open? Such that any filter boxes that are present are kept until they are explicitly deleted with the trash can buttons or by closing the sidebar?

Or check how the old code handles this situation?

EshaanAgg:
Yes sir I'm trying both of these options. Not really sure how it functioned before the upgradation (seems like an unintentional bug which evolved as a feature), but I'll give the previous code a glance again alongside exploring  the option of storing the same in the state.

EshaanAgg:
@PhilippWendler Sir, I tried a couple of options to retain the behavior via state, but the same would not work as the whole `FilterBox` component is rendered as the `filters` are changed for the URL. I even tried multiple approaches involving comparing the values of `prevState` and current `state` in the `ComponentDidUpdate` handlers, but since we have no strict control over the rendering order chosen by React, whatever logic I came up with seemed to fill at one place of the other. In net, I can't figure out any suitable way to achieve the same as per my explorations in the past week.

An easier fix would change the serialization of the URL, as I have demoed in the latest commit. This does introduce an extra `notIn()` clause in the URL, but I think the same is worth having as since we have strongly coupled the state of our filters to the URL, and thus it makes sense that the information about which filters should be displayed remain there. I haven't updated the tests for this serialization change, and thus we have the failing build pipeline. Would like to receive your views on the same.

PhilippWendler:
Hm, are you referring to 1b099909c679cc981686e8e468ae300f0beb6544? I don't really understand what it does and how it introduces `notIn`. Instead, it seems to remove NavSync which I thought was necessary according to your explanations?

I think increasing the filter complexity just to handle some UI issues could bite us later. Is there no way to have some local state in the filter card or the filter sidebar that stores information that a card should be shown? And during rendering we would either show the respective card if the filter tells us to do so or this separate state? And when closing the sidebar this state is reset.

EshaanAgg:
> Hm, are you referring to [1b09990](https://github.com/sosy-lab/benchexec/commit/1b099909c679cc981686e8e468ae300f0beb6544)? I don't really understand what it does and how it introduces `notIn`. Instead, it seems to remove NavSync which I thought was necessary according to your explanations?

Yes sir. I had initially added the same as I felt the same was important to sync the state, but then making the Firefox compatibility changes (like adding `callbacks` arguments to `setURLParameter`), I realized via testing manually that the job of the component was already being well handled and we do not need to introduce the same. 

As for the change with the `notIn`, with the current implementation, there would be a small change in how the URL serializer works. Earlier the same would be like:
- User opens the app and applies a status filter -> `http://localhost:3000/#/?filter=0(0*status*(status(notIn()),category(notIn(correct))))`
- User selects all the options in the status filter -> `http://localhost:3000/#/`
- User deletes the status filter -> `http://localhost:3000/#/`

With the current change, the same would be the following:
- User opens the app and applies a status filter -> `http://localhost:3000/#/?filter=0(0*status*(status(notIn()),category(notIn(correct))))`
- User selects all the options in the status filter -> `http://localhost:3000/#/?filter=0(0*status*(status(notIn()),category(notIn())))`
- User deletes the status filter -> `http://localhost:3000/#/`

There is an extra `status(notIn()),category(notIn())` in the URL bar in step 2 that is basically acting as our state that allows for the syncing to be done and says that "show the status selection filter but with the default value of all checked." The changes corresponding to the same are in the `utils.js` in the `makeStatusColumnFilter` function. 

> I think increasing the filter complexity just to handle some UI issues could bite us later. Is there no way to have some local state in the filter card or the filter sidebar that stores information that a card should be shown? And during rendering we would either show the respective card if the filter tells us to do so or this separate state? And when closing the sidebar this state is reset.

Even this is the approach that I had spent the most time getting to work, but couldn't get the syncing of the state to work as expected. Even I feel that it should be possible, but I have trouble getting it to work and require assistance with the actual code.

PhilippWendler:
> As for the change with the notIn, with the current implementation, there would be a small change in how the URL serializer works.

So no new part of the filter language, just a difference in what filters are put into the URL in which situation?
Then my fear of increased complexity is resolved, because anything that we can express with the existing filter logic is certainly less complexity than new local state in the component.

> With the current change, the same would be the following:
> 
>   * User opens the app and applies a status filter -> `http://localhost:3000/#/?filter=0(0*status*(status(notIn()),category(notIn(correct))))`
> 
>   * User selects all the options in the status filter -> `http://localhost:3000/#/?filter=0(0*status*(status(notIn()),category(notIn())))`
> 
>   * User deletes the status filter -> `http://localhost:3000/#/`

Hm. It would be consistent with how numeric and text filters work. That seems like a good thing actually.

I would have thought that it is stiller nicer if the irrelevant filter is deleted once the sidebar is cleared, but now that I noticed that irrelevant numeric and text filters are not deleted, we can also handle this for all kinds of filters later.

So I would say, yes, let's move forward with your suggestion. Overall I think it is a good choice and probably the best we have. Thanks a lot with coming up with it and for the explanations!

EshaanAgg:
> So no new part of the filter language, just a difference in what filters are put into the URL in which situation? 

Yes sir. Some URLs might be a bit longer now, but it is a good thing in my opinion as now we have two different states:
- Not having a filter
- Having a filter (even though that filter might not actually be "filtering" anything, but still it is active nonetheless)

> Hm. It would be consistent with how numeric and text filters work. That seems like a good thing actually.

That's a nice unintended positive change!

> I would have thought that it is stiller nicer if the irrelevant filter is deleted once the sidebar is cleared, but now that I noticed that irrelevant numeric and text filters are not deleted, we can also handle this for all kinds of filters later.

Sounds great! I will wrap this PR soon, including updating the tests and a final round of manual testing. Hopefully all should be good post that. Thanks for the continuous feedback and reviews on this PR!

EshaanAgg:
I seem to have messed up the git history of this branch (I rebased `main` into it instead of merging it, and thus, now all previous commits show me as an author as well). Will close this PR, and open a new one so that the history is maintained. Apologies for the same!

PR #1055, opened 2024-06-14
Update CPAchecker tool info module

PhilippWendler:
Prepare the tool-info modules of CPAchecker and CPA-Witness2test for upcoming changes:
- https://gitlab.com/sosy-lab/software/cpachecker/-/issues/632
- https://gitlab.com/sosy-lab/software/cpachecker/-/issues/1218

### DIFF
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -142,11 +142,15 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "https://cpachecker.sosy-lab.org/"
 
     def _get_additional_options(self, existing_options, task, rlimits):
+
+        def option_present(option):
+            return f"-{option}" in existing_options
+
         options = []
-        if rlimits.cputime and "-timelimit" not in existing_options:
+        if rlimits.cputime and not option_present("timelimit"):
             options += ["-timelimit", f"{rlimits.cputime}s"]
 
-        if "-stats" not in existing_options:
+        if not option_present("stats"):
             options += ["-stats"]
 
         if task.property_file:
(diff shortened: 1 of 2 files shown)


PhilippWendler:
@lembergerth Would you mind having a quick look whether I overlooked something and whether the logic looks sane?

I want to get this into the next release of BenchExec such that I can then put it into the CPAchecker repository. Thus this is done before the actual change in CPAchecker.

The most important open question is whether some of the options used here would likely be [renamed](https://gitlab.com/sosy-lab/software/cpachecker/-/issues/1218#note_1950486702). But the logic here would be less complex if we avoid that and stick with `--stats`.

benchexec/tools/cpachecker.py
```diff
@@ -121,23 +142,31 @@ def project_url(self):
         return "https://cpachecker.sosy-lab.org/"
 
     def _get_additional_options(self, existing_options, task, rlimits):
+        # If at least one option uses "--" style, we use that for options added here.
+        # Otherwise use "-" to be safe for old versions of CPAchecker.
+        new_option_style = any(option.startswith("--") for option in existing_options)
```
lembergerth, line 147: How about providing `_get_additional_options` with the executable, and use that as an additional case (or the only case?) for `new_option-style`?
That would differentiate between the option style based on the implied CPAchecker version and not require a heuristic.

lembergerth (approved):
Looks good to me :-)
Just one non-blocking idea as an alternative

benchexec/tools/cpachecker.py
PhilippWendler: You mean based on `executable.endswith("cpa.sh")`? I started with this at first, actually, but then switched to the current style to have command lines in a consistent style even if people use old-style arguments with a new CPAchecker version (which I guess will happen a lot because people have a lot of benchmark definitions that they will continue to use, or take a benchmark definition from last SV-COMP).

I also think the current decision is less of a heuristic than one based on the executable: If there is an argument with `--` in the command line, then we certainly have a CPAchecker version that supports them. But one could in principle cherry-pick the new executable in a branch of CPAchecker without also cherry-picking the new command-line arguments.

lembergerth: Ok :-)

PhilippWendler: FTR: https://gitlab.com/sosy-lab/software/cpachecker/-/issues/1218#note_1952602398 is also an argument for the current implementation.

PR #1056, opened 2024-06-14
Add --architecture flag to theta

leventeBajczi:
We have some wrong results regarding the architecture, because we didn't pass this flag to Theta beforehand. This fixes it.

### DIFF
--- a/benchexec/tools/theta.py
+++ b/benchexec/tools/theta.py
@@ -45,7 +45,6 @@ class Tool(benchexec.tools.template.BaseTool2):
         if run.was_timeout:
             return result.RESULT_TIMEOUT
         status = result.RESULT_UNKNOWN
-        
         for line in run.output:
             if "SafetyResult Unsafe" in line:
                 status = result.RESULT_FALSE_REACH
--- a/benchexec/tools/theta.py
+++ b/benchexec/tools/theta.py
@@ -34,10 +34,18 @@ class Tool(benchexec.tools.template.BaseTool2):
         # Theta supports data race and unreach call
         if task.property_file:
             options += ["--property", task.property_file]
+        if isinstance(task.options, dict) and task.options.get("language") == "C":
+            data_model = task.options.get("data_model")
+            if data_model:
+                options += ["--architecture", data_model]
+
         return [executable, task.single_input_file] + options
 
     def determine_result(self, run):
+        if run.was_timeout:
+            return result.RESULT_TIMEOUT
         status = result.RESULT_UNKNOWN
+        
         for line in run.output:
             if "SafetyResult Unsafe" in line:
                 status = result.RESULT_FALSE_REACH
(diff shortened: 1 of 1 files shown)


benchexec/tools/theta.py
```diff
@@ -34,9 +34,16 @@ def cmdline(self, executable, options, task, rlimits):
...
+        if isinstance(task.options, dict) and task.options.get("language") == "C":
+            data_model = task.options.get("data_model")
+            if data_model:
+                options += ["--architecture", data_model]
+
         return [executable, task.single_input_file] + options
 
     def determine_result(self, run):
+        if run.was_timeout:
+            return result.RESULT_TIMEOUT
```
PhilippWendler, line 46: Note that you do not strictly need this change, BenchExec will anyway change the result to `TIMEOUT` if the run is marked as such. You can add this check if you want `determine_result` to not run or return some confusing result like "crashed" or so, though.

PhilippWendler:
Was `--architecture` already available in Theta previously, and just missing from the command line? Or is it new? In the latter case, the change would make old versions of Theta difficult to execute with BenchExec.

benchexec/tools/theta.py
leventeBajczi: Because we have the "Parsing OK" status which is almost always given early (right after parsing the input, before the analysis), I don't want to get many "TIMEOUT (Parsing OK)" results, which I think this fixes. At least that was the main motivation behind this.

leventeBajczi:
> Was `--architecture` already available in Theta previously, and just missing from the command line? Or is it new? In the latter case, the change would make old versions of Theta difficult to execute with BenchExec.

This has been available to Theta since our original entry in SV-COMP 2022. We were debugging some erroneous results and it turned out we have been using ILP32 for every benchmark run, hence the wrong results. I think that all useful versions of Theta (which handle C verification somewhat well) handle this flag, so I'm not really worried about compatibility.

benchexec/tools/theta.py
PhilippWendler: Yes, this is the use case for `was_timeout`. But you might also want to consider `was_terminated` then? Helps you to prevent `OUT OF MEMORY (Parsing OK)`.

Simply returning `result.RESULT_ERROR` from inside `if run.was_terminated:` is intended to work nicely (BenchExec will override the `ERROR` with the appropriate reason without producing something useless like `TIMEOUT (ERROR)`, but it will still produce something like `TIMEOUT (SEGMENTATION FAULT)` (which the current solution does not, besides handling only timeouts)).

leventeBajczi: I see, that's a good point. I've added some code to handle `was_terminated` as well.
Let me know if you think this flow could still be improved!

PhilippWendler: `was_timeout` is a subset of the cases covered by `was_terminated`, I recommend to just remove the former.

leventeBajczi: Oh, sorry, I did not know that. Fixed.

PhilippWendler (approved):

PR #1059, opened 2024-06-28
Upgrade React Router to V6

### LINKED ISSUE #805: Upgrade to react-router 6

EshaanAgg:
Reimplementation of all the code and discussion in #1054, which was closed due to wrong history errors. 

Fixes #805.


benchexec/tablegenerator/react-table/src/components/Overview.js
```diff
@@ -370,62 +365,78 @@ export default class Overview extends React.Component {
...
-                  />
-                </Route>
-                <Route path="/info">
-                  <Info
-                    version={this.props.data.version}
-                    selectColumn={this.toggleSelectColumns}
-                  />
-                </Route>
-              </Switch>
+              {/* Component to sync the navigation state. Does not render anything. */}
```
PhilippWendler, line 368:
```suggestion
```

PhilippWendler (commented):
Filters seem to work nicely.

However, when I try to change anything about a plot configuration, the URL updates but the rendered UI does not. I need to refresh the page to make it work. (Again in Firefox)

Can you reproduce this?

EshaanAgg:
@PhilippWendler Thanks for the catch. I have pushed some changes, and now all the plots seem to be working on my end. Can you confirm the same?

PhilippWendler (approved):

PhilippWendler:
Ah, now I found some more rendering problems: If I use the filters in the table header, it seems to render twice. For example, when I use the drop down, I first see a glimpse of the previous filter before it renders again. And if I use one of the filters with text entry, the rendering causes the text field to loose focus, so I cannot type continuously. Can you please have a look at this?

EshaanAgg:
Can you please attach a video of the same? I tried to apply a `status` filter and then type a query in the `task name` field in the task filter, and it seems to work fine on my end on Firefox. Or am I not replicating it properly?

PhilippWendler:
Sure:

https://github.com/sosy-lab/benchexec/assets/2545335/24f1e37f-7c2c-482d-9227-ad580cc1a823

This is on fee40a6f.

EshaanAgg:
Got it, sir! I am trying the fixes for these and will update here as soon as I make progress.

EshaanAgg:
I tried multiple approaches over the weekend, including:
- Trying to move the filter components to be top-level components and using their memorized versions to prevent rerenders
- Using a state variable called `focusedElement` to check the last focused element, and trying to restore the focus. 
- Changing the `useCallback` hooks used in multiple ways to prevent the re-renders
but I couldn't develop a solution that could preserve the focus without leading to some weird, incorrect filtering-related bugs. I am still trying my best here, but I am running out of ideas of how the same can be tackled. 

PS. I sincerely apologize for the extra time this PR is taking. The migration had posed various changes, which I had not initially envisioned even remotely.

PhilippWendler:
Don't worry, and I think that you have done great work so far! I am sorry that I cannot help you with these technical problems. Unfortunately, the focus loss is kind of a big UI disadvantage, so we need to treat it as a blocker for merging it.

I remember that on the original implementation we also had some trouble to get the focus right, but eventually managed. Maybe you can find the relevant commits in the git history?

Do you think this creates a time problem for your project roadmap?

EshaanAgg:
Yes sir. I am trying some other ways that could fix the issue. Thanks for all the support. I'll try to browse the history and what pops up. I don't think this should cause a problem for the roadmap right now, and I can make up for these weeks in the buffer time planned.

EshaanAgg:
@PhilippWendler Can you please take a look? Fixed the focus problem by defining a state on the `ReactTable` component level and restoring it between re-renders. It also ensures that the same does not cause unnecessary re-renders by memoizing the component at each level.

PhilippWendler:
Nice, thank you very much! I guess it is anyway better to have the `FilterInputFieldComponent`s as independent top-level definitions. Maybe we move them even into a separate file?

The only thing that is not perfect yet is the double re-rendering that happens when the drop-down is changed. Is this easy to fix? Otherwise I would say this is ready to merge. :tada:

EshaanAgg:
Yes sir. Moving them to separate files is a better idea as it reduces the associated complexity of the base `ReactTable` component. I have done the same, as well as memorized the `StatusFilter` as well to avoid re-renders. Can you please confirm if the issue is now resolved (I can't seem to reproduce the re-rendering bug locally on either Chrome or Firefox, and thus I'm uncertain if it's fixed).

PhilippWendler (approved):
Very nice! Thank you a lot, this works really well! Congratulations for your good work! :rocket: 

I am glad that we can finally close this issue and solve this part of our technical debt backlog!

PR #1061, opened 2024-07-01
Create a tool-info module for Pono

Po-Chun-Chien:


PhilippWendler:
Can you confirm that there is no way to get some version number or similar from Pono? Tables without a version number can easily become confusing and lead to wrong interpretations of results, so whenever possible we strive to have this information accessible via the tool-info module.

Po-Chun-Chien:
> Can you confirm that there is no way to get some version number or similar from Pono?

Yes, Pono's executable does not support `--version` (or similar).
AFAIK, the only way to retrieve version is by `git describe`.

PR #1066, opened 2024-07-26
Add key tool info module

ricffb:
This MR adds the tool info module for the KeY CLI (a non interactive mode of the KeY Verifier)

### DIFF
--- a/benchexec/tools/key_cli.py
+++ b/benchexec/tools/key_cli.py
@@ -39,5 +39,8 @@ class Tool(template.BaseTool2):
     def name(self):
         return "KeY"
 
+    def project_url(self):
+        return "https://www.key-project.org/"
+
     def cmdline(self, executable, options, task, rlimits):
-        return super().cmdline(executable, options, task, rlimits)
+        return [executable, *options, task.single_input_file]
--- a/benchexec/tools/key_cli.py
+++ b/benchexec/tools/key_cli.py
@@ -1,3 +1,10 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2024 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
 import re
 from pathlib import Path
 
--- a/benchexec/tools/key_cli.py
+++ b/benchexec/tools/key_cli.py
@@ -1,0 +1,37 @@
+import re
+from pathlib import Path
+
+from benchexec.tools import template
+
+
+class Tool(template.BaseTool2):
+    """
+    This tool-info module runs Cheetah, a parallel portfolio
+    of natively compiled CPAchecker instances.
+    """
+
+    REQUIRED_PATHS = ["bin", "lib"]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("key-cli", subdir="bin")
+
+    def version(self, executable):
+        lib = Path(executable).parent.parent / "lib"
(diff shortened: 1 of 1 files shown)


benchexec/tools/key_cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def version(self, executable):
+        lib = Path(executable).parent.parent / "lib"
+        for file in lib.glob("key-*-exe.jar"):
+            match = re.match(r"key-(.*)-exe.jar", file.name)
+
+            if match:
+                return match.group(1)
+
+        return "unknown"
```
PhilippWendler, line 32: Empty string is better here.

benchexec/tools/key_cli.py
```diff
@@ -0,0 +1,43 @@
...
+    This tool-info module runs KeY CLI, a cli version of the KeY deductive verifier.
+    """
+
+    REQUIRED_PATHS = ["bin", "lib"]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("key-cli", subdir="bin")
+
+    def version(self, executable):
+        lib = Path(executable).parent.parent / "lib"
```
PhilippWendler, line 25: If `key-cli` is in `bin`, this uses `bin/../../lib`. Are you sure that it is correct? Usually I would expect `lib` to be a sibling of `bin`. What is the directory layout of this tool?

benchexec/tools/key_cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+        return "unknown"
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def name(self):
+        return "KeY"
```
PhilippWendler, line 40: Can you also add the method for the webpage of the tool? Certainly KeY has a webpage.

benchexec/tools/key_cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        return super().cmdline(executable, options, task, rlimits)
```
PhilippWendler, line 43: Is this correct? I.e., does `key-cli` accept tasks without input files or with more than one input file?

benchexec/tools/key_cli.py
ricffb: I tested this with `benchexec.test_toolinfo` in an environment where no `bin/../../lib` is present and the test output showed the correct version.
Just like you suggested `lib` is a sibling to `bin`

```
archive/
    bin/key-cli
    lib/key-2.13.0-exe.jar
...
```

My assumption is that `executable` is the path pointing to the executable itself, i.e. `archive/bin/key-cli`
Then `Path(executable).parent` returns `archive/bin` and `Path(exectuable).parent.parent` to `archive` as I expect.

benchexec/tools/key_cli.py
ricffb: Good to know. I will change this.

benchexec/tools/key_cli.py
PhilippWendler: Oh, yes, sure. I don't use `Path` regularly and hadn't thought about what `parent` means for a path to a file, not to a directory.

benchexec/tools/key_cli.py
PhilippWendler: If unsupported, then yes. If you use `task.single_input_file` this will automagically happen.

benchexec/tools/key_cli.py
ricffb: done 👍

PR #1071, opened 2024-08-09
Add tool-info module for super_prove

Po-Chun-Chien:


benchexec/tools/super_prove.py
```diff
@@ -0,0 +1,57 @@
...
+import benchexec.tools.template
+from pathlib import Path
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for super_prove: A portfolio model checker based on ABC
+    - project URL: https://github.com/berkeley-abc/super_prove
+    - build repository: https://github.com/sterin/super-prove-build
+    - tool archive: https://gitlab.com/sosy-lab/benchmarking/coveriteam-archives/hardware-verifiers/-/blob/main/super_prove.zip
```
PhilippWendler, line 18: That's just an internal repository for being able to execute a few tools with our CoVeriTeam service AFAIK. I don't think it is intended as a public and permanent service. I would remove the URL.

Or is this tool-info module specific to that archive?

And is it specific to that super_prove fork in https://github.com/sterin/super-prove-build?

benchexec/tools/super_prove.py
```diff
@@ -0,0 +1,57 @@
...
+        return tool_locator.find_executable("super_prove.sh", subdir="bin")
+
+    def name(self):
+        return "super_prove"
+
+    def project_url(self):
+        return "https://github.com/berkeley-abc/super_prove"
+
+    def version(self, executable):
+        version_file = Path(executable).parent.parent / "VERSION.txt"
```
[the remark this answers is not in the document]
Po-Chun-Chien: > Is there no way to get the version from the executable?

No, unfortunately. That's why I wrote the method this way.
I could get the version of ABC (a component of super_prove) from the executable,
but that version is different from super_prove's version.

benchexec/tools/super_prove.py
Po-Chun-Chien: Sure, I can remove the archive URL.

> And is it specific to that super_prove fork in https://github.com/sterin/super-prove-build?

`super-prove-build` is not a fork.
It is *the* repository that can be used to build the tool.
`super_prove` is a submodule of `super-prove-build`.
Or should the project URL points to `super-prove-build` instead of `super_prove`?

PhilippWendler: It is somewhat weird that `super-prove-build` lives in a different GitHub org than `super-prove`, so it seems more like an (unofficial) fork (not in the strict sense of a GitHub fork but in the broader sense of a project fork) instead of the official way to build `super-prove`. Is there some official information connecting `super-prove-build` to `super-prove` and clarifying its status by the `super-prove` authors? Or is there an author overlap?

I want to avoid having BenchExec treat `super-prove-build` as the canonical `super-prove` representation if that is not official. (Worst case is that `super-prove-build` does something bad or unwanted by the `super-prove` authors, someone uses BenchExec to benchmark `super-prove` and then someone else complains that this is not really `super-prove` although BenchExec claims so.)

If `super-prove-build` is unofficial but the community generally treats it as official (e.g., other papers using it as `super-prove`), that would also be enough for me.

The project URL is part of this question, but also the name of the tool-info module.

Po-Chun-Chien: > Is there some official information connecting super-prove-build to super-prove and clarifying its status by the super-prove authors? Or is there an author overlap?

`sterin` is the main developer of both `super-prove-build` and `super_prove`.

> If super-prove-build is unofficial but the community generally treats it as official (e.g., other papers using it as super-prove), that would also be enough for me.

I believe this is the case.
I was pointed to the `super-prove-build` repository by the maintainer of ABC (https://github.com/berkeley-abc/abc).
ABC in the backend solver of super_prove, and both are from the same research group at UC Berkeley (`berkeley-abc`).

> The project URL is part of this question, but also the name of the tool-info module.

The tool name "super_prove" is used in the literature and also recognized by the community.

PhilippWendler: > `sterin` is the main developer of both `super-prove-build` and `super_prove`.

Ok, then everything is fine.

PR #1072, opened 2024-08-16
Extend LXCFS integration

PhilippWendler:
We already use LXCFS to provide better isolation of the containers, and virtualize `/proc/uptime`. But LXCFS can do more and for example also provide a virtualized view on the CPU-core information provided by the kernel, such that applications see only cores that they are allowed to use. This was incompletely implemented and not working so far.


PhilippWendler:
@schroeding A code review and testing in as many scenarios as possible would be good.

PhilippWendler:
Thanks, also for explaining the LXCFS problem. I think the problem does not look bad enough that we need a workaround or so.

PR #1078, opened 2024-09-16
Handle validation tasks inside CPAchecker

marian-lingsch:
The goal of this MR is to handle validation tasks following the [proposed format](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/witness-benchmark-set/c/loop-invariants/linear-inequality-inv-c.2.witness-validation.yml?ref_type=heads) in SV-Benchmarks. A short description of the new option is given in the [SV-Benchmarks README](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/witness-benchmark-set/README.md?ref_type=heads#task-definitions)

The goal is to leave the previous behavior of CPAchecker untouched i.e. it still being able to validate witnesses given explicitly through extra parameters in a benchmark definition file. If both ways to pass witnesses, through the task-definition and through the benchmark file,  are used then CPAchecker fails with an error, as should be expected.

### DIFF
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -194,8 +194,8 @@ class Tool(benchexec.tools.template.BaseTool2):
                 options += [f"{prefix}witness", possible_witness_files[0]]
             else:
                 raise benchexec.tools.template.UnsupportedFeatureException(
-                    f"You are passing a witness as both an option and through the task definition. "
-                    f"Please remove one of them."
+                    "You are passing a witness as both an option and through the task definition. "
+                    "Please remove one of them."
                 )
 
         return options
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -180,16 +180,40 @@ class Tool(benchexec.tools.template.BaseTool2):
                         f"Unsupported data_model '{data_model}' defined for task '{task}'"
                     )
 
+        if isinstance(task.options, dict) and "witness" in task.options.keys():
+            if not option_present("witness"):
+                possible_witness_files = list(
+                    filter(
+                        lambda x: x.endswith(task.options.get("witness")),
+                        task.input_files_or_empty,
+                    )
+                )
+                assert (
+                    len(possible_witness_files) == 1
+                ), f"Expected exactly one witness file, but found {len(possible_witness_files)}: {possible_witness_files}"
+                options += [f"{prefix}witness", possible_witness_files[0]]
+            else:
+                raise benchexec.tools.template.UnsupportedFeatureException(
+                    f"You are passing a witness as both an option and through the task definition. "
+                    f"Please remove one of them."
+                )
+
         return options
+
+    def _get_input_files(self, task):
+        if "witness" in task.options.keys():
+            return list(
+                filter(
+                    lambda x: not x.endswith(task.options.get("witness")),
+                    task.input_files_or_identifier,
+                )
+            )
+        return list(task.input_files_or_identifier)
 
     def cmdline(self, executable, options, task, rlimits):
         additional_options = self._get_additional_options(options, task, rlimits)
-        return (
-            [executable]
-            + options
-            + additional_options
-            + list(task.input_files_or_identifier)
-        )
+        input_files = self._get_input_files(task)
+        return [executable] + options + additional_options + input_files
 
     def determine_result(self, run):
         """
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -188,9 +188,11 @@ class Tool(benchexec.tools.template.BaseTool2):
                         task.input_files_or_empty,
                     )
                 )
-                assert (
-                    len(possible_witness_files) == 1
-                ), f"Expected exactly one witness file, but found {len(possible_witness_files)}: {possible_witness_files}"
+                if len(possible_witness_files) != 1:
+                    raise benchexec.tools.template.UnsupportedFeatureException(
+                        f"Expected exactly one witness file, but found {len(possible_witness_files)}: {possible_witness_files}"
+                    )
+
                 options += [f"{prefix}witness", possible_witness_files[0]]
             else:
                 raise benchexec.tools.template.UnsupportedFeatureException(
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
+from pathlib import Path
 import sys
 import os
 import re
@@ -182,12 +183,8 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         if isinstance(task.options, dict) and "witness" in task.options.keys():
             if not option_present("witness"):
-                possible_witness_files = list(
-                    filter(
-                        lambda x: x.endswith(task.options.get("witness")),
-                        task.input_files_or_empty,
-                    )
-                )
+                possible_witness_files = self._get_witness_input_files(task)
+
                 if len(possible_witness_files) != 1:
                     raise benchexec.tools.template.UnsupportedFeatureException(
                         f"Expected exactly one witness file, but found {len(possible_witness_files)}: {possible_witness_files}"
@@ -202,19 +199,28 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         return options
 
-    def _get_input_files(self, task):
-        if "witness" in task.options.keys():
-            return list(
-                filter(
-                    lambda x: not x.endswith(task.options.get("witness")),
-                    task.input_files_or_identifier,
-                )
-            )
-        return list(task.input_files_or_identifier)
+    def __partition_input_files(self, task):
+        input_files = task.input_files_or_empty
+        witness_files = []
+        other_files = []
+        for file in input_files:
+            if Path(file).name == task.options.get("witness"):
+                witness_files.append(file)
+            else:
+                other_files.append(file)
+        return witness_files, other_files
+
+    def _get_witness_input_files(self, task):
+        witness_files, _ = self.__partition_input_files(task)
+        return witness_files
+
+    def _get_non_witness_input_files(self, task):
+        _, other_files = self.__partition_input_files(task)
+        return other_files
 
     def cmdline(self, executable, options, task, rlimits):
         additional_options = self._get_additional_options(options, task, rlimits)
-        input_files = self._get_input_files(task)
+        input_files = self._get_non_witness_input_files(task)
         return [executable] + options + additional_options + input_files
 
     def determine_result(self, run):
--- a/benchexec/tools/cpachecker.py
+++ b/benchexec/tools/cpachecker.py
@@ -200,7 +200,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return options
 
     def __partition_input_files(self, task):
-        input_files = task.input_files_or_empty
+        input_files = task.input_files_or_identifier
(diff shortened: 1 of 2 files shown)


benchexec/tools/cpachecker.py
```diff
@@ -180,16 +180,40 @@ def option_present(option):
...
+
     def cmdline(self, executable, options, task, rlimits):
         additional_options = self._get_additional_options(options, task, rlimits)
-        return (
-            [executable]
-            + options
-            + additional_options
-            + list(task.input_files_or_identifier)
-        )
+        input_files = self._get_input_files(task)
```
PhilippWendler, line 215: The overall approach of retrieving only the witness from `task` in `_get_additional_options` and then independently filtering out the witness here is error prone: If there is an inconsistency in the filtering logic, some options could be left out or kept twice.

It would be better to (a) at least implement the filtering only once and call it twice, or (b) partition the list of input files into two lists (witnesses and other) in a single place, such that it is obvious to see that the result is a partitioning.

Care also needs to be taken to handle the difference between `input_files_or_empty` and `input_files_or_identifier`.

benchexec/tools/cpachecker.py
```diff
@@ -180,16 +180,40 @@ def option_present(option):
                         f"Unsupported data_model '{data_model}' defined for task '{task}'"
                     )
 
+        if isinstance(task.options, dict) and "witness" in task.options.keys():
+            if not option_present("witness"):
+                possible_witness_files = list(
+                    filter(
+                        lambda x: x.endswith(task.options.get("witness")),
+                        task.input_files_or_empty,
+                    )
```
PhilippWendler, line 189: This is not really "Pythonic". In general, many developers (including the Python founder) prefer to avoid `map` and `filter` in Python in favor of comprehensions. Sometimes the former can make sense, but here it seems clear that something like the following is better:

```python
possible_witness_files = [input for input in task.input_files_or_empty if input.endswith(witness)]
```

(Same below).

However, `endswith` could lead to wrong matches in principle, couldn't it? Maybe do some path-name matching?

benchexec/tools/cpachecker.py
```diff
@@ -180,16 +180,40 @@ def option_present(option):
...
+                )
+                assert (
+                    len(possible_witness_files) == 1
+                ), f"Expected exactly one witness file, but found {len(possible_witness_files)}: {possible_witness_files}"
```
PhilippWendler, line 193: Given that we produce a proper error message instead of a crash for a different case of unsupported inputs a few lines below, why not do the same here?

PhilippWendler (commented):
I marked the PR as draft as long as the 	`witness` option is still a proposal and not community agreed.

The PR should link to the documentation that has the definition of that option.

marian-lingsch:
Thanks a lot for your comments!

They have now been addressed.

benchexec/tools/cpachecker.py
PhilippWendler: Now the logic is implemented once, this is better. Thanks.

But my last sentence needs to be considered: The current version of the code regresses in this regard, tasks with identifiers instead of input files are not handled correctly.

PhilippWendler:
There are several tool-info modules that build on top of the tool-info module of CPAchecker. Most of them do not override `cmdline`, so should be safe, but I think we should go through each of `grep cmdline $(grep -Rl cpachecker benchexec/tools/)` and check whether they need adjustments. In particular `cpa-witness2test.py` even calls `_get_additional_options()`.

benchexec/tools/cpachecker.py
marian-lingsch: I just changed it. But just to understand how would one go about giving a tasks using the `<withoutfile>` tag to pass the normal options inside a task definition file?

marian-lingsch:
> There are several tool-info modules that build on top of the tool-info module of CPAchecker. Most of them do not override `cmdline`, so should be safe, but I think we should go through each of `grep cmdline $(grep -Rl cpachecker benchexec/tools/)` and check whether they need adjustments. In particular `cpa-witness2test.py` even calls `_get_additional_options()`.

Thanks a lot for pointing this out!

Just went through all the different tools which use CPAchecker's tool info module. Only CPA-witness2test had to be adjusted, the others should not be affected. MetaVal also needs to be adapated, but that will be a problem for another PR.

benchexec/tools/cpachecker.py
PhilippWendler: Not sure I understand the question fully. A `<withoutfile>` task by definition does not have a task-definition *file*.

PhilippWendler:
> > There are several tool-info modules that build on top of the tool-info module of CPAchecker. Most of them do not override `cmdline`, so should be safe, but I think we should go through each of `grep cmdline $(grep -Rl cpachecker benchexec/tools/)` and check whether they need adjustments. In particular `cpa-witness2test.py` even calls `_get_additional_options()`.
> 
> Thanks a lot for pointing this out!
> 
> Just went through all the different tools which use CPAchecker's tool info module. Only CPA-witness2test had to be adjusted, the others should not be affected.

Hm, but now `cpa-witness2test.py` does not fail anymore if there is more than one (non-witness) input file. And is `-witness` the correct way to pass witnesses to `cpa-witness2test`? Did you test this tool-info module with your new tasks as well?

> MetaVal also needs to be adapated, but that will be a problem for another PR.

Hm, what does that mean? How will the tool-info module behave in the meantime? It should still continue work for existing users.

benchexec/tools/cpachecker.py
marian-lingsch: I think I get it, the problem is that the properties can be used to give parameters and were being ignored in the previous implementation.

marian-lingsch:
> > > There are several tool-info modules that build on top of the tool-info module of CPAchecker. Most of them do not override `cmdline`, so should be safe, but I think we should go through each of `grep cmdline $(grep -Rl cpachecker benchexec/tools/)` and check whether they need adjustments. In particular `cpa-witness2test.py` even calls `_get_additional_options()`.
> > 
> > 
> > Thanks a lot for pointing this out!
> > Just went through all the different tools which use CPAchecker's tool info module. Only CPA-witness2test had to be adjusted, the others should not be affected.
> 
> Hm, but now `cpa-witness2test.py` does not fail anymore if there is more than one (non-witness) input file. And is `-witness` the correct way to pass witnesses to `cpa-witness2test`? Did you test this tool-info module with your new tasks as well?

With the latest commit it fails again if there is more than one non-witness file. 
`-witness` is also the correct way to pass the parameter as seen [here](https://gitlab.com/sosy-lab/sv-comp/bench-defs/-/blob/main/benchmark-defs/cpa-witness2test-validate-violation-witnesses-1.0.xml?ref_type=heads). It also works as expected when running it on the benchmark definition files.

> > MetaVal also needs to be adapated, but that will be a problem for another PR.
> 
> Hm, what does that mean? How will the tool-info module behave in the meantime? It should still continue work for existing users.

Currently it will simply not work for the proposed validation task files. It will continue working as intended with the current way of validating witnesses.

PhilippWendler:
> > > MetaVal also needs to be adapated, but that will be a problem for another PR.
> > 
> > 
> > Hm, what does that mean? How will the tool-info module behave in the meantime? It should still continue work for existing users.
> 
> Currently it will simply not work for the proposed validation task files. It will continue working as intended with the current way of validating witnesses.

Ok, this is good to hear. What happens if someone passes a new task-definition file with the `witness` option? Will it just be ignored (like all other tool-info modules would do), or is the behavior worse than ignoring the additional option, i.e., would it fail or do something meaningless if a user passes a new task-definition file?

I think it would be good if we keep it such that every tool-info modules either fully supports some option, or ignores its existence completely. That is the least confusing for users. Otherwise we would have three different behaviors of the Metaval tool-info module for validation tasks: in old versions of BenchExec it just ignores the `witness` option, in future versions it will eventually support it, and in the meantime yet something else.

marian-lingsch:
> Ok, this is good to hear. What happens if someone passes a new task-definition file with the `witness` option? Will it just be ignored (like all other tool-info modules would do), or is the behavior worse than ignoring the additional option, i.e., would it fail or do something meaningless if a user passes a new task-definition file?

Currently MetaVal will just not work for the validation task definition format, since it ignores the `witness` option. Since it will be missing the expected witness option which is passed to it using `--metavalWitness` and will therefore just print its `--help` section when run.

> I think it would be good if we keep it such that every tool-info modules either fully supports some option, or ignores its existence completely. That is the least confusing for users. Otherwise we would have three different behaviors of the Metaval tool-info module for validation tasks: in old versions of BenchExec it just ignores the `witness` option, in future versions it will eventually support it, and in the meantime yet something else.

I agree that keeping the behavior of validators consistent is important, but IMO this issue is orthogonal to this PR, since all validators will need to be adjusted in order to work with the validation format and handle the new option correctly. 

If I understood your proposal correctly, for the next release of BenchExec all validators should have adapted their tool-info module such that there is a specific version of BenchExec from which the validation task definition files start working and previously the `witness` option will just be ignored.

PhilippWendler:
> > Ok, this is good to hear. What happens if someone passes a new task-definition file with the `witness` option? Will it just be ignored (like all other tool-info modules would do), or is the behavior worse than ignoring the additional option, i.e., would it fail or do something meaningless if a user passes a new task-definition file?
> 
> Currently MetaVal will just not work for the validation task definition format, since it ignores the `witness` option. Since it will be missing the expected witness option which is passed to it using `--metavalWitness` and will therefore just print its `--help` section when run.

But not if I use an appropriate benchmark definition that includes the necessary options for witnesses (just as in the previous years), wouldn't it?

> > I think it would be good if we keep it such that every tool-info modules either fully supports some option, or ignores its existence completely. That is the least confusing for users. Otherwise we would have three different behaviors of the Metaval tool-info module for validation tasks: in old versions of BenchExec it just ignores the `witness` option, in future versions it will eventually support it, and in the meantime yet something else.
> 
> I agree that keeping the behavior of validators consistent is important, but IMO this issue is orthogonal to this PR, since all validators will need to be adjusted in order to work with the validation format and handle the new option correctly.
> 
> If I understood your proposal correctly, for the next release of BenchExec all validators should have adapted their tool-info module such that there is a specific version of BenchExec from which the validation task definition files start working and previously the `witness` option will just be ignored.

No, I never said this. I said that every tool-info modules should either support the option or ignore it completely. But I don't want that tool-info module X ignores option Y in versions A, B, C, does something that is broken in versions D and E, and fully supports option Y in versions F and beyond.

Of course it would be *nice* for users if "there is a specific version of BenchExec from which the validation task definition files start working and previously the `witness` option will just be ignored", but I would not require that.

PhilippWendler:
> > > > Ok, this is good to hear. What happens if someone passes a new task-definition file with the `witness` option? Will it just be ignored (like all other tool-info modules would do), or is the behavior worse than ignoring the additional option, i.e., would it fail or do something meaningless if a user passes a new task-definition file?
> > > 
> > > 
> > > Currently MetaVal will just not work for the validation task definition format, since it ignores the `witness` option. Since it will be missing the expected witness option which is passed to it using `--metavalWitness` and will therefore just print its `--help` section when run.
> > 
> > 
> > But not if I use an appropriate benchmark definition that includes the necessary options for witnesses (just as in the previous years), wouldn't it?
> 
> Exactly

Then we should ensure that this PR does not cause a regression for this situation (new task definition with appropriate benchmark definition with manually defined options for Metaval).

PhilippWendler (approved):
Can be merged once the community has agreed on and finalized the specification of the new option in the task-definition files. Would be nice to add a link to the spec to the code once this happens.

PhilippWendler:
Replaced by #1093.

PR #1080, opened 2024-10-11
VeriAbs Updated for Pre-Runs.

ManojAlladawar56:
- Updated veriabs.py with project_url()

### DIFF
--- a/benchexec/tools/veriabs.py
+++ b/benchexec/tools/veriabs.py
@@ -25,9 +25,9 @@ class Tool(benchexec.tools.template.BaseTool2):
         "afl-2.35b",
         "verifuzz",
         "afl-2.35b_v1",
-        "scripts",
         "frama-c-Chlorine-20180502",
         "UAutomizer-linux",
+        "scripts",
         "supportFiles",
     ]
 
@@ -46,7 +46,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "VeriAbs"
 
     def project_url(self):
-        return "https://zenodo.org/records/10243500"
+        return "https://doi.org/10.5281/zenodo.10066250"
 
     def cmdline(self, executable, options, task, rlimits):
         if task.property_file:
--- a/benchexec/tools/veriabs.py
+++ b/benchexec/tools/veriabs.py
@@ -25,8 +25,6 @@ class Tool(benchexec.tools.template.BaseTool2):
         "afl-2.35b",
         "verifuzz",
         "afl-2.35b_v1",
-        "frama-c-Chlorine-20180502",
-        "UAutomizer-linux",
         "scripts",
         "supportFiles",
     ]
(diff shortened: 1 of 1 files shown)


benchexec/tools/veriabs.py
```diff
@@ -25,8 +25,6 @@ class Tool(benchexec.tools.template.BaseTool2):
         "afl-2.35b",
         "verifuzz",
         "afl-2.35b_v1",
-        "frama-c-Chlorine-20180502",
-        "UAutomizer-linux",
```
PhilippWendler, line 29: I would say this should be kept because otherwise the tool-info module will not work correctly anymore for older versions of VeriAbs. If these paths do not exist in the current version of VeriAbs this is not a problem because non-matching paths will simply be ignored.

benchexec/tools/veriabs.py
```diff
@@ -46,7 +44,7 @@ def name(self):
         return "VeriAbs"
 
     def project_url(self):
-        return "https://www.tcs.com/designing-complex-intelligent-systems"
+        return "https://www.tcs.com"
```
PhilippWendler, line 47: Is there a better URL for the VeriAbs project itself? If a user who is interested in VeriAbs clicks on this link, they won't find information about it.

benchexec/tools/veriabs.py
ManojAlladawar56: Thanks for the input @PhilippWendler ! , I will add the paths in the required_paths section.

benchexec/tools/veriabs.py
ManojAlladawar56: we’ve updated the link with the correct URL for accessing the information. We're also working on getting a dedicated project page.

benchexec/tools/veriabs.py
```diff
@@ -46,7 +46,7 @@ def name(self):
         return "VeriAbs"
 
     def project_url(self):
-        return "https://www.tcs.com/designing-complex-intelligent-systems"
+        return "https://zenodo.org/records/10243500"
```
PhilippWendler, line 49: This refers to a specific version of VeriAbs, so this URL will be wrong for all benchmarks done with a different version of VeriAbs, and thus could lead to confusion. Don't you have a project page with information about this tool that could be used here?

If you really wish to link to Zenodo, then at least the meta DOI that always points to the latest version should be used.

benchexec/tools/veriabs.py
```diff
@@ -25,9 +25,9 @@ class Tool(benchexec.tools.template.BaseTool2):
         "afl-2.35b",
         "verifuzz",
         "afl-2.35b_v1",
+        "scripts",
```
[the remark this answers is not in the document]
ManojAlladawar56: Adjusted as suggested, with no unnecessary reordering.

PR #1082, opened 2024-10-15
Add the tool info module for nacpa

ricffb:
This PR adds the tool info for the tool "nacpa"

### DIFF
--- a/benchexec/tools/nacpa.py
+++ b/benchexec/tools/nacpa.py
@@ -11,7 +11,7 @@ from benchexec.tools import template
 
 class Tool(template.BaseTool2):
     """
-    This tool-info module runs nacpa, a parallel portfolio
+    Tool-info module for nacpa, a parallel portfolio
     of natively compiled CPAchecker instances.
     """
 
--- a/benchexec/tools/nacpa.py
+++ b/benchexec/tools/nacpa.py
@@ -49,7 +49,7 @@ class Tool(template.BaseTool2):
             if data_model:
                 options = options + ["--data-model", data_model]
 
-        return super().cmdline(executable, options, task, rlimits)
+        return [executable, *options, task.single_input_file]
 
     def determine_result(self, run):
         for line in reversed(run.output):
--- a/benchexec/tools/nacpa.py
+++ b/benchexec/tools/nacpa.py
@@ -1,0 +1,85 @@
+from benchexec import result
+from benchexec.tools import template
+
+
+class Tool(template.BaseTool2):
+    """
+    This tool-info module runs nacpa, a parallel portfolio
+    of natively compiled CPAchecker instances.
+    """
+
+    REQUIRED_PATHS = ["bin", "lib"]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("nacpa", subdir="bin")
+
+    def version(self, executable):
+        return self._version_from_tool(
+            executable, "--version", line_prefix="nacpa version"
+        )
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def name(self):
+        return "nacpa"
+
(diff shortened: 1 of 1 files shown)


benchexec/tools/nacpa.py
```diff
@@ -0,0 +1,92 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from benchexec import result
+from benchexec.tools import template
+
+
+class Tool(template.BaseTool2):
+    """
+    This tool-info module runs nacpa, a parallel portfolio
+    of natively compiled CPAchecker instances.
```
PhilippWendler, line 15: *nitpick*: tool-info modules do not actively run anything (they just return information), but because this is a common misconception I would prefer to clear this up here.
```suggestion
    Tool-info module for nacpa, a parallel portfolio
    of natively compiled CPAchecker instances.
```

benchexec/tools/nacpa.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
+        )
+
+    def name(self):
+        return "nacpa"
+
```
PhilippWendler, line 35: Could you also add a function `project_url()` that returns the tool's URL? This is quite useful and makes all tables link back to your tool page.

benchexec/tools/nacpa.py
```diff
@@ -0,0 +1,92 @@
...
+        if (
+            "--data-model" not in options
+            and isinstance(task.options, dict)
+            and task.options.get("language") == "C"
+        ):
+            data_model = task.options.get("data_model")
+            if data_model:
+                options = options + ["--data-model", data_model]
+
+        return super().cmdline(executable, options, task, rlimits)
```
PhilippWendler, line 49: Does your tool support tasks with more than one or no input file? If not, this should be changed here to not (indirectly) use `task.input_files_or_identifier`.

benchexec/tools/nacpa.py
```diff
@@ -0,0 +1,92 @@
...
+            if line.lstrip().startswith(identifier):
+                startPosition = line.find(":") + 1
+                endPosition = line.find("(", startPosition)
+                if endPosition == -1:
+                    endPosition = len(line)
+                if match is None:
+                    match = line[startPosition:endPosition].strip()
+                    break
+
+        return match
```
PhilippWendler, line 92: Doesn't this do the same as the following?
```suggestion
                return line[startPosition:endPosition].strip()

        return None
```

If yes, also remove `match` completely.

benchexec/tools/nacpa.py
```diff
@@ -0,0 +1,92 @@
...
+                elif verdict.lower().startswith("false(valid-memcleanup)"):
+                    return result.RESULT_FALSE_MEMCLEANUP
+                elif verdict.lower().startswith("false"):
+                    return result.RESULT_FALSE_PROP
+        return result.RESULT_UNKNOWN
+
+    def get_value_from_output(self, output, identifier):
+        # search for the text in output and get its value,
+        # search the first line, that starts with the searched text
+        # warn if there are more lines (multiple statistics from sequential analysis?)
```
PhilippWendler, line 80: Hm, I did not find any code that corresponds to this claim. Was this overlooked or is the comment wrong?

benchexec/tools/nacpa.py
ricffb: Neat, I was not aware that this is what the project url is used for 😄

benchexec/tools/nacpa.py
ricffb: You are right. This code was copied and adjusted without much care. I refactored it to remove the unnecessary statements.

benchexec/tools/nacpa.py
ricffb: Removed.

PhilippWendler (approved):

PR #1083, opened 2024-10-16
AProVE 2025

nilslommen:


benchexec/tools/aprove.py
```diff
@@ -22,14 +22,14 @@ class Tool(benchexec.tools.template.BaseTool2):
     Only the binary (jar) distribution of AProVE is supported.
     """
 
-    REQUIRED_PATHS = ["aprove.jar", "AProVE.sh", "bin", "newstrategy.strategy"]
+    REQUIRED_PATHS = ["aprove.jar", "AProVE.sh", "bin", "newstrategy.strategy", "lib", "fake_include"]
```
PhilippWendler, line 25: This line would need to be reformatted (you can run `black` on the file or have a look at CI).

benchexec/tools/aprove.py
```diff
@@ -22,14 +22,14 @@ class Tool(benchexec.tools.template.BaseTool2):
...
     BIT_WIDTH_PARAMETER_NAME = "--bit-width"
 
     def executable(self, tool_locator):
         return tool_locator.find_executable("AProVE.sh")
 
     def name(self):
-        return "AProVE"
+        return "AProVE (KoAT + LoAT)"
```
PhilippWendler, line 32: This would mean that whenever one uses a new BenchExec version to benchmark any version of AProVE, the results would claim that the benchmarks were on "AProVE (KoAT + LoAT)", but I guess this would be wrong for old versions?

So I suggest to just keep it as is.

nilslommen: I'm unsure how your BenchExec environment works. I thought the tool name should simply be returned here.

> [...] the results would claim that the benchmarks were on "AProVE (KoAT + LoAT)", but I guess this would be wrong for old versions?

Yes, if this should also be used with old versions, then the name is not correct. But isn't using old configurations a problem anyway if the required paths, flags etc. change?

benchexec/tools/aprove.py
nilslommen: Thanks! This should be fixed now. Do I have to squash the commits?

benchexec/tools/aprove.py
PhilippWendler: The required paths are no problem if they define a union of what all AProVE versions need. Non-matching paths are just ignored.

Flags are typically defined in the benchmark definition, and the user needs to write an appropriate benchmark definition for the version of AProVE that they want to benchmark. But they should be able to use any version of BenchExec, independently of the AProVE version. So all changes to the tool-info module should keep backwards compatibility.

benchexec/tools/aprove.py
PhilippWendler: Thanks, it is fine as it is.

benchexec/tools/aprove.py
PhilippWendler: Overlooked this question:

> I'm unsure how your BenchExec environment works. I thought the tool name should simply be returned here.

Yes, the tool name should be reported here, and BenchExec will take it and print it in the tables.

nilslommen: Then it will probably be best to create completely new entries for the new version of AProVE. Otherwise it will be confusing with the old version, which used different subsystems, and also the tool name which we submitted at easychair. I will adjust this accordingly this afternoon.

PhilippWendler: Hm, not sure. If the tool remains AProVE, I think it should remain the same tool-info module. We do not need to or want to have a separate tool-info module for each version or configuration of a tool, that would just create a maintenance nightmare. The fact that a new version of AProVE is benchmarked is indicated by the version number of the tool, and for the used configuration users see the passed command-line options.

nilslommen: The problem is that the new tool isn't any longer just AProVE. In our new framework, AProVE transforms C programs into transitions systems which are then analyzed for termination and non-termination by KoAT and resp. LoAT. In case of non-termination, AProVE constructs the witness from the output of LoAT. So, the termination and non-termination analysis is essentially performed by KoAT and LoAT (which both are developed independently from AProVE) and not AProVE. This is the reason why I think its important to mention them in the name. Isn't there an intermediate solution in the SV-Comp bylaws to handle something like this?

PhilippWendler: This is not really about any SV-Comp rules, it is just about what does technically solve the requirements and what is most appropriate for users of BenchExec.

First, note that the tool name here is really supposed to be the name of the tool. This does not need to be the same name as the submission to SV-COMP. Sometimes the same tool participates twice in different configurations in SV-COMP, so the submission name can differ from the tool name. You can see this in the SV-COMP results of the last years for example for the tool CoVeriTeam. So if your goal is just to have KoAT and LoAT be visible in SV-COMP, you can do that without changing the tool-info module. But note that it is also common that tools delegate (parts of) the verification to backend tools, and while this should be explained in the system description, it is not necessary (I would not even say common) to add this information to the name of the submission.

The name of the existing tool-info module really can not be changed because it would result in wrong information when plain AProVE is benchmarked with BenchExec.

Whether a new tool-info module should be added depends on a few things. If plain AProVE and "AProVE (KoAT + LoAT)" where incompatible, e.g., produce output that needs to be parsed differently or need different command lines, this would likely mean that we need different tool-info modules. But it seems this is not the case here.

Another question is how and whether users would distinguish between these two. Suppose that we add a second tool-info module, and a user downloads AProVE and wants to benchmark it with BenchExec. How would such a user know which of the two tool-info modules to use? Is "AProVE (KoAT + LoAT)" advertised separately from AProVE, has a different webpage and a different download archive? Or is such that users always just download something named "AProVE", and depending on its version or config or whatever it sometimes happens to execute KoAT and LoAT in the background? More generally: Would a user think of AProVE and "AProVE (KoAT + LoAT)" as different tools, or as the same tool in different configurations?

nilslommen: I now recovered the original name.

>  This does not need to be the same name as the submission to SV-COMP. Sometimes the same tool participates twice in different configurations in SV-COMP, so the submission name can differ from the tool name.

On the SV-Comp page (https://sv-comp.sosy-lab.org/2025/submission.php) it says "(iii) a tool-info module for BenchExec, [..] the name of which should be mentioned in the registration record (i)". Then I find that a little confusing. Maybe it should be adapted for competitions in the future.

PhilippWendler: Yes, I think that requirement could just be removed. The name of the tool-info module is part of the benchmark definition anyway. Feel free to suggest this in the SV-COMP context.

PR #1084, opened 2024-10-18
Add SVF tool info

Lasagnenator:


benchexec/tools/svf.py
```diff
@@ -0,0 +1,62 @@
...
+
+from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+import benchexec.tools.template
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for SVF: a framework for static value-flow analysis.
+    Specifically this tool is a wrapper around SVF to make it work with SV-COMP.
```
PhilippWendler, line 16: I am wondering what the relation between SVF and svf-svc-svcomp is. Is the latter an official project by the SVF team? Or is it a third-party wrapper that implements something on top of SVF?

Right now this tool-info module claims to be for SVF, but it looks as if one cannot actually execute SVF with it, only svf-svc-svcomp. I guess then the tool-info module's name should probably reflect that. If later someone comes and wants to benchmark proper SVF with BenchExec, they should be able to submit a tool-info module named `svf.py`.

benchexec/tools/svf.py
```diff
@@ -0,0 +1,62 @@
...
+    - Project URL: https://github.com/Lasagnenator/svf-svc-comp
```
PhilippWendler, line 17: This is specified anyway below, it can be removed.

benchexec/tools/svf.py
```diff
@@ -0,0 +1,62 @@
...
+            elif line.startswith("Incorrect"):
+                good_exit |= True
+                verified &= False
+
+        if not good_exit:
+            return result.RESULT_UNKNOWN
+        elif verified:
+            return result.RESULT_TRUE_PROP
+        else:
+            return result.RESULT_FALSE_PROP
```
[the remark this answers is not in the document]
Lasagnenator: Originally the plan was to use this logic since SVF can check for multiple violations at once and so it would need to check all the output for any violations. I suppose we could modify the wrapper to do the same logic instead.

Right now there is no distinction between no verdict and error since SVF itself uses 'error' return codes to indicate some meaning rather than errors during execution. With some more time and planning we could definitely make it possible to distinguish between them.

benchexec/tools/svf.py
```diff
@@ -0,0 +1,62 @@
...
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for SVF: a framework for static value-flow analysis.
+    Specifically this tool is a wrapper around SVF to make it work with SV-COMP.
+    - Project URL: https://github.com/Lasagnenator/svf-svc-comp
+    - SVF: https://github.com/SVF-tools/SVF
+    - SVF version used: SVF-3.0
```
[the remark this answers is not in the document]
Lasagnenator: Done.

benchexec/tools/svf.py
Lasagnenator: Done.

benchexec/tools/svf.py
yuleisui: The  submission is built on top of SVF and use SVF as a library. I would say the naming svf-svc is more appropriate name as the submission is tailored for this competition.

Lasagnenator: Thanks Yulei. I've updated the name to reflect this.

benchexec/tools/svf.py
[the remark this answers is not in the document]
Lasagnenator: For now the wrapper has a very simple error checking method so I've added the distinction between error and unknown.

I also made the early returns for the verdicts and will implement the logic of only outputting one verdict in the tool soon.

benchexec/tools/svf.py
PhilippWendler: Thanks. Note that you can also name your tool-info module `svf-svc` if you prefer, even though the dash is normally not allowed in Python module names.

benchexec/tools/svf_svc.py
```diff
@@ -40,23 +38,20 @@ def cmdline(self, executable, options, task, rlimits):
...
-        elif verified:
-            return result.RESULT_TRUE_PROP
-        else:
-            return result.RESULT_FALSE_PROP
+                return result.RESULT_FALSE_PROP
+            elif line.startswith("Error"):
+                # The line should contain the error message with details.
+                return line
+
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 57: Typically, if the tool outputs nothing or something completely unexpected, this is because of an error (for example if the tool crashes and dumps a stacktrace or segfaults or so). So I would recommend to return `RESULT_ERROR` here. If your tool gives up willingly because it does not know the answer, it would be nice for users as well if the output would indicate this, and this could be matched here for the `UNKNOWN` case.

benchexec/tools/svf_svc.py
```diff
@@ -40,23 +38,20 @@ def cmdline(self, executable, options, task, rlimits):
...
-        if not good_exit:
-            return result.RESULT_UNKNOWN
-        elif verified:
-            return result.RESULT_TRUE_PROP
-        else:
-            return result.RESULT_FALSE_PROP
+                return result.RESULT_FALSE_PROP
+            elif line.startswith("Error"):
+                # The line should contain the error message with details.
+                return line
```
PhilippWendler, line 55: For consistency with other tools and BenchExec itself it would be good if the result here would conform to the same structure as is typical: Either return plain `RESULT_ERROR` or return `RESULT_ERROR + "(" + reason + ")"`.

Furthermore, the reason should not be too long, because otherwise the tables become impractical to use. Something like "exception", "segmentation fault", "parsing failed", etc. as the reason is good and really useful, but something like `Exception in thread "main" java.util.NoSuchElementException: No value present` would be too long in practice. Is the "error message with details" expected to be reasonably short in practice?

benchexec/tools/svf_svc.py
Lasagnenator: Done.

benchexec/tools/svf_svc.py
```diff
@@ -0,0 +1,61 @@
...
+                return result.RESULT_FALSE_PROP
+            elif line.startswith("Error: "):
+                # The line should contain a short error message.
+                # The tool returns errors in the format Error: info.
+                error = line[line.index(":") + 2 :]
+                return result.RESULT_ERROR + "(" + error + ")"
+            elif line.startswith("Unknown"):
+                return result.RESULT_UNKNOWN
+
+        return result.RESULT_ERROR + "(Unknown error)"
```
PhilippWendler, line 61: Thanks for your changes!

```suggestion
        return result.RESULT_ERROR
```

Last nitpick: If you just return `RESULT_ERROR`, BenchExec detects common cases of failures (such as segfaults, aborts, etc.) and enhances the result value with that information. For limit violations or if the tool exits with an error code the result will also be better.

PhilippWendler (approved):

PR #1085, opened 2024-10-18
SV-COMP 25 Hornix tool

jankofron:
This is a part of the Hornix SV-COMP'25 submission.

### DIFF
--- a/benchexec/tools/hornix.py
+++ b/benchexec/tools/hornix.py
@@ -1,0 +1,44 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+# SPDX-FileCopyrightText: 2024 Jan Kofron <jan.kofron@d3s.mff.cuni.cz>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec
+import benchexec.result as result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Info object for the tool Hornix.
+    """
+
+    REQUIRED_PATHS = ["."]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("hornix")
+
+    def version(self, executable):
+        return self._version_from_tool(executable, arg="--version")
+
+    def name(self):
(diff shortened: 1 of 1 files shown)


benchexec/tools/hornix.py
```diff
@@ -0,0 +1,40 @@
...
+        return "https://github.com/d3sformal/hornix"
+
+    def determine_result(self, run):
+        for line in run.output:
+            if "sat" == line.strip():
+                return result.RESULT_TRUE_PROP
+            elif "unsat" == line.strip():
+                return result.RESULT_FALSE_REACH
+            else:
+                return result.RESULT_UNKNOWN
```
PhilippWendler, line 40: Note that this will return unknown at the first line of the output. Is this intended? Then you do not need a loop at all.

Additionally, do you have a way to distinguish between successful tool execution but without a verdict found and an error? Then I recommend to distinguish these in the results as well. This is much more helpful for analyzing the results.

jankofron: Thanks for the comment. True, we do not need the loop here (we have adapted the korn.py script and this has slipped through). Nonetheless, we assumed that the output that is obtained by `determine_result()` is the output of our tool, which we restrict to a single line (sat, unsat, or unknown), so this should be correct. Sorry for the troubles, we are trying to submit for the first time. And thanks a lot once more!

blishko: We have tried to update the result translation to be a bit more informative.

benchexec/tools/hornix.py
```diff
@@ -0,0 +1,46 @@
...
+    def name(self):
+        """The human-readable name of the tool."""
+        return "Hornix"
+
+    def project_url(self):
+        return "https://github.com/d3sformal/hornix"
+
+    def determine_result(self, run):
+        if len(run.output) == 0:
+            return result.RESULT_TIMEOUT
```
PhilippWendler, line 35: This assumption is in general likely wrong. For example, the tool could also be terminated because of an out-of-memory situation. Furthermore, the tool-info module does not need to care about timeouts and other limit violations, because BenchExec handles that. So just return `RESULT_ERROR` if the output is not of length 1.

PhilippWendler (approved):

benchexec/tools/hornix.py
blishko: Thanks for the notice! Should be fixed now.

PR #1086, opened 2024-10-18
Add toolinfo module for MetaVal++

marian-lingsch:
This PR adds the toolinfo module for a new tool called MetaVal++.  It is a witness validator, which is based on the ideas of transforming a task such that a backend tool can validate the witness.

Currently it mostly supports deductive verifiers like Frama-C as backend.

### DIFF
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -85,8 +85,8 @@ class Tool(BaseTool2):
         # only the first group will be returned.
         match = None
         for line in output:
-            matches = re.search(identifier, line).group()
+            matches = re.search(identifier, line)
             if matches:
-                match = matches
+                match = matches.group()
                 break
         return match
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -22,11 +22,6 @@ class Tool(BaseTool2):
     def executable(self, tool_locator: BaseTool2.ToolLocator):
         return tool_locator.find_executable("metaval++.py")
 
-    def program_files(self, executable):
-        return [executable] + self._program_files_from_executable(
-            executable, self.REQUIRED_PATHS, parent_dir=True
-        )
-
     def version(self, executable):
         return self._version_from_tool(executable)
 
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -17,10 +17,10 @@ class Tool(BaseTool2):
     Tool info for Deductive Validator.
     """
 
-    REQUIRED_PATHS = ["lib", "src", "metaval_2.0.py"]
+    REQUIRED_PATHS = ["lib", "src", "metaval++.py"]
 
     def executable(self, tool_locator: BaseTool2.ToolLocator):
-        return tool_locator.find_executable("metaval_2.0.py")
+        return tool_locator.find_executable("metaval++.py")
 
     def program_files(self, executable):
         return [executable] + self._program_files_from_executable(
@@ -31,10 +31,10 @@ class Tool(BaseTool2):
         return self._version_from_tool(executable)
 
     def name(self):
-        return "MetaVal 2.0"
+        return "MetaVal++"
 
     def project_url(self):
-        return "https://gitlab.com/sosy-lab/software/metaval-2.0"
+        return "https://gitlab.com/sosy-lab/software/metavalpp"
 
     def cmdline(self, executable, options, task, rlimits):
         if task.property_file:
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -54,26 +54,31 @@ class Tool(BaseTool2):
         )
 
     def determine_result(self, run):
+        separator = ":"
         if not run.output:
             return benchexec.result.RESULT_ERROR
         lastline = run.output[-1]
         if lastline.startswith("Witness is correct"):
             return benchexec.result.RESULT_TRUE_PROP
         elif lastline.startswith("Witness could not be validated"):
-            if ":" in lastline:
+            if separator in lastline:
                 return (
                     benchexec.result.RESULT_ERROR
-                    + self.substring_after_identifier(lastline, ":").strip()
+                    + "("
+                    + self.substring_after_identifier(lastline, separator).strip()
+                    + ")"
                 )
             else:
                 return benchexec.result.RESULT_ERROR
         elif lastline.startswith(
             "There was an error validating the witness in the backend verifier"
         ):
-            if ":" in lastline:
+            if separator in lastline:
                 return (
                     benchexec.result.RESULT_ERROR
-                    + self.substring_after_identifier(lastline, ":").strip()
+                    + "("
+                    + +self.substring_after_identifier(lastline, separator).strip()
+                    + ")"
                 )
             else:
                 return benchexec.result.RESULT_ERROR
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -57,11 +57,9 @@ class Tool(BaseTool2):
             status = benchexec.result.RESULT_TRUE_PROP
         elif lastline.startswith("Witness could not be validated"):
             if ":" in lastline:
-                status = (
-                    benchexec.result.RESULT_UNKNOWN + lastline.split(":")[1].strip()
-                )
+                status = benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
             else:
-                status = benchexec.result.RESULT_UNKNOWN
+                status = benchexec.result.RESULT_ERROR
         elif lastline.startswith(
             "There was an error validating the witness in the backend verifier"
         ):
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -54,24 +54,23 @@ class Tool(BaseTool2):
             return benchexec.result.RESULT_ERROR
         lastline = run.output[-1]
         if lastline.startswith("Witness is correct"):
-            status = benchexec.result.RESULT_TRUE_PROP
+            return benchexec.result.RESULT_TRUE_PROP
         elif lastline.startswith("Witness could not be validated"):
             if ":" in lastline:
-                status = benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
+                return benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
             else:
-                status = benchexec.result.RESULT_ERROR
+                return benchexec.result.RESULT_ERROR
         elif lastline.startswith(
             "There was an error validating the witness in the backend verifier"
         ):
             if ":" in lastline:
-                status = benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
+                return benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
             else:
-                status = benchexec.result.RESULT_ERROR
+                return benchexec.result.RESULT_ERROR
         elif lastline.startswith("Witness file does not exist"):
-            status = benchexec.result.RESULT_ERROR + "(no witness)"
+            return benchexec.result.RESULT_ERROR + "(no witness)"
         else:
-            status = benchexec.result.RESULT_ERROR
-        return status
+            return benchexec.result.RESULT_ERROR
 
     def get_value_from_output(self, output, identifier):
         # search for the text in output and get its value,
--- a/benchexec/tools/metaval++.py
+++ b/benchexec/tools/metaval++.py
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
+import re
 
 import benchexec.result
 from benchexec.tools.sv_benchmarks_util import ILP32, LP64, get_data_model_from_task
@@ -68,22 +69,21 @@ class Tool(BaseTool2):
             return benchexec.result.RESULT_ERROR
 
     def get_value_from_output(self, output, identifier):
-        # search for the text in output and get its value,
-        # search the first line, that starts with the searched text
-        # warn if there are more lines with the searched text
+        # Search the text line per line using the regex passed as identifier
+        # and return the first match found. If the regex has groups, 
+        # only the first group will be returned.
         match = None
         for line in output:
-            if line.lstrip().startswith(identifier):
-                startPosition = line.find(":") + 1
-                endPosition = line.find("(", startPosition)
-                if endPosition == -1:
-                    endPosition = len(line)
-                if match is None:
-                    match = line[startPosition:endPosition].strip()
-                else:
+            matches = re.findall(identifier, line)
+            if matches is not None and len(matches) > 0:
+                if isinstance(matches[0], tuple):
                     logging.warning(
-                        "skipping repeated match for identifier '%s': '%s'",
-                        identifier,
-                        line,
+                        "The regex '{}' has groups, but only the first group will be returned".format(
+                            identifier
+                        )
                     )
+                    match = matches[0][0]
+                else:
+                    match = matches[0]
+                break
         return match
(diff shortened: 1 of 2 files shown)


PhilippWendler:
Why should this have a new tool-info module? Can't the existing tool-info module for MetaVal handle this or be extended to handle this?

marian-lingsch:
> Why should this have a new tool-info module? Can't the existing tool-info module for MetaVal handle this or be extended to handle this?

Because they are two completely different tools. In particular there is no commonality in their code bases nor with how they are called, since they are based on completely different workflows and frameworks.

The only commonality is that they are based on the same principle i.e. instrument the program with the witness and pass it along to a backend verifier, but everything else differs completely. The name was suggested by Dirk to indicate that they are based on the same principle and to not have too many tool names.

marian-lingsch:
@PhilippWendler are the CI problems due to the new toolinfo module, I am unsure?

Apart from that due to the renaming from MetaVal 2.0 to MetaVal++ should I create a new branch and PR or keep this one?

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    REQUIRED_PATHS = ["lib", "src", "metaval++.py"]
+
+    def executable(self, tool_locator: BaseTool2.ToolLocator):
+        return tool_locator.find_executable("metaval++.py")
+
+    def program_files(self, executable):
+        return [executable] + self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
```
PhilippWendler, line 27: It seems the executable is in the main directory, so basing `REQUIRED_PATHS` on the parent directory does not make sense. After fixing this, the method is identical with the one it overrides, so it can be removed.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+from benchexec.tools.sv_benchmarks_util import ILP32, LP64, get_data_model_from_task
+from benchexec.tools.template import BaseTool2
+
+
+class Tool(BaseTool2):
+    """
+    Tool info for Deductive Validator.
+    """
+
+    REQUIRED_PATHS = ["lib", "src", "metaval++.py"]
```
PhilippWendler, line 20:
```suggestion
    REQUIRED_PATHS = ["lib", "src"]
```
The executable is added to the program files anyway by `program_files()`.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+            options = options + ["--property", task.property_file]
+
+        data_model_param = get_data_model_from_task(
+            task, {ILP32: "ILP32", LP64: "LP64"}
+        )
+
+        if data_model_param and "--data-model" not in options:
+            options += ["--data-model", data_model_param]
+
+        return [executable] + options + list(task.input_files_or_identifier)
```
PhilippWendler, line 50: Does your tool support tasks with more than one input file or without any input files? If not, please use `task.input_files` or `task.single_input_file` instead because these will produce a proper error message for the user if an unsupported task is used.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+        ):
+            if ":" in lastline:
+                status = benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
+            else:
+                status = benchexec.result.RESULT_ERROR
+        elif lastline.startswith("Witness file does not exist"):
+            status = benchexec.result.RESULT_ERROR + "(no witness)"
+        else:
+            status = benchexec.result.RESULT_ERROR
+        return status
```
PhilippWendler, line 76: This method uses a mixture of early return and a result variable. I think consistently using early returns would make it easier to understand.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+    def determine_result(self, run):
+        if not run.output:
+            return benchexec.result.RESULT_ERROR
+        lastline = run.output[-1]
+        if lastline.startswith("Witness is correct"):
+            status = benchexec.result.RESULT_TRUE_PROP
+        elif lastline.startswith("Witness could not be validated"):
+            if ":" in lastline:
+                status = (
+                    benchexec.result.RESULT_UNKNOWN + lastline.split(":")[1].strip()
```
PhilippWendler, line 61: For `RESULT_UNKNOWN`, BenchExec currently does not support suffixes. Anything else except `RESULT_UNKNOWN` is treated as category error, e.g., labeled pink.

If you want the reason, I think you should use `RESULT_ERROR` here. (And maybe below as well? `RESULT_UNKNOWN` is for cases where the tool could not compute the answer and gave up.)

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+                )
+            else:
+                status = benchexec.result.RESULT_UNKNOWN
+        elif lastline.startswith(
+            "There was an error validating the witness in the backend verifier"
+        ):
+            if ":" in lastline:
+                status = benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
```
PhilippWendler, line 69: It is not strictly necessary, but is common practice and certainly nice for users if the suffix is within parentheses? Is this always the case here?

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,97 @@
...
+        elif lastline.startswith("Witness file does not exist"):
+            status = benchexec.result.RESULT_ERROR + "(no witness)"
+        else:
+            status = benchexec.result.RESULT_ERROR
+        return status
+
+    def get_value_from_output(self, output, identifier):
+        # search for the text in output and get its value,
+        # search the first line, that starts with the searched text
+        # warn if there are more lines (multiple statistics from sequential analysis?)
```
PhilippWendler, line 81: Is this the case for MetaVal++, or was it just copied from CPAchecker? In the latter case, does this behavior make sense here?

PhilippWendler:
> Apart from that due to the renaming from MetaVal 2.0 to MetaVal++ should I create a new branch and PR or keep this one?

No reason to create a new PR. After all, it is still a tool-info module for the same tool as before.

benchexec/tools/metaval++.py
marian-lingsch: This was copied from CPAchecker, I have removed it.

benchexec/tools/metaval++.py
marian-lingsch: Thanks for pointing this out!

The reason is usually important, so I will change it to `RESULT_ERROR`.

benchexec/tools/metaval++.py
marian-lingsch: Thanks for poining this out!

Has been changed

benchexec/tools/metaval++.py
marian-lingsch: Thanks for pointing this out!

It only supports a single file so I changed it accordingly

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for pointing this out!

I was not aware of it, has been fixed now

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for pointing this out!

Has been fixed

marian-lingsch:
Thanks a lot for your comments!

They have been addressed and the code improved

benchexec/tools/metaval++.py
PhilippWendler: Well, just removing a comment that accurately describes the code does not make sense, now it is worse than before.

I have asked whether you actually want the described behavior. If not, then please change the code, not just the comment.

benchexec/tools/metaval++.py
marian-lingsch: Yes that is always the case here, the tool returns the error message inside parenthesis in the last line

benchexec/tools/metaval++.py
marian-lingsch: Okay I see, just changed the code to actually represent what is useful to analyze the output from the tool

benchexec/tools/metaval++.py
PhilippWendler: The tool really outputs something like `foo: (bar)` with colon and parentheses and will continue to do so in the future? Just asking because this is a really uncommon output format.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,88 @@
...
+            return benchexec.result.RESULT_ERROR
+
+    def get_value_from_output(self, output, identifier):
+        # Search the text line per line using the regex passed as identifier
+        # and return the first match found. If the regex has groups,
+        # only the first group will be returned.
+        match = None
+        for line in output:
+            matches = re.findall(identifier, line)
+            if matches is not None and len(matches) > 0:
```
PhilippWendler, line 78:
```suggestion
            if matches:
```

benchexec/tools/metaval++.py
PhilippWendler: Now the code clearly does something different than the comment claims: last vs. first match.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            return benchexec.result.RESULT_ERROR
+
+    def get_value_from_output(self, output, identifier):
+        # Search the text line per line using the regex passed as identifier
+        # and return the first match found. If the regex has groups,
+        # only the first group will be returned.
+        match = None
+        for line in output:
+            matches = re.findall(identifier, line)
```
PhilippWendler, line 77: Why do you use `findall` if you only use the first match anyway?

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,88 @@
...
+                if isinstance(matches[0], tuple):
+                    logging.warning(
+                        "The regex '%s' has groups, but only the first group will be returned",
+                        identifier,
+                    )
+                    match = matches[0][0]
+                else:
+                    match = matches[0]
+                break
+        return match
```
PhilippWendler, line 88: It is certainly up to you how to implement the matching, but my recommendation would be to use something different. For the common case of finding a key-value pair and extracting only the value, the use of a regexp means that the user has to pass a regexp with a non-capturing group, with likely plenty of people do not know how to do.

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for your suggestion, has been implemented!

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def determine_result(self, run):
+        if not run.output:
+            return benchexec.result.RESULT_ERROR
+        lastline = run.output[-1]
+        if lastline.startswith("Witness is correct"):
+            return benchexec.result.RESULT_TRUE_PROP
+        elif lastline.startswith("Witness could not be validated"):
+            if ":" in lastline:
+                return benchexec.result.RESULT_ERROR + lastline.split(":")[1].strip()
```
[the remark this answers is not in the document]
marian-lingsch: Thanks a lot for pointing this out!

I will change it to match the first `:`

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for your comment!

I have changed it to use `re.search` which matches only the first one

benchexec/tools/metaval++.py
marian-lingsch: That is a very good point!

I implemented it like this initially in order to have an easier integration with the BenchExec toolinfo module. I just changed it in the tool and adjusted the toolinfo module accordingly.

benchexec/tools/metaval++.py
```diff
@@ -45,22 +44,33 @@ def cmdline(self, executable, options, task, rlimits):
 
         return [executable] + options + list(task.single_input_file)
 
+    @staticmethod
+    def substring_after_identifier(string, identifier, occurrences=1):
+        return (
+            identifier.join(string.split(identifier)[occurrences:])
+            if occurrences > 0
+            else string
+        )
```
PhilippWendler, line 53: Are you just looking for the `maxsplit` parameter of `str.split()`?

Why define an optional parameter that you are never using?

benchexec/tools/metaval++.py
```diff
@@ -74,15 +84,8 @@ def get_value_from_output(self, output, identifier):
...
-            if matches is not None and len(matches) > 0:
-                if isinstance(matches[0], tuple):
-                    logging.warning(
-                        "The regex '%s' has groups, but only the first group will be returned",
-                        identifier,
-                    )
-                    match = matches[0][0]
-                else:
-                    match = matches[0]
+            matches = re.search(identifier, line).group()
```
PhilippWendler, line 87: This will crash because `search()` returns `None` if it does not find anything.

benchexec/tools/metaval++.py
```diff
@@ -74,15 +84,8 @@ def get_value_from_output(self, output, identifier):
         # only the first group will be returned.
```
PhilippWendler, line 84: I do not think this is currently true. I think the code currently always returns the full match.

benchexec/tools/metaval++.py
PhilippWendler: The comment claims that the code returns the first match, but the code returns the last match.

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for your comments!

`maxsplit` was exactly what I was looking for

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for pointing this out!

Has been fixed

benchexec/tools/metaval++.py
marian-lingsch: Thanks a lot for your comment!

I improved the documentation

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_value_from_output(self, output, identifier):
+        # Search the text line per line using the regex passed as identifier
+        # and return the first match found.
+        match = None
+        for line in output:
+            matches = re.search(identifier, line)
+            if matches:
+                match = matches.group()
+                break
+        return match
```
PhilippWendler, line 82: Last suggestion: use early return instead of break. There is no reason not to and I find it easier to follow.

dbeyer: @marian-lingsch I agree, early return reduces complexity (according to DepDegree).

marian-lingsch: Thanks a lot for the suggestion, improved in https://github.com/sosy-lab/benchexec/pull/1086/commits/02088e04e78ea01af4c5b5a96736af5585b3f07b

marian-lingsch:
Thanks a lot for all your suggestions!

This should now hopefully be ready to be merged.

benchexec/tools/metaval++.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def get_value_from_output(self, output, identifier):
+        # Search the text line per line using the regex passed as identifier
+        # and return the first match found.
+        match = None
+        for line in output:
+            matches = re.search(identifier, line)
+            if matches:
+                return matches.group()
+        return match
```
PhilippWendler, line 81:
```suggestion
        for line in output:
            matches = re.search(identifier, line)
            if matches:
                return matches.group()
        return None
```

marian-lingsch: Thanks a lot for the suggestion!

Has been applied in https://github.com/sosy-lab/benchexec/pull/1086/commits/e6dc8832853614f3b75d2447c8cc67a8e905325e

PhilippWendler (approved):

PR #1087, opened 2024-10-19
Added SV-COMP 2025 marker.

rmetta:


PhilippWendler:
What purpose does this have? Tool-info modules are not specific to particular competitions.

rmetta:
> What purpose does this have? Tool-info modules are not specific to particular competitions.

I see. Didn't realise that. Thanks for letting me know

PR #1088, opened 2024-10-19
Test-comp 2025: Update of script for Fizzer 

trtikm:


PhilippWendler:
This removes a lot of code, and without following it in detail, I would assume that this changes the behavior of the tool-info module in a backwards-incompatible way. If correct, this would mean that users who benchmark a previous version of your tool with a newer BenchExec get wrong results. This is something that we want to avoid. Can you please clarify if the changes keep the behavior of the tool-info module as before, or ensure that all changes are backwards compatible?

PhilippWendler:
@trtikm Are you still interested in getting this merged?

benchexec/tools/fizzer.py
```diff
@@ -70,8 +70,15 @@ def cmdline(self, executable, options, task, rlimits):
         data_model = get_data_model_from_task(task, {ILP32: ["--m32"], LP64: []})
         if data_model is None:
             data_model = []
+        time_options = []
+        if rlimits.cputime is not None:
+            time_options += [ "--max_seconds", str(int((865.0/900.0) * rlimits.cputime))]
+            time_options += [ "--optimizer_max_seconds", str(int((30.0/900.0) * rlimits.cputime))]
+        memory_options = []
+        if rlimits.memory is not None:
+            time_options += [ "--max_exec_megabytes", str(int((13.0/15.0) * (rlimits.memory / (1024.0 * 1024.0))))]
```
PhilippWendler, line 79: There are some magic numbers here such as 865, 900, 30, etc. Where do these come from and what do they mean?
Is this how your tool is supposed to be executed in general, and will this remain true for future versions as well?
Note that this tool-info module is not specific to SV-COMP, people will likely also use it to benchmark Fizzer in other circumstances, so it should not hard-code something that is specific to SV-COMP. And it should not hard-code anything that you likely do not want for future versions.

benchexec/tools/fizzer.py
```diff
@@ -92,57 +99,23 @@ def determine_result(self, run):
...
-
         if termination_type in [
             "SERVER_INTERNAL_ERROR",
             "CLIENT_COMMUNICATION_ERROR",
             "UNCLASSIFIED_ERROR",
         ]:
             result_code = result.RESULT_ERROR
-            termination_reason = (
-                error_message if error_message is not None else termination_reason
-            )
```
PhilippWendler, line 145: Here a lot of code is removed, what is the reason for this?

Is the code necessary for using the tool-info module with older versions of your tool? Then it should stay, because the tool-info module should remain backwards compatible.

PhilippWendler:
@trtikm If you want this merged, please respond to my questions. Otherwise I am going to close this.

PhilippWendler:
Closing due to lack of response.

PR #1089, opened 2024-10-20
SV-COMP 25: Add tool info for RacerF

TDacik:


benchexec/tools/racerf.py
```diff
@@ -0,0 +1,57 @@
...
+        return "https://github.com/TDacik/Deadlock_Racer"
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("racerf-sv.py")
+
+    def cmdline(self, executable, options, task, rlimits):
+        input_files = list(task.input_files_or_identifier)
```
PhilippWendler, line 33: Does your tool accept verification tasks with no input file or with more than one input file? If not, I would recommend to adjust this line accordingly.

benchexec/tools/racerf.py
```diff
@@ -0,0 +1,57 @@
...
+        if run.output.any_line_contains("The result may be imprecise"):
+            return result.RESULT_UNKNOWN
+
+        if run.output.any_line_contains("Data race (must)"):
+            return result.RESULT_FALSE_PROP
+
+        if not run.output.any_line_contains("Data race (may)"):
+            return result.RESULT_TRUE_PROP
+
+        return result.RESULT_UNKNOWN
```
PhilippWendler, line 57: Typically, if the tool outputs nothing or something completely unexpected, this is because of an error (for example if the tool crashes and dumps a stacktrace or segfaults or so). So I would recommend to return `RESULT_ERROR` here.

benchexec/tools/racerf.py
```diff
@@ -0,0 +1,57 @@
...
+        machdep = get_data_model_from_task(
+            task, {ILP32: "gcc_x86_32", LP64: "gcc_x86_64"}
+        )
+        machdep_opt = ["-machdep", machdep] if machdep is not None else []
+
+        return [executable] + machdep_opt + options + input_files
+
+    def determine_result(self, run):
+        if run.was_terminated:
+            return result.RESULT_UNKNOWN
```
PhilippWendler, line 43: This is not necessary, BenchExec takes care of adjusting the result as required for the case of a limit violation etc. Furthermore, `UNKNOWN` would be wrong, because a timeout etc. counts like an error and not like unknown. You can just remove this.

benchexec/tools/racerf.py
TDacik: The tool accepts tasks with one or more file. What is the purpose of task with **no file**? I am not sure how such a situation should be handled.

PhilippWendler: Some tool accepts this if the task is in different format, for example like when you call a Java program you give a classpath and an entry point, but not necessarily a list of files. So we provide support for this as well in BenchExec. In these cases instead of input files a task identifier is given.

If your tool does not support this, you can simply choose to not handle these cases. But if you use `task.input_files_or_identifier` and a user attempts to pass tasks without input files, a useless command line would be created and your tool would be executed and just fail. If instead you use `task.input_files`, BenchExec will report an error to the user directly.

TDacik: Thank you for explanation.

PhilippWendler (approved):

PR #1090, opened 2024-10-20
Add toolinfo module for meta-verifier CoOpeRace

vesalvojdani:
This adds the tool info module for our _meta verifier_ specialized on data races.

(I am a bit confused about whether one has to override `program_files` or not. Seems that coveriteam and a few other tools do this, but many tools that depend on stuff within the package do not define this and still seem to work fine.)

### DIFF
--- a/benchexec/tools/cooperace.py
+++ b/benchexec/tools/cooperace.py
@@ -1,0 +1,46 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import benchexec.tools.template
+import benchexec.result as result
+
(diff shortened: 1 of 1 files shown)


benchexec/tools/cooperace.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def version(self, executable):
+        return self._version_from_tool(executable, line_prefix="CoOpeRace")
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options += ["--prop", task.property_file]
+        if task.options is not None and "data_model" in task.options:
+            options += ["--arch", task.options.get("data_model")]
+        return [executable, *options, *task.input_files]
```
PhilippWendler, line 34: Does your tool support verification tasks with more than one input file? If yes, this is fine, otherwise I would recommend to adjust the last part accordingly.

vesalvojdani: Yes, the intent is to support that!

PhilippWendler (approved):

PR #1091, opened 2024-10-20
Theta tool-info module update

leventeBajczi:
This PR modifies how Theta interprets its output (esp. regarding before- and after-parsing errors).

### DIFF
--- a/benchexec/tools/theta.py
+++ b/benchexec/tools/theta.py
@@ -45,39 +45,44 @@ class Tool(benchexec.tools.template.BaseTool2):
         if run.was_terminated:
             return result.RESULT_ERROR
         status = result.RESULT_UNKNOWN
+        parsing_status = "before"
         for line in run.output:
             if "SafetyResult Unsafe" in line:
                 status = result.RESULT_FALSE_REACH
             elif "SafetyResult Safe" in line:
                 status = result.RESULT_TRUE_PROP
             elif "ParsingResult Success" in line:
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Just as confirmation: Can you clarify that this just changes the wording of ERROR results and does not otherwise change the semantics, i.e., is backwards compatible?

leventeBajczi:
> Just as confirmation: Can you clarify that this just changes the wording of ERROR results and does not otherwise change the semantics, i.e., is backwards compatible?

The following changes have been made:
* Previously, in the case of an error after parsing, we discarded the actual error (this was a bug introduced in ba7d6132e59baee7c1d3dd325bf3e772570c1009), because after finding the `ParsingResult Success` content in a line, we set `status` to `Parsing OK`, and therefore, skip the error detection based on exit code. This **will** affect how the output of Theta is reported (changing `Parsing OK` results to actual errors), but I consider this a bugfix, rather than a breaking change.
* We added a new exit code, 209, for unsupported source elements (this won't affect older versions, they simply do not produce this exit code)
* We clarified in the exit code-based statuses if they happened _before_ or _after_ parsing (this will slightly affect very old versions of Theta (prior to SV-COMP'23), as they don't print `ParsingResult Sucess`, so every error status will state it is _before_ parsing finished. I don't consider this problematic, as these versions also didn't use the current exit-code based determination of error cause, so most likely, they will remain unspecified errors.)

PhilippWendler (approved):

PR #1093, opened 2024-10-24
Update all validators to handle Validation tasks

marian-lingsch:
As decided in the SV-COMP community meeting on the 2024-10-22 we would like all validators to support witness task definition files.

This PR implements that, by creating some helper functions and adjusting all tool info modules of current validators to handle this.


marian-lingsch:
For testing these changes I used the scripts and examples contained here:
[test-validators.zip](https://github.com/user-attachments/files/17509366/test-validators.zip)

Unpacking this in a BenchExec clone will allow you to run all make targets which will download the tool and have the target name `toolname` and run the tool on the benchmark xml with the target `run-toolname`

benchexec/tools/validation_utils.py
```diff
@@ -0,0 +1,67 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
marian-lingsch, line 1: @PhilippWendler I am unsure on how to handle the utils in this file. Currently I put them here such that it is easier to refactor them into a place you find more suitable.

benchexec/tools/ultimate.py
danieldietsch, line 1: Could you describe in the PR what the goal of these changes is? Ultimate already handles validation tasks, and it looks like this also filters options passed to us. It also looks like it restricts the module to receive only one witness.

marian-lingsch: Just updated the description of the PR, hopefully this makes it clearer what is being meant.

danieldietsch:
Thank you for the clarification and thank you for adding the support for that. As I see it, this involves no API change and tool-info-module authors could easily change the implementation to their own as long as they support the `witness` option from the task definition file, correct?

benchexec/tools/validation_utils.py
PhilippWendler: We already have `sv_benchmarks_utils.py`. Seems like the perfect place for it.

benchexec/tools/validation_utils.py
```diff
@@ -0,0 +1,67 @@
...
+This module contains some useful functions for handling Validation Tasks
+"""
+
+from pathlib import Path
+
+from benchexec.tools.template import UnsupportedFeatureException
+import benchexec.tools.template
+
+
+def __partition_input_files(task):
```
PhilippWendler, line 18: The Python way to name private stuff is with a single underscore as prefix.

benchexec/tools/validation_utils.py
```diff
@@ -0,0 +1,67 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This module contains some useful functions for handling Validation Tasks
```
PhilippWendler, line 9: All public functions here need documentation.

benchexec/tools/validation_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+def get_non_witness_input_files(task):
+    _, other_files = __partition_input_files(task)
+    return other_files
+
+
+def get_unique_non_witness_input_files(task):
+    other_files = get_non_witness_input_files(task)
+    if len(other_files) > 1:
+        raise UnsupportedFeatureException("Tool does not support multiple input files")
```
PhilippWendler, line 52: As written, a user would get a different error message depending on whether they pass a task with two input files or a task with two input files plus a witness. Please make sure to use the same error message as in the standard case, and adjust the other error messages to be written in the same style.

PhilippWendler (changes requested):
Without documentation this PR is not yet reviewable.

benchexec/tools/validation_utils.py
marian-lingsch: Thanks a lot!

Moved the helper functions to there

marian-lingsch:
> Thank you for the clarification and thank you for adding the support for that. As I see it, this involves no API change and tool-info-module authors could easily change the implementation to their own as long as they support the `witness` option from the task definition file, correct?

That is correct, since this would be more effort per tool, this PR transforms the new way of passing witnesses into the previous one by assembling the command line. This makes it possible for tools to handle the validation task definition files without changing their internal structure.

benchexec/tools/validation_utils.py
marian-lingsch: Thanks a lot for the comment. Has been changed.

benchexec/tools/validation_utils.py
marian-lingsch: Sorry must have missed adding it, have added documentation to every function.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
...
+
+    @return: List of additional options to handle witnesses. They are constructed
+        based on witness_options. For example if witness_options is ["-w"], then
+        the return value will be ["-w", "witness_file"]
+    """
+
+    additional_options = []
+    if isinstance(task.options, dict) and "witness" in task.options.keys():
+        if not any(witness_option in options for witness_option in witness_options):
+            for witness_option in witness_options:
```
PhilippWendler, line 148: Is it really intended that if `witness_options` is a list, the tool gets the witness passed with each of them? I don't expect that a tool needs the witness more than once on the command line. The documentation also does not explain this case.

I would suggest to replace with `witness_options` with a single value.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
...
+
+
+def get_witness_options(options, task, witness_options):
+    """
+    This function returns the additional options to handle witnesses.
+    It checks if the witness is passed as an option or through the task definition.
+    If the witness is passed through both, it raises an exception.
+
+    @param options: List of existing options to see weather the witness
+        has already been passed to the tool
```
PhilippWendler, line 134: What precisely the function uses this list for is not necessary to know for the caller, I would say. Removing this avoids clutter.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
...
+    @param task: An instance of a task
+    @param witness_options: List of options which need to be set to handle witnesses
+        e.g. if the options are ["-w"], then the witness should be passed as "-w witness_file"
+        to the tool
+
+    @return: List of additional options to handle witnesses. They are constructed
+        based on witness_options. For example if witness_options is ["-w"], then
+        the return value will be ["-w", "witness_file"]
```
PhilippWendler, line 142: The documentation of this function needs to explain whether it is allowed to call it for regular tasks without any witness, and if yes, what the result will be in this case.

benchexec/tools/fshell-witness2test.py
```diff
@@ -62,7 +66,10 @@ def cmdline(self, executable, options, task, rlimits):
         if data_model_param and data_model_param not in options:
             options += [data_model_param]
 
-        return [executable] + options + list(task.input_files_or_identifier)
+        input_files = get_non_witness_input_files(task)
+        witness_options = ["--graphml-witness"]
```
PhilippWendler, line 70: Hm, this seems to hard-code an assumption about what specific format the witness is in. But of course this is not guaranteed. What to do here?

marian-lingsch: The option was taken from the benchmark definition file for `fshell-witness2test`. Therefore I think this is tool dependent on how to handle this. Currently they also only accept witnesses in version 1.0 for which the option makes sense.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
...
+    other_files = []
+    for file in input_files:
+        if Path(file).name == task.options.get("witness"):
+            witness_files.append(file)
+        else:
+            other_files.append(file)
+    return witness_files, other_files
+
+
+def get_witness_input_files(task):
```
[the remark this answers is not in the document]
marian-lingsch: The goal of this function is to have an abstraction layer between `_partition_input_files` and `get_unique_witness`. If you prefer I can remove it since it is only called in `get_unique_witness`.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
...
+    The witness file is identified by the option "witness" in the task options.
+
+    @param task: An instance of a task
+    @return: List of non-witness files
+    """
+    _, other_files = _partition_input_files(task)
+    return other_files
+
+
+def get_unique_non_witness_input_files(task):
```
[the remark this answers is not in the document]
marian-lingsch: Sorry about that, I was unsure about the nomenclature. Changed it to use `single`

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Makes sense. My goal was to indicate which options should be passed, since many tools adjust their options during the creation of the command line.

Has been changed.

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Thanks for pointing this out!

Just improved the documentation to reflect this

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: For all tools except goblint this is correct. Goblint requires two parameters to be set to validate a correctness witness in version 2.0. Due to this I chose to implement the more general case which uses a list as `witness_options`.

Should I replace it with a single value and just call it twice inside goblint's toolinfo module?

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,114 @@ def get_data_model_from_task(task, param_dict):
                     f"Unsupported data_model '{data_model}' defined for task '{task}'"
                 )
     return None
+
+
+def _partition_input_files(task):
+    """
+    This function partitions the input files into witness and other files.
```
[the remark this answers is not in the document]
marian-lingsch: Thanks a lot for pointing this out!

I added new functions to keep it consistent with the previous behavior of the tools.

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: Maybe this is better yes.

> Goblint requires two parameters to be set to validate a correctness witness in version 2.0.

Because you are referring specifically to version 2.0 here: Does this mean that it will fail for non-2.0 witnesses? If differences between formats are important here, we need to think about how to handle this.

benchexec/tools/sv_benchmarks_util.py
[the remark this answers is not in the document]
PhilippWendler: It should at least not clutter the public interface. But even internally I do not see that it is worth it.

benchexec/tools/sv_benchmarks_util.py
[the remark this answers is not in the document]
PhilippWendler: This should be done consistently, of course.

But for witnesses, the format extension that you propose only supports a single witness anyway. So why distinguish between witnesses and single witnesses at all? Why not simply use for example `get_witness`?

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,132 @@ def get_data_model_from_task(task, param_dict):
...
+def get_non_witness_input_files(task):
+    """
+    This function returns the non-witness input files from the task.
+    They consist of all files which do not match the witness file name.
+    The witness file is identified by the option "witness" in the task options.
+
+    @param task: An instance of a task
+    @return: List of non-witness files
+    """
+    _, other_files = _partition_input_files(task.input_files, task.options)
```
PhilippWendler, line 121: I think here we should throw if there is no input file, similarly to how `task.input_files` throws if it would be empty.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,132 @@ def get_data_model_from_task(task, param_dict):
...
+def get_witness_input_files(task):
+    """
+    This function returns the witness input files from the task.
+    They are identified by the option "witness" in the task options.
+
+    @param task: An instance of a task
+    @return: List of witness files
+    """
+    witness_files, _ = _partition_input_files(
+        task.input_files_or_identifier, task.options
```
PhilippWendler, line 75: This is kind of a hack: It works because it is highly unlikely that someone creates a task with no input files but a witness. It also contradicts the documentation and naming of `_partition_input_files`, which only refers to files.

But I think we can keep it as is, if we add a comment that says that we can just pass through the task identifier as if it were a file name because it will not match a witness name.

benchexec/tools/dartagnan.py
```diff
@@ -26,7 +30,10 @@ def project_url(self):
     def cmdline(self, executable, options, task, rlimits):
         if task.property_file:
             options += [task.property_file]
-        return [executable] + options + list(task.input_files_or_identifier)
+        input_files = get_non_witness_input_files_or_identifier(task)
+        witness_options = ["-witness"]
+        additional_options = get_witness_options(options, task, witness_options)
```
PhilippWendler, line 35: I can understand that this is easier to add, but it creates inconsistent tool-info modules, for example this one. The property file gets added to `options`, but then an additional `additional_options` is also created. Future developers will likely wonder why this is the case and whether it is relevant.

Can we attempt to adjust the changes to each module such that the style of the code in the module remains consistent?

benchexec/tools/concurrentwitness2test.py
```diff
@@ -26,7 +30,10 @@ def version(self, executable):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, task, rlimits):
-        return [executable, task.single_input_file] + options
+        witness_options = ["--witness"]
+        additional_options = get_witness_options(options, task, witness_options)
+        input_file = get_unique_non_witness_input_files(task)
```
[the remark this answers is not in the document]
marian-lingsch: That is a very good idea!

Currently I used an enum as a compromise to have a single function but still be able to distinguish between the three options to access the files of the task. Hopefully that is appropriate for this case.

benchexec/tools/sv_benchmarks_util.py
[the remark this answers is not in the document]
marian-lingsch: Okay, remove it :ok_hand:

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +42,132 @@ def get_data_model_from_task(task, param_dict):
...
+
+
+def get_witness_options(options, task, witness_options):
+    """
+    This function returns the additional options to handle witnesses.
+    It checks if the witness is passed as an option or through the task definition.
+    If the witness is passed through both, it raises an exception.
+    If no witness is given in the task, it returns an empty list.
+    Therefore, this function can be used with any task regardless if it
+    has a witness option or not.
```
PhilippWendler, line 139: I think we should add some more high-level / general documentation. Right now I do not think that a future developer of a tool-info module would know how to implement support for the witness option in task definitions except by copying it from an existing tool-info module.

So we should write something like "To handle the witness keyword in task.options, ... append the result of get_witness_option to the options passed to the tool ... use get_..._non_witness_input_files to retrieve the list of remaining input files for your tool ..."

Maybe the best place to add is here, at least I do not know a better place.

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: > Maybe this is better yes.
Okay, will change the code to reflect this :ok_hand: 

> > Goblint requires two parameters to be set to validate a correctness witness in version 2.0.
> 
> Because you are referring specifically to version 2.0 here: Does this mean that it will fail for non-2.0 witnesses? If differences between formats are important here, we need to think about how to handle this.

If the options for goblint are different for non-2.0 witnesses it will fail. The difference between formats makes a difference, I tried to adjust the tool info modules based on the options given in the benchmark definition files. It would be easy to add support for other witnesses, by checking the witness version, for example based on the file ending or by parsing it and then adjust the options correspondingly.

benchexec/tools/concurrentwitness2test.py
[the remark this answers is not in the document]
marian-lingsch: > But why use the additional variable for the constant witness option and not just inline it? Then we at least have only two additional lines in many cases.

Since it was currently a list, I introduced a variable to make it easier to change, but you are right it makes little sense. I adjusted this with commit fcba870b6534a0dea0df438d3a16903fdfbdd756

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Sounds good, just added the comment.

PhilippWendler:
I would prefer if you get another pair of eyes to look over the final state of the PR to check that the adjustments made to each tool-info module look correct, i.e., no mixup of single input file / input files / input files or identifier, correct addition of the options, etc. It is easy to overlook such mistakes after looking at the code so much.

benchexec/tools/dartagnan.py
marian-lingsch: Hopefully fcba870b6534a0dea0df438d3a16903fdfbdd756 present an improvement with respect to this comment. Here I tried to keep the `witness_options` sepparate from the other options and consistent accross tool info modules.

PhilippWendler: Hm, yes, with the additional benefit from naming it `witness_options` it seems worth it to just keep it always separate.

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: Now this should be part of `handle_witness_of_task`, of course. And it can be shorter because everything is now much easier to understand from the point of view of a caller.

benchexec/tools/concurrentwitness2test.py
```diff
@@ -26,7 +30,10 @@ def version(self, executable):
         return self._version_from_tool(executable)
 
     def cmdline(self, executable, options, task, rlimits):
-        return [executable, task.single_input_file] + options
+        witness_options = ["--witness"]
+        additional_options = get_witness_options(options, task, witness_options)
+        input_file = get_single_non_witness_input_file(task)
```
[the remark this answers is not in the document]
marian-lingsch: Thanks a lot for pointing this out!

Must have missed it, when going through all validators

benchexec/tools/cpachecker.py
```diff
@@ -180,16 +183,21 @@ def option_present(option):
...
                     )
 
+        if isinstance(task.options, dict) and "witness" in task.options.keys():
+            if not option_present("witness"):
+                options += [f"{prefix}witness", get_unique_witness(task)]
+            else:
+                raise benchexec.tools.template.UnsupportedFeatureException(
+                    "You are passing a witness as both an option and through the task definition. "
+                    "Please remove one of them."
+                )
```
PhilippWendler, line 193: Can `get_witness_options` not be used here?

marian-lingsch: Thanks for pointing out this simplification!

Just implemented it

marian-lingsch:
> I would prefer if you get another pair of eyes to look over the final state of the PR to check that the adjustments made to each tool-info module look correct, i.e., no mixup of single input file / input files / input files or identifier, correct addition of the options, etc. It is easy to overlook such mistakes after looking at the code so much.

Thanks a lot for all your help!

@danieldietsch, @sim642 would you be willing to take a look as participants in the validation track or do you know of someone who could do it?

danieldietsch (changes requested):
I only looked at sv_benchmarks_utils.py and ultimate.py, and this looks good. 

However, I also tried to run the examples from https://github.com/user-attachments/files/17509366/test-validators.zip and got 
```
  File "/storage/repos/benchexec/bin/../benchexec/localexecution.py", line 285, in run
    self.execute(currentRun)
  File "/storage/repos/benchexec/bin/../benchexec/localexecution.py", line 304, in execute
    args = run.cmdline()
           ^^^^^^^^^^^^^
  File "/storage/repos/benchexec/bin/../benchexec/model.py", line 1055, in cmdline
    self._cmdline = cmdline_for_run(
                    ^^^^^^^^^^^^^^^^
  File "/storage/repos/benchexec/bin/../benchexec/model.py", line 207, in cmdline_for_run
    args = tool.cmdline(rel_executable, list(options), task, rlimits)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/storage/repos/benchexec/bin/../benchexec/containerized_tool.py", line 91, in proxy_function
    return self._forward_call(method_name, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/storage/repos/benchexec/bin/../benchexec/containerized_tool.py", line 83, in _forward_call
    return self._pool.apply(_call_tool_func, [method_name, list(args), kwargs])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/multiprocessing/pool.py", line 360, in apply
    return self.apply_async(func, args, kwds).get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/multiprocessing/pool.py", line 774, in get
    raise self._value
AttributeError: 'dict' object has no attribute 'options'
```

Did you rerun your tests with the latest version?

marian-lingsch:
> I only looked at sv_benchmarks_utils.py and ultimate.py, and this looks good.
> 
> However, I also tried to run the examples from https://github.com/user-attachments/files/17509366/test-validators.zip and got
> 
> ```
> ....
> ```
> 
> Did you rerun your tests with the latest version?

Sorry about that, I forgot to run the tests. The problems have been addressed with commit cc08b49908087610ed1d646d09a0932cbfdef567

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +43,187 @@ def get_data_model_from_task(task, param_dict):
...
+
+    @param task: An instance of a task
+    @return: Unique witness file
+    """
+    witness_files, _ = _partition_input_files(
+        task.input_files_or_identifier, task.options
+    )
+    if len(witness_files) > 1:
+        raise UnsupportedFeatureException(
+            "Tool does not support tasks with more than one witness file"
```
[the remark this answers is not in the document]
marian-lingsch: Thanks a lot for pointing this out!

Since currently the proposed format for validation tasks supports only a single witness, this is not yet relevant IMO. For an extension of the validation tasks to multiple witnesses, this definitely needs to be extended, but I would do this once it is relevant.

To make it easier to locate the the reason for a possible failure in the future I added a more explicit warning as you suggested in commit b864071578082f1b7c7c98eb3c877688dc4f4d9a

danieldietsch (approved):

danieldietsch:
> > I only looked at sv_benchmarks_utils.py and ultimate.py, and this looks good.
> 
> Thanks a lot!
> 
> > However, I also tried to run the examples from https://github.com/user-attachments/files/17509366/test-validators.zip and got
> > ```
> > ....
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Did you rerun your tests with the latest version?
> 
> Sorry about that, I forgot to run the tests. The problems have been addressed with commit [cc08b49](https://github.com/sosy-lab/benchexec/commit/cc08b49908087610ed1d646d09a0932cbfdef567)

I tried again and it works.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+):
+    """
+    This function returns the input files and witness options for a task.
+    The input files are based on the task_files_considered parameter.
+    The witness options are based on the witness_options parameter.
+
+    In validation task definition files, there is a key "witness" inside the
+    options of a task. To handle this type of tasks correctly in the tool info module
+    you should do the following two steps, if your tool does not already support
+    them natively:
```
PhilippWendler, line 204: What do you mean with "support natively"? A witness validator in the context of SV-COMP never even gets the task-definition files.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+    2. Filter the input files to only get those tasks which are not witnesses. This
+        can be done by calling one of the following functions, depending on the
+        type of input files you want:
+            * `get_non_witness_input_files_or_identifier`: If you want to consider
+                both input files and identifiers
+            * `get_non_witness_input_files`: If you want to consider only input files
+            * `get_single_non_witness_input_file`: If you want to consider only a single
+                input file
+    This function takes care of the above two steps and returns the input files and
+    witness options.
```
PhilippWendler, line 216: This is written in the wrong order. A developer does not need to implement something that takes 9 lines to describe. They just need to call this function *and* (this is missing) appropriately use the result, i.e., append the options to the command line and use the returned input files instead of the fields of the task instance. *After* this is described, we can give additional information on what this does for developers who want to know more or can not make use of `handle_witness_of_task` and need to implement something on their own.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+
+    @param task: An instance of a task
+    @param options: List of existing options
+    @param witness_options: List of options which need to be set to handle witnesses
+        e.g. if the options are ["-w"], then the witness should be passed as "-w witness_file"
```
PhilippWendler, line 221: This is written in a wrong way, the "if ... then ... should" goes into the other direction.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+                both input files and identifiers
+            * `get_non_witness_input_files`: If you want to consider only input files
+            * `get_single_non_witness_input_file`: If you want to consider only a single
+                input file
+    This function takes care of the above two steps and returns the input files and
+    witness options.
+
+    @param task: An instance of a task
+    @param options: List of existing options
+    @param witness_options: List of options which need to be set to handle witnesses
```
PhilippWendler, line 220: Didn't we discuss that this should not be a list? If it remains a list, the documentation at least needs to explain what happens if there are more than one entries.

Same for `get_witness_options`, of course.

In principle it would also be possible to support only one option in the high-level API and more possibility in the low-level API, if this is useful.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+    witness_files, _ = _partition_input_files(
+        task.input_files_or_identifier, task.options
+    )
+    if len(witness_files) != 1:
+        logging.warning(
+            "Tool uses benchexec default validation task processing. "
+            "It does not support tasks with more than one witness"
+        )
+        raise UnsupportedFeatureException(
+            "Tool only supports tasks with exactly one witness file"
```
PhilippWendler, line 87: First, logging *and* throwing is an anti-pattern, one needs to do either of them, not both. If one does both, the two parts of the user message will get split and might be logged with several lines in between, or one line might not be logged at all. Furthermore, suppose we implement a wrapper that takes care of this situation and handles the exception, then we would not want any message to be shown to the user. But the wrapper could not avoid the log message.

Second, there is no "benchexec default validation task processing". This wording will lead people to think as if BenchExec somehow defines and implements how to handle validation tasks, but this is not true. BenchExec itself does not care or know or enforce anything about validation tasks. It is just the tool-info modules (most of which happen to be bundled with BenchExec) that do this.

Third, according to your [description](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1572) of these tasks, the `witness` option needs to be a single string with a path. So it is not even the tool-info module that does not support more than one witness, but your format extension.

Fourth, what happens here is not really about "one witness" or "more than one witness". As mentioned above, we only have one witness in the options. What has happened if we enter this case is that we encountered a task definition where there is one witness that occurs more than once in the input files, at least with the same base name:
```yaml
input_files: ["foo/witness.yml", "bar/witness.yml", "program.c"]
options:
  witness: "witness.yml"
```
Now, we could in principle allow this (just append each of the witness files to the command line), but I do not really see a reason why we would want this. It seems this is more likely an error in the task definition, and it is nicer for the user if we detect this.

Fifth, the current error message about "more than one witness" is also produced when the witness is missing from the input files. This needs to be taken into account when writing the message and in the documentation (where it is currently also wrong).

I think we should probably have a message that says something like the witness in the task definition did not occur exactly once in the input files of the task definition.

Most of these points also apply below.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,196 @@ def get_data_model_from_task(task, param_dict):
...
+        if Path(file).name == task_options.get("witness"):
+            witness_files.append(file)
+        else:
+            other_files.append(file)
+    return witness_files, other_files
+
+
+def get_witness(task):
+    """
+    This function returns the unique witness file from the task.
```
PhilippWendler, line 71: Please remove all occurrences of "unique" unless you are actually talking about uniqueness, e.g., when eliminating duplicates.

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: My goal was to indicate that tool developers could interpret the options in the task internally instead of changing their command line. I made this more precise in 15eba32e7cd6a91604fad75e91f449b14c865765

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Sorry about that. Changed the order of the elements and made it more precise what happens in commit 944d21e0f00de377ed952a1e1a8ab4805457d365

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Thanks!

Fixed in 55a066037999a637ec701b77a6c2447cb535821b

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Sorry I forgot to implement the change when we discussed it.

I like the compromise to have a high-level API `handle_witness_of_task` which takes only a single option and the low level API `get_witness_options` taking a list. Addressed in 1c4bcbdbc9b9c9a94fc2cf191ce5d1babed69090

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Thanks a lot for the amazing explanation!

To address the concerns I removed the extra logging and adjusted the error message into one similar to your suggestion in commits 833b33a9ebafe3b337c47000f28704ce8e7b0724 and 21021b244a2e6bc96408566994fb7fb94c0ded95

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Sorry forgot about fixing this in the documentation, improved in 267f9266f09fdcf2c480f07606020d827efb49ab

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: Sorry, but "if your tool does not already support them by parsing the options correctly" still reads if there was a way for the tool to parse the options (in this context these are the options defined in the task definition) directly. The subject of the sentence would need to be `your tool-info module`. And it is better to use `task options`, because tool-info modules also get the options from the benchmark definition. Or just remove everything after the comma, I do not think it adds much value.

marian-lingsch: Makes sense, the comment was deleted in 944d21e0f00de377ed952a1e1a8ab4805457d365

PhilippWendler:
The code is fine for me after the small comment tweak discussed above.

However, I would still prefer if someone independent has a quick look at all the tool-info modules and compares the old and new code to check whether for example the correct enum value is passed, whether the options are inserted correctly, and whether the input files are taken correctly from the result of the new function. The reviewer would not need to be familiar with the existing code nor would they need to look at the code of `sv_benchmarks_utils.py`. I don't think it is much effort, I just prefer to not do it myself because I think it is likely that people involved in writing such code will easily overlook mistakes.

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: Right, overlooked.

PhilippWendler (approved):

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: @danieldietsch Which case specifically are you aiming at?

I would like to again note (cf. item 4 of my long comment above) that the original error message about "more than one witness" did not describe the situation correctly at all and thus lead the discussion into a wrong direction based on wrong assumptions. The case handled with this error message did *not* occur if more than one witness was present in the task definition, instead it occurred if someone created a task definition that lists the single witness twice in the input files (cf. my example above). This simply does not make sense, we do not want to support it, and it is the responsibility of the user to fix their task definition. The current error message correctly describes that.

For your current comment, I am sorry but I don't really understand all of it. What I can say is:

BenchExec has no responsibility here at all. Tool-info modules for validators that want to compete in the SV-COMP validation track need to handle task definitions with a `witness` key in the options, this is the responsibility of the tool-info module. In order to reduce the implementation and maintenance effort, we provide utility methods that make it easy to implement this, and tool-info modules can and should make use of them. And in order to push this forward, @marian-lingsch has volunteered to adapt all existing tool-info modules. Thus, after this PR the developers of the tool-info module do not need to do anything unless they change how their tool needs to be called, so I do not see any point where some responsibility is left. Furthermore, there is also no way to support a superset or a subset of the features currently being handled: The format extension proposed by @marian-lingsch supports task definitions with exactly one witness, a tool-info module can either support this format extension or not, but not partially. And tool-info modules also cannot support more than one witness file in tasks with @marian-lingsch's format extension, because that extension only uses a single string to define the witness.

I hope this helps. If you have further questions, please clarify.

danieldietsch: No, that helps. I guess my confusion comes from the naming and parts of the docs, e.g., https://github.com/sosy-lab/benchexec/blob/addcaa97e5181e029bc10d58edffc948a4019d96/benchexec/tools/sv_benchmarks_util.py#L70

You are right, this _only_ deals with witnesses specified with yaml `options/witness:` structure (how would one write that in one line?). 

The API and the current docs make it somewhat hard to distinguish these witnesses from the witnesses contained in the `input_files` part. The exception messages also do not distinguish these cases, e.g. 
```
         "Witness given in the task-definition did not occur exactly once "
            "in input files of the task definition"
```
The witnesses given via input_files are "witness given in the task-definition" as well. 

So, basically, distinguishing _the_ "option-witness" from the sequence of "input_files-witnesses" is the problem.

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +43,190 @@ def get_data_model_from_task(task, param_dict):
...
+    is identified by the option "witness" in the task options.
+
+    @param input_files: List of input files
+    @param task_options: Dictionary of task options
+    @return: Tuple of witness files and other files
+    """
+    witness_files = []
+    other_files = []
+    for file in input_files:
+        if Path(file).name == task_options.get("witness"):
```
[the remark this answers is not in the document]
PhilippWendler: Yes, and the reason for this is also the reason for why there even is this duplication at all.

The problem is that the witness is a relative path, and needs to be treated like all other input paths, i.e., BenchExec needs to adjust it to make it a path that is relative to the working directory of the tool. But all the stuff inside the `options` keyword is (from BenchExec's point of view) just arbitrary additional information that it passes through to the tool-info module. We do not want to hard-code information about specific keys in this dict. (If we would for example require in BenchExec that `witness` is a string, this would prevent other communities from using the same key for something else.) So we decided to list the witness as an input file as well. For them, BenchExec knows that it needs to do the proper path adjustment. So the tool-info module gets all the correct relative paths, and needs to filter out the one for the witness. For this only the base name of the witness file is usable.

benchexec/tools/sv_benchmarks_util.py
PhilippWendler: I have attempted to clarify this in 25ca9b85.

dbeyer:
Dear All, Many thanks for the great work done here.
I am now designing and considering the changes to the rest of the competition infrastructure,
including the category structure, the CI checks, and the scoring schema in SV-COMP.
I would like to ask you to wait with merging until I am done with the design of the rest of the components.

PhilippWendler:
This PR depends on the format in https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1572, so unless that is accepted it should not be merged anyway.

benchexec/tools/sv_benchmarks_util.py
danieldietsch: This looks very helpful and much better, thank you!

benchexec/tools/sv_benchmarks_util.py
```diff
@@ -40,3 +44,195 @@ def get_data_model_from_task(task, param_dict):
...
+    if isinstance(task.options, dict) and "witness" in task.options.keys():
+        if not any(witness_option in options for witness_option in witness_options):
+            additional_options = []
+            for witness_option in witness_options:
+                additional_options += [witness_option, get_witness(task)]
+            return additional_options
+        else:
+            logging.warning(
+                "You are passing a witness as both an option and through the task definition. "
+                "Using the witness passed as option."
```
PhilippWendler, line 174: Maybe something stronger, like "The witness in the task definition will be ignored."?

benchexec/tools/witnessmap.py
```diff
@@ -0,0 +1,53 @@
...
+            if line_without_log_prefix.startswith(identifier):
+                return line_without_log_prefix.split(":", maxsplit=1)[-1].strip()
+        return None
+
+    def determine_result(self, run):
+        exit_code = run.exit_code.value
+        if exit_code != 0:
+            return result.RESULT_ERROR
+
+        return result.RESULT_DONE
```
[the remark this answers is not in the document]
marian-lingsch: That is good to know!

Fixed in 29bc80a9638b02ec300dbb7639270ace1c77c61f

benchexec/tools/witnessmap.py
```diff
@@ -0,0 +1,53 @@
...
+        return "https://gitlab.com/sosy-lab/software/witnessmap"
+
+    def version(self, executable):
+        version_string = self._version_from_tool(executable)
+        return version_string.partition("version")[2].strip().split(" ")[0]
+
+    def cmdline(self, executable, options, task, rlimits):
+        mapping_options = get_witness_options(options, task, ["--input"])
+
+        return [executable, *options, *mapping_options]
```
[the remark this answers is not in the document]
marian-lingsch: Thanks for pointing this out!

Done in 0bad285a89baf1e7d4a93604c4d404d43b205870

benchexec/tools/sv_benchmarks_util.py
marian-lingsch: Tried to improve the formulation in 3dd7093997a752a253ca792fd65c88fc0276c110, hopefully that works.

benchexec/tools/witnessmap.py
```diff
@@ -0,0 +1,53 @@
...
+import benchexec.result as result
+import benchexec.tools.template
+from benchexec.tools.sv_benchmarks_util import (
+    get_witness_options,
+)
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for the witness mapper (witnessmap)
```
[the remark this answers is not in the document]
marian-lingsch: That makes a lot of sense, made it explicit in e4e628bdc4d9bfeb4cedcd52927850debca91375.

Thanks for making the requirement for backwards compatibility explicit, I will take care that it conforms to this.

PhilippWendler:
@marian-lingsch Did you find a reviewer for https://github.com/sosy-lab/benchexec/pull/1093#issuecomment-2444263922 already?

marian-lingsch:
> @marian-lingsch Did you find a reviewer for [#1093 (comment)](https://github.com/sosy-lab/benchexec/pull/1093#issuecomment-2444263922) already?

I saw that @sim642 and @danieldietsch did some pass over the files, but will look for someone to do a full review

marian-lingsch:
> What is the URL for the description of this `witness` key and how it is supposed to be used? We should add the link to the docs of `handle_witness_of_task`.

Thanks for pointing this out!

Addressed in https://github.com/sosy-lab/benchexec/pull/1093/commits/b550e76081461c2da6d3ffbd8bb23cb7bb4d33b5

matthiaskettl:
I went over the changes and could not find any inconsistencies.

marian-lingsch:
> I went over the changes and could not find any inconsistencies.

Thanks a lot @PhilippWendler this should take care of https://github.com/sosy-lab/benchexec/pull/1093#issuecomment-2444263922 right?

PhilippWendler:
> > I went over the changes and could not find any inconsistencies.
> 
> Thanks a lot @PhilippWendler this should take care of [#1093 (comment)](https://github.com/sosy-lab/benchexec/pull/1093#issuecomment-2444263922) right?

Thanks @matthiaskettl, yes.

dbeyer:
@marian-lingsch Could you please update the format of the witness in the description? It currently does not match version 2.1.
(Once you are at it, please also fix 'unsert' and use 'validation task file' instead of 'witness task-definition file' (mind the hyphen).)

dbeyer:
> Dear All, Many thanks for the great work done here. I am now designing and considering the changes to the rest of the competition infrastructure, including the category structure, the CI checks, and the scoring schema in SV-COMP. I would like to ask you to wait with merging until I am done with the design of the rest of the components.

This is finished now and I believe all necessary changes are made (in MRs so far).

dbeyer (approved):

dbeyer (approved):

dbeyer:
@marian-lingsch Also, the description of the action in the table "Exception (Two witnesses)" should be adjusted to the implementation: "witness in task definition ignored".

marian-lingsch:
> @marian-lingsch Could you please update the format of the witness in the description? It currently does not match version 2.1. (Once you are at it, please also fix 'unsert' and use 'validation task file' instead of 'witness task-definition file' (mind the hyphen).)

@dbeyer many thanks for catching the inconsistency in the description. This was left over from the previous validation task file format. Has been updated.

PR #1101, opened 2024-11-06
Fix Korn toolinfo module, include data_model

gernst:


benchexec/tools/korn.py
```diff
@@ -34,8 +37,21 @@ def version(self, executable):
...
+
+        data_model = task.options.get("data_model")
+
+        if data_model == "ILP32":
+            cmd = cmd + ["-32"]
+
+        if data_model == "LP64":
+            cmd = cmd + ["-64"]
+
+        cmd = cmd + [task.single_input_file]
```
PhilippWendler, line 52: Compared to the previous implementation, this restricts the tool-info module to tasks with exactly one input file. But I guess that the tool does not support other tasks anyway?

benchexec/tools/korn.py
```diff
@@ -34,8 +37,21 @@ def version(self, executable):
     def name(self):
         return "Korn"
 
-    def project_url(self):
-        return "https://github.com/gernst/korn"
```
PhilippWendler, line 38: Let's keep this? Then you can remove it from the comment again.

benchexec/tools/korn.py
```diff
@@ -34,8 +37,21 @@ def version(self, executable):
...
+        cmd = [executable]
+        cmd = cmd + options
+
+        data_model = task.options.get("data_model")
+
+        if data_model == "ILP32":
+            cmd = cmd + ["-32"]
+
+        if data_model == "LP64":
+            cmd = cmd + ["-64"]
```
PhilippWendler, line 50: Many tool-info modules check whether the option is already present, and if yes, do not add it again. This keeps old benchmark definitions working. Not crucial, but maybe nice to have?

There is also a utility function in `sv_benchmarks_util`, which you could use, for example like this:
```python
            task, {ILP32: "-32", LP64: "-64"}
        )

        if data_model_param and data_model_param not in options:
            cmd += data_model_param
```

gernst: Thanks! Sounds good.

benchexec/tools/korn.py
gernst: oh right

benchexec/tools/korn.py
gernst: Yes, that is intended

PhilippWendler (approved):

PhilippWendler:
In the future, please use proper commit messages. For now I am just going to squash the commits.

PR #1102, opened 2024-11-07
Update Ultimate tool info module

maul-esel:
This PR updates the tool info modules for Ultimate tools such that they are able to run the upcoming Ultimate version 0.3.0.

The second commit also fixes a bug in the detection of the Java version to be used by the tool info module (`glob.glob("java")` always returns `[]`).

### DIFF
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -30,6 +30,7 @@ _ULTIMATE_VERSION_REGEX = re.compile(r"^Version is (.*)$", re.MULTILINE)
 _LAUNCHER_JARS = [
     "plugins/org.eclipse.equinox.launcher_1.5.800.v20200727-1323.jar",
     "plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar",
+    "plugins/org.eclipse.equinox.launcher_1.6.800.v20240513-1750.jar",
 ]
 
 
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -467,6 +467,8 @@ class UltimateTool(benchexec.tools.template.BaseTool2):
         return None
 
     def get_java_installations(self):
+        # The code in this method is (and should remain) consistent with the method `get_java` in
+        # <https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/default/adds/Ultimate.py>.
         candidates = [
             "java",
             "/usr/bin/java",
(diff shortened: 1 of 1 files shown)


benchexec/tools/ultimate.py
```diff
@@ -474,11 +474,16 @@ def get_java_installations(self):
...
+            if "*" in c:
+                candidates_extended += glob.glob(c)
+            else:
+                candidates_extended += [c]
         pattern = r'"(\d+\.\d+).*"'
 
         rtr = {}
-        for c in candidates:
+        for c in candidates_extended:
             candidate = shutil.which(c)
```
PhilippWendler, line 487: With your explanation, I could understand what the code is doing, but not without it. Maybe it is better to refactor this completely?

Something like `candidates = [shutil.which("java"), "/usr/bin/java"] + glob.glob("/opt/oracle-jdk-bin-*/bin/java") + ...` should make much more clear what is being done.

benchexec/tools/ultimate.py
```diff
@@ -71,15 +73,13 @@ def project_url(self):
...
     def executable(self, tool_locator):
         exe = tool_locator.find_executable("Ultimate.py")
-        dir_name = os.path.dirname(exe)
-        logging.debug("Looking in %s for Ultimate and plugins/", dir_name)
-        for _, dir_names, file_names in os.walk(dir_name):
-            if "Ultimate" in file_names and "plugins" in dir_names:
-                return exe
-            break
+        dir_name = Path(os.path.dirname(exe))
+        logging.debug(f"Checking if {dir_name} contains a launcher jar")
```
PhilippWendler, line 77: We prefer `%s` for logging because the logging library has builtin support for that and can handle it lazily.

benchexec/tools/ultimate.py
```diff
@@ -71,15 +73,13 @@ def project_url(self):
...
+        if any([(dir_name / rel_launcher).exists() for rel_launcher in _LAUNCHER_JARS]):
```
PhilippWendler, line 78: Does the new check also work for all the Ultimate versions for which the old check works?

maul-esel: Yes, all the Ultimate versions should have a launcher JAR. Otherwise, even if the check would work, the actual invocation of Ultimate would fail.

benchexec/tools/ultimate.py
maul-esel: I simply copied this code from Ultimate's own wrapper script `Ultimate.py` to really be sure both methods resolve the Java version in exactly the same way. We've had some issues when benchmarking recently, because depending on how you invoke benchexec, the tool info module and our wrapper script find different Java versions and java crashes with rather cryptic errors.

But if you think it's important, I could have a look at refactoring this code (in both scripts).

PhilippWendler: Ok, I see. I can understand that you prefer it to be similar. Maybe you even want to cross-reference the code then?

While I do recommend considering the refactoring, of course this does not need to be done now.

benchexec/tools/ultimate.py
maul-esel: :+1: fixed

PhilippWendler (approved):

benchexec/tools/ultimate.py
maul-esel: :+1: I've added a comment to cross-reference the wrapper script.

PR #1104, opened 2024-11-13
Modify the project_url in aise.py

liar1007:


benchexec/tools/aise.py
```diff
@@ -29,7 +29,7 @@ def name(self):
         return "AISE"
 
     def project_url(self):
-        return "https://github.com/ZhenWang233/AISE"
+        return "https://github.com/zbchen/aise-verifier.git"
```
PhilippWendler, line 32: First, this should probably be the URL without `.git`.

Second, I noticed that the new GitHub repository contains even less content than the old one, and not even the tool. Are you sure you want to designated this as the main URL for your tool?

PhilippWendler: @ZhenWang233 do you agree with this?

liar1007: Hello, [PhilippWendler](https://github.com/PhilippWendler). I apologize for any confusion. Since AISE doesn’t have a tool's webpage, we would like to provide a simple introduction to our tool via GitHub. To facilitate future updates, we would like to modify the GitHub link.

liar1007: I have already added aise.zip on [GitHub](https://github.com/zbchen/aise-verifier), and I will update it with information for the SV-COMP 2025 competition in the future.

ZhenWang233: > @ZhenWang233 do you agree with this?

Yes, It's ok, I agree with this change.

PhilippWendler: Then please just remove the `.git` from the URL.

liar1007: OK，the `.git`  has been deleted.

PR #1105, opened 2024-11-13
Add constant for 'false(no-data-race)'

dbeyer:
Contributes to solving the following issue:
https://gitlab.com/sosy-lab/benchmarking/competition-scripts/-/issues/37


PhilippWendler:
Just to state this here for the record as well: Adding the constant is a convenience for tool-info modules. It will not change the behavior of BenchExec in any way and has no influence on https://gitlab.com/sosy-lab/benchmarking/competition-scripts/-/issues/37.

PR #1109, opened 2024-11-14
Adding (sub)property to violations

leventeBajczi:
This change adds subproperties for violation results, parsed from the output of Theta. We expect `(Property <property>)` to exist in the output for this. As older versions do not support this change, we keep the `false(unreach-call)` for outputs not containing the matching string.

We also added a way to annul results by printing `SafetyResult Unknown` after the original (which is useful for using bounded or otherwise incomplete methods and only accepting certain outcomes). We deliberately want to overwrite the `status` flag by later values.

### DIFF
--- a/benchexec/tools/theta.py
+++ b/benchexec/tools/theta.py
@@ -55,7 +55,7 @@ class Tool(benchexec.tools.template.BaseTool2):
                 status = result.RESULT_TRUE_PROP
             elif "ParsingResult Success" in line:
                 parsing_status = "after"
-            elif "Property" in line and property:
+            elif "Property" in line and not property:
                 match = re.search("\(Property ([a-z-]*)\)")
                 if match:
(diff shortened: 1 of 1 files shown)


benchexec/tools/theta.py
```diff
@@ -30,14 +31,24 @@ def project_url(self):
     def version(self, executable):
         return self._version_from_tool(executable)
 
+    def min_version(self, current, limit):
+        for i in range(len(current)):
+            if int(current[i]) < int(limit[i]):
+                return False
+        return True
```
PhilippWendler, line 38: I think if you just use `[int(x) for x in current] >= limit` it does the same and works even if `limit` is a prefix of `current`, for which the current version crashes.

PhilippWendler (approved):
Thanks for taking so much care about backwards compatibility!

PR #1118, opened 2024-11-21
Hornix: Get only version number, without the name of the tool

blishko:
This seems to be the requirement of fm-tools.
The pipeline check in https://gitlab.com/sosy-lab/benchmarking/fm-tools/-/merge_requests/478 fails because the output from version query contains the name of the tool.

### DIFF
--- a/benchexec/tools/hornix.py
+++ b/benchexec/tools/hornix.py
@@ -6,6 +6,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import re
+
 import benchexec
 import benchexec.result as result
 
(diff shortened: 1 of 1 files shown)


benchexec/tools/hornix.py
```diff
@@ -21,7 +23,10 @@ def executable(self, tool_locator):
         return tool_locator.find_executable("hornix")
 
     def version(self, executable):
-        return self._version_from_tool(executable, arg="--version")
+        output = self._version_from_tool(executable, arg="--version")
+        version_pattern = r"\d+\.\d+\.\d+"
```
PhilippWendler, line 27: Is this complete and future-proof? I.e., could it happen that your tool eventually produces version numbers with not exactly 3 digits? Something like `10.0.0` or `3.1.4.2` or `3.3.3-beta1`?

I would recommend to make this more general.

PhilippWendler:
Yes, the version number of a tool inside BenchExec should only be the version number. Otherwise this leads to redundant output in tables etc.

benchexec/tools/hornix.py
blishko: First example should be covered by this regex already (note the `+`).
From the second and third this would extract `3.1.4` and `3.3.3`, respectively. That is not ideal.
But we plan to stick to the main semantic versioning.

It would help us to get this merged as soon as possible, so we could actually get a pre-run today.

PhilippWendler: Please provide a follow-up PR until tomorrow such that it can get into the next release of BenchExec.

PR #1122, opened 2024-11-22
Hornix: Strip tool name from the version output

blishko:
This is a follow up to #1118.

### DIFF
--- a/benchexec/tools/hornix.py
+++ b/benchexec/tools/hornix.py
@@ -6,8 +6,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import re
-
 import benchexec
 import benchexec.result as result
 
@@ -23,10 +21,9 @@ class Tool(benchexec.tools.template.BaseTool2):
         return tool_locator.find_executable("hornix")
 
     def version(self, executable):
-        output = self._version_from_tool(executable, arg="--version")
-        version_pattern = r"\d+\.\d+\.\d+"
-        match = re.search(version_pattern, output)
-        return match.group(0) if match else ""
+        return self._version_from_tool(
+            executable, arg="--version", line_prefix="Hornix"
+        )
 
     def name(self):
         """The human-readable name of the tool."""


benchexec/tools/hornix.py
```diff
@@ -24,8 +24,10 @@ def executable(self, tool_locator):
 
     def version(self, executable):
         output = self._version_from_tool(executable, arg="--version")
-        version_pattern = r"\d+\.\d+\.\d+"
-        match = re.search(version_pattern, output)
+
+        # regex for semantic versioning taken from https://semver.org
+        semantic_versioning_pattern = r"(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?"
+        match = re.search(semantic_versioning_pattern, output)
```
PhilippWendler, line 30: This looks like a real beast of a regexp :-)

Most tool-info module simply implement the strategy that they take the n-th word of the tool output as the version. Maybe you would also want to do that?
Or, if you think that the tool output will always look like `Hornix $VERSION` (which I think is reasonable to assume), you can just call `self._version_from_tool(executable, arg="--version", line_prefix="Hornix")`, which will return everything after the `Hornix`.

blishko: That looks like a better solution.
What about the space between Hornix and the version? Is that going to be stripped automatically? Or should `line_prefix` be `Hornix `?

PhilippWendler: There is a `.strip()` call made already.

Note that you can (and should) test this if you have an installed BenchExec with `python -m benchexec.test_tool_info hornix --tool-directory path/to/hornix` (`--read-only-dir /` might be necessary on your machine, add it if it complains).

blishko: Thanks for the help! I updated the PR according to your suggestions, and verified that it works.

PhilippWendler (approved):

benchexec/tools/hornix.py
PhilippWendler: Thanks for taking the time to improve this!

PR #1123, opened 2024-11-22
Update swat.py

nils-loose:


benchexec/tools/swat.py
```diff
@@ -16,9 +16,13 @@ class Tool(BaseTool2):
     """
 
     REQUIRED_PATHS = [
-        "knife-fuzzer",
-        "local_z3_installation",
```
PhilippWendler, line 20: You can keep the old paths in the list, otherwise you loose compatibility with old versions. If there is a pattern that does not match, it does not hurt.

PhilippWendler (approved):

PR #1124, opened 2024-11-22
Update the toolinfo of WitnessMap for version 0.0.2-dev

marian-lingsch:
Currently WitnessMap only copied the witness from a validation benchmark file into the target directory. In general, this is insufficient, since the contents of the witness need to be adjusted to match those of the input task.

Version 0.0.2-dev of WitnessMap adds this support and the toolinfo needs to also be able to interact with it.

### DIFF
--- a/benchexec/tools/witnessmap.py
+++ b/benchexec/tools/witnessmap.py
@@ -7,6 +7,8 @@
 
 import benchexec.tools.template
 from benchexec.tools.sv_benchmarks_util import (
+    handle_witness_of_task,
+    TaskFilesConsidered,
     get_witness_options,
 )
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Do you need to be able to execute old WitnessMap versions with BenchExec? If not, you can remove the outdated code.

dbeyer (approved):
Bug-fix releases should increase the 3rd number, feature additions (like this) should increase the 2nd number of the version number.

Backwards compatibility is required, but not for bug-fix releases (those versions from a day ago are not yet widespread in use ;-) ).

I would say you should rewrite this once 1.0 is release, and then only support that version and remove all other code.

marian-lingsch:
> Bug-fix releases should increase the 3rd number, feature additions (like this) should increase the 2nd number of the version number. (-dev releases should not deserve an own Tool-Info Module.)
> 
> Backwards compatibility is required, but not for versions from a day ago that are not yet widely in use.
> 
> I would say you should rewrite this once 1.0 is release, and then only support that version and remove all other code.

Sounds good, I will do the required changes once the release of version 1.0 is done

dbeyer:
@PhilippWendler Thanks for your comment also, which I did not see because I took so long to write mine.

PhilippWendler:
@dbeyer This should not have been merged. For example, besides introducing unnecessary complexity, there is also the problem that it does not cache the version.

PhilippWendler:
The rule for backwards compatibility of tool-info modules is that if a tool-info module was part of a BenchExec release, then future versions of BenchExec should aim at supporting all the tool versions that the older BenchExec supported as well. So right now we have no expectation of backwards compatibility, but as soon as this code gets into a BenchExec release, it should be kept there.

The version number of the tool does not really has an influence on this.

So if you want to get rid of code for supporting old versions, then do so before the next BenchExec release, which is at the latest tomorrow morning.

dbeyer:
> @dbeyer This should not have been merged. For example, besides introducing unnecessary complexity, there is also the problem that it does not cache the version.

We needed it merged for a documented pre-run (i.e., with a main-git-hash of BenchExec),
while in parallel @marian-lingsch is already rewriting this completely.

All this code will be gone in a few hours.

dbeyer:
> The version number of the tool does not really has an influence on this.

Yes, that part of my review was placed here although mentioned for WitnessMap.

marian-lingsch:
> @dbeyer This should not have been merged. For example, besides introducing unnecessary complexity, there is also the problem that it does not cache the version.

Sorry about that I though that your comment https://github.com/sosy-lab/benchexec/pull/1093#discussion_r1824652042 referred to the kind of backwards compatibility I implemented here.

PhilippWendler:
> Sorry about that I though that your comment [#1093 (comment)](https://github.com/sosy-lab/benchexec/pull/1093#discussion_r1824652042) referred to the kind of backwards compatibility I implemented here.

You are correct about this. But this is only relevant for released versions of BenchExec.

marian-lingsch:
> The rule for backwards compatibility of tool-info modules is that if a tool-info module was part of a BenchExec release, then future versions of BenchExec should aim at supporting all the tool versions that the older BenchExec supported as well. So right now we have no expectation of backwards compatibility, but as soon as this code gets into a BenchExec release, it should be kept there.
> 
> The version number of the tool does not really has an influence on this.
> 
> So if you want to get rid of code for supporting old versions, then do so before the next BenchExec release, which is at the latest tomorrow morning.

Sorry about that, should be fixed by https://github.com/sosy-lab/benchexec/pull/1125

marian-lingsch:
> > Sorry about that I though that your comment [#1093 (comment)](https://github.com/sosy-lab/benchexec/pull/1093#discussion_r1824652042) referred to the kind of backwards compatibility I implemented here.
> 
> You are correct about this. But this is only relevant for released versions of BenchExec.

Thanks for the explanation!

PR #1126, opened 2024-11-22
Update concurrentwitness2test.py

leventeBajczi:
Previously, tests that did not produce the desired reach_error() call were classified as 'true'. This corrects this to be error (with "ineffective witness" as the specific error)


benchexec/tools/concurrentwitness2test.py
```diff
@@ -40,7 +40,7 @@ def determine_result(self, run):
             if "Verdict: SOMETIMES" in line or "Verdict: ALWAYS" in line:
                 return result.RESULT_FALSE_REACH
             elif "Verdict: NEVER" in line:
-                return result.RESULT_TRUE_PROP
+                return result.RESULT_ERROR + "(ineffective  witness)"
```
PhilippWendler, line 43:
```suggestion
                return result.RESULT_ERROR + "(ineffective witness)"
```

PhilippWendler (approved):

PR #1132, opened 2024-12-03
Add feature to use podman container when determining version for vcloud-benchmark

ricffb:
This MR adds extra functionality for the vcloud-benchmark.py.

Some tools like PeSCo crash under newer linux versions even while obtaining the version.
And in this case the crash even causes the entire benchmark execution to fail (since it causes a index out of bounds directly in the tool info module). These tools are run inside of a podman container in the benchcloud, denoted with the `--vcloudContainerImage` option.

I propose to go ahead and also use this image while obtaining the version of the tool.

I manually tested this code with cpachecker, PeSCo, VeriAbs and nacpa. I also tested the fallback by renaming "podman" to "podmans".

I don't know, if we want to make this way of obtaining a Version optional via yet another flag (or if the `--no-container` should serve double duty here).

@dbeyer is strongly interested in some form of containering the version retrieval for the Hors Concours runs of SV-COMP25.

I am grateful for feedback how to improve, so that we can swiftly incorporate this feature.

PeSCo also needs the full tool_directory mounted to the container. I expect this from other python based tools aswell.
This is why I added the find_tool_base_dir function that mimics how the find_executable works but tries to determine the candidate that actually won.

### DIFF
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -16,6 +16,7 @@ import os
 import signal
 import socket
 import tempfile
+from abc import ABCMeta
 
 from benchexec import (
     BenchExecException,
@@ -26,12 +27,11 @@ from benchexec import (
     util,
 )
 
-
 tool: tooladapter.CURRENT_BASETOOL = None
 
 
 @tooladapter.CURRENT_BASETOOL.register  # mark as instance of CURRENT_BASETOOL
-class ContainerizedTool(object):
+class ContainerizedToolBase(object, metaclass=ABCMeta):
     """Wrapper for an instance of any subclass of one of the base-tool classes in
     benchexec.tools.template.
     The module and the subclass instance will be loaded in a subprocess that has been
@@ -45,7 +45,7 @@ class ContainerizedTool(object):
     But the use of containers in BenchExec is for safety and robustness, not security.
     """
 
-    def __init__(self, tool_module, config):
+    def __init__(self, tool_module, config, initializer):
         """Load tool-info module in subprocess.
         @param tool_module: The name of the module to load.
             Needs to define class named Tool.
@@ -57,13 +57,13 @@ class ContainerizedTool(object):
 
         container_options = containerexecutor.handle_basic_container_args(config)
         temp_dir = tempfile.mkdtemp(prefix="Benchexec_tool_info_container_")
-
+        self.container_id = None
         # Call function that loads tool module and returns its doc
         try:
-            self.__doc__ = self._pool.apply(
+            self.__doc__, self.container_id = self._pool.apply(
                 _init_container_and_load_tool,
-                [tool_module, temp_dir],
-                container_options,
+                [initializer] + self.mk_args(tool_module, config, temp_dir),
+                self.mk_kwargs(container_options),
             )
         except BaseException as e:
             self._pool.terminate()
@@ -74,8 +74,15 @@ class ContainerizedTool(object):
             with contextlib.suppress(OSError):
                 os.rmdir(temp_dir)
 
+    def mk_args(self, tool_module, config, tmp_dir):
+        return [tool_module, config, tmp_dir]
+
+    def mk_kwargs(self, container_options):
+        return container_options
+
     def close(self):
         self._forward_call("close", [], {})
+        self._cleanup()
         self._pool.close()
 
     def _forward_call(self, method_name, args, kwargs):
@@ -109,7 +116,7 @@ for member_name, member in inspect.getmembers(
 ):
     if member_name[0] == "_" or member_name == "close":
         continue
-    ContainerizedTool._add_proxy_function(member_name, member)
+    ContainerizedToolBase._add_proxy_function(member_name, member)
 
 
 def _init_worker_process():
@@ -125,15 +132,16 @@ def _init_worker_process():
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
-def _init_container_and_load_tool(tool_module, *args, **kwargs):
+def _init_container_and_load_tool(initializer, tool_module, *args, **kwargs):
     """Initialize container for the current process and load given tool-info module."""
+    container_id = None
     try:
-        _init_container(*args, **kwargs)
+        container_id = initializer(*args, **kwargs)
     except OSError as e:
         if container.check_apparmor_userns_restriction(e):
             raise BenchExecException(container._ERROR_MSG_USER_NS_RESTRICTION)
         raise BenchExecException(f"Failed to configure container: {e}")
-    return _load_tool(tool_module)
+    return _load_tool(tool_module), container_id
 
 
 def _init_container(
@@ -288,3 +296,11 @@ def _call_tool_func(name, args, kwargs):
     except SystemExit as e:
         # SystemExit would terminate the worker process instead of being propagated.
         raise BenchExecException(str(e.code))
+
+
+class ContainerizedTool(ContainerizedToolBase):
+    def __init__(self, tool_module, config, initializer):
+        super().__init__(tool_module, config, initializer, initializer=_init_container)
+
+    def _cleanup(self):
+        pass
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -54,28 +54,28 @@ class ContainerizedTool(object):
         # We use multiprocessing.Pool as an easy way for RPC with another process.
         self._pool = multiprocessing.Pool(1, _init_worker_process)
 
-        self.container_options = containerexecutor.handle_basic_container_args(config)
-        self.temp_dir = tempfile.mkdtemp(prefix="Benchexec_tool_info_container_")
-
         # Call function that loads tool module and returns its doc
         try:
-            self._setup_container(tool_module)
-
+            self._setup_container(tool_module, config)
         except BaseException as e:
             self._pool.terminate()
             raise e
+
+    def _setup_container(self, tool_module, config):
+        container_options = containerexecutor.handle_basic_container_args(config)
+        temp_dir = tempfile.mkdtemp(prefix="Benchexec_tool_info_container_")
+
+        try:
+            self.__doc__, _ = self._pool.apply(
+                _init_container_and_load_tool,
+                [_init_container, tool_module, temp_dir],
+                container_options,
+            )
         finally:
             # Outside the container, the temp_dir is just an empty directory, because
             # the tmpfs mount is only visible inside. We can remove it immediately.
             with contextlib.suppress(OSError):
-                os.rmdir(self.temp_dir)
-
-    def _setup_container(self, tool_module):
-        self.__doc__, _ = self._pool.apply(
-            _init_container_and_load_tool,
-            [_init_container, tool_module, self.temp_dir],
-            self.container_options,
-        )
+                os.rmdir(temp_dir)
 
     def close(self):
         self._forward_call("close", [], {})
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -299,8 +299,8 @@ def _call_tool_func(name, args, kwargs):
 
 
 class ContainerizedTool(ContainerizedToolBase):
-    def __init__(self, tool_module, config, initializer):
-        super().__init__(tool_module, config, initializer, initializer=_init_container)
+    def __init__(self, tool_module, config):
+        super().__init__(tool_module, config, initializer=_init_container)
 
     def _cleanup(self):
         pass
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -62,7 +62,7 @@ class ContainerizedToolBase(object, metaclass=ABCMeta):
         try:
             self.__doc__, self.container_id = self._pool.apply(
                 _init_container_and_load_tool,
-                [initializer] + self.mk_args(tool_module, config, temp_dir),
+                [initializer] + self.mk_args(tool_module, temp_dir),
                 self.mk_kwargs(container_options),
             )
         except BaseException as e:
@@ -74,8 +74,8 @@ class ContainerizedToolBase(object, metaclass=ABCMeta):
             with contextlib.suppress(OSError):
                 os.rmdir(temp_dir)
 
-    def mk_args(self, tool_module, config, tmp_dir):
-        return [tool_module, config, tmp_dir]
+    def mk_args(self, tool_module, tmp_dir):
+        return [tool_module, tmp_dir]
 
     def mk_kwargs(self, container_options):
         return container_options
--- a/benchexec/containerized_tool.py
+++ b/benchexec/containerized_tool.py
@@ -16,7 +16,6 @@ import os
 import signal
 import socket
 import tempfile
-from abc import ABCMeta
 
 from benchexec import (
     BenchExecException,
@@ -31,7 +30,7 @@ tool: tooladapter.CURRENT_BASETOOL = None
 
 
 @tooladapter.CURRENT_BASETOOL.register  # mark as instance of CURRENT_BASETOOL
-class ContainerizedToolBase(object, metaclass=ABCMeta):
+class ContainerizedTool(object):
     """Wrapper for an instance of any subclass of one of the base-tool classes in
     benchexec.tools.template.
     The module and the subclass instance will be loaded in a subprocess that has been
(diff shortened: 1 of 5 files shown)


dbeyer:
@ricffb @PhilippWendler I fully support this and confirm that this feature is required for competition execution.  This solution looks elegant.

PhilippWendler:
`_version_from_tool` is not public API of tool-info module (as indicated by the dash prefix). It is a utility method that is used by some tool-info modules, but not necessarily by all of them. And of course the tool-info module might overwrite `_version_from_tool` with  a method that does something else (such as accepting more parameters). In summary, this currently only works for some tool-info modules and depends on their internal implementation details. Only the `version` method must be used from the outside.

I am also surprised that this currently works at all (and do not understand why). The `_version_from_tool` method should never actually be called on the `benchmark.tool` object, because the latter is a wrapper (`ContainerizedTool`) around the actual tool-info module. The call to `version` happens on the wrapper and is then forward to the actual tool-info module, and all internal calls should happen strictly inside its class and not involve the wrapper at all. What is the call chain here if you say it succeeds?

If a container image is given as parameter, I think we should never execute the tool outside of this image. So there should be no fallback. And I would actually say that the whole tool-info module should live inside the Podman container (just like it currently lives inside a BenchExec container). The tool-info module's methods are free to do what they want, and for example can call `version` from inside other methods (several tool-info modules actually require that). And this would fail because Podman cannot be executed from within the containerized tool-info module. Furthermore, the tool-info module can even execute the tool from other methods, such as for example from within `executable` or `cmdline`.

ricffb:
@PhilippWendler You are right. Both setups has a `--no-container` lingering around in the command line.
Removing it will not work (gives an error from multiprocessin).

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,297 @@
...
+        .strip()
+    )
+
+    try:
+        logging.debug("Joining user namespace of container %s", container_id)
+
+        # The user namespace must be joined first
+        user_ns = f"/proc/{container_pid}/ns/user"
+        with open(user_ns, "rb") as f:
+            libc.setns(f.fileno(), 0)
```
PhilippWendler, line 228: Put this into a little helper function?

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,297 @@
...
+
+        for namespace in os.listdir(f"/proc/{container_pid}/ns"):
```
PhilippWendler, line 230: If you use `scandir`, you can avoid the need for manual `join` AFAIS.

PhilippWendler (approved):
Thanks, the new solution is much better than the previous one. I think it now works in all cases if `--tool-directory` is passed and the Python installations inside and outside of the container do not clash.

I have some recommendations on how the code and the podman container can be improved further, but I do not see any crucial problems. So if it is required it could be merged and used, and improvements be added later.

PhilippWendler:
> I don't know, if we want to make this way of obtaining a Version optional via yet another flag (or if the `--no-container` should serve double duty here).

Passing `--no-container` and `--vcloudContainerImage` should probably produce an error, because it is just confusing. And if `--vcloudContainerImage` is passed, we always want to use it for the tool-info module.

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,163 @@
...
+
+            except OSError as e:
+                logging.debug(
+                    "Failed to join namespace %s: %s", namespace, os.strerror(e.errno)
+                )
+
+        # The mount namespace must be joined so we want
+        # to fail if we cannot join the mount namespace.
+        # mnt must be joined last because after joining it,
+        # we can no longer access /proc/<container_pid>/ns
```
PhilippWendler, line 114: This could actually be solved by opening `/proc/<container_pid>/ns` as a directory fd and making all subsequent accesses relative to that fd. Not sure if it makes things easier, but it is possible.

benchexec/containerized_tool.py
```diff
@@ -57,13 +57,13 @@ def __init__(self, tool_module, config):
...
+        self.container_id = None
         # Call function that loads tool module and returns its doc
         try:
-            self.__doc__ = self._pool.apply(
+            self.__doc__, self.container_id = self._pool.apply(
                 _init_container_and_load_tool,
-                [tool_module, temp_dir],
-                container_options,
+                [initializer] + self.mk_args(tool_module, temp_dir),
+                self.mk_kwargs(container_options),
```
PhilippWendler, line 66: This
- adds code that is specific to the new subclass to the parent class (`container_id`),
- needs two hooks for letting the subclass overwrite the behavior, and
- still has code that is not needed for the new subclass (`temp_dir`, `container_options`).

It looks like this is not the best way to distribute the code over the inheritance hierarchy.

I would recommend that the whole `apply` call plus setup and cleanup is moved into a method that can be overwritten. `_init_container_and_load_tool` can be extended to pass through the result value, but it does not need to know anymore that it is the `container_id`.

The split into three classes looks similarly not-so-useful: We have the base class, which contains the common code but also specific code for the regular containerized tool-info module. Then we have `ContainerizedTool`, which just adds one little bit of information (what the `initializer` is). Either the base class should have nothing that is specific, or we just have two classes, the class that we had so far plus one new subclass. I would go with the latter choice.

test/tasks/benchmark.xml
```diff
@@ -10,7 +10,7 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 <!DOCTYPE benchmark PUBLIC "+//IDN sosy-lab.org//DTD BenchExec benchmark 1.18//EN" "https://www.sosy-lab.org/benchexec/benchmark-1.18.dtd">
-<benchmark tool="dummy">
+<benchmark tool="locksmith">
```
PhilippWendler, line 13: Please be careful what to commit.

benchexec/containerized_tool.py
ricffb: Thanks for the feedback. I was so focused on the arguments to the pool that I didn't realize, that i could just also call pool.apply from the subclass.

I reverted to the main ContainerizedTool again, this time just replacing the setup call to pool with a method `_setup_container`. The _init_and_load still resturns a tuple, but this is just ignored in the ContainerizedTool class. Only the subclass saves and uses the info.

test/tasks/benchmark.xml
ricffb: Fixed.

benchexec/containerized_tool.py
```diff
@@ -74,8 +74,15 @@ def __init__(self, tool_module, config):
...
 
+    def mk_args(self, tool_module, tmp_dir):
+        return [tool_module, tmp_dir]
+
+    def mk_kwargs(self, container_options):
+        return container_options
+
     def close(self):
         self._forward_call("close", [], {})
+        self._cleanup()
```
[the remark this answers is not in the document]
ricffb: Yes. This is indeed a much better option.

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,297 @@
...
+    logging.debug(
+        "Command to start container: %s",
+        " ".join(map(str, command)),
+    )
+    res = subprocess.run(
+        command,
+        stdout=subprocess.PIPE,
+    )
+
+    res.check_returncode()
```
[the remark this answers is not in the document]
ricffb: I added explicit DEVNULLs and checks to all subprocess.run calls in the podman_containerized_tool

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,297 @@
...
+            "--volume",
+            f"{tool_directory}:{TOOL_DIRECTORY_MOUNT_POINT}:O",
+            # :O creates an overlay mount. The tool can write files in the container
+            # but they are not visible outside the container.
+            "--workdir",
+            "/mnt",
+        ]
+    )
+
+    # Create a container that does nothing but keeps running
```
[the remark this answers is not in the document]
ricffb: Yes. It works super well with your suggestion 👍

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -35,12 +37,95 @@ def set_vcloud_jar_path(p):
...
+
+        # If the tool info does not call find_executable, we don't know if the
+        # executable path is containing the mount point.
+        # In this case we can check whether the path is relative
+        # and continue with the assumption that it is relative to the provided
+        # tool directory.
+        try:
+            executable_relative_to_mount_point = Path(
+                executable_for_version
+            ).relative_to(TOOL_DIRECTORY_MOUNT_POINT)
```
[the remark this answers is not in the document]
ricffb: Here is what I considered:

1. executable_for_version is created with the find_tool helper

Then the path will be `TOOL_DIRECTORY_MOUNT_POINT/subdir/executable` and the
relative method will retrieve the `subdir/executable` part. We then know `tool_dir/subdir/executable` is the path to the executable on the host.

2. find tool helper is not called

a. The ti just returns a relative path, eg., ./sub/executable 
The relative_to will fail with ValueError; since ./sub/executable is not absolute,
it is assumed to be relative to the tool directory and we proceed like in 1.

b. The ti returns an absolute path, that is not calculated in the container (e.g. /opt/my/tool).
The relative_to() call with fail with ValueError. There is no hope of recovering from here.

c. There could be the an edge case, where someone hardcodes "/mnt/__benchexec_tool_dir/something" . If that something is invoking '..' to a point further up the path than "/mnt/__benchexec_tool_dir" will cause a ValueError again.

d. If the tool returns `../../something` this is currently not handled. If the executable indeed exists it will probably be uploaded to the BenchCloud.

If d. is a concern to us, then we could sanitize the path coming from the ti or be a lot stricter in what we acceptas valid return from tool.executable()

benchexec/containerized_tool.py
PhilippWendler: Why not move the setup and cleanup (`handle_basic_container_args`, `mkdtemp`, etc.) into `_setup_container` as well like I recommended?

contrib/vcloud/podman_containerized_tool.py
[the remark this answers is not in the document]
PhilippWendler: The `inspect` call misses `check=True`.

benchexec/containerized_tool.py
ricffb: I misunderstood your recommendation.
There is a `finally:` that handles removal of the tempdir. Is it save to move this finally to the `_setup_container aswell`?

PhilippWendler: If your code does not need the temp directory, you can move its creation into `_setup_container`. And then you can also move the cleanup. And to me it looks as if your code does not need it.

PhilippWendler:
Please do not mark comment threads that were created by others as "resolved". The respective reviewer should close it once they agree that their comments are sufficiently addressed. Any other process has a high risk of missing open problems and creates a high mental load for reviewers for keeping track of what they still need to check.

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,297 @@
...
+    command = (
+        ["podman", "run", "--entrypoint", "tail", "--rm", "-d"]
+        + volumes
+        + [image, "-F", "/dev/null"]
```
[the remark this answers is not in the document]
PhilippWendler: This still needs to be addressed.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -35,12 +37,95 @@ def set_vcloud_jar_path(p):
...
+            )
+
+        msg = (
+            f"Could not find executable '{executable_name}'. "
+            f"The searched directories were: " + "".join("\n  " + d for d in dirs)
+        )
+        if not self.tool_directory:
+            msg += "\nYou can specify the tool's directory with --tool-directory."
+
+        raise ToolNotFoundException(msg)
```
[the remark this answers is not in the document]
PhilippWendler: The question whether this class is necessary at all is still open.

contrib/vcloud/benchmarkclient_executor.py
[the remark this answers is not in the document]
PhilippWendler: > Here is what I considered:
> 
> 1. executable_for_version is created with the find_tool helper

What is the "find_tool helper"?

> Then the path will be `TOOL_DIRECTORY_MOUNT_POINT/subdir/executable` and the relative method will retrieve the `subdir/executable` part. We then know `tool_dir/subdir/executable` is the path to the executable on the host.

What is "the relative method"?

> 2. find tool helper is not called
> 
> a. The ti just returns a relative path, eg., ./sub/executable The relative_to will fail with ValueError; since ./sub/executable is not absolute, it is assumed to be relative to the tool directory and we proceed like in 1.
>
> b. The ti returns an absolute path, that is not calculated in the container (e.g. /opt/my/tool). The relative_to() call with fail with ValueError. There is no hope of recovering from here.

So we want to always return relative paths as they are, and for absolute paths call `relative_to`?
Wouldn't the following convey the intent more clearly?
```
executable_for_version = Path(executable_for_version)
# ensure executable_for_version is relative
if executable_for_version.is_absolute():
    try:
        executable_for_version = executable_for_version.relative_to(TOOL_DIRECTORY_MOUNT_POINT)
    except ValueError:
        raise ...
```

> d. If the tool returns `../../something` this is currently not handled. If the executable indeed exists it will probably be uploaded to the BenchCloud.
> 
> If d. is a concern to us, then we could sanitize the path coming from the ti or be a lot stricter in what we acceptas valid return from tool.executable()

If `executable_for_version` points to somewhere outside of the tool directory, this means it comes from the container. I think that causes trouble no matter whether the path exists outside of the container or not, because we could for example upload a file that is not the one we want to benchmark. And it is trivial to detect this case and reject it, so this should be done.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -35,12 +37,85 @@ def set_vcloud_jar_path(p):
...
+                    " and is not containing the expected container to the mount point"
+                    " {TOOL_DIRECTORY_MOUNT_POINT}"
+                ) from None
+            executable_relative_to_mount_point = executable_for_version
+
+        # The vcloud uses the tool location later to determine which files need to be uploaded
+        # So this needs to point to the actual path where the executable is on the host
+        benchmark.executable = (
+            Path(config.tool_directory) / executable_relative_to_mount_point
+        )
```
PhilippWendler, line 112: I just noticed that this leaks a `Path` instance into the general BenchExec data structures, where it will end up being used in locations that might not expect or support such objects. The migration of paths in BenchExec to `Path` instances must be done in a coordinated matter and for example planned with regards to how it affects backwards compatibility for public APIs. Until this is done please make sure that any `Path` instance is only used locally and not leaked to unaffected code.

contrib/vcloud/podman_containerized_tool.py
```diff
@@ -0,0 +1,187 @@
...
+@tooladapter.CURRENT_BASETOOL.register  # mark as instance of CURRENT_BASETOOL
+class PodmanContainerizedTool(ContainerizedTool):
+    tool_directory: str
+    image: str
+    container_id: Optional[str]
+
+    def __init__(self, tool_module, config, image):
+        assert (
+            config.tool_directory
+        ), "Tool directory must be set when using podman for tool info module."
```
PhilippWendler, line 151: This is currently the only place where this is checked and thus causes a crash of `vcloud-benchmark.py`. That a tool directory is given should be checked higher up (where other config options are also handled) and with a proper error message for the user instead of a crash.

ricffb: Hi, I now check for this in the `hook_load_tool_info` function.

contrib/vcloud/podman_containerized_tool.py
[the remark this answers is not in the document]
ricffb: Thanks. I did oversee that.

contrib/vcloud/benchmarkclient_executor.py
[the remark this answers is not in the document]
ricffb: Hi,
I changed the code to the one you suggested. I also added a check for d.

PhilippWendler (commented):
Can be merged as soon as CI is green.

Would still better to get a more strictly isolated container, though, for example without network access (currently tools do not have network access, but this container allows it).

dbeyer:
Many thanks @ricffb and @PhilippWendler ! I merged it now because I needed it already now. 
I would like to ask you to make further improvements in a new MR.

PR #1133, opened 2024-12-04
Add tool-info modules for MoXI model checkers

Po-Chun-Chien:


benchexec/tools/moxi-mc-flow.py
```diff
@@ -0,0 +1,52 @@
...
+        return tool_locator.find_executable("modelcheck.py")
+
+    def name(self):
+        return "MoXI-MC-Flow"
+
+    def project_url(self):
+        return "https://github.com/ModelChecker/moxi-mc-flow"
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(executable, self.REQUIRED_PATHS)
```
PhilippWendler, line 36: Is it deliberate that this overwrites our default implementation and only changes it to skip adding the `executable`?

benchexec/tools/moxichecker.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def project_url(self):
+        return "https://gitlab.com/sosy-lab/software/moxichecker"
+
+    def version(self, executable):
+        return self._version_from_tool(executable)
+
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
```
PhilippWendler, line 33: This also misses the executable and `bin`, is this intended?

benchexec/tools/moxi-mc-flow.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        if rlimits.cputime and "--timeout" not in options:
+            options += ["--timeout", str(ceil(rlimits.cputime * 1.05))]
```
PhilippWendler, line 40: Hm, what is the goal here? Passing a higher time limit than the one of BenchExec seems weird. Also, hard-coding a magic number like 1.05 in the tool-info module is not recommend. Would this number be appropriate in all situations?

Po-Chun-Chien: (I posted a comment on this line yesterday but just noticed that it is still "pending")

I added 5% more CPU time here to prevent MoXI-MC-Flow from terminating its backend verification engines prematurely (for example, [these cases](https://www.cip.ifi.lmu.de/~chien/benchexec-results/moxichecker/incr-solving-900s/results.2024-07-15_09-38-17.table.html#/table?hidden=0,1,2,3,5&filter=4(0*status*(status(in(ERROR%20%281%29)),category(notIn())),1*cputime*(value(897%3A)))&sort=4_cputime_1,asc)).

The tool uses `timeout` to limit the time on the verification engines internally, which is not as accurate as using cgroups.
(And IIRC, `timeout` imposes limit on walltime instead of CPU time (?).)

What would be your suggested solution?

- Pass additional time limit to the tool (the current way) to ensure that it runs into "real" timeout
- Check whether the log contains `timeout` in `determine_result` although this would mean that the result and measured CPU time does not correlate well (e.g., `TIMEOUT` vs `897 s`)

PhilippWendler: Do you even need to handle the timeout in the tool? What happens if you do not pass this parameter? BenchExec will be able to terminate the tool and designate the run as timeout anyway.

I would also assume that `timeout` uses wall time, so it doesn't really match.

PhilippWendler: Hm, I see, tricky. So we need to pass a value, but we actually never want the timeout in the tool to trigger, right? Because it would either trigger too early, which makes the tool not be able to use its full time limit, or trigger too later, when the tool is already terminated by BenchExec anyway.

So I don't see a better solution that to pass a high timeout value, but I would increase the factor, for example to 2. And please add a comment explaining this, such that readers do not need to wonder whether this magic number is somehow relevant for the tool execution.

benchexec/tools/moxi-mc-flow.py
Po-Chun-Chien: Fixed in e1fee24

benchexec/tools/moxichecker.py
Po-Chun-Chien: You're right, the executable (`bin/moxichecker`) is also needed.
Fixed in e1fee24.

benchexec/tools/moxi-mc-flow.py
Po-Chun-Chien: I've increased the factor to 2 and added some explanations in bac42d5.

PR #1134, opened 2024-12-06
Improve avr tool info

Po-Chun-Chien:
When preparing #1133 , I noticed that AVR has a similar issue that it sometimes terminates itself too early (for example, [these cases](https://www.cip.ifi.lmu.de/~chien/benchexec-results/func-encoding/svcomp24/avr.table.html#/table?hidden1=2,3,4&hidden=0,2,3,4,5,6&sort=1_cputime_1,desc&filter=1(0*status*(status(in(ERROR)),category(notIn())),1*cputime*(value(899%3A))))).
Therefore, I applied the same solution as discussed.

This PR also brings some small improvements to the tool-info modules I wrote.

### DIFF
--- a/benchexec/tools/abc.py
+++ b/benchexec/tools/abc.py
@@ -51,8 +51,6 @@ class Tool(benchexec.tools.template.BaseTool2):
         """
         @return: status of ABC after executing a run
         """
-        if run.was_timeout:
-            return result.RESULT_TIMEOUT
         for line in run.output[::-1]:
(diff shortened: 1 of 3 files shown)


PhilippWendler:
Of course this changes the behavior of the tool-info module in a way that makes benchmarks not reproducible anymore across old and new BenchExec versions. But given that this is a fix for clearly unwanted behavior, we should still do this.

PR #1138, opened 2024-12-18
update spf to BaseTool2

ricffb:
Migrate SPF to BaseTool2 so that it won't complain about the --tool-directory parameter necessary for podman containerized execution.

### DIFF
--- a/benchexec/tools/spf.py
+++ b/benchexec/tools/spf.py
@@ -38,9 +38,9 @@ class Tool(benchexec.tools.template.BaseTool2):
         first_line = output.splitlines()[0]
         return first_line.strip()
 
-    def cmdline(self, executable, options, tasks, rlimits):
-        options = options + ["--propertyfile", tasks.property_file]
-        return [executable] + options + list(tasks.input_files)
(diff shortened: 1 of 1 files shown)


benchexec/tools/spf.py
```diff
@@ -39,11 +38,12 @@ def version(self, executable):
         first_line = output.splitlines()[0]
         return first_line.strip()
 
-    def cmdline(self, executable, options, tasks, propertyfile, rlimits):
-        options = options + ["--propertyfile", propertyfile]
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, tasks, rlimits):
+        options = options + ["--propertyfile", tasks.property_file]
+        return [executable] + options + list(tasks.input_files)
```
ricffb, line 43: I tried to see if the tool would support identifiers. I don't think it does.

PhilippWendler, line 43:
```suggestion
    def cmdline(self, executable, options, task, rlimits):
        options = options + ["--propertyfile", task.property_file]
        return [executable] + options + list(task.input_files)
```

PhilippWendler (approved):

PR #1143, opened 2025-01-01
Error counts in generated table summary tab #974

Mamatha1718:

PR #1147, opened 2025-01-29
Replace escape with encodeURIComponent

Mamatha1718:


PhilippWendler:
What is the goal of this PR? Could you please provide a description?

It also seems that the PR contains a lot of unrelated changes together with formatting changes (like #1143 as well). Please clean this up, in the current state it is not reviewable.

PR #1148, opened 2025-01-29
`Replace` escape with `escapeURIComponent`

Mamatha1718:
Restored deleted files and made the following changes:
- Replaced 'escape' with 'encodeURIComponent' in `stats.worker.js` `abc.py` , `utils.js`


PhilippWendler:
This PR shows all 700 files in our repository as new files to be added. This needs to be cleaned up.

When filing PRs, please have a look at the changes *before* creating the PR and review that everything looks as it should be.

PR #1149, opened 2025-01-31
Replace `escape` with `encodeURIComponent`

Mamatha1718:
Replace `escape` with `encodeURIComponent` in utils.test.js , utils.js , abc.py files.


PhilippWendler:
Does `encodeURIComponent` do exactly the same as `escape`? Or are there differences that we need to address?

How did you create the changes in this PR? I am surprised to see Python changes.

PR #1156, opened 2025-03-21
Make the mount bind target directories before executing fuse-overlayfs

t0hsumi:
Contributes to solving the following issue: https://github.com/sosy-lab/benchexec/issues/1136
Based on the discussion: https://github.com/sosy-lab/benchexec/pull/1155
`get_bind_mount_points()` is based on https://unix.stackexchange.com/questions/18048/list-only-bind-mounts.
However, I'm unsure whether the assumption holds that the oldest mount point with the shortest relative path to the root of the mounted file system is the original mount.


benchexec/util.py
```diff
@@ -123,6 +123,20 @@ def decode_to_string(toDecode):
         return toDecode
 
 
+def decode_path(path):
```
PhilippWendler, line 126: I would prefer to keep `decode_path` in `container.py` (likely as `_decode_path`). It is good to keep utilities close to where they are used, if they are not globally useful. `util.py` is already quite some kitchensink of various stuff.

benchexec/container.py
```diff
@@ -754,24 +754,56 @@ def get_mount_points():
...
+            )
+
+    for device_id, mounts in device_id_to_mounts.items():
+        # Skip single mounts
+        if len(mounts) <= 1:
+            continue
+        # Sort list to get the first mount of the device's root dir (if still mounted)
+        mounts.sort(key=lambda x: len(x[1]))
+        # Assume that the mount point with the shortest relative path to the root of
+        # the mounted file system is the original mount.
```
PhilippWendler, line 794: Hm, it is somewhat unfortunate that we have to rely on such heuristics and do not get definitive information. Maybe we can make it a little bit more strict to be on the safe side?

What would happen if one bind mounts the root of a file system to some other place? Like, if I bind-mount `/run` to some other place? Or if I unmount the original mount points? Can we expect `root_source` to be always `/` for example? If we find a situation where we don't know what to do, it might be safer to do nothing.

At least the above is what I would say for a general-purpose `get_bind_mount_points` that returns source and target. But we actually do not need that, right? We only need to find those mount points that are bind mounts, not the information about the source. So if we make the function less powerful and generic, then it becomes easier to implement it in a safe way.

Or, maybe a wild thought: What happens if we just pre-create directories for *all* mount points below `temp_base`? Would this work?

PhilippWendler (commented):
Thanks a lot!

Overall the code looks good, I liked it!

While reviewing I realized that maybe we can simplify this considerably, so I only did a quick test and not tried out all cases extensively. Could you think about this (creating directories for all mount points) and try it out?

benchexec/container.py
t0hsumi: > Hm, it is somewhat unfortunate that we have to rely on such heuristics and do not get definitive information. Maybe we can make it a little bit more strict to be on the safe side?
What would happen if one bind mounts the root of a file system to some other place? Like, if I bind-mount /run to some other place? Or if I unmount the original mount points? Can we expect root_source to be always / for example? If we find a situation where we don't know what to do, it might be safer to do nothing.

Indeed, in some cases, this assumption does not hold. I agree.

The complexity arises from the fact that the kernel does not treat bind mounts differently from normal mounts. So, I gave up on listing all the bind mount points and made a wild choice: pre-create all the mount points.

Since creating empty directories in upperdir that already exist in lowerdir does no harm—I just checked—it seems like a simple and effective solution.

t0hsumi:
Thanks for the detailed review. I just replied the comment and add the commit!

benchexec/container.py
PhilippWendler: Great! And thanks!

I am currently on vacation, will have a closer look afterwards, but so far this looks ready.

benchexec/container.py
```diff
@@ -747,31 +747,73 @@ def determine_directory_mode(dir_modes, path, fstype=None):
...
+            device_id, mountpoint = fields[2], fields[4]
+            device_id_to_mounts.setdefault(device_id, []).append(mountpoint)
+
+    for device_id, mounts in device_id_to_mounts.items():
+        # Skip single mounts
+        if len(mounts) <= 1:
+            continue
+        # Make `upperdir/target` directory
+        for target in mounts:
+            if not os.path.isdir(target):
```
PhilippWendler, line 813: Probably around here we should call `_decode_path, shouldn't we? I assume that `mountinfo` uses the same escaping as `mounts`.

PhilippWendler (approved):
The current solution does nothing more than creating a few directories in the upper dir of the overlayfs, which AFAIK has no side effects for us and should be completely safe. This is nice and we do not need to worry if it breaks something as long as the code does not crash during execution. But I don't think this can happen, even for inaccessible mount points (which are typically a problem) because `isdir` just returns `False` for them.

And in my tests it works fine as a solution for #1136.

Also, the code is nicely readable and well explained, thanks!

So this looks ready to merge, only one question about path decoding remains.

Thank you a lot! Your thorough work on this problem and all the iterations and tests have paid off!

benchexec/container.py
t0hsumi: Yes, we should. Thanks for telling me! I just updated.

PhilippWendler: Now there is `_decode_path` for `isdir`, but `makedirs` is still called on the raw path.

Maybe it is even better to call `_decode_path` right after reading the path from `mountinfo`?

t0hsumi: Oh yes, it's better to use the _decode_path() function. It's also more consistent with other os.makedirs() calls, such as [this one](https://github.com/sosy-lab/benchexec/blob/dc9f7e03653eef83b7fd838ec5725b69ef131b60/benchexec/container.py#L546), where mountpoint is already decoded.

PhilippWendler (approved):
Great, thanks a lot again!

PR #1157, opened 2025-03-25
Getting version from Pono

Po-Chun-Chien:
Since `./pono --version` has non-zero exit code,
a new option `expected_exitcode` is introduced in `_version_from_tool()`.

If such a change is not desirable,
I could write a custom function for Pono,
instead of reusing the existing one.

### DIFF
--- a/benchexec/tools/template.py
+++ b/benchexec/tools/template.py
@@ -140,6 +140,7 @@ class BaseTool2(object, metaclass=ABCMeta):
         use_stderr=False,
         ignore_stderr=False,
         line_prefix=None,
+        *,
         expected_exitcode=0,
     ):
         """


benchexec/tools/template.py
```diff
@@ -140,6 +140,7 @@ def _version_from_tool(
         use_stderr=False,
         ignore_stderr=False,
         line_prefix=None,
+        expected_exitcode=0,
```
PhilippWendler, line 143: In principle, I agree with this addition. I would, however, suggest that we make it a [keyword-only argument](https://peps.python.org/pep-3102/). We would never want someone to pass a single int without explicitly naming its purpose here. (The same would be true for the other parameters, but we cannot change it for them.)

Po-Chun-Chien: Done in 826ef032c3b6daf79df4f9800b8cd60392389051

PhilippWendler (approved):

PR #1158, opened 2025-04-17
Make `<option>` `name` attribute optional in benchmark.dtd

sim642:
These are supported and used by:
- doc/benchmark-example-rand.xml,
- doc/benchmark-example-true.xml,
- SV-COMP bench-defs repository.


PhilippWendler (approved):
Thanks a lot for noticing! It is indeed intended that either `<option name="--foo"/>` or `<option>--foo</option>` can be used and have the same effect.

PR #1160, opened 2025-04-24
Added Support for Walltime limit in Vcloud

MaxBarth95:
Backwards compatible support for walltime.
The optional parameter walltime has to be the 5th parameter.  
If no corelimit is given we write "-" instead.

Older versions will simply not support walltime since they do only provide 4 parameters at most.


Po-Chun-Chien:
The changes LGTM.

@Nipo773 Since you reviewed the corresponding BenchCloud MR, do you want to double-check?

@MaxBarth95 I assume you've already tested this on the currently deployed BenchCloud version, right?
And do you know if people from SAT- or SMT-COMP have already used this new feature?

PR #1162, opened 2025-05-10
add testcoca

cvldrt:
Add TestCoCa verification tool for Test-Comp

https://gitlab.com/sosy-lab/test-comp/bench-defs/-/merge_requests/150
https://gitlab.com/sosy-lab/benchmarking/fm-tools/-/merge_requests/663
https://github.com/sosy-lab/benchexec/pull/1162
https://zenodo.org/records/15380985

### DIFF
--- a/benchexec/tools/testcoca.py
+++ b/benchexec/tools/testcoca.py
@@ -25,7 +25,7 @@ class Tool(benchexec.tools.template.BaseTool2):
         return "My Toolname"
         @return a non-empty string
         """
-        return "TestCoca"
+        return "TestCoCa"
 
     def project_url(self):
         return "https://github.com/staticafi/TestCoCa"
--- a/benchexec/tools/testcoca.py
+++ b/benchexec/tools/testcoca.py
@@ -74,11 +74,10 @@ class Tool(benchexec.tools.template.BaseTool2):
 
         if task.property_file:
             options += ["--goal", task.property_file]
-        
+
         return (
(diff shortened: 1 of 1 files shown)


benchexec/tools/testcoca.py
```diff
@@ -0,0 +1,125 @@
...
+    REQUIRED_PATHS = ["lib", "lib32", "tools"]
+
+    def name(self):
+        """
+        Return the name of the tool, formatted for humans.
+        This method always needs to be overriden, and typically just contains
+        return "My Toolname"
+        @return a non-empty string
+        """
+        return "TestCoca"
```
PhilippWendler, line 28: Do you want `TestCoca` or `TestCoCa` here?

benchexec/tools/testcoca.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def project_url(self):
+        return "https://github.com/staticafi/TestCoCa"
```
PhilippWendler, line 31: I noticed that this repository is empty - is this intended?

Is the tool's command-line interface already expected to be stable, or will it still change in potentially breaking ways? It doesn't really make sense to merge a tool-info module if nothing is stable yet.

cvldrt: Thanks for the feedback. The implementation was previously on a different branch, but that has now been resolved. The tool is stable at this point, and I don’t anticipate any further changes to the CLI.

Issue #1164, opened 2025-05-26
Solution from error message is not persistent


leventeBajczi:
This solution suggestion:

https://github.com/sosy-lab/benchexec/blob/5c07c9d501955dfea37f68693af1a50ba8f84341/benchexec/container.py#L125-L130

is not persistent, which might not be obvious to users, and therefore, can lead to repeated problems.
Instead, this should suggest `echo 'kernel.apparmor_restrict_unprivileged_userns = 0' | sudo tee /etc/sysctl.d/99-local.conf` or something similar, _or_, warn that this is a temporary solution only.

PhilippWendler:
I am not really in favor of adding a command like the one you suggested because it sometimes depends on the distribution of what exactly to do and it is hard to provide a suggestion in a safe way. For example, the suggested command inevitably overrides `/etc/sysctl.d/99-local.conf` and might delete other deliberate settings the user has. The command is also *only* a persistent solution that becomes active on the next boot, not immediately, so we would have to explain even more and tell the user to do both. Furthermore, here the recommended solution would actually be to install BenchExec from the `.deb` package. But our [installation documentation](https://github.com/sosy-lab/benchexec/blob/main/doc/INSTALL.md#kernel-requirements) covers this and also mentions `sysctl.conf` for manual persistency. So I think this is fine.

But of course you are correct that we should improve the error message and for example point to the installation instructions. Would you like to make a suggestion?

PR #1165, opened 2025-05-26
Improve user ns error message

leventeBajczi:
As discussed in #1164, the error message for forbidden unprivileged user namespaces provides a solution to the problem, but does not say that it is only a temporary one (until the next reboot).


PhilippWendler (approved):
Thanks a lot!

Issue #535, opened 2019-12-08
Prevent truncation of output file by tool


PhilippWendler:
When a tool truncates the file descriptors that it has for `stdout` and `stderr`, this leads to truncation of the log file with the output. So a tool can undo output that it has previously written, and it can remove and overwrite the first few lines of the log that are written by BenchExec (with the command line).

While we could argue that making a tool to remove its own output would be acceptable, it should not be able to overwrite the lines that BenchExec produces.

This can even happen unintentionally if `/dev/stdout` or `/dev/stderr` is used for output redirection with `>` instead of `>>`.

Example:
```
runexec -- bash -c 'echo "echo_stdout1" ; echo "echo_stderr1" 1>/dev/stderr ; echo "echo_stdout2"'
```
The following happens:
1. BenchExec writes the command line.
2. `echo_stdout1` is appended.
3. The file content is deleted completely, and `echo_stderr1` is written to the beginning of the file.
4. `echo_stdout2` is appended to the file, so it appears after the position where the output from 2. ended. The hole in the file between the output from 3. and 4. is filled with null bytes.

We could
- Use a pipe as target for the output. This would solve this completely, but with overhead during the run.
- Add the command line to the log after the run (which involves copying the whole file). This has no overhead during the run but does not solve the tool overwriting its own output and the null bytes.
- Investigate whether there is some option that prevents a file descriptor from being truncated (maybe `chattr +a` helps).

PhilippWendler:
`chattr` is not possible as a regular user.

I hope someone will answer this [question](https://unix.stackexchange.com/q/556348/15398).

PR #1170, opened 2025-06-20
[WIP] Add output selector

t0hsumi:
This branch is intended to address https://github.com/sosy-lab/benchexec/issues/535 using `selectors` module.

### DIFF
--- a/benchexec/containerexecutor.py
+++ b/benchexec/containerexecutor.py
@@ -773,13 +773,16 @@ class ContainerExecutor(baseexecutor.BaseExecutor):
                     grandchild_proc = subprocess.Popen(
                         args,
                         stdin=stdin,
-                        stdout=subprocess.PIPE,
-                        stderr=subprocess.PIPE,
+                        stdout=None if stdout is None else subprocess.PIPE,
+                        stderr=None if stderr is None else subprocess.PIPE,
                         env=env,
                         close_fds=False,
                         preexec_fn=grandchild,
                     )
-                    self._stream_output_with_selector(grandchild_proc, stdout, stderr)
+                    if stdout is not None or stderr is not None:
+                        self._stream_output_with_selector(
+                            grandchild_proc, stdout, stderr
+                        )
                 except (OSError, RuntimeError) as e:
                     logging.critical("Cannot start process: %s", e)
                     return CHILD_OSERROR
--- a/benchexec/containerexecutor.py
+++ b/benchexec/containerexecutor.py
@@ -773,12 +773,13 @@ class ContainerExecutor(baseexecutor.BaseExecutor):
                     grandchild_proc = subprocess.Popen(
                         args,
                         stdin=stdin,
-                        stdout=stdout,
-                        stderr=stderr,
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE,
                         env=env,
                         close_fds=False,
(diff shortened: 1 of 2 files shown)


benchexec/containerexecutor.py
```diff
@@ -773,12 +773,16 @@ def child():
...
+                        stdout=None if stdout is None else subprocess.PIPE,
+                        stderr=None if stderr is None else subprocess.PIPE,
                         env=env,
                         close_fds=False,
                         preexec_fn=grandchild,
                     )
+                    if stdout is not None or stderr is not None:
+                        self._stream_output_with_selector(
+                            grandchild_proc, stdout, stderr
+                        )
```
t0hsumi, line 785: If we call self._stream_output_with_selector() here and the tool does not send an EOF but is instead terminated by a signal, the procedure gets blocked in this function (more precisely, in selector.select() within self._stream_output_with_selector()), and runexec does not terminate. (See [this pytest error case](https://github.com/sosy-lab/benchexec/blob/b07c314d74ed9a3471917d73600d8a3e61a4ca63/benchexec/test_runexecutor.py#L836).)

To handle this properly, we need to process stdout and stderr while also checking for walltime limits and signals. In the implementation, we need to integrate self._stream_output_with_selector() with [wait_for_child_and_forward_signals()](https://github.com/sosy-lab/benchexec/blob/b07c314d74ed9a3471917d73600d8a3e61a4ca63/benchexec/container.py#L1194). However, as far as I can tell, there are two possible solutions—neither of which seems ideal:

1. Set a timeout in selector.select()
2. Use an additional thread

t0hsumi:
The methods of selectors module may raise the exceptions in some edge cases (e.g., ValueError on an invalid event mask or file descriptor during `selectors.register()`). In our current usage, these cases should not occur, so I’m unsure whether we need to add a try-catch block for safety.

benchexec/containerexecutor.py
PhilippWendler: Can you make select wait for termination of a process? I think that might be possible and should solve the problem.

PhilippWendler: I think the documentation talks about the case if the select()ing process receives a signal. But this is not our case.

I would assume that the pipe is closed if the only listening process terminates, and that this causes select() to return. Can you check if there are any other processes that still have the same pipe open, for example BenchExec itself? BenchExec should only have the read ends of the pipe open, and the benchmarked process only the write ends.

Another way could be to add a file descriptor that signals if the process ends. Maybe we can do this with a pidfd. Or the part of the code that calls wait could close the pipes, such that select() returns?

t0hsumi: Sorry for the late reply.
> I think the documentation talks about the case if the select()ing process receives a signal. But this is not our case.

Thanks for pointing that out — I misunderstood. I've now updated the code to retry select() when it returns an empty list.

> Can you check if there are any other processes that still have the same pipe open, for example BenchExec itself?

Which pipe do you mean? I assume you're referring to `proc.stdout` and `proc.stderr`, but I couldn't find any indication that BenchExec is reading from those.

> Another way could be to add a file descriptor that signals if the process ends. Maybe we can do this with a pidfd. Or the part of the code that calls wait could close the pipes, such that select() returns?

I’m now using `pidfd_open()` together with the `select` module. It works fine in my environment.

PR #1171, opened 2025-06-25
Add tool-info module for LTSmin

dbeyer:


ricffb:
@PhilippWendler 

A short explanation to this weird looking tim:
In ltsmin all different "configs" are separate binaries. The corresponding yaml file in fm-tool selects the desired one by setting the first argument in the benchexec_options to the desired binary. E.g. i want to convert petri nets, then I give as first argument `pnml2lts-sym` so that the petrinet module is selected. This can very conveniently be done with fm-weck.

benchexec/tools/ltsmin.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Tool(BaseTool2):
+    """
+    Wrapper for LTSmin.
+    """
+
+    REQUIRED_PATHS = ["bin/", "include/", "share/"]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("ltsmin-reduce", subdir="bin")
```
PhilippWendler, line 20: Is `ltsmin-reduce` a meaningful default? I.e., if users specify only the tool-info module, will they expect `ltsmin-reduce` as default?

benchexec/tools/ltsmin.py
```diff
@@ -0,0 +1,43 @@
...
+    def project_url(self):
+        return "https://ltsmin.utwente.nl/"
+
+    def version(self, executable):
+        return self._version_from_tool(executable, line_prefix="v")
+
+    def cmdline(self, executable, options, task: BaseTool2.Task, rlimits):
+        from pathlib import Path
+
+        real_executable = options.pop(0) if len(options) > 0 else executable
```
PhilippWendler, line 34: Requiring this as first option could be inconvenient if users have a larger benchmark definition and for example want to specify the tool inside some `<rundefinition>`. It is also somewhat hard to discover that a user has to specify the tool name as first name, and if they don't, BenchExec will try to execute an unexpected executable name like `--foo` (if that happens to be the first option). Should this logic be made for flexible?

benchexec/tools/ltsmin.py
```diff
@@ -0,0 +1,43 @@
...
+        real_executable = str(Path(executable).parent / real_executable)
+
+        return [real_executable] + options + [task.single_input_file]
+
+    def determine_result(self, run):
+        if run.exit_code.signal != 0:
+            return result.RESULT_ERROR
+
+        return result.RESULT_DONE
```
PhilippWendler, line 43: This is actually redundant, as it is the default behavior. BenchExec will overwrite `DONE` with an error string.

PhilippWendler:
@ricffb Is this something that you need in the next BenchExec version (tomorrow)?

PhilippWendler:
@ricffb Any intent on following up with this? Following from #1266 it might be better to split up the tool-info module for each executable?

PhilippWendler:
No response after several months, closing.

PR #1179, opened 2025-09-30
Optimize run definition checks for -r flag

diiegoNacho:
BenchExec is currently executing several checks for every runDefinition indicated in the `.xml` file, no matter if the runDefinition was included `-r` flag in the benchmark command.

An example: Let's say that in the run definition file there are only two files included in the tasks, and there are three runDefinitions. Then BenchExec will execute the mentioned check (among others) for all three runDefinition, for both files -> 6 checks.
In general, in the case of a run definition file, with `x` run definitions, and `y` files included in the task definition, BenchExec makes the checks `x * y` times, which for big `x` and `y` causes a considerable and unnecessary overhead in the pre-run stage.

This pull request adds a condition that restricts these checks to the runs indicated in the `-r` flag of the execution command.

Closes https://gitlab.com/sosy-lab/software/benchcloud/-/issues/706

### DIFF
--- a/benchexec/model.py
+++ b/benchexec/model.py
@@ -597,82 +597,86 @@ class RunSet(object):
         blocks = []
 
         for index, sourcefilesTag in enumerate(sourcefilesTagList):
-            sourcefileSetName = sourcefilesTag.get("name")
-            matchName = sourcefileSetName or str(index)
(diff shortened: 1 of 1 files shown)


baierd:
@PhilippWendler you can add me as assignee additionally to Diego.

PhilippWendler:
Replaced by #1182.

PR #1186, opened 2025-10-16
Add JLiSA tool integration in benchexec

giacomozanatta:


benchexec/tools/jlisa.py
```diff
@@ -0,0 +1,60 @@
...
+        return self._version_from_tool(executable, arg="version")
+
+    def name(self):
+        return "JLiSA"
+
+    def project_url(self):
+        return "https://github.com/lisa-analyzer/jlisa"
+
+    def cmdline(self, executable, options, task, rlimits):
+        cmd = [executable, "check", "--inputs", " ".join(task.input_files_or_empty)]
```
PhilippWendler: Does your tool support tasks with more than one input file or without any input files? If not, please use `task.input_files` or `task.single_input_file` instead because these will produce a proper error message for the user if an unsupported task is used.

PR #1188, opened 2025-10-19
add muval

Iwancof:
Add MuVal verification.

### DIFF
--- a/benchexec/tools/muval.py
+++ b/benchexec/tools/muval.py
@@ -82,4 +82,4 @@ class Tool(benchexec.tools.template.BaseTool2):
         return result.RESULT_ERROR
 
     def project_url(self):
-        return "https://bitbucket.org/uhiro/fptprove"
+        return "https://github.com/hiroshi-unno/coar"


benchexec/tools/muval.py
```diff
@@ -0,0 +1,89 @@
+"""
```
PhilippWendler: Please add a license and copyright header using the [REUSE format](https://reuse.software/), like we have in all files in the repository.

benchexec/tools/muval.py
```diff
@@ -0,0 +1,89 @@
...
+BenchExec tool-info module for MuVal.
+
+MuVal is a verification tool based on modal μ-calculus for C programs.
+"""
+
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
```
PhilippWendler: Please add a function called `project_url()` that returns the URL of your project.

PhilippWendler: The BitBucket URL that is now linked is broken (error 404).

PR #1190, opened 2025-10-20
Update the option to print the version for MLB

xiaogvo:


PhilippWendler:
This looks as if it would not work anymore for older versions of the tool, is this correct? Then it would not be possible to benchmark these older versions with BenchExec. Thus, can you please ensure backwards compatibility in the tool-info module?

xiaogvo:
Yes, the new commands don't work for the older versions. But I don't understand what I'm supposed to do to "ensure backwards compatibility", could you please provide more details?

PhilippWendler:
The goal is that one can take a current BenchExec version and use it to benchmark older versions of MLB (as well as newer versions, of course). This is important not only for reproducibility studies, but also simply in any case where developers or other researchers want or need to experiment with the old version.

What needs to be done depends on your tool, particularly on how the versions differ. If the only difference is the parameter for the version number, you can maybe simply try both? Of course, if your tool offers backwards compatibility and still supports the old argument, then the easiest way is to keep the old code.

PR #1193, opened 2025-10-21
Update of the script regarding Fizzer's registration to TestComp2026

trtikm:


trtikm:
> This looks as if it would not work anymore for older versions of the tool, is this correct?

This question does not make sense. This is the first version of Fizzer I am submitting to Test-Comp 2026.

I am just applying the point 3 of the tool submission procedure presented here:

https://test-comp.sosy-lab.org/2026/submission.php

Observe, there is no mention about backward compatibility or tool version handling on the page whatsoever. So, I kindly ask you to not extend official requirements published on the web page.

> Then it would not be possible to benchmark these older versions with BenchExec.

I do not think that is my problem. I am only concerned in Test-Comp 2026 and related official submission procedure. Your statement is clearly irrelevant to the procedure.

> Thus, can you please ensure backwards compatibility in the tool-info module?

Unfortunately, I cannot. Fizzer itself does not do versioning. And again, the procedure does not require to provide different versions of the tool. Only some version string must be printed; and Fizzer does so (prints 0.1). It is unfortunate that this file is called `fizzer.py` and internally it was called `sbt-fizzer`. So, I cannot even create a new Python script, because the name is already taken (by what you call "old version" of the tool).

PhilippWendler:
BenchExec is a benchmarking framework that is independent of any competition (though it happens to be used by quite a few). And benchmarking is not only done for competitions, but also for research experiments, evaluations in papers, etc. So some people may have used or may want to use BenchExec for benchmarking Fizzer, and we do not want to break their use cases and workflows, even if they want to benchmark a previous version of Fizzer (which can be important for example for a reproducibility study). So we always ask that changes to tool-info modules in BenchExec keep old tool versions supported. 

If a project moves and its URL changes, this can of course always be updated. The URL is just meant to give people who look at result tables a convenient way to access the project webpage.

If the entry point of a tool changes, this can also be accommodated. The `executable()` method can for example search for both old and new entry point and take the one that it finds. Functions like `cmdline` and `determine_result` can also support changes as long as there is some way to distinguish between the old and new version, e.g., by looking at the version number or the entry point, or in case of `determine_result` by looking at the tool output. There are plenty of examples for this in our existing tool-info modules.

Side note: If (substantially) different versions of your tool output the same version, that seems highly confusing and prone to leading to errors and wrong conclusions. For example, suppose someone writes a paper comparing a new approach with state of the art, and uses Fizzer 0.1 as state of the art. Reviewers and readers would not know which version was actually used, or even worse, just assume that it is one particular version of the tool that happens to output version number 0.1 when in fact it was a different one. So, completely independently from this PR, I would recommend to use different version numbers for substantially different versions (or even base the version number on a commit hash or so) in order to avoid such problems.

trtikm:
Although I think the issue is resolved (although the solution is not nice), I'd like to mention @dbeyer  just be informed that I tried to make the MR.

PhilippWendler:
This is BenchExec's issue tracker and discussions about competition rules etc. are off-topic here, please keep it restricted to BenchExec topics.

BenchExec does not force you to change your tool, it is one of its design goals that this should not be necessary. Adopting the tool-info module to a new entry point and tool output would be possible as described above. Of course, if you prefer to rename the entry point of the tool instead of adopting the tool-info module, feel free to do so.

As you said that you do not want this MR to be merged, I will close it.

PR #1195, opened 2025-10-21
Update to new SVF-SVC files.

Lasagnenator:
We are using a new main file for execution.


PhilippWendler:
This looks as if it would not work anymore for older versions of the tool, is this correct? Then it would not be possible to benchmark these older versions with BenchExec. Thus, can you please ensure backwards compatibility in the tool-info module?

Lasagnenator:
Ok, I've made our new code backwards compatible.

PR #1196, opened 2025-10-21
Add iekke to SV-COMP26

paolo-di-biase:


benchexec/tools/iekke.py
```diff
@@ -0,0 +1,79 @@
...
+                return True
+            if int(version1[i]) < int(version2[i]):
+                return False
+        return True
+
+    def cmdline(self, executable, options, task, rlimits):
+        data_model_param = get_data_model_from_task(task, {ILP32: "--32", LP64: "--64"})
+        if not data_model_param:
+            data_model_param = "--32"
+        if self.version_geq(self.ver, "2.2"):
```
PhilippWendler: I didn't find a tool named iekke when googling for it, so I am wondering: Are there really older versions of the tool (< version 2.2) that did not support this parameter, and you do want to have support for them in this tool-info module? If yes, this is fine, I am just wanting to make sure that this is intended.

paolo-di-biase: Yeah you right, I'm doing it

PR #1201, opened 2025-10-28
added features for runsolver/starexec compatibility

geoffgeoffgeoff3:
Hi Philip, I'll happily take feedback rather than a merge. I have done the work for only the --no-container case. I run with ...
bin/runexec --no-container --read-only-dir / --no-output-header --timestamp --add-eof --

### DIFF
--- a/benchexec/containerexecutor.py
+++ b/benchexec/containerexecutor.py
@@ -21,7 +21,6 @@ import subprocess
 import sys
 import tempfile
 import traceback
-import time
 
 from benchexec import __version__
 from benchexec import baseexecutor


geoffgeoffgeoff3:
The --timelimit and --walltimelimit are working. I was confused by the amount of WC time it takes to print a lot of tracing from my test program (recursive Fibonacci).

PhilippWendler:
In general this looks quite promising, nice, and thank you! We can still discuss naming a little bit.

I can also somewhat understand the request for configuring output files, though maybe you can also simply redirect the output?

Maybe we should not add the overhead of piping the tool output through BenchExec for everyone and disable the new code if no timestamp is requested? Should be easy.

> When you are happy with what I have done give me a push, and I'll work on the container version.

Would be great!

I just remembered that we have a start of the GSoC attempt on this here: https://github.com/sosy-lab/benchexec/pull/1170/files It is a more complicated because it keeps stdout and stderr separate and thus needs to work on two streams in parallel. But actually I think your approach should also work, I don't see any disadvantage right now.

geoffgeoffgeoff3:
Yes, last night I also thought about disabling the code for timestamping. I'll do it this afternoon. I like the symmetry of having "-" to pipeline both stdin and stdout/stderr ... it's easy for users to remember when it's the same in both directions.

geoffgeoffgeoff3:
The container version has got me stumped for now. I might need you (Philip) to nudge me in the right direction.

geoffgeoffgeoff3:
I see the code ...

    def _set_termination_reason(self, reason):
        if not self._termination_reason:
            self._termination_reason = reason

If I have a soft limit, typically that means the solver will catch the SIGTERM and continue. The code sets the termination reason with self.callback("cputime-soft"). If the solver hits the hard limit while continuing, and that really stops the solver, then the code tries to set the termination reason with self.callback("cputime"). However, as it was set previously the real new reason is not recorded and the output says "terminationreason=cputime-soft". I suggest taking out "if not self._termination_reason:". Whatcha think?

geoffgeoffgeoff3:
Strangely, the non-container "solver" runs slower than the container version.

PR #1210, opened 2025-11-07
Fizzer: New version

trtikm:
There is however an issue with `REQUIRED_PATHS`. While old version uses `["lib", "lib32", "tools"]`, the new one `["fizzer"]`. I resolved this issue by merging both to `["lib", "lib32", "tools", "fizzer"]` and hoping that if some dir is not present, then it is ignored.

### DIFF
--- a/benchexec/tools/fizzer.py
+++ b/benchexec/tools/fizzer.py
@@ -103,11 +103,10 @@ class Tool(benchexec.tools.template.BaseTool2):
             return result.RESULT_UNKNOWN
 
         mark = "--- TestCompResult ---"
-        txt:str = run.output.text
+        txt = run.output.text
         mark_idx = txt.find(mark)
         if mark_idx != -1:
             return txt[mark_idx + len(mark) :].strip().splitlines()[-1]
-
 
         compilation = None
         instrumentation = None


benchexec/tools/fizzer.py
```diff
@@ -92,6 +100,13 @@ def determine_result(self, run):
         if not run.output:
             return result.RESULT_UNKNOWN
 
+        try:
+            mark = "--- TestCompResult ---"
+            txt = run.output.text
+            return txt[txt.index(mark) + len(mark) :].strip()
```
PhilippWendler: This returns everything after the mark, which could be a multi-line string, for example if for some reason the tool crashes after printing the output. That would probably break the tables. Could it be improved such that only the actually desired result is extracted, for example by restricting it to the remainder of the line or the next line, depending on how the output is structured?

dbeyer:
Regarding `REQUIRED_PATHS`: Isn't it more robust to use the default `['.']`, instead of adjusting the tool-info module every time the tool uses a new directory name?

benchexec/tools/fizzer.py
trtikm: Ok, fixed.

trtikm:
Regarding REQUIRED_PATHS: I'll put there whatever version you want. For now I keep the original version, i.e., `["lib", "lib32", "tools", "fizzer"]`, but if it should be `["."]`, then just tell me.

PR #1226, opened 2025-11-14
Migrate tool-info module for Jayhorn to BaseTool2

ma1eck:


benchexec/tools/jayhorn.py
```diff
@@ -27,15 +33,22 @@ def name(self):
...
-        return [executable] + options + tasks
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file:
+            options = options + ["--propertyfile", task.property_file]
+
+        return [executable] + options + list(task.input_files_or_identifier)
+    
+    def program_files(self, executable):
+        return self._program_files_from_executable(
+            executable, self.REQUIRED_PATHS, parent_dir=True
```
PhilippWendler: Are the files that you listed in `REQUIRED_PATHS` in the same directory as the `jayhorn` executable, or in its parent directory? This method here seems to indicate the latter, but from `REQUIRED_PATHS` it seems that everything is just in the same directory?

PR #1234, opened 2025-11-30
Refactor input handling to include witness options

leventeBajczi:
Theta couldn't handle witnesses passed from the task definition files.
This is to handle the handcrafted witnesses category at SV-COMP.
**my intention is** that the changes here only affect this case, when a .yml task definition defines both the input .c file and the corresponding witness. In every other case, the cmdline remains the same as before this current change.

### DIFF
--- a/benchexec/tools/theta.py
+++ b/benchexec/tools/theta.py
@@ -8,6 +8,10 @@ import benchexec.result as result
 import benchexec.tools.template
 import functools
 import re
+from benchexec.tools.sv_benchmarks_util import (
+    handle_witness_of_task,
+    TaskFilesConsidered,
+)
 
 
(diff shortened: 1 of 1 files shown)


PhilippWendler:
> > The tool-info module should simply remove the extra input if the file extension is `.yml`.
> > Optionally, it could use the input if no parameter `--witness` is present.
> 
> I think this is exactly what `handle_witness_of_task` ends up doing in the SV-COMP setup where the witness parameter is also given by bench-defs XML.

The logic is: If the task definition defines a witness file in its options, then `handle_witness_of_task` either adds it to the command line or ignores it, depending on whether the command line already contains the witness parameter. In both cases the witness file is removed from the input files. In no cases the file extension is relevant.

PR #1238, opened 2025-12-10
goblint: extend `get_value_from_output` for last match

sim642:
This is a clever/dirty way to choose which match is returned when multiple are present: the single used regex capture group can be named `first`/`last`.


PhilippWendler:
Nice idea, thanks for sharing! Maybe this will be helpful for others as well.

PR #1245, opened 2026-01-03
Add PyInstaller-based bundling support for runexec

### LINKED ISSUE #1243: Investigate bundling BenchExec+Python
For situations where a (modern) Python is not available, it would be convenient if one can execute BenchExec (or at least `runexec`) as a tool with as few dependencies as possible. Not sure if this is really feasible, but we should investigate.

Possible variants of this:
- A ZIP with a bundled Python and BenchExec, potentially even with a static build of Python, or one that only depends on an (old) libc.
- BenchExec compiled to native code.

saurabh12nxf:
This PR adds the ability to create a standalone bundled version of `runexec` using PyInstaller, addressing #1243.

## Implementation
- **Approach**: PyInstaller
- **Binary Size**: 17 MB
- **Dependencies**: Only standard system libraries (libc, libdl, libz, libpthread)
- **Build Time**: ~30 seconds

## Features
✅ All core runexec features work:
- Command execution
- Resource limits (--memlimit, --timelimit)
- Cgroups support
- Container mode (on native Linux)
- ctypes syscalls work correctly

## Files Added
- `bundling_investigation/runexec.spec` - PyInstaller configuration
- `bundling_investigation/build_bundle.sh` - Production build script
- `bundling_investigation/test_pyinstaller.sh` - Testing script
- `bundling_investigation/test_nuitka.sh` - Alternative approach (reference)
- `bundling_investigation/README.md` - Investigation results
- `doc/bundling.md` - User documentation

## Testing
Tested on Ubuntu 24.04:
- ✅ Version check
- ✅ Command execution
- ✅ Resource limits
- ✅ Cgroups
- ✅ Container mode (except WSL overlay limitation)

## Usage
```bash
cd bundling_investigation
./build_bundle.sh
./dist/runexec --version
```

## Known Limitations
- Container mode may not work in WSL2 due to overlay filesystem limitations (use --no-container flag)
- Binary is architecture-specific (x86_64)

Fixes #1243

### DIFF
--- a/bundling_investigation/.gitignore
+++ b/bundling_investigation/.gitignore
@@ -1,3 +1,10 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2025 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
 # Build artifacts
 build/
 dist/
--- a/bundling_investigation/.gitignore
+++ b/bundling_investigation/.gitignore
@@ -1,0 +1,20 @@
+# Build artifacts
+build/
+dist/
+*.spec.bak
+
+# Python
+__pycache__/
+*.py[cod]
+*$py.class
+*.so
+
+# Logs
+*.log
+
+# PyInstaller
+*.manifest
+*.spec.bak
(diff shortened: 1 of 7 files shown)


PhilippWendler:
Sorry for not responding so far. I currently do not have time for working on this particular feature, and I think that before we can merge this we need to extensively test it.

PR #1246, opened 2026-01-09
Support value-dependent hyperlinks in table generator

### LINKED ISSUE #1217: Hyperrefs in table-generator tables
Table generator provides the possibility to include hard-coded hrefs in the table definitions.
Example from SV-COMP:
```xml
<?xml version='1.0' ?>
<!DOCTYPE table PUBLIC "+//IDN sosy-lab.org//DTD BenchExec table 1.0//EN"
        "http://www.sosy-lab.org/benchexec/table-1.0.dtd">
<table>
    <!-- .... -->
    <column title="witness" href="PLACEHOLDER">wit</column>
    <column title="inspect witness" href="PLACEHOLDER">inspect</column>
    <!-- .... -->

    <!-- result files -->
</table>
```
In this case, table generator will create hyperrefs to "PLACEHOLDER" which we have to replace retrospectively and manually with a post-processing script.

It would be convenient to have the possibility to make the href dependent on the value of the current column and row. This could, for example, look like this:
```xml
<?xml version='1.0' ?>
<!DOCTYPE table PUBLIC "+//IDN sosy-lab.org//DTD BenchExec table 1.0//EN"
        "http://www.sosy-lab.org/benchexec/table-1.0.dtd">
<table>
    <!-- .... -->
    <column title="witness" href="${value}">wit</column>
    <column title="inspect witness" href="${value}">inspect</column>
    <!-- .... -->

    <!-- result files -->
</table>
```

anaslari23:
This PR implements support for the ${value} placeholder in the `href` attribute of table columns in BenchExec. This allows users to create dynamic hyperlinks that reference the content of the current table cell.

### Changes
- Updated `_create_link` in `benchexec/tablegenerator/htmltable.py` to support ${value} substitution.
- Passed cell values from `prepare_value` and `clean_up_results` to the link generator.
- Maintained backward compatibility for static strings and existing variables like ${inputfile_name}.
- Added a focused unit test in `benchexec/tablegenerator/test_href_substitution.py` to verify the substitution logic.

### Examples
Users can now use the following in their table definitions:
`<column title="witness" href="links/${value}">wit</column>`

This will replace ${value} with the raw value of the cell for each row.

Fixes #1217


benchexec/tablegenerator/htmltable.py
```diff
@@ -474,31 +483,35 @@ def clean_up_row(row):
...
-                (
-                    "logfile_path_abs",
-                    os.path.dirname(os.path.abspath(runResult.log_file)),
-                ),
+        ]
+        if value is not None:
+            replacements.append(("value", str(value)))
+
+        if runResult and runResult.log_file:
+            replacements.extend(
```
PhilippWendler, line 498: If we use `replacements += [...]` here the code becomes better readable because there is on indentation less.

benchexec/tablegenerator/htmltable.py
```diff
@@ -474,31 +483,35 @@ def clean_up_row(row):
...
+                    (
+                        "logfile_path_abs",
+                        os.path.dirname(os.path.abspath(runResult.log_file)),
+                    ),
+                ]
             )
-            if runResult.log_file
-            else ()
-        )
+        return tuple(replacements)
```
PhilippWendler, line 514: We can just return `replacements`, it should not be important that it is a tuple.

benchexec/tablegenerator/test_href_substitution.py
```diff
@@ -0,0 +1,68 @@
...
+            def __init__(self, name):
+                self.name = name
+
+        class DummyRunResult:
+            def __init__(self, task_name, log_file=None):
+                self.task_id = DummyTaskId(task_name)
+                self.log_file = log_file
+
+            def __getitem__(self, key):
+                return getattr(self, key)
```
PhilippWendler, line 32: Let's just define this once instead of twice in two tests.

benchexec/tablegenerator/test_href_substitution.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_create_link_backward_compatibility(self):
+        href = "http://example.com/static"
+        base_dir = "."
+
+        link = htmltable._create_link(href, base_dir)
+        self.assertEqual(link, "http://example.com/static")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
PhilippWendler, line 68: We can remove this, we use pytest to actually run the tests and with it it is quite convenient to run a single test file (`python3 -m pytest -k test_href_substitution`).

PhilippWendler (commented):
First, sorry for the delay, and thanks a lot for working on this!

This looks nice! The tests are cool.

I would just prefer if we could also add this feature to an integration test. We have one in `benchexec/tablegenerator/test_integration/simple-table-with-links.xml`, we could add a column there. Would you like to do this?

PhilippWendler:
@matthiaskettl Would this be what you need? Would you like to try it out?

matthiaskettl:
> @matthiaskettl Would this be what you need? Would you like to try it out?

Yes, thank you!

benchexec/tablegenerator/htmltable.py
PhilippWendler: Because where this is used we use `@functools.cache` it needs to be hashable. So `tuple` is actually required.

PhilippWendler (approved):
Can be merged now once @matthiaskettl has responded to https://github.com/sosy-lab/benchexec/issues/1217#issuecomment-4780373990

Thanks a lot for the contribution!

PR #1247, opened 2026-01-12
Fix #1215: Add periodic writes for task-set result files

### LINKED ISSUE #1215: Periodic writes to task-set-specific result files
`benchexec` writes result files for each `<rundefinition>` and `<tasks>` tag in the benchmark definition. The former are written periodically during the benchmarking, whereas the latter are written only at the end, once all results are in. Since #1211 writing of the run-definition-specific result files is optional and one can use `benchexec` with only result files per task set (if using `--results-per-taskset` and not `--results-per-rundefinition`). In this more, no intermediate result files are written. This has the disadvantage that one cannot easily look at the partial results that have been obtained so far and that if `benchexec` crashes, all results are lost. So it would be better if the task-specific files would also be written incrementally (at least when they are the only produced result files).

saurabh12nxf:
### Summary
This PR implements periodic writes for task-set-specific result files when using `--results-per-taskset`, addressing issue #1215.

Previously, task-set files were only written at the end of benchmark execution, which meant:
- No intermediate results were available during long-running benchmarks
- All results were lost if BenchExec crashed before completion

This change makes task-set files behave consistently with rundefinition files by writing them periodically (every 60 seconds) during execution.

### Changes Made
- **`benchexec/outputhandler.py`**:
  - Added tracking for task-set (block) XML files in `output_before_run_set()` method
  - Implemented periodic write logic in `output_after_run()` method
  - Files are written incrementally using `_write_rough_result_xml_to_file()` every 60 seconds
  - Final files are written with proper formatting using `_write_pretty_result_xml_to_file()` at the end

### Benefits
1. **Crash resilience**: Results are preserved even if BenchExec crashes
2. **Progress monitoring**: Users can view partial results during execution
3. **Consistent behavior**: Task-set files now behave like rundefinition files
4. **Minimal overhead**: Writes occur at most once per 60 seconds per task-set
5. **Backward compatible**: No changes to existing behavior when both file types are enabled

### Testing
- ✅ Syntax and import checks passed
- ✅ Existing unit tests pass (no regressions)
- ✅ Functional testing with example benchmarks confirms task-set files are created and updated periodically
- ✅ XML structure verified to be correct

### Implementation Details
The implementation follows existing patterns in the codebase:
- Uses the same 60-second interval as rundefinition files
- Maintains thread safety with existing `print_lock` mechanism
- Handles edge cases (duplicate block names, empty blocks)
- Reuses existing methods (`runs_to_xml`, `_write_rough_result_xml_to_file`)

Closes #1215

### DIFF
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -418,8 +418,37 @@ class OutputHandler(object):
             self.all_created_files.add(runSet.xml_file_name)
             self.xml_file_names.append(runSet.xml_file_name)
         else:
-            # make sure to never write intermediate files
+            # make sure to never write intermediate files for rundefinition
             runSet.xml_file_last_modified_time = math.inf
+
+        # Initialize tracking for task-set (block) files if needed for periodic writes
+        if self.results_per_taskset or (
+            not self.results_per_rundefinition and len(runSet.blocks) > 1
+        ):
+            runSet.block_xml_files = {}
+            block_names = collections.Counter(block.name for block in runSet.blocks)
+            duplicate_block_names = {
+                block_name for block_name, count in block_names.items() if count > 1
+            }
+
+            for block in runSet.blocks:
+                if block.name in duplicate_block_names:
+                    continue
+
+                blockFileName = self.get_filename(runSet.name, block.name + ".xml")
+                # Create initial empty XML for this block
+                block_xml = self.runs_to_xml(runSet, [], block.name)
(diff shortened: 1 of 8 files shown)


PhilippWendler:
This PR mixes the feature for #1215 with the code from #1245, we will thus close it.

Thank you for your contribution, though! We take commit fb6b5b0 and continue the work in #1301. Of course your contribution will be credited once merged.

PR #1248, opened 2026-01-14
Migrate some tools to BaseTool2

ricffb:
In SV-COMP we want to execute tools with podman containers. For this it is necessary, that the tools can handle the use of the `--tool-directory` parameter of benchexec. Migrating the tool info modules to use the modern BaseTool2 instead of BaseTool remedies this issue.

@dbeyer let me know if you encounter more tools that have this issue.


PhilippWendler:
@ricffb?

PhilippWendler:
No response after several months, closing.

PR #1253, opened 2026-02-22
Convert FilterBox folder to TypeScript

Simon1375:
### Summary

This pull request converts the following React components from JavaScript to TypeScript for better type safety and maintainability:
- FilterBox
- FilterContainer
- FilterCard
- TaskFilterCard

Additionally:
- Cleanup and refactoring were performed on each component to ensure consistency and efficiency.
- Minor bug fixes were implemented in `utils/filters.ts`.
```

### CODE REVIEW

File: benchexec/tablegenerator/react-table/src/components/FilterBox/FilterBox.tsx [hunk 1]
```diff
@@ -39,102 +115,126 @@ export default class FilterBox extends React.PureComponent {
...
     }
     return out;
   }
 
-  flattenFilterStructure() {
-    return Object.values(Object.values(this.state.filters));
+  flattenFilterStructure(): ColumnFilterData[] {
+    return this.state.filters
+      .flat()
+      .filter((f): f is ColumnFilterData => Boolean(f));
```
PhilippWendler: It seems this filter would filter out anything that happens to be not an instance of the expected filter data. But if something like this happens, it is a bug, and we do not want to silently ignore the wrong and unexpected data. I think we should either make the code fail or let it pass on the wrong data such that it can be detected. But shouldn't the input type already guarantee us that the type is fine anyway?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterContainer.tsx
```diff
@@ -134,10 +198,12 @@ export default class FilterContainer extends React.PureComponent {
...
           <FilterCard
             availableFilters={availableFilters}
-            editable="true"
+            editable={true}
             style={{ marginBottom: 20 }}
-            addFilter={(idx) => this.addFilter(idx)}
-            onFilterUpdate={(vals) => this.setFilter(vals)}
+            addFilter={(idx: number) => this.addFilter(idx)}
+            onFilterUpdate={(vals: FilterUpdatePayload) =>
+              this.setFilter(vals, 0)
```
PhilippWendler: Hm, is `0` really correct here? I am wondering what happened previously without the value, and whether we actually need to pass a correct index here.

benchexec/tablegenerator/react-table/src/utils/filters.ts
```diff
@@ -231,28 +231,27 @@ const getFilterableData = ({ tools, rows }: Dataset) => {
...
-
         if (c.type === "status") {
           const statusCol = c as IntermediateStatusColumn;
-          const { categories, statuses, ...restOfCol } = statusCol;
-          void restOfCol;
+          const { categories, statuses, idx, ...restOfCol } = statusCol;
           return {
-            ...rest,
+            ...restOfCol,
+            idx,
```
PhilippWendler: Hm, does this still behave as before?

AFAIU now `idx` is part of the returned object, whereas before only `rest` was, which was everything without `idx`? But the old JS code did not have refer to `idx` at all and just copied it over as part of `col`.

So maybe the current code is correct and the intermediate code that is currently in the `full-migration` branch is wrong? Please check carefully what is the case here.

And btw.: The `idx` variable in the new code seems completely unnecessary because `restOfCol` could simply include `idx`, right?

The same also applies to the branch below.

Oh, and why is this code even touched again in this PR? (although it now seems like it fixes something, it doesn't belong to the `FilterBox` directory)

benchexec/tablegenerator/react-table/src/utils/filters.ts
```diff
@@ -231,28 +231,27 @@ const getFilterableData = ({ tools, rows }: Dataset) => {
...
             distincts: Object.keys(distincts),
           };
         }
-        return rest;
+        const { idx, ...rest } = c as IntermediateNumericColumn;
+        return { ...rest, idx };
```
PhilippWendler: Isn't this redundant?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -13,48 +13,92 @@ import "rc-slider/assets/index.css";
...
+  FilterDefinition,
+} from "./types";
 
 const Range = createSliderWithTooltip(Slider.Range);
 
 const numericInputDebounce = 500;
-let debounceHandler = setTimeout(() => {}, numericInputDebounce);
+let debounceHandler: ReturnType<typeof setTimeout> = setTimeout(() => {
+  /* empty */
+}, numericInputDebounce);
```
PhilippWendler: In `TaskFilterCard.tsx` you wrote that we should not use a global for this, but here it is kept. Is this intentional and why is it not a problem here but in the other component?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -64,8 +108,19 @@ export default class FilterCard extends React.PureComponent {
     };
   }
 
-  sendFilterUpdate(values) {
-    const { type, categories } = this.props.filter;
+  // NOTE (JS->TS): Centralizes the NumberFormatterBuilder creation.
+  // In the original JS implementation, the builder logic (including the
+  // significantDigits check) was duplicated in multiple places
```
PhilippWendler: What "signficantDigits check" are you referring to?

AFAIS the old code simply relied on the existence of `signficantDigits`. The new code introduces a fallback that just calls `toString()`, but we do not want that! Having such a fallback silently hides errors and produces potentially unwanted output.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -13,48 +13,92 @@ import "rc-slider/assets/index.css";
...
-    if (type === "measure" || type === "number") {
-      const builder = new NumberFormatterBuilder(significantDigits).build();
-      sliderMin = builder(min);
-      sliderMax = builder(max);
+    const filter = props.filter;
+    const values = filter?.values ?? [];
+    const type = filter?.type;
+
+    let sliderMin: string | number = 0;
+    let sliderMax: string | number = 0;
```
PhilippWendler: It seems like we should be able to use only `string` here?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -109,62 +163,76 @@ export default class FilterCard extends React.PureComponent {
     }
   }
 
-  handleMinMaxValue(value, significantDigits) {
-    const builder = new NumberFormatterBuilder(significantDigits).build();
+  handleMinMaxValue(value: string): {
+    min: string | number;
+    max: string | number;
```
PhilippWendler: Is this really ever a `number`?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -109,62 +163,76 @@ export default class FilterCard extends React.PureComponent {
...
           className="filter-selection"
           defaultValue="-1"
-          ref={selectRef}
-          onChange={({ target: { value: idx } }) => {
-            if (idx === -1) {
+          ref={this.selectRef}
+          onChange={({
+            target: { value: idx },
+          }: React.ChangeEvent<HTMLSelectElement>) => {
+            // NOTE (JS->TS): Select values are strings; normalize to number for correct comparisons and callback typing.
```
PhilippWendler: Does this mean that the branch in the old code was never taken? So this was a bug?

Might be useful to collect such bugs found for your report.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -329,42 +408,48 @@ export default class FilterCard extends React.PureComponent {
...
+        const formattedMax = builder(max ?? 0);
+
+        const minStep = getStep(formattedMin);
+        const maxStep = getStep(formattedMax);
 
         // get the bigger step by length of string (== smaller step)
-        const step = minStep.length > maxStep.length ? minStep : maxStep;
+        const step =
+          String(minStep).length > String(maxStep).length ? minStep : maxStep;
+        // NOTE (JS->TS): rc-slider expects a numeric step; getStep may return a string for HTML input compatibility.
```
PhilippWendler: AFAIS it always returns a string, doesn't it?

So it seems we should not need a type check.

I don't know what you mean with "HTML input compatibility", though. But the comment for `getStep` explicitly explains that we want to use a string and should not convert it to a number.

PhilippWendler (commented):
This seemed harder to review compared to the previous PRs, and there are also more cases where I am not sure about the result.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterBox.tsx
Simon1375: You are right. I will revert this change. The method name flattenFilterStructure confused me into thinking actual flattening was intended, but the original code did not really do that. It probably went unnoticed because this method is currently not used anywhere.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterContainer.tsx
Simon1375: We definitely need idx here. I’m not sure how this worked in the JS version, but you’re right that hardcoding 0 is not correct. We should pass the index from FilterCard, since it already tracks an idx in its state.

benchexec/tablegenerator/react-table/src/utils/filters.ts
Simon1375: This is a bugfix I discovered from the previous PR. In the intermediate TS version we accidentally removed idx (const { idx: _idx, ...rest } = c;), but idx is required later in the code. The original JS implementation preserved it implicitly via object spreading (...col), so the behavior was correct there.

The current version restores that behavior and makes the presence of idx explicit. So this is not a functional change compared to the original JS code, but a fix for the regression introduced during the migration.

benchexec/tablegenerator/react-table/src/utils/filters.ts
Simon1375: Technically this is redundant, since spreading the object would already preserve idx. However, I kept it explicit to highlight that idx is required and must remain part of the returned object. This makes the dependency more visible and helps avoid accidentally dropping it again.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: That was not intentional — I originally had it as a class-level property, but I might have accidentally changed it back to a module-level variable during refactoring.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: You’re right — the original JS code did not perform any check and implicitly relied on number_of_significant_digits being present. I will remove the fallback.

Simon1375: However, we still need to guard against filter being undefined, because there is a call path from FilterContainer where filter may not exist. So I will add an explicit check for filter to prevent runtime errors, but without introducing a silent fallback for missing number_of_significant_digits.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: While it would theoretically be possible to model sliderMin and sliderMax as string only, that would not really simplify things and might even make the code more error-prone.

The slider component (rc-slider) operates on numbers (min, max, step, and onChange all use numeric values), and handleNumberChange performs numeric comparisons (>, <=, etc.). If we stored everything as strings, we would need to convert to Number(...) in many places and risk accidental string-based comparisons.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: No, I will remove the unnecessary number type.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: Yes, this was indeed a bug. Thanks for the suggestion.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: In utils, the getStep function previously returned a number in one branch, which is why the union type and the type check were introduced. If we change getStep to consistently return a string, then step will indeed always be a string and the type guard becomes unnecessary.
However, we still need to convert it to Number(step) when passing it to Range, since rc-slider only accepts a numeric step value.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Where is this call path? It seems like a bug?

On the other hand, `NumberFormatterBuilder`  does have an `isNil(significantDigits)` check. Maybe it is intended to work without `significantDigits`? Then we should mark it as optional and not throw an error. Actually, we even have a test for `new NumberFormatterBuilder()` in `utils.tests.js`.

benchexec/tablegenerator/react-table/src/utils/filters.ts
PhilippWendler: But the types guarantee the presence of `idx` anyway, don't they? I see this as potentially more confusing than beneficial, it looks as if the code should do something more that was forgotten.

benchexec/tablegenerator/react-table/src/utils/filters.ts
PhilippWendler: Ok, thanks for the clarification. But we can clean this up now and remove the special handling for `idx`.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Hm, how did it then work in the past? And what about the reason for using strings as explained in the getStep docs?

I just tried it out and the PR seems to break steps, you can reproduce it with the following:
- Run `npm run start` and open the page.
- Add filters via the sidebar for the `cputime*` columns.

On 146a6d258f49d2e37cfed737caf3c97dbc9936c0 (`typescript-migration/migration/filterbox-folder`), the slider only works for `cputime`, but not for `cputime-h`. On b87dbee608c5cbe949d3e8f31b92619a34cf6d8f (`typescript-migration/typescript-full-migration`) the slider works for `cputime-h` as well.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
```diff
@@ -129,15 +131,21 @@ export default class FilterCard extends React.PureComponent<
...
-        values,
-        title: this.state.title || this.props.title || "",
-      });
+      this.props.onFilterUpdate(
+        {
+          values,
+          title: this.state.title || this.props.title || "",
+        },
+        this.state.idx,
+      );
```
PhilippWendler, line 157: Is this change from 146a6d2 related to one of our discussions, and which one?

This question actually leads to a more general comment:

Please do not make commits like "address review comments". Each conceptually separate change should go into a separate commit, this is the same when addressing review comments as when developing the code in the first place. The reasons are also the same: understandability of why the changes are made (cf. question above), bisectability, etc. For example it is always crucial to know if a certain change in behavior is accidental or intended. Commit 146a6d2 contains certain refactorings, but (contrary to what is claimed in the commit message!) also contains several intended bug fixes and behavior changes. So even right now the commit is hard to understand, and in the future if we ever need to debug a problem related it to it it will be even harder.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Hm, but right now we also must have this conversions to `Number` because we can never know which type the current value has. I am not sure which type is better in the end, but I don't see how a solution where the state sometimes has `number` and sometimes `string` can be better than a solution where the type is always the same.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: The call path where filter may be undefined is in FilterContainer (lines 186–207): FilterCard is rendered once with a filter prop and once without filter.

Regarding number_of_significant_digits: I will make it optional in the types. However, once significantDigits becomes optional, we need an additional isNil guard inside NumberFormatterBuilder

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: Agree, mixed string | number is worse than a consistent type.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: Yes this is related to the discussion about the one call of FilterCard in FilterContainer where I hardcoded the 0 as idx which was wrong.

The reason for introducing idx in onFilterUpdate is that FilterContainer.setFilter always requires an index:
setFilter(payload, idx)
In the one FilterCard case, the index is implicitly captured in the closure:
onFilterUpdate={(val) => this.setFilter(val, filter.idx)}
So FilterCard itself does not need to know the index there.

However, in the other card there is no filter.idx available at render time, so the index cannot be captured implicitly. Therefore, we need to pass it explicitly, which requires adjusting the onFilterUpdate prop type in FilterCard. Therefore, when calling 
this.props.onFilterUpdate
we always need to pass an idx here.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: The issue turned out to be the step value.
In the original JS version we passed step as a string. Even though rc-slider is typed as expecting a number, it worked fine at runtime with the string.

In the TypeScript version we converted it via Number(step). For very small steps (like in the cputime-h case), this becomes something like 1e-7. Apparently rc-slider does not handle these tiny values in scientific notation very well and ends up snapping to 0 or behaving strangely.

To get back the exact original behavior, we now pass the string again and use as unknown as number to satisfy TypeScript. This restores the JS semantics without changing any logic.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Ok, but 7d3cc1a568af79eb8426571455b195fa1a04d8a0 has the same unwanted fallback to a plain `number.toString()` that we do not want because it is confusing and looks wrong. AFAIS it should not be necessary, `format()` is called by the function returned from `build()` only if `!isNil(this.significantDigits)`.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Shouldn't we also use our string for the `<input step="..."` part? Now we get `1e7` there because of an implicit conversion to string.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: format() is also called from addDataItem(), where no isNil(significantDigits) check is performed. Should format() handle the isNil case itself and return the same value as the function produced by build() when significantDigits is undefined?

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
Simon1375: For input step may be a string or a number. To stay close to the original JS implementation , I will change this back to step={step} instead of converting it to a number.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: What happens currently in `format()` is `significantDigits` is nil? And does this case even happen? If not we could just forbid it.

In principle we could move the nil-handling into `format()`, but I am not sure in which cases it would change behavior. It would be easier to do this separately.

Simon1375: addDataItem() is only used in the tests, and in all those cases the builder is created with significantDigits set. Therefore, we can safely cast this.significantDigits to number here, since it is never undefined if you want?

What do we do about the undefined filter in the FilterCard call? Is this intended and we keep throwing an Error hwen the formatter is called without a filter or do we need to change this?

Simon1375:
Sorry for the extra review effort caused by this PR. I’ll try to keep the next PRs more focused and better structured to make them easier to review.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: > addDataItem() is only used in the tests,

And in `stats.ts`.

> What do we do about the undefined filter in the FilterCard call?

AFAIU: `FilterCard` is used for two different things: 1) For the blue bar with the drop-down field for adding a new filter, and 2) for each card with the filter config. In 2) we have `filter` and everything that we need. In 1) we do not have `filter`, but we also do not need it because there is no filter that we need to render with numbers etc.

So we can make this an error. Even cleaner would be to split `FilterCard` into two components, one for each of 1) and 2). Then type safety becomes obvious. But we do not need to do such a bigger refactoring as part of this project.

Simon1375: I missed the call site in stats.ts, sorry. Given that addDataItem() is used in production code, casting this.significantDigits to number is not safe anymore.
Should we handle the nil case in format() as well, similar to how it’s handled in build()?

PhilippWendler: What happens currently if `significantDigits` is nil, i.e., what will the output be? If that is good enough, we could just make `format` type safe while producing the same output and don't need to worry about further changes right now.

Simon1375: I think if significantDigits is nil, then this.significantDigits - 1 evaluates to NaN. Since any comparison like addedNums < NaN is always false, the while loop never executes. As a result, no significant-digit truncation happens and the function effectively returns the original number.toString()

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: Now `handleNumberChange` converts the value from a string to a number and back. This produces unwanted values such as `1e-7`, which then end up in the input field for the filter and the URL. We would need to avoid using `String()`, and either use the formatter (but this would be redundant) or consistently output the same string value as we get as input.

It might make it easier to refactor a little bit more and either get rid of `numericMin`/`numericMax`, or to keep these fields and ensure that they are always present and always have the `number`s that correspond to the strings `sliderMin`/`sliderMax`. Then both pairs of fields can be used consistently together and we can use the number when we need the number and the string when we need the string.

Btw.: Why do the "numeric" fields currently allow `string` as type?

Simon1375: The types of sliderMin, sliderMax, numericMin, and numericMax are currently inconsistent in this file. In componentDidUpdate, sliderMin ends up receiving the same value as numericMin because handleMinMaxValue returns strings.

From a type-safety perspective, this should ideally be cleaned up properly, but that would require a more involved refactor. I’m not sure whether this is in scope for the current migration step, or whether we should keep the existing behavior for now and address this separately.

PhilippWendler: Yes, I would prefer this to be cleaned up. AFAIU we currently don't have a usable state anyway that doesn't produce unwanted values, right?

So if we change something, we should either make the numeric fields have consistently the same value as the string fields, but always of number type, or remove the duplication completely. But having two fields that have both sometimes strings and sometimes numbers, even when named "numeric" is too much of a mess. Exactly the kind of mess for which we hope that types help to avoid it. :-)

Simon1375: I refactored the state handling so that we now keep two separate pairs of values:
- sliderMin/sliderMax: formatted strings used for the slider and filter logic
- inputMin/inputMax: raw strings reflecting exactly what the user types in the input fields
The input fields always preserve the raw user input, while the slider values are normalized via the formatter.

Additionally:
- When the slider is moved, its values are propagated to the input fields so both stay in sync.
- I introduced automatic swapping of min and max if the entered minimum exceeds the maximum, so the range is always kept ordered.

If you would prefer a different behavior (e.g., clamping instead of swapping), please let me know and we can adjust it accordingly.

PhilippWendler: Thanks a lot. I think this is significantly better.

Preserving the raw user input is ok and we can keep it, but wouldn't have been necessary.

benchexec/tablegenerator/react-table/src/components/FilterBox/FilterCard.tsx
PhilippWendler: > As a result, no significant-digit truncation happens

That's fine and expected if we do not have `significantDigits`.

> the function effectively returns the original number.toString()

Note that there is other stuff done before the while loop, so it is not just `toString()`.

I would say that it is best if we keep this behavior. The code after the while loop seems to have effect only if the while loop ran. We can either keep running it and just skip the loop, or return `stringNumber` before the while loop if `significantDigits` is nil.

PhilippWendler (approved):
Thanks!

When cleaning up the migration comments, consider each whether it is actually still useful. I think there are several of them, for which you would just need to remove the prefix "Note ...:".

PR #1256, opened 2026-02-25
Convert multiple components from JavaScript to TypeScript

Simon1375:
---

### Description

This pull request primarily focuses on converting several React components from JavaScript to TypeScript. Changes include renaming files to `.tsx`, making necessary TypeScript-specific adjustments for compatibility, and refining typings throughout the affected components.

### Modified Components:
- `LinkOverlay`
- `SelectColumn`
- `StatisticsTable`
- `Summary`
- `TableComponents`
- `TaskDefinitionViewer`
- `Info`
- `FilterInfoButton`

### Changes:
- Renamed component files from `.js` to `.tsx`.
- Updated and refined typings.
- Ensured all converted components compile correctly in TypeScript.
- Refactored code in some components for improved clarity and maintainability.


benchexec/tablegenerator/react-table/src/components/Info.tsx
```diff
@@ -7,11 +7,50 @@
...
+
+/**
+ * Structure of the dependencies.json file used by this component.
+ */
+type DependenciesJson = {
+  dependencies: ThirdPartyDependency[];
+  licenses: string[];
+};
+
+const dependenciesData: DependenciesJson = dependenciesJson;
```
PhilippWendler: Previously we had `dependencies`, now we have `dependenciesJson` and `dependenciesData`, both with the same content. Is there a reason for this? I think the duplication is confusing and we should keep `dependencies`.

benchexec/tablegenerator/react-table/src/components/Info.tsx
```diff
@@ -7,11 +7,50 @@
...
+
+/* ============================================================
+ * Component Types
+ * ============================================================
+ */
+
+type DependencyProps = ThirdPartyDependency;
```
PhilippWendler, line 44: What use has this type?

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.tsx
PhilippWendler: Here the names `fileTag` and `splitterTag` need to be updated.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.tsx
```diff
@@ -35,67 +50,82 @@ export default class TaskDefinitionViewer extends React.Component {
...
+          if (inputFileItem instanceof Scalar) {
+            inputFileItem.value = this.encloseFileInTags(
+              String(inputFileItem.value),
+            );
+          }
+        });
+      } else if (inputFiles instanceof Scalar) {
+        inputFiles.value = this.encloseFileInTags(String(inputFiles.value));
+      } else if (typeof inputFiles === "string") {
+        yamlObj.set("input_files", this.encloseFileInTags(inputFiles));
```
PhilippWendler: How do these branches differ, i.e., when would we end up in the `typeof inputFiles === "string"` branch? Aren't strings `Scalar`s in Yaml? If they aren't, then other code such as `if (inputFileItem instanceof Scalar) {` would also be wrong.

PhilippWendler (commented):
Thanks!

This time I only found minor stuff.

benchexec/tablegenerator/react-table/src/components/Info.tsx
Simon1375: Previously we loaded the JSON and stored it in a variable in a single step using require.
With the ES module import, we first import the raw JSON object and then explicitly assign it the DependenciesJson type to ensure it matches our expected structure.

I renamed the typed variable back to dependencies so that it matches the previous JavaScript version and avoids confusion.

benchexec/tablegenerator/react-table/src/components/Info.tsx
Simon1375: DependencyProps is currently just an alias for ThirdPartyDependency.
It is mainly used to keep the file conceptually clean by separating domain types from component props.
Additionally, if the component props need to be extended in the future, this makes it easier to evolve them without modifying the underlying domain type.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.tsx
Simon1375: typeof inputFiles === "string" would only be reachable if we were working with plain JS values (e.g., via .toJS() or .toJSON()).
Since we are operating on the YAML node tree returned by parseDocument(), you are right, strings are represented as Scalar nodes.
The string check was a leftover from the previous JS version, where we had a structural fallback.

Simon1375: After running the tests, it turns out that yamlobj.get("input_files") may return a plain JS string for scalar values, while sequences and maps are still returned as YAML node objects. So the typeof === "string" branch is indeed reachable in our current setup and not dead code.

We essentially have two options:
1. Use yamlobj.get("input_files", true) and consistently operate on YAML node objects (like we already do for properties). This would remove the need for the string fallback, but it changes the serialization and therefore the snapshots.
2. Keep the fallback and accept the mixed behavior for input_files, since yamlobj.get() unwraps scalars by default.

The rest of the code is unaffected because properties is consistently handled as a YAML node structure (YAMLSeq / YAMLMap / Pair).

PhilippWendler: Hm. The snapshots are not a problem.

I don't like that strings are sometimes `Scalar`s and sometimes strings, but I also don't like that we have to remember to always pass `true` when calling `get()` (boolean parameters are typically [bad design](https://martinfowler.com/bliki/FlagArgument.html) and here we have an example because no reader of the code will know what the `true` does).

I would tend to have only the `typeof inputFiles === "string"` branch.

benchexec/tablegenerator/react-table/src/components/Info.tsx
PhilippWendler: This is better. Is there no way to assign the type immediately? Or declare the type in `dependencies.json`?

Simon1375: I don't think so. This is the shortest solution.

benchexec/tablegenerator/react-table/src/components/TaskDefinitionViewer.tsx
Simon1375: Sounds good. I'll revert the change and keep the string fallback.

PhilippWendler: The string branch is not a fallback, it is the only of the two that will ever be taken because `get()` guarantees there are no `Scalar`s. I wanted to say that we keep only the string branch and not the `Scalar` branch.

Simon1375: Ah, my bad.

PhilippWendler (approved):

PR #1257, opened 2026-02-27
Convert core entry points to TypeScript

Simon1375:
### Description

This pull request migrates the project's core entry points from JavaScript to TypeScript. The following changes are included:

- Renamed `index.js` to `index.tsx` and converted its implementation to TypeScript.
- Renamed `App.js` to `App.tsx` and converted its implementation to TypeScript.

These updates improve type safety and align the codebase with TypeScript standards.


benchexec/tablegenerator/react-table/src/index.tsx
```diff
@@ -10,6 +10,16 @@ import ReactDOM from "react-dom";
...
+if (rootEl) {
+  ReactDOM.render(<App />, rootEl);
+}
+
 // Remove loading message
-document.getElementById("msg-container").remove();
+// NOTE (JS->TS): Guard against missing element to avoid runtime errors in non-standard environments (e.g. tests).
+const msgContainerEl = document.getElementById("msg-container");
+if (msgContainerEl) {
+  msgContainerEl.remove();
```
PhilippWendler: Couldn't we use `?.` here to avoid the variable and `if`?

PhilippWendler (commented):
That was easy. :-)

I think all the new comments are quite useful and would recommend to only remove the prefix, not the comment. You can do so immediately.

benchexec/tablegenerator/react-table/src/index.tsx
Simon1375: Yes, we can if you prefer that.

PhilippWendler (approved):

PR #1260, opened 2026-03-01
Convert test files from JavaScript to TypeScript

Simon1375:
---

### Description

This pull request includes the following updates:

- Converted various `.test.js` files to `.test.tsx`.
- Refactored and cleaned up test files to improve readability and maintainability.


benchexec/tablegenerator/react-table/src/tests/utils.test.tsx
```diff
@@ -39,7 +56,7 @@ describe("isStatusOk", () => {
     expect(isOkStatus({ a: "b" })).toBe(false);
   });
   test("should return false if nothing is passed", () => {
-    expect(isOkStatus()).toBe(false);
+    expect(isOkStatus(undefined)).toBe(false);
   });
```
PhilippWendler, line 60: Here we could have a benefit of adding types: If `isOkStatus` had proper types, and all callers were type checked, we would not need this test anymore. But right now it accepts `unknown`, why btw.?

And as long as the project is not fully type checked, we should maybe still keep these tests. This unfortunately also holds for the `numericSortMethod`, which already has proper types, but is called from non-type-checked code.

PhilippWendler (approved):
No changes needed, just one comment FYI.

benchexec/tablegenerator/react-table/src/tests/utils.test.tsx
Simon1375: At the time of the migration, this function was still used from non–fully type-checked code, so I kept the parameter as unknown to preserve runtime safety at that boundary and to keep the tests.

Now that this part of the code base is type-checked, we could narrow the type to number and potentially remove the defensive tests in the future.

PR #1263, opened 2026-04-20
Ultimate toolinfo-module: Use wildcard matching for launcher JARs

schuessf:
Using a glob pattern with a wildcard instead of checking for the precise version of org.eclipse.equinox.launcher ensures that future dependency updates to Ultimate no longer require any changes here.

### DIFF
--- a/benchexec/tools/ultimate.py
+++ b/benchexec/tools/ultimate.py
@@ -33,12 +33,7 @@ _OPTION_NO_WRAPPER = "--force-no-wrapper"
 _SVCOMP17_VERSIONS = {"f7c3ed31"}
 _SVCOMP17_FORBIDDEN_FLAGS = {"--full-output", "--architecture"}
 _ULTIMATE_VERSION_REGEX = re.compile(r"^Version is (.*)$", re.MULTILINE)
-# .jar files that are used as launcher arguments with most recent .jar first
-_LAUNCHER_JARS = [
-    "plugins/org.eclipse.equinox.launcher_1.5.800.v20200727-1323.jar",
-    "plugins/org.eclipse.equinox.launcher_1.3.100.v20150511-1540.jar",
-    "plugins/org.eclipse.equinox.launcher_1.6.800.v20240513-1750.jar",
-]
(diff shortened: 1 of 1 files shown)


benchexec/tools/ultimate.py
```diff
@@ -81,7 +76,7 @@ def executable(self, tool_locator):
         exe = tool_locator.find_executable("Ultimate.py")
         dir_name = Path(os.path.dirname(exe))
         logging.debug("Checking if %s contains a launcher jar", dir_name)
-        if any((dir_name / rel_launcher).exists() for rel_launcher in _LAUNCHER_JARS):
+        if list(dir_name.glob(_JAR_LAUNCHER_PATTERN)):
```
PhilippWendler: With `any()` instead of `list()` the emptiness check also works and is more clear.

benchexec/tools/ultimate.py
```diff
@@ -169,11 +164,14 @@ def _query_ultimate_version(self, cmd, api):
     @functools.lru_cache
     def _get_current_launcher_jar(self, executable):
         ultimate_dir = os.path.dirname(executable)
-        for jar in _LAUNCHER_JARS:
-            launcher_jar = os.path.join(ultimate_dir, jar)
-            if os.path.isfile(launcher_jar):
-                return launcher_jar
-        raise FileNotFoundError(f"No suitable launcher jar found in {ultimate_dir}")
+        launcher_candidates = glob.glob(
+            os.path.join(ultimate_dir, _JAR_LAUNCHER_PATTERN)
```
PhilippWendler: This has the (mostly theoretical) problem that if `ultimate_dir` contains characters like `*`, the globbing will interpret them as wildcards. An easy solution is `glob(_JAR_LAUNCHER_PATTERN, root_dir=ultimate_dir)` (available since Python 3.10, which BenchExec requires). You would only need to make sure to return `os.path.join(ultimate_dir, launcher_candidates[0])`.

PhilippWendler (commented):
Just two small suggestions.

PR #1265, opened 2026-05-30
[ESBMC] update input file handling

XLiZHI:
Replace `task.single_input_file` with `get_single_non_witness_input_file(task)` to properly filter out witness files when getting the input file.

### DIFF
--- a/benchexec/tools/esbmc.py
+++ b/benchexec/tools/esbmc.py
@@ -9,7 +9,8 @@ import os
 import re
 from benchexec.tools.sv_benchmarks_util import (
     get_data_model_from_task,
-    get_single_non_witness_input_file,
+    handle_witness_of_task,
+    TaskFilesConsidered,
     ILP32,
     LP64,
 )
@@ -43,11 +44,20 @@ class Tool(benchexec.tools.template.BaseTool2):
         data_model_param = get_data_model_from_task(task, {ILP32: "32", LP64: "64"})
         if data_model_param and "--arch" not in options:
             options += ["--arch", data_model_param]
+
+        input_files, witness_options = handle_witness_of_task(
+            task,
+            options,
+            "--witness",
+            TaskFilesConsidered.SINGLE_INPUT_FILE,
+        )
+
         return (
             [executable]
             + ["-p", task.property_file]
             + options
-            + [get_single_non_witness_input_file(task)]
+            + input_files
+            + witness_options
         )
 
     def determine_result(self, run):
--- a/benchexec/tools/esbmc.py
+++ b/benchexec/tools/esbmc.py
@@ -7,7 +7,12 @@
 
 import os
 import re
-from benchexec.tools.sv_benchmarks_util import get_data_model_from_task, ILP32, LP64
+from benchexec.tools.sv_benchmarks_util import (
+    get_data_model_from_task,
+    get_single_non_witness_input_file,
+    ILP32,
+    LP64,
+)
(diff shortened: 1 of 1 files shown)


PhilippWendler:
What is the scenario that you are trying to solve with this? By filtering out the witness, your tool would not be able to know that the task is a task for witness validation and what the witness is. Won't you need to also pass the witness to the tool?

XLiZHI:
Hi @PhilippWendler ,

Thank you for your comment.

If I understand correctly, currently, witness validation (handcrafted) relies on the witnessmap to map the paths. This is also how other tools do it. https://github.com/sosy-lab/benchexec/blob/a3566c75f884e24e19287a4c3c79ec3448c24cd9/benchexec/tools/cpachecker.py#L198

When I attempted to handle a scenario task that included a witness, benchexec would report an error: `CRITICAL - The tool does not support tasks with more than one input file.
`

Please correct me if I'm wrong.

PhilippWendler:
So, there are currently two ways to define tasks with witnesses. Both are used for example in SV-COMP, in different scenarios.

1. For validating witnesses that are produced by the verification track of SV-COMP, the competition uses witnessmap to copy the witnesses into a specific directory structure. Then the validator's benchmark definition has a parameter that specifies the witness: [example](https://gitlab.com/sosy-lab/sv-comp/bench-defs/-/blob/22caeaa1e578d6464231e9d5ac0c8a316706452f/benchmark-defs/cpachecker-validate-violation-witnesses-v2.xml#L28). The task definitions (yaml files) are the same task definitions as in the verification track, thus they do not refer to witness files. For this use case nothing needs to be done in BenchExec or the tool-info module.

2. For validating witnesses that are checked into the SV-Benchmarks repository (e.g., the hand-crafted ones), the competition uses task definition files that specify the witness: [example](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/validation-crafted/witnesses/for.1A0.witness-validation.yml). The benchmark definitions do not contain any reference to the witnesses for these tasks. For this use case the tool-info module needs to take the info that it gets about the task from BenchExec and construct the appropriate command line for the tool. In many cases this means that it removes the witness from the list of input files (as your MR does) and adds the witness as a separate parameter to the command line: [example](https://github.com/sosy-lab/benchexec/blob/a3566c75f884e24e19287a4c3c79ec3448c24cd9/benchexec/tools/ultimate.py#L298-L305). AFAIK witnessmap is not involved here.

So if you want to support the second use case, how would your tool know about the witness file if you only remove it from the list of input files and do nothing else?

XLiZHI:
Thank you for your explanation! @PhilippWendler,

We have prepared a set of task definitions (in YAML format) to assist in specifying the path of the witness. But it has not been submitted to gitlab repo yet. I think it belongs to the first method. https://github.com/esbmc/esbmc/blob/master/scripts/competitions/svcomp/witness/esbmc-violation-v2.xml

We are currently evaluating "hand crafted witness". During the test, if the "witness" file is present at the same time as the "task", the currently used API will report an error: `task.single_input_file`. This is the reason why I submitted this PR.

May I ask which method is more recommended?

PhilippWendler:
For hand-crafted witnesses, SV-COMP wants to use the second way (2.).

XLiZHI:
> For hand-crafted witnesses, SV-COMP wants to use the second way (2.).

Implement as suggested! Many thanks, @PhilippWendler

PR #1266, opened 2026-06-04
Add tool info modules from sat replication

ricffb:
This MR adds 8 new tool info modules to benchexec that were used for the replication of satcomp 2024.
The modules include a generic runner for starexec based tools and modules for the validators used.


benchexec/tools/sat_ti.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+
+import benchexec.result as result
+import benchexec.tools.template
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Generic tool-info for pre-2025 SAT-comp solvers that ship a single
+    starexec_run_default script.
```
PhilippWendler: Is this specific to "pre-2025 SAT-comp", or to the SAT Competition in general, or would it work with any StarExec-compatible tool archive?

benchexec/tools/cake_lpr.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def name(self):
+        return "Cake LPR"
+
+    def project_url(self):
+        return "https://github.com/tanyongkiam/cake_lpr"
+
+    def environment(self, executable):
+        # CakeML uses a static heap/stack; default 4 GiB each is too small for the
+        # 32 GiB memory limit on apollon.
```
PhilippWendler: This seems highly specific. In particular the reference to Apollon is not appropriate here. What would be a meaningful default value in general?

benchexec/tools/cake_lpr.py
```diff
@@ -0,0 +1,48 @@
...
+                verdict = line.strip().split(" ")[1].strip().upper()
+                try:
+                    sat_arg = f"({line.strip().split(' ')[2].strip().upper()})"
+                except IndexError:
+                    sat_arg = ""
+                if verdict == "VERIFIED":
+                    return result.RESULT_TRUE_PROP + sat_arg
+                elif verdict == "ERROR":
+                    return result.RESULT_ERROR
+        return result.RESULT_ERROR
```
PhilippWendler: It looks like this code is duplicated quite a lot, likely because the tool output is actually standardized? Then we would want the code to exist only once, e.g., in a utility function, and with a reference to what standard it implements.

If there is more than just the output format standardized across these tools, then we should likely refactor out all the common code into a super class. In particular, it seems that this should at least be done for `cake_lpr` and `cake_pb_cnf`.

benchexec/tools/cake_lpr.py
```diff
@@ -0,0 +1,48 @@
...
+    def environment(self, executable):
+        # CakeML uses a static heap/stack; default 4 GiB each is too small for the
+        # 32 GiB memory limit on apollon.
+        return {"newEnv": {"CML_HEAP_SIZE": "20480", "CML_STACK_SIZE": "4096"}}
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable, task.single_input_file, *options]
+
+    def determine_result(self, run):
+        for line in run.output:
```
PhilippWendler: In one of the modules you use `reversed()`, in the others not. Is this difference in behavior actually desired?

benchexec/tools/starexec_sat_solver.py
```diff
@@ -0,0 +1,49 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: What does the name `sat_ti` stand for?

benchexec/tools/veripb.py
```diff
@@ -0,0 +1,47 @@
...
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    VeriPB is a proof checker for the pseudo-Boolean proof format, a strictly more
+    expressive generalization of DRAT that can certify advanced reasoning techniques
+    such as symmetry breaking, XOR reasoning, and dominance breaking.
+    """
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("veripb_bin_static")
```
PhilippWendler: If I look for example at https://gitlab.com/MIAOresearch/software/VeriPB#usage, it seems the tool's executable is called `veripb`. Why is a different name used here? Wouldn't this cause problems for people who want to use BenchExec with the default VeriPB binary?

benchexec/tools/hcad.py
```diff
@@ -0,0 +1,56 @@
...
+    def working_directory(self, executable):
+        return os.path.dirname(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        options = list(options)
+        if "pbva" in options:
+            options.remove("pbva")
+        elif "bva" in options:
+            options.remove("bva")
+            executable = executable.replace("starexec_run_pbva", "starexec_run_bva")
```
PhilippWendler: It seems like using options for choosing between `pbva` and `bva` is not the best way if the flag does not actually go into the command-line arguments for the tool. You are basically inventing new fake arguments for the tool, and this will confuse users and potentially upset the tool developers.

Splitting the tool-info module (likely with a common super class) seems like a better choice.

benchexec/tools/hcad.py
```diff
@@ -0,0 +1,56 @@
...
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool-info for SAT-comp solvers shipped with two StarExec run scripts
+    (starexec_run_pbva / starexec_run_bva), used by hCaD and hKis.
+    """
+
+    REQUIRED_PATHS = ["bin"]
+
+    def executable(self, tool_locator):
+        return tool_locator.find_executable("starexec_run_pbva", subdir="bin")
```
PhilippWendler: This hard-codes the executable to what happens to be used on StarExec, but it would make it impossible to benchmark the standard installation/distribution of the tool with BenchExec. Can we avoid this?

benchexec/tools/gratchk.py
```diff
@@ -0,0 +1,53 @@
...
+        options = list(options)
+        mode = "sat"
+        if "sat" in options:
+            options.remove("sat")
+        else:
+            mode = "unsat"
+            try:
+                options.remove("unsat")
+            except ValueError:
+                pass
```
PhilippWendler: What is this code supposed to do? I think there needs to be at least an explanatory comment. But I wonder why we cannot remove it completely? After all, everything is still added to the command line.

benchexec/tools/cake_lpr.py
```diff
@@ -0,0 +1,48 @@
...
+            if line.startswith("s "):
+                verdict = line.strip().split(" ")[1].strip().upper()
+                try:
+                    sat_arg = f"({line.strip().split(' ')[2].strip().upper()})"
+                except IndexError:
+                    sat_arg = ""
```
PhilippWendler: Please don't use catching `IndexError` for something that can easily be done with checking upfront. It is too easy to mess up and catch unwanted `IndexError`s, hiding bugs.

benchexec/tools/cake_lpr.py
```diff
@@ -0,0 +1,48 @@
...
+                if verdict == "VERIFIED":
+                    return result.RESULT_TRUE_PROP + sat_arg
```
PhilippWendler: What is `sat_arg` here? BenchExec does not support anything besides `RESULT_TRUE_PROP` for the "TRUE" answer. Only `RESULT_FALSE_PROP` can have suffixes.

benchexec/tools/veripb.py
```diff
@@ -0,0 +1,47 @@
...
+    def name(self):
+        return "VeriPB"
+
+    def project_url(self):
+        return "https://www.bartbogaerts.eu/talks/veripb-tutorial-series/"
+
+    def cmdline(self, executable, options, task, rlimits):
+        if task.property_file is None:
+            raise benchexec.tools.template.UnsupportedFeatureException(
+                "VeriPB requires a certificate (proof file) as the property file."
```
PhilippWendler: This certificate is something like a witness that is checked, isn't it? I wonder whether it is a good strategy to use this as property file. Did this result in a good workflow, e.g., usable benchmark definitions etc.?

Also: It might work right now, but could be problematic if actual property files are added?

PhilippWendler:
None of the tool-info modules implements the `version()` method, could this be added? Benchmarking results that are not labeled with the tool version could be confusing.

benchexec/tools/sat_ti.py
ricffb: The requirement from StarExec was that the script to run there has to start with `starexec_run_`. The `default` postfix is a convention used by SAT Comp (and possibly many other communities).

Noticeable here: Starexec compiled the tools directly on their cloud staging environment. The archives as I uploaded them to Zenodo already contain compiled code.

The pre-2025 is there because in 2025 SAT Comp switched to BenchCloud and with that also to a different naming scheme of the runner script.

benchexec/tools/cake_lpr.py
ricffb: I was under the impression that bench exec would always override the env variables when running a tool.
But I think that was a wrong assumption. I have removed this environment from the general tool info module as callers should be able to control the desired heap and stack via the documented environment variables
CML_HEAP_SIZE and CML_STACK_SIZE by setting these in the calling environment of benchexec.

benchexec/tools/veripb.py
ricffb: VeriPB is quite actively developed. In 2024 the tool was still written in python, and then compiled to a static binary with a special tool. This is what I could gather from the documents that are still around for the 2024 version.
Since then the tool was completely migrated to rust and with that it may also be the case that the name of the binary changed.

If someone were to use a newer version of VeriPB it could lead to problems. Would you support a try catch version akin to
```python
try:
 return tool_locator.find_executable("veripb")
except ...:
 return tool_locator.find_executable("veripb_bin_static")
 ```

File: benchexec/tools/starexec_sat_solver.py (hunk 5 above)
Comment (ricffb) [reply benchexec/tools/starexec_sat_solver.py:1]: It stands for sat tool info. I have no attachements to the name, we could name it to
`starexec_sat_solver`. This would reflect well what is the expected execution format:
 `./starexec_run_default task path/to/proof.out`

File: benchexec/tools/hcad.py (hunk 8 above)
Comment (ricffb) [reply benchexec/tools/hcad.py]: hCad and hKis do not exist outside of the sat competition. There is no repo or website describing them.
The only defined way of running these two solvers is to call them via the `starexec_run_*` script.

These are the only ones in satcomp that come with two different configurations in one archive. The configuration is chosen by either calling `starexec_run_bva` or `starexec_run_pbva`. This is the standard convention for starexec configurations. Unfortunately, that ment that I could not reuse the sat_ti module that works for the rest of the sat solvers, despite these two being just a plain sat solvers.

I believe that creating a more complicated `starexec_sat_solver` tool info module that accomplishes the correct parsing of starexec run scripts is not in our interests: the benchexec workflow is not designed to have the executable name encode the configuration and I think this is a good thing.

When splitting the tool info modules I also need to duplicate the entries in the FM-Tools catalog. We would then declare the pbva and bva configurations of hCad and hKis as two different tools.

You have more experience with this than me. What would your suggestion be?

File: benchexec/tools/hcad.py (hunk 7 above)
Comment (ricffb) [reply benchexec/tools/hcad.py]: I adress a good chunk of it in the comment below (https://github.com/sosy-lab/benchexec/pull/1266#discussion_r3383263810).

My understanding was that the endings in the `starexec_run_*` scripts kind of act like very low level options.
I understand that the way I encoded it clashes with the expectations of a lot of other tool info modules in benchexec and I am open to refactor. (-> see more on this in the other comment)

File: benchexec/tools/veripb.py (hunk 12 above)
Comment (ricffb) [reply benchexec/tools/veripb.py]: There are no property files for sat solvers and I doubt that they will be added in the near future.
Regarding the workflow: I did only use this tool info module in Coveriteam and there it made working with veripb easier. But for the sake of uniformity I would change it to be similar to how the other proof transformers like gratgen and dpr_trim handle the proof (similar approach to witnesses in SV-COMP witness validators).

**Comment by ricffb** [comment]
First: I want to thank you for the thorough review. The files certainly worked for my evaluation but lack some polishing to be more general usable.

Now to the version: For the starexec based solver it is not possible to extract the version from the tool and executable.
The only reasonable thing that I can imagine is that we take the hash of the contents of the tool directory.
The tools themselves offer no standard interface to extract a version (e.g. the winning long established tool kissat is even specifically compiled without any option or version flags for the competition).
The only validator that reveals its version in some way is VeriPB. I extract the version there now from the help message. Version 3 of VeriPB uses a rust library printing the verision in a different manner. I use the name of the executable to distinguish.

File: benchexec/tools/cake_lpr.py (hunk 11 above)
Comment (ricffb) [reply benchexec/tools/cake_lpr.py]: After the refactor I removed it. I thought I might use the info later on in my reproduction, but I actually never did.

File: benchexec/tools/gratchk.py (hunk 9 above)
Comment (ricffb) [reply benchexec/tools/gratchk.py]: I hope the new comment above explains a little better what happens.
So, I want to be able to provide two arguments to gratchk, similar to when, e.g., cpachecker does witness validation.

The first argument is either "sat" or "unset" this tells grdtchk whether to solve for satisfiability or unsatisfiability in the proof file. The second argument is the actual proof to check. The `task.single_inputfile` is used to transport the actual DIMACS sat task.

E.g., the two versions in fm-tools look like this: 
```yaml
versions:
  - version: "1.3.1-sat"
    doi: 10.5281/zenodo.19881683
    benchexec_toolinfo_options: ["sat", "${certificate}"]
    required_ubuntu_packages: []
    base_container_images:
      - docker.io/ubuntu:22.04
  - version: "1.3.1-unsat"
    doi: 10.5281/zenodo.19881684
    benchexec_toolinfo_options: ["unsat", "${certificate}"]
    required_ubuntu_packages: []
    base_container_images:
      - docker.io/ubuntu:22.04
```

Now the crux: gratchk arguments must follow the exact order
`gratchk ( sat | unsat ) task.single_input certificate_file` .
A duplicate appearance of "sat" or "unsat" in the command line would lead to an error.
This is why I use this ugly bit of code to plug the two options apart and insert the task.single_input_file in between.

benchexec/tools/cake_lpr.py
PhilippWendler: I think specifying environment variables in BenchExec is necessary (only) for the BenchCloud.

benchexec/tools/starexec_sat_solver.py
PhilippWendler: Yes, `ti` in the tool-info module's name is not a common suffix. `starexec_sat_solver` is much better, or something with SAT Competition. We can also uses dashes instead of underscores.

benchexec/tools/veripb.py
PhilippWendler: I see. In this case we can keep the existing code and add support for current versions of the tool when it is actually needed.

PhilippWendler:
> First: I want to thank you for the thorough review. The files certainly worked for my evaluation but lack some polishing to be more general usable.
> 
> Now to the version: For the starexec based solver it is not possible to extract the version from the tool and executable. The only reasonable thing that I can imagine is that we take the hash of the contents of the tool directory.

I see. If the tool has no version that we cannot do anything here. Without further knowledge BenchExec has no information on what files belong to the tool, and we consider this out of scope. Also, inventing a version format for a third-party tool is not a good idea.

benchexec/tools/hcad.py
PhilippWendler: From a BenchExec point of view, the answer is that different executables should have different tool-info modules (if it is not an "only one of the executables is present" situation as for tool versions with different executable names). I would also think that most people also consider this as different tools.

The sad thing is that likely the developers did not even want to write two tools, just one tool with two different configurations, but StarExec forced them to have two entry points, right?

Now, in principle the following would be nice: Given that StarExec entry points have no arguments IIRC, there is no confusion if we invent a command-line argument for choosing between the available entry points: a tool-info module for StarExec in general where the suffix after `starexec_run_` is chosen by an `<option>`. But this doesn't work currently.

So I would say that still two tool-info modules are the best what we can currently do. Having one that chooses one executable in `executable()` and then attempts to replace it later on in a way that was not intended by BenchExec seems risky.

benchexec/tools/gratchk.py
PhilippWendler: I see. I would tend to prefer that we do not pick `sat` and `unsat` from arbitrary positions in the options, though. It seems less risky to only use `[executable, options[0], task.single_input_file] + options[1:]` if `options[0]` exists and is one of `sat` or `unsat`.

And we don't want to hard-code a default.

benchexec/tools/dpr-trim.py
```diff
@@ -0,0 +1,24 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
PhilippWendler, line 1: We can use `dpr-trim` as module name if that is what the tool is called.

benchexec/tools/sat_proof_checker_base.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Base class for SATnproof checker tools that use the standardized output format
+    from the SAT Competition: a line starting with "s VERIFIED" on success or
+    "s ERROR" on failure.
+    """
+
+    def determine_result(self, run):
+        for line in reversed(run.output):
+            if line.startswith("s "):
+                verdict = line.strip().split(" ", 1)[1].strip().upper()
```
PhilippWendler: The splitting seems like a quite complicated way of `removeprefix("s ")`, isn't it?
`line.removeprefix("s ").strip().upper()` should be enough.

benchexec/tools/sat_proof_checker_base.py
```diff
@@ -0,0 +1,32 @@
...
+                if verdict.startswith("VERIFIED"):
```
PhilippWendler, line 23: Is `startswith` a good idea here? Are tools supposed to add something arbitrary after the verdict that is safe to ignore?

Same for the solvers' module.

benchexec/tools/starexec_sat_solver.py
```diff
@@ -0,0 +1,50 @@
...
+    def working_directory(self, executable):
+        return os.path.dirname(executable)
+
+    def cmdline(self, executable, options, task, rlimits):
+        return [executable, task.single_input_file, *options]
+
+    def determine_result(self, run):
+        for line in reversed(run.output):
+            if line.startswith("s "):
+                verdict = line.strip().partition(" ")[-1].strip().upper()
```
PhilippWendler: Yet another way of writing `removeprefix("s ")`?

benchexec/tools/veripb.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def version(self, executable):
+        if os.path.basename(executable) == "veripb_bin_static":
+            # Old Python binary prints "Running VeriPB version X.X.X" on every run
+            return self._version_from_tool(
+                executable, arg="-h", line_prefix="Running VeriPB version "
+            )
+        # New Rust binary (v3+) uses clap; --version prints "veripb X.X.X".
+        version = self._version_from_tool(executable)
+        return version.removeprefix("veripb ").strip()
```
PhilippWendler: Wouldn't `line_prefix` be an option here as well, like above?

PhilippWendler (commented):
Thanks for the addressing the comments! I responded to most threads, but GitHub hides them by default because they are resolved.

Only minor stuff left now.

benchexec/tools/gratchk.py
ricffb: That is a very good suggestion. I will implement it like that

benchexec/tools/dpr-trim.py
ricffb: Yes, we can :)

benchexec/tools/veripb.py
ricffb: Yes. This slop slipped my observation.

benchexec/tools/starexec_sat_solver.py
ricffb: Actually was not aware that this function exists. But it is exactly what I want to achieve

benchexec/tools/sat_proof_checker_base.py
ricffb: Here I remain unsure. There are at least two outcomes: VERIFIED SAT and VERIFIED UNSAT .

benchexec/tools/hcad.py
ricffb: I have created the hCad and hKis tool info modules as 4 separate "tools". It is not much new code, since they just all inherit from the starexec solver module.

benchexec/tools/starexec_sat_solver.py
PhilippWendler: I learned from your code in this PR that this function exists :-) (it was used in the initial `veripb.py`).

PR #1270, opened 2026-06-22
Migrate to YAML input format for sending benchmarking data to BenchCloud

AkshayWarrier:
The PR migrates the format for sending benchmarking data to Benchcloud from the legacy text form to YAML. It also allows users to pass multiple result file patterns, since the new format now supports it.

Must be merged after https://gitlab.com/sosy-lab/software/benchcloud/-/merge_requests/183 is deployed.


PhilippWendler:
Would it be possible to avoid making all paths absolute? I consider this an anti-pattern. Absolute paths may help to avoid path misunderstandings between components, if they are applied everywhere and consistently. But if we have a consistent path handling everywhere, then relative paths with a properly defined base¹ work just as well, so I do not see a real benefit of actual paths. It feels more like "it helped once so now we use absolute paths everywhere". Disadvantages of absolute paths are that eventually they inevitably end up somewhere where they are inconvenient or undesired, for example in logs or result files, and disturb comparisons when debugging, leak private data (user name) on the web, etc. BenchExec for example takes care that it does not introduce absolute paths into output when the user did not use absolute paths as input, and that it never forces the user to use absolute paths as input. I hope BenchCloud does the same.

---
¹ Standard is that relative paths that are given as command-line arguments or on stdin/stdout are relative to the current working directory, and relative paths given in a file are relative to that file's directory, as long as the involved APIs do not specify something explicitly.

AkshayWarrier:
Hi Philipp, so in BenchCloud, all run specific files, and the toolpaths are relativized against the basedir. And I think to relativize them, they both need to be the same kind of path; maybe that's why they were both made absolute. But maybe we can make all these paths absolute on the BenchCloud client side before relativizing them against the base directory?
 
An exception is that the output directory is sent as it is to BenchCloud and is not relativized, so maybe we also just make this absolute on the BenchCloud client side then?

PhilippWendler:
> But maybe we can make all these paths absolute on the BenchCloud client side before relativizing them against the base directory?

If you require absolute paths there, why not?

An alternative would also be to define for this format that all paths in the file are relative to the given base dir. Though if there is the output directory mixed in, this might not make that much sense.

But maybe it is also a good idea to check whether the grouping of inputs is meaningful? E.g., why is the output directory mixed in with the things that define the runs? IIRC there are some things that are passed to the BenchmarkClient as command-line options and some are passed in this file. How is it decided which is where so far?

AkshayWarrier:
> How is it decided which is where so far?

I'm not sure how it has been decided so far, but it might make sense to pass the output-dir as a command-line option, separate from the workload. I guess this would also require me to update the legacy format as such.

> An alternative would also be to define for this format that all paths in the file are relative to the given base dir.

Do you mean I enforce benchexec to always send all files relative to the base dir (are they already relative?), and on benchcloud assert all paths are relative to base-dir?

AkshayWarrier:
> I guess this would also require me to update the legacy format as such.

Oh, actually, for backward compatibility, the legacy format would accept the output directory from the input file, but starting with the new format, we would have to expect the output directory to be passed by command-line option. Right? 🤔

PhilippWendler:
> > An alternative would also be to define for this format that all paths in the file are relative to the given base dir.
> 
> Do you mean I enforce benchexec to always send all files relative to the base dir (are they already relative?), and on benchcloud assert all paths are relative to base-dir?

Would be a possibility. In BenchExec the paths are not relative to the base dir, relative paths are relative to the current working directory.

> Oh, actually, for backward compatibility, the legacy format would accept the output directory from the input file, but starting with the new format, we would have to expect the output directory to be passed by command-line option.

Yes. At least, it is a possibility, not saying you have to do it. I have no overview of which information is passed to the BenchCloud and how/where it is done, so I also do not know what is the best way to pass the output directory. I am just saying it would be good to have a look at all this information and decide where it fits best when one is anyway completely redesigning the input format. Note that in the past there was no clear design of this, information was just put somewhere where it fit.

AkshayWarrier:
Hi Philipp, I've made new changes after our discussion. We decided to pass ``output-dir`` through the command-line just for the YAML input format. And all paths passed through the format must be relative, so there's no danger of leaking absolute paths.

The ``basedir`` is relative to the current working directory, the ``logfile`` is relative to the ``output-dir``, and everything else is relative to ``basedir``.

The code can be reviewed again :)

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -358,6 +302,50 @@ def getToolDataForCloud(benchmark):
...
+    numberOfRuns = 0
+    for runSet in activeRunSets(benchmark):
+        for run in runSet.runs:
+            if os.path.exists(run.identifier):
+                absSourceFiles.extend(map(os.path.abspath, run.sourcefiles))
+            numberOfRuns += 1
+
+    absBaseDir = benchexec.util.common_base_dir(absSourceFiles + absToolpaths)
+    if absBaseDir == "":
+        sys.exit("No common base dir found.")
```
PhilippWendler: I think a user would not understand this message, wouldn't they?

When would base dir even be empty?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -358,6 +302,50 @@ def getToolDataForCloud(benchmark):
...
 
 
+def computeBaseDir(benchmark, absToolpaths):
+    absSourceFiles = []
+    numberOfRuns = 0
+    for runSet in activeRunSets(benchmark):
+        for run in runSet.runs:
+            if os.path.exists(run.identifier):
+                absSourceFiles.extend(map(os.path.abspath, run.sourcefiles))
+            numberOfRuns += 1
```
PhilippWendler: I would say the calculation of `numberofRuns` should be removed from this method. It is not really related to `computeBaseDir`, and trivial especially using an expression like `sum(len(runSet.runs) for ...)`.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
-    sourceFiles = []
-    runDefinitions = []
-    for runSet in benchmark.run_sets:
-        if not runSet.should_be_executed():
-            continue
-        if STOPPED_BY_INTERRUPT:
-            break
+    # get limits
+    rlimits = benchmark.rlimits
+    timeLimit = int(rlimits.cputime_hard)
```
PhilippWendler: Limits should be ints anyway. But even if we encounter a limit that is not an int, we would not want to silently truncate the value and use a wrong value! So please remove all the `int()` calls.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
+    if rlimits.cpu_cores is not None:
+        limits_cpu["cores"] = rlimits.cpu_cores
 
-            # we assume, that VCloud-client only splits its input at tabs,
-            # so we can use all other chars for the info, that is needed to run the tool.
-            argString = json.dumps(cmdline)
-            assert "\t" not in argString  # cannot call toTabList(), if there is a tab
+    limits = {"cpu": limits_cpu}
+    if rlimits.walltime is not None:
+        limits["walltime"] = int(rlimits.walltime)
```
PhilippWendler: Is it really correct that the time limits look like this in the yaml?
```
limits:
  cpu:
    time:
      hard: 1
  walltime: 2
```
Looks quite inconsistent.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
+    cloud_input["corereq"] = (
+        r.cpu_cores if r.cpu_cores is not None else DEFAULT_CLOUD_CPUCORE_REQUIREMENT
+    )
+    if r.cpu_model:
+        cloud_input["cpumodels"] = [r.cpu_model]
+
+    if benchmark.result_files_patterns:
+        cloud_input["resultFilePatterns"] = list(benchmark.result_files_patterns)
+
+    cloud_input["runsets"] = runDefinitions
```
PhilippWendler: Does the BenchCloud distinguish between different run sets in a single submission?

AFAIU there is just a list of runs per run collection, and all have the same settings such as limits. So it seems redundant to split the input into different run sets and submit the limits for each of them.

Or is the intent to change this? Then we should discuss how to best map the BenchExec structures to what the BenchCloud wants to support.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -358,6 +302,50 @@ def getToolDataForCloud(benchmark):
...
+        for run in runSet.runs:
+            cmdline = list(map(vcloudutil.force_linux_path, run.cmdline()))
+            log_file = os.path.relpath(run.log_file, benchmark.log_folder)
+            run_def = {"logfile": log_file, "command": cmdline}
+            run_files = []
+            if os.path.exists(run.identifier):
+                run_files.extend(run.sourcefiles)
+            run_files.extend(run.required_files)
+            if run_files:
+                run_def["files"] = [os.path.relpath(f, absBaseDir) for f in run_files]
```
PhilippWendler, line 325: Does this mean that every required file is added to the list for every tasks for which it is required?

IIRC we had problems in the past when the metadata for runs got too big for the BenchCloud's master's memory and we had to change something such that common required files are added only once to the metadata (on the level of the run collection or so), at least if the user was clever enough to specify them on `<benchmark>` or `<rundefinition>` tag and not the `<tasks>` tag in the benchmark definition.

We should make sure not to make this worse now.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -358,6 +302,50 @@ def getToolDataForCloud(benchmark):
     return (workingDir, validToolpaths)
 
 
+def computeBaseDir(benchmark, absToolpaths):
```
PhilippWendler, line 297: Wouldn't this need to take all files into account that are part of the input, e.g., also the required files?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
+        r.memory if r.memory is not None else DEFAULT_CLOUD_MEMORY_REQUIREMENT
+    )
+    cloud_input["corereq"] = (
+        r.cpu_cores if r.cpu_cores is not None else DEFAULT_CLOUD_CPUCORE_REQUIREMENT
+    )
+    if r.cpu_model:
+        cloud_input["cpumodels"] = [r.cpu_model]
+
+    if benchmark.result_files_patterns:
+        cloud_input["resultFilePatterns"] = list(benchmark.result_files_patterns)
```
PhilippWendler: Maybe here it should be explained what the special case for relative patterns is compared to the other relative paths.

And doesn't the BenchCloud forbid absolute patterns? Is this handled somewhere else?

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
+        cloud_input["priority"] = benchmark.config.cloudPriority
+
+    cloud_input["memoryreq"] = (
+        r.memory if r.memory is not None else DEFAULT_CLOUD_MEMORY_REQUIREMENT
+    )
+    cloud_input["corereq"] = (
+        r.cpu_cores if r.cpu_cores is not None else DEFAULT_CLOUD_CPUCORE_REQUIREMENT
+    )
+    if r.cpu_model:
+        cloud_input["cpumodels"] = [r.cpu_model]
```
PhilippWendler: I find it inconsistent that limits are specified in some nested yaml structure, but requirements are flattened into top-level keys, some of which have the name suffix `req` and one of which doesn't.

Furthermore, core limits are given as `limits: cpu: cores:` and core requirements are given as `corereq:`, which is inconsistent both with respect to plural/singular and with respect to whether "CPU" is an explicit part of the naming.

PhilippWendler: Why is `cpumodels` a list and what are the semantics if there is more than one entry?

Also note that `r.cpu_model` can already be a list of CPU models (e.g., 'i9-9900, i7-6700'), so we can end up with a list of list of CPU models, which could be confusing and error prone.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
     return ";".join(f"{k}={v}" for k, v in environment.get("newEnv", {}).items())
 
 
-def toTabList(items):
-    return "\t".join(map(str, items))
-
-
 def getCloudInput(benchmark):
```
PhilippWendler, line 216: We need a link here that points to a place where the format for the cloud-input file is described.

PhilippWendler (commented):
Could you please provide a benchmark definition that exercises all relevant features (runs without input files, with more than one input file, required files specified on all levels of the XML, all limits and requirements, result-file patterns, etc.) and both the resulting cloud input in the old and new format? I think this would greatly help to see that there is nothing forgotten in the new format and that everything is consistent.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -358,6 +302,50 @@ def getToolDataForCloud(benchmark):
...
+            runs.append(run_def)
+        if runs:
+            runDefinitions.append({"limits": limits, "runs": runs})
+    return runDefinitions
+
+
+def activeRunSets(benchmark):
+    for runSet in benchmark.run_sets:
+        if STOPPED_BY_INTERRUPT:
+            break
```
PhilippWendler: This is not useful in a tight loop. We need to check `STOPPED_BY_INTERRUPT` often enough that we catch Ctrl+C "soon enough", but not in this very short loop. Maybe in bigger loops or after larger steps of the preparation.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: I discussed with the team, and I agree that it looks inconsistent in the current state. We agreed to update it to maybe something like this:
```yaml
formatVersion: "1.0"
files: ["tool/path"]
basedir: "transfer/files/dir"
execdir: "tool/execution/dir"
priority: "HIGH"
resultFilePatterns: ["*.log", "*.txt"]
requirements:
    cores: 1
    memory: 4000000000
    cpumodels:
      - "AMD"
      - "Intel"
limits:
    cores: 1
    memory: 500000000
    walltime: 10
    hardtime: 10
runs:
  - logfile: "outputFile"
    command: ["command1", "param", "param", "file"]
    files: ["file1", "file2"]
  - logfile: "outputFile"
    command: ["command2", "param", "param", "file"]
    files: ["file1", "file2"]
```

What do you think?

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: > Or is the intent to change this?

Yes, it was leftover remnants from the last YAML format, which we initially decided to keep in case we supported this, but we currently don't, so I would just transmit a flat list of runs as I described in the format below:

```yaml
runs:
  - logfile: "outputFile"
    command: ["command1", "param", "param", "file"]
    files: ["file1", "file2"]
  - logfile: "outputFile"
    command: ["command2", "param", "param", "file"]
    files: ["file1", "file2"]
```

But benchexec still supports multiple runsets, right? How does it handle them, and how would it like it to be handled for BenchCloud? Should all runsets be part of one run collection, or only allow one run-definition when submitting to BenchCloud?

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: Are you suggesting finding files common to all ``run.required_files`` and transmitting them in the top-level ``files`` field?

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: I guess it should, earlier it didn't

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: > Maybe here it should be explained what the special case for relative patterns is compared to the other relative paths.

Do you mean to say to explain that it is relative to the output directory and not the basedir?

> And doesn't the BenchCloud forbid absolute patterns? Is this handled somewhere else?

It just checks if it is a valid glob pattern using ``PathMatcher``. Is that not good enough?

Edit: I checked again and found out yes, it also prevents absolute result file patterns :)

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: > Why is cpumodels a list and what are the semantics if there is more than one entry?

If the user specifies multiple, the scheduler checks the cpu models in the order specified to see if any of them would allow a run collection to run on a particular processing unit.

> Also note that r.cpu_model can already be a list of CPU models (e.g., 'i9-9900, i7-6700'), so we can end up with a list of list of CPU models, which could be confusing and error prone.

I think I assumed that benchexec only let users pass a single cpu model, but we can explicitly mention that multiple cpu models can be passed as comma-separated values. (And I can update the code to split the input and pass it)

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: The updated format as described below would fix it: https://github.com/sosy-lab/benchexec/pull/1270#discussion_r3613723833

AkshayWarrier:
> Could you please provide a benchmark definition that exercises all relevant features (runs without input files, with more than one input file, required files specified on all levels of the XML, all limits and requirements, result-file patterns, etc.) and both the resulting cloud input in the old and new format? I think this would greatly help to see that there is nothing forgotten in the new format and that everything is consistent.

Do you mean to add it as integration tests?

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: Sure.

Btw.: I already commented [here](https://gitlab.com/sosy-lab/software/benchcloud/-/work_items/220#note_646565844) with the suggestion that key names could contain units. Did you check this thread for suggestions? I guess we should continue this discussion either in the BenchCloud MR or issue.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: So far the integration of the BenchmarkClient puts all runs from the different run sets into a single BenchCloud run collection. (For the webclient, every run set gets its own run collection.) For me, this can be kept. It is more a question for the BenchCloud whether there are things that it needs to be uniform across a run collection but which could differ between run sets.

I also don't see a need to support more than one run collection.

But check the old discussions in the BenchCloud project, maybe there was a reason.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: I guess the code does not need to be more clever than the existing code (of which I do not know exactly how it handles this situation). But I wanted to raise the question whether the new code creates more inefficient run metadata with more duplication than before, because this should be avoided.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: > > Maybe here it should be explained what the special case for relative patterns is compared to the other relative paths.
> 
> Do you mean to say to explain that it is relative to the output directory and not the basedir?

The relative pattern should be relative to the directory where the run executed. This is definitively not the basedir. Not sure what you mean with "output directory".

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: > > Why is cpumodels a list and what are the semantics if there is more than one entry?
> 
> If the user specifies multiple, the scheduler checks the cpu models in the order specified to see if any of them would allow a run collection to run on a particular processing unit.

Is this a new feature? Do we have a use case for it?

> I think I assumed that benchexec only let users pass a single cpu model, but we can explicitly mention that multiple cpu models can be passed as comma-separated values. (And I can update the code to split the input and pass it)

The existing behavior of what happens when one specifies `i9-9900, i7-6700` as CPU model is important, though, and any changes here must not result in a change of behavior.

I think what needs to be done is to get a complete overview of what the BenchCloud supports as combinations of CPU models, and then design a consistent and understandable Yaml structure for it.

PhilippWendler:
> > Could you please provide a benchmark definition that exercises all relevant features (runs without input files, with more than one input file, required files specified on all levels of the XML, all limits and requirements, result-file patterns, etc.) and both the resulting cloud input in the old and new format? I think this would greatly help to see that there is nothing forgotten in the new format and that everything is consistent.
> 
> Do you mean to add it as integration tests?

A test for this would of course also be great.

I would think it does not make sense as an integration test that requires a BenchCloud instance to execute. But we can have a test that tests that the Yaml structure is as expected for a given benchmark definition.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: I kept it because it existed before. I've removed it now because the base dir should never be empty.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: > But check the old discussions in the BenchCloud project, maybe there was a reason.

I checked, and this was really only part of the format on the parser side, introduced from an old unreviewed merge request. I decided to stick to a flat list of runs as mentioned earlier.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: It considers required files now as well.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: Yes that my was bad, you're right. I've added a comment to mention it.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: I've added units now as well. We can continue the discussion under the MR in case you have other questions about the current format :)

AkshayWarrier: > Is this a new feature? Do we have a use case for it?

I checked, and it looks like it always existed; it was just undocumented. The legacy format parsed CPU models on the benchmark client side, expecting a string with comma-separated values.

And also, to be precise about the behaviour, the CPU model is only decided once for the run collection when it is initially added to the scheduler, where it assigns the first one from the list that is currently usable. (or left empty which allows it to run anywhere)

I think a use case might be if the user perhaps only wants to select CPU models that only have 8 cores. But I see that the use case is very niche. Generally, users can just leave the list empty if they are fine executing the run.

> The existing behavior of what happens when one specifies i9-9900, i7-6700 as CPU model is important, though, and any changes here must not result in a change of behavior.

Does benchexc already use multiple CPU models for something?

AkshayWarrier:
> > > Could you please provide a benchmark definition that exercises all relevant features (runs without input files, with more than one input file, required files specified on all levels of the XML, all limits and requirements, result-file patterns, etc.) and both the resulting cloud input in the old and new format? I think this would greatly help to see that there is nothing forgotten in the new format and that everything is consistent.
> > 
> > 
> > Do you mean to add it as integration tests?
> 
> A test for this would of course also be great.
> 
> I would think it does not make sense as an integration test that requires a BenchCloud instance to execute. But we can have a test that tests that the Yaml structure is as expected for a given benchmark definition.

Okay, I'll add a test once the format is finally decided on :)

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: I removed it from ``getCloudInput`` entirely because the variable was only set after ``KeyboardInterrupt`` was caught outside it anyway.

contrib/vcloud/benchmarkclient_executor.py
```diff
@@ -212,122 +215,63 @@ def formatEnvironment(environment):
...
-    # get Runs with args and sourcefiles
-    sourceFiles = []
-    runDefinitions = []
-    for runSet in benchmark.run_sets:
-        if not runSet.should_be_executed():
-            continue
-        if STOPPED_BY_INTERRUPT:
-            break
+    rlimits = benchmark.rlimits
+    limits = {"hardtime_s": rlimits.cputime_hard}
```
PhilippWendler: A CPU-time limit is optional for BenchExec. So if it is required for the BenchCloud, a missing limit must be handled gracefully.

But wasn't the plan that only either CPU time or wall time are required?

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: So the "list of CPU models" here translates to the feature https://gitlab.com/sosy-lab/software/benchcloud/-/work_items/113?

This is the feature that we regularly use, and it must remain supported. But in the BenchCloud there have been errors already were suddenly a list of CPU models was interpreted differently, so I want to make sure.

Does the WebClient also expect a CPU model with a comma-separated list that is interpreted as fallback? I guess so. If yes, would it maybe make sense to centralize the parsing and interpretation of the string given by the user as CPU model in a single place in the Java code of the BenchCloud? With this MR, there would be at least two places where the CPU-model string is parsed and that would need to be updated if the format is extended.

> Does benchexc already use multiple CPU models for something?

BenchExec does not care about CPU models at all, it just allows passing them through to the BenchCloud.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: > Edit: I checked again and found out yes, it also prevents absolute result file patterns :)

Is this handled gracefully? Or do we need to check here and provide an error message?

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: Done.

> But wasn't the plan that only either CPU time or wall time are required?

It looks like cputime was never optional, and walltime always was.

contrib/vcloud/benchmarkclient_executor.py
AkshayWarrier: > Is this handled gracefully? Or do we need to check here and provide an error message?

Benchcloud throws an error to the user with the offending patterns, that's enough right?

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: What was optional in the past is not really relevant, right? What does the BenchCloud currently require or should require?

AkshayWarrier: I asked Max, and he said there was never a plan for this. It seems reasonable to need only one of them as required, but I would not put it in the scope of the YAML format MR.

PhilippWendler: The check for existence of a CPU-time limit already exists here: https://github.com/AkshayWarrier/benchexec/blob/d395082322cd7b265de3ee9edf694a0ea532b008/contrib/vcloud/benchmarkclient_executor.py#L43

So the conversion code can simply use it with a comment like "existence guaranteed above" or so.

PhilippWendler:
Thanks for your continued work on this. I have no further comments, I guess it can be finished and tests can be added.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+        self.assertNotIn("memory_b", cloud_input["limits"])
+        self.assertNotIn("cores", cloud_input["limits"])
+        self.assertNotIn("priority", cloud_input)
+        self.assertIsNone(cloud_input["requirements"]["cores"])
+        self.assertIsNone(cloud_input["requirements"]["memory_b"])
+        self.assertNotIn("cpumodels", cloud_input["requirements"])
+        self.assertEqual(cloud_input["resultFilePatterns"], ["."])
+
+        self.assertEqual(n_runs, 1)
+        self.assertEqual(len(cloud_input["runs"]), 1)
```
PhilippWendler, line 141: How about replacing these with the full datastructure of expected values, and comparing `cloud_input` to it? I guess that would be easier to see and nothing could be forgotten.

Likely it would be even nicer to store input XML and expected YAML as actual files in the repository, instead of having them in strings. Then you can easily view and edit them with syntax highlighting and auto-formatting, and one can even use the benchmark definitions as inputs for manual tests.

We use this structure already for example for the integration tests of `benchexec` itself (`benchexec/test_integration/`) and `table-generator`. Maybe the tests here can even share some benchmark definitions with `benchexec`'s integration tests.

In `table-generator`'s integration tests we even have a nice feature where we can use the test to write the expected-value files instead of comparing them against it. This makes it trivial to update the tests when there is a deliberate change.

And I would recommend to take the same Yaml files and reuse them as tests for the BenchmarkClient in the BenchCloud repo as well.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+class TestCloudInput(unittest.TestCase):
+    """
+    Tests for benchmarkclient_executor.getCloudInput() to check that the runs and required files generated for
+    the cloud are correct.
+    """
+
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
+        self.tool_file = os.path.join(TEST_TASKS_DIR, "test.sh")
+        open(self.tool_file, "w").close()
```
PhilippWendler, line 78: Please don't write to files in the repository directory. It is always a hassle if such files are left over after a canceled test run, and using temporary files with hard-coded names is also a problem for running tests in parallel.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+
+class TestInit(unittest.TestCase):
+    """Tests for benchmarkclient_executor.init(), which validates
+    the configuration before a cloud run is started."""
+
+    def test_missing_cputime_hard_limit_exits(self):
+        config = MagicMock(reprocessResults=False)
+        benchmark = MagicMock()
+        benchmark.rlimits.cputime_hard = None
+        with self.assertRaises(SystemExit):
```
PhilippWendler, line 55: I think it would be good to check for some mention of the time limit in the exception message. Otherwise the mock might cause some other error and it would not be noticed.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+            benchmarkclient_executor.init(config, benchmark)
+
+    def test_unsupported_environment_configuration_exits(self):
+        config = MagicMock(
+            reprocessResults=False, containerImage=None, tool_directory=None
+        )
+        benchmark = MagicMock()
+        benchmark.rlimits.cputime_hard = 30
+        benchmark.environment.return_value = {"keepEnv": {"PATH": "/usr/bin"}}
+        with self.assertRaises(SystemExit):
```
PhilippWendler, line 65: Same as above: please check for some relevant part of the exception message.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+                benchmark = Benchmark(xml_path, config, util.read_local_time())
+            finally:
+                os.remove(xml_path)
+        benchmark.executable = self.tool_file
+        return benchmark
+
+    def _get_cloud_input(self, xml_content, config=DEFAULT_CONFIG, tool=None):
+        benchmark = self._parse_benchmark(xml_content, config, tool=tool)
+        yaml_str, n_runs = benchmarkclient_executor.getCloudInput(benchmark)
+        return yaml.safe_load(yaml_str), n_runs
```
PhilippWendler, line 116: Maybe the fact that we have to parse here is an indication that `getCloudInput()` is not so well designed and we should decouple creating the data structure from turning it into yaml. `getCloudInput()` could simply return the data structure and the caller would turn it into yaml. Then also the weird tuple as return value would be not necessary.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2026 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import tempfile
+import unittest
+from dataclasses import dataclass, field, replace
```
PhilippWendler: `replace` is a quite generic name and this utility is not well known. So where it is used it is not obvious what `replace(DEFAULT_CONFIG, additional_files=[extra_file])` does or where it comes from. Thus please avoid importing `replace`.

For `field` it is not as bad as to where it is used, but still I see little reason to import it and clutter the namespace with it given that it is used only once.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+        tool = self._make_mock_tool(working_directory="/no/such/directory")
+        with self.assertRaises(SystemExit):
+            self._get_cloud_input(
+                """
+                <benchmark tool="dummy" hardtimelimit="30">
+                  <rundefinition>
+                    <tasks><withoutfile>task1</withoutfile></tasks>
+                  </rundefinition>
+                </benchmark>
+                """,
```
PhilippWendler, line 190: You seem to reuse this several times, so please do not copy it but share it.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+        self.assertIsNone(cloud_input["requirements"]["memory_b"])
+        self.assertNotIn("cpumodels", cloud_input["requirements"])
+        self.assertEqual(cloud_input["resultFilePatterns"], ["."])
+
+        self.assertEqual(n_runs, 1)
+        self.assertEqual(len(cloud_input["runs"]), 1)
+
+    def test_input_files_and_dirs(self):
+        extra_file = os.path.join(TEST_TASKS_DIR, "other.prp")
+        config = replace(DEFAULT_CONFIG, additional_files=[extra_file])
```
PhilippWendler, line 145: Why use `replace()` instead of `VCloudConfig(additional_files=[extra_file])`? I find the latter clearer.

`DEFAULT_CONFIG` probably does not even provide value compared to using `VCloudConfig()`, so it could be removed completely.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+        # one task with no inputs, no result files, only cputime_hard set
+        cloud_input, n_runs = self._get_cloud_input("""
+            <benchmark tool="dummy" hardtimelimit="30">
+              <rundefinition>
+                <tasks><withoutfile>task1</withoutfile></tasks>
+              </rundefinition>
+            </benchmark>
+        """)
+        self.assertEqual(cloud_input["formatVersion"], "1.0")
+        self.assertCountEqual(cloud_input["files"], ["test.sh"])
```
PhilippWendler, line 128: While it is clear that the order of files should not influence the benchmarking, we want all our tools to behave deterministically as far as possible. This is a huge help for debugging/bisecting, e.g., when comparing the output of a tool (like `cloudInput.yml`) across different runs.

So we should aim to make `cloudInput.yml` fully stable across runs, even for order of list elements. And then we can make use of this in the tests as well and at the same time use the tests to detect such non-determinism. So please avoid using `assertCountEqual`.

contrib/vcloud/test_benchmarkclient_executor.py
```diff
@@ -0,0 +1,355 @@
...
+    def test_minimal(self):
+        # one task with no inputs, no result files, only cputime_hard set
+        cloud_input, n_runs = self._get_cloud_input("""
+            <benchmark tool="dummy" hardtimelimit="30">
+              <rundefinition>
+                <tasks><withoutfile>task1</withoutfile></tasks>
+              </rundefinition>
+            </benchmark>
+        """)
+        self.assertEqual(cloud_input["formatVersion"], "1.0")
```
PhilippWendler, line 127: It is ok to use it, but note that it is not necessary in BenchExec to use the `unittest` utilities like `self.assert*`. We use pytest as testing framework and thus test assertions can also be written with plain `assert`.

PhilippWendler (commented):
Thanks for the tests! This certainly goes into the right direction, my main question is about how to store input and expected values.

PhilippWendler:
I only now noticed that you submitted this PR from outside of the repository. This means that the tests can not get executed in CI. This is unfortunate because now we have the PR already with lots of history and comments but need to create a new PR for seeing the tests executed. I guess it is best if you finish the current round of comments and then we switch over to a new PR.

Please, for all our repos never work from a personal fork due to CI and other issues. We always want all the work to be done in the official repos.

contrib/vcloud/benchmarkclient_executor.py
PhilippWendler: I stumbled about this when looking at something else, and now I can clarify: There are required files per `Benchmark` instance and per `Run` instance. The former typically contain the tool and should not be listed redundantly for every run in order to save memory. In the old and new code, `Benchmark.required_files()` ends up in `absToolPaths` and is added once to the cloud-input file, so it works as it should.

contrib/vcloud/test_benchmarkclient_executor.py
AkshayWarrier: > Likely it would be even nicer to store input XML and expected YAML as actual files in the repository, instead of having them in strings.

This is a good idea, but how should I handle checking ``basedir`` and ``execdir``, which depend on the current working directory? Do I leave those out of the expected YAML files?

PhilippWendler: You see now why it is not a good idea to use absolute paths. :-)

I guess relativizing the respective paths in the test before checking against the expected value is good.

AkshayWarrier: Done.

> Maybe the tests here can even share some benchmark definitions with benchexec's integration tests.

I decided not to do this because I didn't find a lot in common with the files here, since the goals of the benchexec tests are different. I also wasn't sure about coupling the test files, since someone updating tests for the core functionality of benchexec might have to update the benchcloud executor's test, or vice versa.


> And I would recommend to take the same Yaml files and reuse them as tests for the BenchmarkClient in the BenchCloud repo as well.

I can do this, but we'll then have to keep a copy of these files in BenchCloud and update them when they are updated here.  I guess it is fine, since it won't be often that we will have to update them?

contrib/vcloud/test_benchmarkclient_executor.py
AkshayWarrier: I removed calls to replace, but for:
> DEFAULT_CONFIG probably does not even provide value compared to using VCloudConfig(), so it could be removed completely.

ruff prevents me to do so:
```
B008 Do not perform function call `VCloudConfig` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable
  --> contrib/vcloud/test_benchmarkclient_executor.py:92:51
   |
90 |         return tool
91 |
92 |     def _parse_benchmark(self, input_file, config=VCloudConfig(), tool=None):
   |                                                   ^^^^^^^^^^^^^^
93 |         mock_tool = tool or self._make_mock_tool()
94 |         with patch(
   |

B008 Do not perform function call `VCloudConfig` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable
   --> contrib/vcloud/test_benchmarkclient_executor.py:104:51
    |
102 |         return benchmark
103 |
104 |     def _get_cloud_input(self, input_file, config=VCloudConfig(), tool=None):
    |                                                   ^^^^^^^^^^^^^^
105 |         benchmark = self._parse_benchmark(input_file, config, tool=tool)
106 |         return benchmarkclient_executor.getCloudInput(benchmark)
    |

Found 2 errors.
```

contrib/vcloud/test_benchmarkclient_executor.py
AkshayWarrier: Okay, good to know :) I used it since I saw some other tests in benchexec use it as well.

AkshayWarrier:
> I only now noticed that you submitted this PR from outside of the repository. This means that the tests can not get executed in CI.

Ah, I couldn't understand why these tests weren't already being run. This explains it. Sorry about that; I should've asked you to invite me as a collaborator before I started :)

contrib/vcloud/test_benchmarkclient_executor.py
PhilippWendler: Note: ruff does not prevent you from doing anything, nor does any other static-analysis tool. These tools only show potential problems, and it is up to the developers to decide what to do (suppress, adopt suggestion of tool, do a bigger rewrite) after understanding the trade offs. Tool should not be blamed for code-design decisions, the human always has the responsibility and needs to stand behind their code.

Here the solution `DEFAULT_CONFIG` looks like the best alternative, so we keep it. But it is our decision and not something that a tool forces on us.

contrib/vcloud/test_benchmarkclient_executor.py
PhilippWendler: > I also wasn't sure about coupling the test files, since someone updating tests for the core functionality of benchexec might have to update the benchcloud executor's test, or vice versa.

This would be a good thing, because if someone updates/adds tests, both components profit without more effort.

But if the tests do not match, then we don't need to reuse files.

PhilippWendler (approved):
Looks good!

I pushed the same branch to the GitHub repo, so now the full CI ran, and everything is green. Because we likely do not need further changes, we can keep this and do not need a new PR.

(How) do we need to coordinate the release of this? Can we merge it already or would it break BenchCloud benchmarking for people using the BenchExec repository?

It is planned that the BenchmarkClient will still accept the old format and thus work with for example old CPAchecker checkouts/branches with old BenchExec, right? This would mean that we just merged and release this PR after the currently released and used BenchCloud supports the new input format.

Can you please ping me once this is the case?

AkshayWarrier:
> I pushed the same branch to the GitHub repo, so now the full CI ran, and everything is green. Because we likely do not need further changes, we can keep this and do not need a new PR.

Thanks!

>(How) do we need to coordinate the release of this? Can we merge it already or would it break BenchCloud benchmarking for people using the BenchExec repository?

Yes, I guess it would.

> It is planned that the BenchmarkClient will still accept the old format and thus work with for example old CPAchecker checkouts/branches with old BenchExec, right? This would mean that we just merged and release this PR after the currently released and used BenchCloud supports the new input format.

Yes, I guess we have to wait until the new BenchCloud gets released to merge this. I'll ping you when it is the case. The old input format's parser is untouched and should still work (I quickly tested it on our test system).

PR #1283, opened 2026-07-23
new energy measurement approach using intel-rapl kernel module

ronin145:
This new approach aims to create an alternative to cpu energy meter, directly inside of benchexec.
The intel-rapl kernel module allows to read the total amount of energy consumed by each package/domain from a file 
This is part of my bachelor thesis


benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,92 @@
...
-    for pkg, domains in energy.items():
-        for domain, value in domains.items():
-            if domain == DOMAIN_PACKAGE:
-                cpuenergy += value
-            result[f"cpuenergy-pkg{pkg}-{domain}"] = value
-    result["cpuenergy"] = cpuenergy
+    total = Decimal(0)
+    for package in measurement.packages:
+        p_energy = Decimal(package.energy) / Decimal(1000000)
+        if not package.name.__eq__("psys"):
```
PhilippWendler: Why use `__eq__`?

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,92 @@
...
-    cpuenergy = Decimal(0)
-    for pkg, domains in energy.items():
-        for domain, value in domains.items():
-            if domain == DOMAIN_PACKAGE:
-                cpuenergy += value
-            result[f"cpuenergy-pkg{pkg}-{domain}"] = value
-    result["cpuenergy"] = cpuenergy
+    total = Decimal(0)
+    for package in measurement.packages:
+        p_energy = Decimal(package.energy) / Decimal(1000000)
```
PhilippWendler: I think we should have better unit handling, i.e., clear documentation of what is in what unit and dedicated conversions instead of ad-hoc adjustments like this one.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,92 @@
...
-class EnergyMeasurement(object):
-    def __init__(self, executable):
-        self._executable = executable
-        self._measurement_process = None
+@dataclass
+class Package:
+    name: str
+    path: Path
+    energy: int
+    domains: list[Domain]
```
PhilippWendler: Does it make sense to store the energy both per package and per domain? It seems there is a lot of code that because of this needs to handle both cases. How about we store energy in a flat structure such that a simple loop suffices for reading values?

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,92 @@
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
-            bufsize=10_000,
-            preexec_fn=os.setpgrp,  # Prevent delivery of Ctrl+C to subprocess
-        )
+        for package in self.packages:
+            package.energy = int((package.path/"energy_uj").read_text().strip())
+            for domain in package.domains:
+                domain.energy = int((domain.path/"energy_uj").read_text().strip())
```
PhilippWendler: This code in `start()` is basically duplicated in `stop()`. I would recommend to deduplicate it by moving the reading functionality into a common helper.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,92 @@
...
-DOMAIN_CORE = "core"
-DOMAIN_UNCORE = "uncore"
-DOMAIN_DRAM = "dram"
+rapl_path = Path("/sys/class/powercap/intel-rapl/")
 
+@dataclass
+class Domain:
+    name: str
+    path: Path
+    energy: int
```
PhilippWendler: I am not sure whether the mixing of the structure of the machine, which is created in the beginning and then immutable, and the energy measurements, which are changed over time, is such a good idea. Maybe separate static and changeable data?

PhilippWendler (commented):
The code is hard to read, please run the code formatter ruff.

There should also be more comments, why did you even remove the previously existing comments?

How does the code handle errors like permission problems?

I would recommend to think about how to make the code testable.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
+                string += f"    {domain.name}: {domain.energy.value} uj\n"
+        return string
+
+
+def get_path_content(path):
+    try:
+        content = path.read_text().strip()
+        return content
+    except OSError as error:
+        print(f"cannot read {path}: {error}")
```
PhilippWendler: Will this error handling work? Or wouldn't it lead to crashes when `None` is returned as value?

Additionally, if we want to output a message, we need to use the logging framework and write a message that is understandable and useful for a user of BenchExec (or turn it into a debug message if it is only relevant for developers).

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
+
+
+def update_values(measurement):
+    """this updates the values of the measurement either to the initial values at start or the delta values at stop
+    we can use the same code because the energy values are initialized with 0"""
+    for package in measurement.packages:
+        new_energy = int(get_path_content(package.path / "energy_uj"))
+        if new_energy < package.energy.value:
+            new_energy += int(get_path_content(package.path / "max_energy_range_uj"))
```
PhilippWendler: This captures overflows with exactly one wraparound. But the measurement would be wrong if the measurement interval is long enough that the measured value is larger than twice `max_energy_range_uj`. Can we handle this?

Also, please always consider whether there should be explanatory comments. For example here, it would make the code much quicker understandable if there would be some comment explaining that this is for handling overflows. (I got it only because I saw the commit message before.)

benchexec/intel_cpu_energy.py
PhilippWendler: Even now, the code for reading energy measurements and handling overflows needs to exist twice, which is suboptimal because for example if we find a bug in the overflow handling, we need to remember to fix it twice. Can this be improved?

benchexec/intel_cpu_energy.py
PhilippWendler: The current solution makes some of the fields of `Domain` immutable, but there are still mutable fields (nested) inside it, so conceptually little changed (just with technical enforcement of some immutability). I am not saying that we can not use the current solution, but we could also consider alternatives. How about a map from domain to values?

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
-    result["cpuenergy"] = cpuenergy
+    total = Decimal(0)
+    for package in measurement.packages:
+        p_energy = convert_to_joules(package.energy.value)
+        # psys describes energy usage of the entire system and is therefore not relevant for cpuenergy
+        if not package.name == "psys":
+            total += p_energy
+            result[f"cpuenergy-{package.name}"] = p_energy
+        else:
+            result["psys"] = p_energy
```
PhilippWendler: I think the name `psys` needs to be improved. Consider that for example `runexec` returns lots of measurements like CPU time, memory, etc. Would users have an idea what a `psys` measurement in the middle of these is?

PhilippWendler: This line made me wonder what happens if there is more than one package named `psys`. I think the code should document that package names are unique.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
+
+def get_path_content(path):
+    try:
+        content = path.read_text().strip()
+        return content
+    except OSError as error:
+        print(f"cannot read {path}: {error}")
+
+
+def update_values(measurement):
```
PhilippWendler: Why is this a module-level function and not a method in `EnergyMeasurement`?

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
+        ):
+            p_name = get_path_content(package / "name")
+            domains = []
+            for domain in sorted(
+                d for d in package.glob("intel-rapl:*") if d.name.count(":") == 2
+            ):
+                d_name = get_path_content(domain / "name")
+                domains.append(Domain(d_name, domain, EnergyWrapper(0)))
+
+            self.packages.append(Package(p_name, package, EnergyWrapper(0), domains))
```
PhilippWendler, line 96: It would be nice to have a comment here with an example of how these paths look and what files are expected to exist. This makes it easier to understand that/whether the parsing of file names is correct.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
-        """Returns True if there is currently an instance of the external measurement program running, False otherwise."""
-        return self._measurement_process is not None
-
-
-def format_energy_results(energy):
-    """Take the result of an energy measurement and return a flat dictionary that contains all values."""
-    if not energy:
+        """Stop the measurement if it hasn't been stopped already and calculate difference between end and start values
+        This method has to return self because of the way the old cpu-energy-meter was implemented,
+        changing this would require changing the readout in every other file"""
```
PhilippWendler, line 119: First, I am not sure whether the current code even works as expected because previously `stop()` returned a dictionary and now it returns an `EnergyMeasurement` object, which for example cannot be iterated over.

However anyway, changing outside code is completely fine and should be done if it leads to an improvement. This is an internal API and we can simply change all callers instead of having to live with suboptimal code due to backwards compatibility. So please design the API in the best possible way, irregardless of how it was before, and then adapt the existing code.

benchexec/intel_cpu_energy.py
```diff
@@ -7,97 +7,148 @@
...
+    def __str__(self):
+        string = ""
+        for package in self.packages:
+            string += f"{package.name}: {package.energy.value} uj\n"
+            for domain in package.domains:
+                string += f"    {domain.name}: {domain.energy.value} uj\n"
+        return string
+
+
+def get_path_content(path):
```
PhilippWendler, line 183: Btw.: The `benchxec.util` package already has `read_file` and `try_read_file`.

PhilippWendler (commented):
Please also merge the `main` branch into your branch in order to resolve the merge conflicts and get the latest CI improvements.

ronin145:
I am just now getting errors from ruff check that I didn't get before. Like EnergyMeasurement inheriting from object, which was the case even with the old implementation. I don't know why these errors didn't show up before, and I am not sure whether they should show up.

PhilippWendler:
> I am just now getting errors from ruff check that I didn't get before. Like EnergyMeasurement inheriting from object, which was the case even with the old implementation. I don't know why these errors didn't show up before, and I am not sure whether they should show up.

Yes, this is due the CI improvements on the main branch, which I recommended to merge. Ruff's checks help to get the code clean and readable and also find many common errors. You can fix the warnings manually or run `ruff check` locally, for many cases it even offers an automated fix.

benchexec/intel_cpu_energy.py
```diff
@@ -5,100 +5,200 @@
...
+
+    def update_values(self):
+        """this method is run by a thread to constantly sample energy values for overhead protection"""
+        self.update_all()
+        while not self.stop_event.wait(
+            timeout=1.0
+        ) and not EnergyMeasurement.error_event.wait(timeout=1.0):
+            self.update_all()
+        if EnergyMeasurement.error_event.is_set():
+            print("Energy measurement failed, defaulting to 0")
```
PhilippWendler: This is not acceptable. I explained already that the worst thing that can happen are wrong measurement values. If there is an error in the measurement, then the respective value must not be produced at all instead of being reported with a wrong value.

benchexec/intel_cpu_energy.py
```diff
@@ -5,100 +5,200 @@
...
+        """this updates the counter for total energy consumed across all packages and domains"""
+        for package in self.packages:
+            package.update_value()
+
+    def update_values(self):
+        """this method is run by a thread to constantly sample energy values for overhead protection"""
+        self.update_all()
+        while not self.stop_event.wait(
+            timeout=1.0
+        ) and not EnergyMeasurement.error_event.wait(timeout=1.0):
```
PhilippWendler: Do we need to sample every second? We want to keep the overhead as low as possible.

ronin145: This is just arbitrary number I chose for testing. I wanted to talk this over with you since I don't really have a lot of experience with benchmarking. What  interval would you recommend?

PhilippWendler: The important question is: What interval do we need to correct measurements and handle all overflows?

Then we want to sample as rarely as possible while still often enough to get correct measurements.

ronin145: I implemented a way to calculate a measurement interval from the cpus short term power consumption limit. When testing it came out close to what I had found through testing (for my CPU). I added a default interval of 1000s. This should cover pretty much any CPU, but is unnecessarily often for lower end CPUs. Please let me know what you think.

PhilippWendler:
Can you please clean up the code for a new review and ping me once this is done? Like, remove debug statements, finish the error handling, add tests, etc.? It would also be nice if you could go through the comments again and make sure that you addressed them and respond with a short explanation that it is done (any maybe where/how such that I can easily see it) or why not.

benchexec/intel_cpu_energy.py
PhilippWendler: Interesting, can you please link to documentation that explains these constraints?

At least on some of my machines, the respective files are not there. What are the requirements for CPUs and kernel versions to have them?

Why 1000s? Can you please explain how you derived this value? (in a comment next to it)

It is not only important that the code and the measurements are correct, it is also important that it is possible to see from the code and its documentation that it is correct.

ronin145:
I am not really experienced in writing unit tests, I gave it a try, hopefully it's useable. I assume I can just copy the licensing/copyright informations for the new file?

benchexec/intel_cpu_energy.py
ronin145: This should be fixed with the AreaOfMeasurement class

benchexec/intel_cpu_energy.py
ronin145: I think we talked about this in person. I feel like a mapping would introduce more potentially confusing code, and I don't see a huge benefit. Additionally I think it's good that the code mirrors the file structure of the RAPL Kernel Module.

benchexec/intel_cpu_energy.py
ronin145: This is fixed with the new threaded approach

benchexec/intel_cpu_energy.py
ronin145: I added a comment at the top of the function clarifying that package names are unique

benchexec/intel_cpu_energy.py
ronin145: I need the function to return "0" on error, so I have to write my own additional function anyways. I could use try_read_file inside my function, but then I get a circular error import during testing, so I just left it as is. I get that error because I am running the file directly instead of e.g. runexec from bin/. But this is something that I could easily change at the end if you want me to.

benchexec/intel_cpu_energy.py
ronin145: Now the measurement returns None on Error in the stop() function.

benchexec/intel_cpu_energy.py
PhilippWendler: First, returning `"0"` is bad. On errors we should not return invalid data that is not distinguishable from real data. Even with the error event it is just error prone to return wrong data that could accidentally be used somewhere else.

Second, I don't think you get circular imports, but if you run `python benchexec/intel_cpu_energy.py` then yes, local imports would not work because this is not the correct way to run a Python module. The correct way is `python -m benchexec.intel_cpu_energy` (in a virtual environment where the local checkout of BenchExec was installed with `pip install -e .`). Then the imports should work, we have several utilities inside BenchExec where we do exactly the same (e.g. `benchexec.test_tool_info` and `benchexec.check_cgroups`).

benchexec/intel_cpu_energy.py
PhilippWendler: How come that 250W lead to 1000s in all cases? Doesn't this depend on the value in `max_energy_range_uj` as well?

And why 250W as worst case? How did you get this value?

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TestEnergyMeasurement(unittest.TestCase):
+    def test_run(self):
+        measurement = energy.EnergyMeasurement.create_if_supported()
+        if measurement is None:
+            self.skipTest("Energy Measurements not available")
+        measurement.start()
+        sleep(1)
+        measurement.stop()
```
PhilippWendler: This test does not actually check something, so it basically never fails. This means it is not a test. You need to add some assertions for example.

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,53 @@
...
+from time import sleep
+
+from benchexec import intel_cpu_energy as energy
+
+
+class TestEnergyMeasurement(unittest.TestCase):
+    def test_run(self):
+        measurement = energy.EnergyMeasurement.create_if_supported()
+        if measurement is None:
+            self.skipTest("Energy Measurements not available")
```
PhilippWendler: The tests here (except for the one that only tests a utility function) will not actually be useful if they only work on a machine where energy measurements are available. This is because tests are most useful in CI, which always runs on virtual machines.

So for useful tests, you need to mock the kernel interface. My suggestion is to mock the parts of the code that are used for reading the available domains and values and let them return values that are defined by the test. If this is not easily doable, then it is usually a good idea anyway to refactor the code and separate I/O from actual logic, and then mocking I/O becomes easier.
The alternative would be to create a directory with a dummy file hierarchy that simulates the files that the kernel provides and let the measurement code access that directory instead of the real one. But I think this will likely be more effort than Python-level mocking. We use pytest for our tests, so you can use utilities from both the `unittest` and pytest modules for mocking.

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_error_event(self):
+        """this test is expected to print an error message, it's not actually an error"""
+        measurement = energy.EnergyMeasurement.create_if_supported()
+        if measurement is None:
+            self.skipTest("Energy Measurements not available")
+        measurement.start()
+        sleep(0.5)
+        energy.EnergyMeasurement.error_event.set()
+        sleep(0.5)
```
PhilippWendler: It is always nicer if there is no need to sleep in a test, because this basically means that every developer of BenchExec has to wait 1s longer for tests everytime they run. If every test does this, it sums up. Furthermore, on an overloaded CI runner or so it could mean that the test fails because the code was run too slowly.

Sometimes there is no way around it, but please consider rewriting the test such that there is no sleep.

PhilippWendler:
> I am not really experienced in writing unit tests, I gave it a try, hopefully it's useable.

I added some comments to the tests. Additionally, I think the tests cover only a minimum right now and should be extended. For example, given how much we discussed about overflows and how important correct overflow handling is, I guess it is clear that we must test this as well.

> I assume I can just copy the licensing/copyright informations for the new file?

Yes.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,228 @@
...
+        super().update_value()
+        for domain in self.domains:
+            domain.update_value()
 
 
 class EnergyMeasurement:
-    def __init__(self, executable):
-        self._executable = executable
-        self._measurement_process = None
+    error_event = threading.Event()
```
PhilippWendler: Why do we need this `error_event`? Signaling errors out-of-band like this is error-prone because it means that some code can easily forget to check for errors, and then return wrong values mistakenly. It is better to return the error via the usual channels (like the actual data) or use exceptions (which typically crash the program if they are unhandled instead of letting the program silently continue).

ronin145: Yes I know, I originally had something else in mind, but I realised it was nonsense. I changed it to work without the error_event, I also made it so that all measurements are deleted if an error occurs.

benchexec/test_energy_measurement.py
ronin145: I removed it as far as possible, but sometimes we need a very short sleep to let the thread read values before we change them

benchexec/test_energy_measurement.py
ronin145: I made it so that every test has it's own mock RAPL directory that is patched over the real rapl_path. Now it should work on a virtual machine.

benchexec/test_energy_measurement.py
ronin145: with the new mock directory I was able to add assertions for the expected values

ronin145:
I am getting an error from ruff check, I think because I am setting write permissions in one of the tests. Unfortunately I need write permissions.

PhilippWendler:
> I am getting an error from ruff check, I think because I am setting write permissions in one of the tests. Unfortunately I need write permissions.

Why do you need to give write permissions to *all* users on the system? I think this is unnecessary and unintended.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
-DOMAIN_UNCORE = "uncore"
-DOMAIN_DRAM = "dram"
+
+@dataclass
+class EnergyWrapper:  # This wrapper is needed to keep immutability in the Domain and Package classes while still allowing energy values to be updated
+    last_value: int
+    total: int = 0
+
+
+class AreaOfMeasurement:
```
PhilippWendler, line 27: The design of this class hierarchy does not follow best practices.

`AreaOfMeasurement.update_value` accesses `self.energy` and `self.path`, but `AreaOfMeasurement` does not define these fields. It only works because it happens to be the case that all subclasses of `AreaOfMeasurement` do have these values. But `AreaOfMeasurement` is not self contained.

Additionally, `Domain` and `Package` have several fields that are defined in both, and have the same semantics: `name`, `path`, and `energy`. Why the redundancy? If you already have a class hierarchy, why not use it do deduplicate this?
Right now you are using the class hierarchy only to share the code in `update_values` between `Domain` and `Package`. But this is not really a meaningful reason for inheritance, as `update_value` could also be a standalone function that takes an object as parameter.

So I suggest to reconsider the class hierarchy. Likely it is a good idea to start moving the shared fields into `AreaOfMeasurement` and then do further redesigning.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+    def update_value(self):
+        """update the energy values of current domain or package, this checks for overflows as well"""
+        new_energy = int(get_path_content(self.path / "energy_uj"))
+        if self.energy.total == 0 and self.energy.last_value == 0:  # first measurement
+            self.energy.last_value = new_energy
+
+        elif new_energy <= self.energy.last_value:  # overflow
```
PhilippWendler, line 34: Is the case where both values are the same really an overflow?

Note: corner cases like this are important to have tests for!

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+            overflow_border = int(get_path_content(self.path / "max_energy_range_uj"))
```
PhilippWendler, line 35: Not sure if it is better to read this once and store it?

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+            self.energy.total += overflow_border - self.energy.last_value + new_energy
+            self.energy.last_value = new_energy
+
+        else:
+            self.energy.total += new_energy - self.energy.last_value
+            self.energy.last_value = new_energy
```
PhilippWendler, line 41: I would recommend to improve this code by removing the redundancy between the branches. For example, updating `total` and `last_value` is done on both branches, and the code would be easier to understand if shared actions exist only once in the code.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
-            bufsize=10_000,
-            preexec_fn=os.setpgrp,  # Prevent delivery of Ctrl+C to subprocess  # noqa: PLW1509
-        )
+        """Start the measurement"""
+        self.update_thread = threading.Thread(target=self.update_values)
+        self.stop_event.clear()
+        self.update_thread.start()
```
PhilippWendler, line 114: This code delays the first measurement until the thread has been started and scheduled by the OS. But we want to get the measurement to be as close as possible to the point where `start` is called. So I would recommend to do the first measurement here.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+        for package in sorted(
+            p for p in rapl_path.glob("intel-rapl:*") if p.name.count(":") == 1
+        ):
+            p_name = get_path_content(package / "name")
+            domains = []
+            for domain in sorted(
+                d for d in package.glob("intel-rapl:*") if d.name.count(":") == 2
+            ):
+                d_name = get_path_content(domain / "name")
+                domains.append(Domain(d_name, domain, EnergyWrapper(0)))
```
PhilippWendler, line 94: The initialization of `EnergyWrapper` and the other data classes feels inconsistent. Why are both fields of `EnergyWrapper` initialized differently? Why do callers of `Domain()` need to care to pass a new `EnergyWrapper` instance, this feels like it should be the responsibility of `Domain` (same for `Package`)?

Also right now `EnergyWrapper`'s `last_value` field is initialized with `0` to indicate a new instance. But this value could also be an actual reading, leading to potential confusion. It is clearer and safer to distinguish between actual possible values and special values for special conditions. I would initialize `last_value` with `None`.

Maybe the code for the first measurement should actually be separate from the code for further measurements in an `init` or `start` method even.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+def get_path_content(path):
+    """if reading file fails the error event signals the thread that something went wrong
+    and stops further measurement"""
+    try:
+        content = path.read_text().strip()
+        return content
+    except OSError as error:
+        message = f"cannot read {path}: {error}"
+        logging.debug(message)
+        raise error
```
PhilippWendler, line 192: Logging an exception and re-throwing it is an anti-pattern, because this typically leads to duplicate and confusing log entries. Consider what happens if the code that finally handles the exception also (rightfully) logs it, or if every piece of intermediate code would also log messages of exceptions that pass through it.

So please remove the logging here.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+    # because the int() function throws a seperate error on None values
+
+
+def convert_to_joules(energy):
```
PhilippWendler, line 196: You added some type annotations in this class, but not consistently. I would recommend to do so and add type annotations for all parameters and return values of functions (except for boring `-> None` cases). This can help find bugs.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
-        if executable is None:  # not available on current system
+        if not rapl_path.exists():
             logging.debug(
-                "Energy measurement not available because cpu-energy-meter binary could not be found."
+                'Intel RAPL Kernel module for energy measurement not available, try "modprobe intel-rapl-msr"'
             )
             return None
-
-        return cls(executable)
+        return cls()
```
PhilippWendler, line 106: Can an I/O error happen here already? If yes, we would not want to crash BenchExec but log it and return `None` as well.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+        for package in self.packages:
+            string += f"{package.name}: {package.energy.total} uj\n"
+            for domain in package.domains:
+                string += f"    {domain.name}: {domain.energy.total} uj\n"
+        return string
+
+
+def get_path_content(path):
+    """if reading file fails the error event signals the thread that something went wrong
+    and stops further measurement"""
```
PhilippWendler, line 185: Please make sure to also update respective comments when changing code. Basically all of this comment is outdated.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
-    """Take the result of an energy measurement and return a flat dictionary that contains all values."""
-    if not energy:
+        return self
+
+    def update_all(self):
+        """this updates the counter for total energy consumed across all packages and domains"""
+        for package in self.packages:
+            package.update_value()
+
+    def update_values(self):
```
PhilippWendler, line 133: I think the naming of methods in this class should be improved. For a developer who wants to use the method it might not be clear if they need to call `start()`, `update_values()`, or `update_all()`.

Please prefix internal methods with `_` as it is common in Python code.

Also I find `update_values()` and `update_all()` confusing in itself: Both methods update *all* *values*. Please find more descriptive names that allow to distinguish the methods.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+                    )
+                    if constraint_value == 0 or max_range == 0:
+                        logging.debug(
+                            "failed to read a constraint value for EnergyMeasurement"
+                        )
+                        return
+                    intervals.append(max_range / constraint_value)
+
+        if intervals != []:
+            self.interval = min(intervals)
```
PhilippWendler, line 172: The assignment here is not good design. When a developer looks at the class `EnergyMeasurement` and wants to find out how `interval` is set, they will see the field with the default value of `1000`. Then they might check the `__init__` method, but they will not see an assignment to `self.interval` there, so they might conclude that `interval` is always `1000`.

The code is much easier to understand if initialization of object fields is in the `__init__` method, so please refactor it such that `calculate_interval` returns the value and `__init__` assigns it.
For further clarity, the assignment of the computed value and the default value could even be done in the same place such that the logic for `interval`'s default value is easily understandable.

Another, more minor, issue is that `calculate_interval()` is a method that is called by `__init__` while `self` is not fully initialized. This works, but can easily lead to errors or misunderstandings if `calculate_interval` accesses a field that does not yet have its final value. Thus it is typically clearer to avoid method calls from `__init__`. Of course it is a good idea to have the code for `calculate_interval` in a separate function, though.
But it is easy to make `calculate_interval` a plain function that takes the `packages` and returns `int | None`. Then it is immediately obvious what the code does and all the discussed potential errors are prevented.

And btw., such a design would also simplify testing `calculate_interval`, which is important but still missing.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+                        get_path_content(package.path / "max_energy_range_uj")
+                    )
+                    if constraint_value == 0 or max_range == 0:
+                        logging.debug(
+                            "failed to read a constraint value for EnergyMeasurement"
+                        )
+                        return
+                    intervals.append(max_range / constraint_value)
+
+        if intervals != []:
```
PhilippWendler, line 171: Pythonic would be:
```suggestion
        if intervals:
```

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+        except (OSError, ValueError, TypeError):
+            logging.error("Energy measurement failed")
+            self.packages = None  # to prevent accidental accessing
+            return
+
+    def calculate_interval(self):
+        """calculate measurement interval from short term power limit
+        each package has various constraints numbered from 0, therefore we have to check
+        each constraint name to find the number correlating to the short term limit
+        Example limit file name would be "constraint_1_power_limit_uw
```
PhilippWendler, line 149: What is the reason for choosing the short-term power limit and ignoring the others?

Would it be better to use the highest power limit of all?
Should power limits that are possible only for a short time ignored because the CPU can not consume so much power for a long time anyway? (If the meaning of "short time" is less than the max time until overflow?)
Should `power_limit_uw` or `max_power_uw` be used or both?

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+            package.update_value()
+
+    def update_values(self):
+        """this method is run by a thread to constantly sample energy values for overhead protection"""
+        try:
+            self.update_all()
+            while not self.stop_event.wait(timeout=self.interval):
+                self.update_all()
+            self.update_all()
+        except (OSError, ValueError, TypeError):
```
PhilippWendler, line 140: Catching `ValueError` and `TypeError` here is quite broad and can easily hide bugs due to unintended occurrences of such exceptions.

It is important to distinguish between two different sources of exceptions: Some exceptions happen because of expected error cases (like `ValueError` from `int()` if we read an empty file or so). These exceptions must be handled gracefully.
But some exceptions happen due to bugs in the code. These exceptions must be allowed to propagate such that they crash BenchExec and can be detected and fixed.

Unfortunately in Python `ValueError` and `TypError` are common exception types that occur in both situations. So we cannot just distinguish between the two cases by exception type, and instead we must make sure to only catch `ValueError` and `TypeError` precisely in those locations where we want to handle them.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+            logging.error("Energy measurement failed")
```
PhilippWendler, line 141: This error message completely hides the reason of the error and leaves the user with no information about what the problem is and what they could do. This is also the case for several other logging statements.

Please go through all logging statements and check that they are useful, informative, and targeted at the correct audience:
Messages that are useful for users should be info or warning messages (error is for when BenchExec needs to stop completely) and be written in a way that they are useful for plain users.
Other messages should be debug logs.
Both should contain relevant information.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+                    )
+                    max_range = int(
+                        get_path_content(package.path / "max_energy_range_uj")
+                    )
+                    if constraint_value == 0 or max_range == 0:
+                        logging.debug(
+                            "failed to read a constraint value for EnergyMeasurement"
+                        )
+                        return
+                    intervals.append(max_range / constraint_value)
```
PhilippWendler, line 169: Note that you don't even need to store a list with all intervals, you can with the same effort simply compute the (current) minimum on the fly.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
-        for domain, value in domains.items():
-            if domain == DOMAIN_PACKAGE:
-                cpuenergy += value
-            result[f"cpuenergy-pkg{pkg}-{domain}"] = value
-    result["cpuenergy"] = cpuenergy
+    total = Decimal(0)
+    for package in measurement.packages:
+        p_energy = convert_to_joules(package.energy.total)
+        # psys describes energy usage of the entire system and is therefore not relevant for cpuenergy
+        if package.name != "psys":
```
PhilippWendler, line 212: Code is typically easier to understand if there are fewer negations. Here we have an `if ... else ...` with a negated condition that could just as well have both branches swapped instead.

The above comment would then best be placed inside the `psys` branch.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+        raise error
+    # because the int() function throws a seperate error on None values
+
+
+def convert_to_joules(energy):
+    """The values read from the energy_uj file are in microjoules and need to be converted to joules"""
+    return Decimal(energy) / Decimal(1000000)
+
+
+def format_energy_results(measurement):
```
PhilippWendler, line 201: If this function requires an `EnergyMeasurement` instance anyway, wouldn't it be cleaner to make it a method in this class?

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,113 @@
...
+        (self.package_mock / "constraint_0_power_limit_uw").write_text("1000")
+
+        (self.domain_mock / "name").write_text("domain")
+        (self.domain_mock / "energy_uj").write_text("200")
+        (self.domain_mock / "max_energy_range_uj").write_text("10000")
+        (self.domain_mock / "constraint_0_name").write_text("short_term")
+        (self.domain_mock / "constraint_0_power_limit_uw").write_text("100")
+
+    def test_run(self):
+        with patch.object(energy, "rapl_path", self.rapl_mock):
```
PhilippWendler, line 44: In cases like this you can use `@patch` as decorator for the test, saves you having to indent the whole method.

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,113 @@
...
+            (self.package_mock / "energy_uj").write_text("4000")
+            (self.domain_mock / "energy_uj").write_text("400")
+            result = measurement.stop()
+            self.assertIsNotNone(result)
+
+            expectedResult = {}
+            expectedResult["cpuenergy"] = energy.convert_to_joules(2000)
+            expectedResult["cpuenergy-package"] = energy.convert_to_joules(2000)
+            expectedResult["cpuenergy-package-domain"] = energy.convert_to_joules(200)
+            expectedResult = collections.OrderedDict(sorted(expectedResult.items()))
```
PhilippWendler, line 94: You can just use a simple dict literal for this. You do not even need `OrderedDict`.

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,113 @@
...
+            sleep(
+                0.1
+            )  # we need a short sleep so that writing does not interfere with thread reading
+            (self.package_mock / "energy_uj").write_text("4000")
+            (self.domain_mock / "energy_uj").write_text("400")
+            result = measurement.stop()
+            self.assertIsNotNone(result)
+
+            expectedResult = {}
+            expectedResult["cpuenergy"] = energy.convert_to_joules(2000)
```
PhilippWendler, line 91: If you call `convert_to_joules` here, bugs in it would not be detected by the tests because both the tests and the real code would use the same calculation.

Either add tests for `convert_to_joules` or avoid using it in tests and hard-code the expected values.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,223 @@
...
+@dataclass
+class EnergyWrapper:  # This wrapper is needed to keep immutability in the Domain and Package classes while still allowing energy values to be updated
+    last_value: int
+    total: int = 0
+
+
+class AreaOfMeasurement:
+    def update_value(self):
+        """update the energy values of current domain or package, this checks for overflows as well"""
+        new_energy = int(get_path_content(self.path / "energy_uj"))
```
PhilippWendler, line 30: Maybe it would be a good idea to check that the value is in the expected range?

PhilippWendler (commented):
Please continue adding tests for all relevant cases, the current tests only cover basic cases.

Also, please structure your commits in a meaningful way and do not mix unrelated changes in a single commit as for example done in e0f63d8d.

benchexec/test_energy_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+
+            (second_domain_mock / "name").write_text("domain2")
+            (second_domain_mock / "energy_uj").write_text("100")
+            (second_domain_mock / "max_energy_range_uj").write_text("1000")
+            (second_domain_mock / "constraint_0_name").write_text("short_term")
+            (second_domain_mock / "constraint_0_power_limit_uw").write_text("100")
+
+            measurement = energy.EnergyMeasurement.create_if_supported()
+            if measurement is None:
+                self.skipTest()
```
PhilippWendler, line 135: Please remove all `skipTest()` calls for all cases that should not occur in tests. Otherwise the test could be accidentally skipped and bugs be hidden.

benchexec/test_energy_measurement.py
PhilippWendler: This is still problematic as explained.

A better strategy is to not rely on threads in (most of) the tests. After all, most tests test just the reading code and calculations, and not the code for measuring in a background thread. So the tests should not start any thread but trigger measurements manually at the required points of time.

Separate tests would then test specifically the measurements in a background thread, and these may use `sleep` if there is no better alternative.

But it is much more robust to separate the goals of tests and not use the threads in every test where only some calculation is tested.

benchexec/intel_cpu_energy.py
```diff
@@ -7,98 +7,234 @@
...
-            [self._executable, "-r"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
-            bufsize=10_000,
-            preexec_fn=os.setpgrp,  # Prevent delivery of Ctrl+C to subprocess  # noqa: PLW1509
-        )
+        """Start the measurement"""
+        for package in self.packages:
+            package.reset_value()
```
PhilippWendler, line 111: What is the goal of this?

And why do we need separate code instead of `update_values` handling this?

ronin145: We need this to reset the total counter in between runs.

benchexec/test_energy_measurement.py
ronin145: The idea was that there already is a test that confirms if the measurements are created correct. If I don't skip this in the other tests than a bug in this function would fail all tests, and I thought that perhaps that would make finding it harder. But I can change it if that is the desired behavior

benchexec/intel_cpu_energy.py
ronin145: Neither am I. It could happen that for some reason this file disappears during runtime. That wouldn't prevent us from continuing measurements, but I think for that to happen something has to have broken in the Kernel module, so maybe it is better to check the file just to be sure?

benchexec/intel_cpu_energy.py
ronin145: I think that it is more likely to be an overflow than not. Especially the packages, which are the only values that contribute to cpuenergy, grow very quickly. I did some tests with the ls command and it never detected a wrong overflow.
To be honest this is a lose-lose situation, we can only handle either a full wraparound or no energy consumed at all, and I think a full wraparound is much more likely in our usecase

benchexec/intel_cpu_energy.py
ronin145: On some systems (mine as well) the long time power limit is absurdly high, and serves no purpose. I think it is just a lazy way of configuring the machine to only impose short time limits. If we read those we get a interval that is shorter than it could be. In a normal configuration the short time limit should be the highest one anyway. If the long time power limit is higher than the sort time limit, the short time limit will always be the bottleneck, because it triggers first.

benchexec/intel_cpu_energy.py
ronin145: done

benchexec/intel_cpu_energy.py
ronin145: I removed the entire function and replaced it with the benchexec read_file

benchexec/intel_cpu_energy.py
ronin145: done

benchexec/intel_cpu_energy.py
ronin145: I am not sure what you mean. I added error handling to the init function, but I don't see how an error could occur in this function

benchexec/intel_cpu_energy.py
ronin145: done

benchexec/intel_cpu_energy.py
ronin145: Did that, and also changed the default interval to 500 seconds. Peak energy consumption for modern CPUs is closer to 500W. I don't know what happend, but I wrote 500W in my notes and somehow used 250W in the code

benchexec/intel_cpu_energy.py
ronin145: done

benchexec/test_energy_measurement.py
ronin145: I wrote a small test

benchexec/test_energy_measurement.py
PhilippWendler: Yes, the risk for accidentally skipping tests because something does not is more important than making sure that there is only one failing test. For each of the failing tests a developer can easily see that the cause is already the creation of the measurement object.

benchexec/intel_cpu_energy.py
PhilippWendler: This is not really an argument. If you fear that the kernel breaks and some files start disappearing, then you would have to check also all the other files (like` name`) all the time. But this does not make sense. As long as there is no other indication, we can assume that the file remains available as long as all the others. Whether we read it every time or not can depend on what is best for the code, overhead, etc.

benchexec/intel_cpu_energy.py
PhilippWendler: I think this is a completely wrong approach. We do not want to consider "likely" when we want correct measurements.

Why should it be relevant if some value contributes to the cpuenergy value?

Testing can be a good idea: If it shows us that 0 µJ differences can happen on subsequent reads, we must treat such differences as non-overflows. But of the course the inverse does not hold!

First, packages are the fastest growing values, so if we want to test this, we need to look at other domains (the slowest growing ones). 

Second, `ls` already does a lot of work. Even if we would need a dummy command-line utility to test this, we could use `true`, which does less.

But we don't have to use any command at all. We can just read the same value twice with `cat energy_uj; cat energy_uj` or (even closer together) `cat energy_uj energy_uj`. I can easily produce subsequent reads that return the same value with both commands for the uncore domain on my notebook (more than 10% of tries for the first command, almost every try for the second command). This shows we need to correctly handle this.

But there is also no lose-lose situation here, not even conceptually, and we do not need to rely on testing this. After all, we need to handle overflows properly by making sure that the measurement interval is small enough. And so we need to (and can) make sure that the measurement interval is small enough that the CPU never consumes `>= max_energy_range_uj` between two measurements, and then there is no problem at all with this case.

benchexec/intel_cpu_energy.py
PhilippWendler: So, if you are saying that some systems have dummy power limits here that are too high to make sense, why can we trust these values at all? What is the reason that allows us to rely on short-term power limits to always make sense if we know already that the long-term power limits sometimes do not make sense? Is this documented somewhere?

What is "absurdly high", and to which measurement interval would it lead? Would it cause problems if we measure that often, e.g., because it happens every few milliseconds and the overhead would be too high?

Do we have to guard against such dummy values?

If there are dummy values, how can we rely on dummy values being always higher, and not (too) low?

You really need to make a convincing argument here (and in the thesis) that this is a correct approach, and ideally back this with concrete guarantees from documentation etc.

benchexec/intel_cpu_energy.py
PhilippWendler: First, I still do not understand how you derive a measurement interval of 500s from an assumed maximum power of 500W.

Second, how did you arrive at a peak power for modern CPUs that is close to 500W? I think the chosen approach was wrong, because even quickly checking a list processors for a single vendor on Wikipedia already shows me counterexamples.

Third, even if we assume that peak power for modern CPUs is close to 500W, that should certainly not mean that we hard-code this value as an assumption and produce wrong values immediately above 500W power. After all, it is entirely plausible that CPUs consume more power in the future and we need to make a realistic estimate that leaves room for future grow.

benchexec/intel_cpu_energy.py
PhilippWendler: What are your arguments that all the I/O in this function is guaranteed to succeed, if you don't see how it can fail?

Also, please follow the same as what we have discussed before for what to log and how.

The current error handling still makes the code crash if an error in `__init__` occurs.

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
+Powercap files are only readable with root permissions by default, but we can configure read acces for benchexec with a udev rule (this should be done automatically with the Debian software package):
+`SUBSYSTEM=="powercap", KERNEL=="intel-rapl*", RUN+="/usr/bin/chgrp benchexec /sys/%p/energy_uj", RUN+="/usr/bin/chmod g+r /sys/%p/energy_uj"`
+Benchexec has to execute as the `benchexec` group.
+If the Powercap files aren't readable it might be neccesary to retrigger udev rules:
+`sudo udevadm control --reload-rules
+sudo udevadm trigger`
+If that doesn't work, try unloading and reloading the `intel-rapl-msr` module.
+In any case, rebooting should fix problems with udev configuration.
+
+Benchexec will measure up to four values for each of the CPUs:
```
PhilippWendler, line 100: Shouldn't this explain the system-energy value as well?

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
-the tool [cpu-energy-meter](https://github.com/sosy-lab/cpu-energy-meter) needs to be installed.
-It will measure up to four values for each of the CPUs:
-
-- `cpuenergy-pkg<i>-package` is the energy consumption of the CPU `<i>` (whole "package").
-- `cpuenergy-pkg<i>-core` is only the consumption of the CPU cores.
-- `cpuenergy-pkg<i>-uncore` is the consumption of the so-called "uncore" parts of the CPU (this may include an integrated graphics card).
-- `cpuenergy-pkg<i>-dram` is the consumption related to memory attached to CPU `<i>` (unclear what exactly this covers and might vary across systems).
+(not the whole system), and only for modern Intel CPUs (since SandyBridge) and AMD CPUs (since Zen).
+
+BenchExec uses the Power Capping Framework to access RAPL energy counters.
```
PhilippWendler, line 88: Is there a webpage or docs for powercap that we can link?

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
-
-- `cpuenergy-pkg<i>-package` is the energy consumption of the CPU `<i>` (whole "package").
-- `cpuenergy-pkg<i>-core` is only the consumption of the CPU cores.
-- `cpuenergy-pkg<i>-uncore` is the consumption of the so-called "uncore" parts of the CPU (this may include an integrated graphics card).
-- `cpuenergy-pkg<i>-dram` is the consumption related to memory attached to CPU `<i>` (unclear what exactly this covers and might vary across systems).
+(not the whole system), and only for modern Intel CPUs (since SandyBridge) and AMD CPUs (since Zen).
+
+BenchExec uses the Power Capping Framework to access RAPL energy counters. 
+The necessary Kernel module can be loaded by executing `modprobe intel-rapl-msr`. 
+The Power Capping Framework exposes the RAPL over sysfs.
```
PhilippWendler, line 90: "the RAPL" is undefined here. Please explain/link.

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
-- `cpuenergy-pkg<i>-package` is the energy consumption of the CPU `<i>` (whole "package").
-- `cpuenergy-pkg<i>-core` is only the consumption of the CPU cores.
-- `cpuenergy-pkg<i>-uncore` is the consumption of the so-called "uncore" parts of the CPU (this may include an integrated graphics card).
-- `cpuenergy-pkg<i>-dram` is the consumption related to memory attached to CPU `<i>` (unclear what exactly this covers and might vary across systems).
+(not the whole system), and only for modern Intel CPUs (since SandyBridge) and AMD CPUs (since Zen).
+
+BenchExec uses the Power Capping Framework to access RAPL energy counters. 
+The necessary Kernel module can be loaded by executing `modprobe intel-rapl-msr`. 
+The Power Capping Framework exposes the RAPL over sysfs. 
+Powercap files are only readable with root permissions by default, but we can configure read acces for benchexec with a udev rule (this should be done automatically with the Debian software package):
```
PhilippWendler, line 91: Here you use the short name "powercap" without introducing it before.

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
-It will measure up to four values for each of the CPUs:
-
-- `cpuenergy-pkg<i>-package` is the energy consumption of the CPU `<i>` (whole "package").
-- `cpuenergy-pkg<i>-core` is only the consumption of the CPU cores.
-- `cpuenergy-pkg<i>-uncore` is the consumption of the so-called "uncore" parts of the CPU (this may include an integrated graphics card).
-- `cpuenergy-pkg<i>-dram` is the consumption related to memory attached to CPU `<i>` (unclear what exactly this covers and might vary across systems).
+(not the whole system), and only for modern Intel CPUs (since SandyBridge) and AMD CPUs (since Zen).
+
+BenchExec uses the Power Capping Framework to access RAPL energy counters. 
+The necessary Kernel module can be loaded by executing `modprobe intel-rapl-msr`.
```
PhilippWendler, line 89: Does this mean one has to manually load it? I guess not, but the text sounds like it. Please clear up this confusion.

doc/resources.md
PhilippWendler, line 91: "should be done" does not really make it clear whether it *is* done, or a todo.

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
-
-For energy measurements to work,
-the tool [cpu-energy-meter](https://github.com/sosy-lab/cpu-energy-meter) needs to be installed.
-It will measure up to four values for each of the CPUs:
-
-- `cpuenergy-pkg<i>-package` is the energy consumption of the CPU `<i>` (whole "package").
-- `cpuenergy-pkg<i>-core` is only the consumption of the CPU cores.
-- `cpuenergy-pkg<i>-uncore` is the consumption of the so-called "uncore" parts of the CPU (this may include an integrated graphics card).
-- `cpuenergy-pkg<i>-dram` is the consumption related to memory attached to CPU `<i>` (unclear what exactly this covers and might vary across systems).
+(not the whole system), and only for modern Intel CPUs (since SandyBridge) and AMD CPUs (since Zen).
```
PhilippWendler, line 86: Are there any other requirements? Like kernel versions? Then this should be documented as well.

doc/resources.md
```diff
@@ -81,18 +81,27 @@ though the wall-time limit cannot be disabled completely if a CPU-time limit is
...
+
+BenchExec uses the Power Capping Framework to access RAPL energy counters. 
+The necessary Kernel module can be loaded by executing `modprobe intel-rapl-msr`. 
+The Power Capping Framework exposes the RAPL over sysfs. 
+Powercap files are only readable with root permissions by default, but we can configure read acces for benchexec with a udev rule (this should be done automatically with the Debian software package):
+`SUBSYSTEM=="powercap", KERNEL=="intel-rapl*", RUN+="/usr/bin/chgrp benchexec /sys/%p/energy_uj", RUN+="/usr/bin/chmod g+r /sys/%p/energy_uj"`
```
PhilippWendler, line 92: I don't think someone who does not know udev already will know what do to with this line.

PhilippWendler (commented):
Comments on the new documentation.

Please also do not hide the documentation on how to get energy measurements in `doc/resources.md`, add it to the installation documentation. `doc/resources.md` should still be the place where the measurements are explained.

benchexec/intel_cpu_energy.py
martinpycha: Addressed in 89446b70.

Changed the comparison so that two cases with the same values are not considered an overflow anymore.

The interval has been halved, so that now we read often enough to prevent missing the case where the counter went around more than one full time between two reads.

Added a test (`test_unchanged_value_is_not_overflow`) - when the counter did not change between two reads, the correct answer is that 0 energy was used.

benchexec/intel_cpu_energy.py
martinpycha: Addressed in e64eae3.

The `max_energy_range_uj` is now read only once during initialisation and stored.

benchexec/intel_cpu_energy.py
martinpycha: Changed in 9ba20c4.

PR #1291, opened 2026-07-26
Migrate `flake8` to `ruff`

### LINKED ISSUE #1290: Suggestion: use just `ruff` for `flake8`
`ruff` implements all useful rules of the `flake8-*` tools.
The maintenance is also better.

I would recommend to migrate to just `ruff` instead of using both.

Rotzbua:
## Changes

Remove the legacy `.flake8` file and move its lint behavior into `pyproject.toml` under `ruff`.
Rules are updated.

## Reference

fixes #1290


PhilippWendler:
Thanks! I fully agree that ruff is much nicer than flake8. I just wasn't sure whether it indeed is a complete replacement for flake8 or whether there is value in having both. But if you can confirm that ruff supersedes flake8, that is nice.

To fully understand the PR: Why do we need to change the ruff config? After all, we use ruff already with the existing config. If we can enable more checks, that is great, but I do not see the reason for porting over flake8 configs that are not needed for ruff.

Independently I noticed that the next ruff version (0.16.0) enables a lot more checks, and I have started on fixing these. So if there are checks that are on by default in 0.16.0, we do not need to enable them now. I expect to merge the improvements from ruff 0.16.0 soon as well.

PhilippWendler:
I replaced this now with 0ed10486ec87ac8ee2b32fcb1e5e881e2419ec9b because after #1297 it was easier than updating this PR. Still thanks for pushing this topic! If you happen to have any suggestions for improving our ruff config, this is always welcome.

PR #1292, opened 2026-07-26
Fix Python version label in Gitlab CI job

Rotzbua:
Rename the cgroupsv2 test job from `python-3.15` to `python-3.13` so the job name matches the configured `PYTHON_VERSION` value.


PhilippWendler:
Thanks!

PR #1294, opened 2026-07-30
Add different ways to handle missing <requiredfiles>

### LINKED ISSUE #785: Allow skipping runs for which <requiredfiles> does not match
In uses cases like SV-COMP, we have benchmark definitions with lots of runs, but for some of the runs a required input file does not exist, and we are not interested in executing these runs (they would just fail anyway). We can make BenchExec avoid executing these useless runs by adding an attribute like "strictly required" or so to `<requiredfiles>`. If this parameter was set and the pattern would not match for a specific task, BenchExec would silently skip this run instead of logging `"Pattern %s in requiredfiles tag did not match any file for task %s."` and keeping the run.

martinpycha:
Fixes #785 

Adding different modes to the `<requiredfiles>` tag:
- Fail - raises BenchExecException immediately, aborts.
- Warn - warning per affected run.
- Ignore - run completed, completely silent.
- Skip (strict) - run skipped and, single summary at the end logged at once.


benchexec/model.py
```diff
@@ -1033,18 +1070,33 @@ def __init__(
...
+                )
+            elif mode == "warn":
                 logging.warning(
                     "Pattern %s in requiredfiles tag did not match any file for task %s.",
                     pattern,
                     self.identifier,
                 )
-            self.required_files.update(this_required_files)
-
+            elif mode == "skip":
```
PhilippWendler: After thinking more about it, I think that `skip` could be misunderstood as "skip this file" (i.e., what `ignore` does) instead of "skip this run".

So I think we better rename `skip` to `skip-run`.

benchexec/model.py
```diff
@@ -220,6 +220,20 @@ def relpath(path):
     return args
 
 
+_REQUIREDFILES_MODES = ("fail", "warn", "ignore", "skip")
+_REQUIREDFILES_MODES_WITHOUT_SKIP = ("fail", "warn", "ignore")
+
+
+def _get_requiredfiles_mode(tag, allowed_modes=_REQUIREDFILES_MODES):
+    mode = tag.get("mode", "warn")
```
PhilippWendler: `mode` feels quite generic when thinking more about it. Something like `ifmissing="warn"` etc. would be quite a lot more descriptive, do you agree?

In the code we could use the term "missing-files mode" instead of just "mode".

benchexec/model.py
```diff
@@ -220,6 +220,20 @@ def relpath(path):
     return args
 
 
+_REQUIREDFILES_MODES = ("fail", "warn", "ignore", "skip")
+_REQUIREDFILES_MODES_WITHOUT_SKIP = ("fail", "warn", "ignore")
```
PhilippWendler: Why do we sometimes not allow `skip`?

benchexec/model.py
```diff
@@ -1033,18 +1070,33 @@ def __init__(
...
+            matched = self.runSet.expand_filename_pattern(
+                pattern, runSet.benchmark.base_dir, sourcefile=rel_sourcefile
             )
-            if not this_required_files:
+
+            if matched:
+                self.required_files.update(matched)
+                continue
+
+            if mode == "fail":
```
PhilippWendler: Please use `elif` instead of `continue` and `if`. This makes the control flow more obvious.

benchexec/model.py
```diff
@@ -1033,18 +1070,33 @@ def __init__(
 
         self.required_files = set(required_files)
         rel_sourcefile = os.path.relpath(self.identifier, runSet.benchmark.base_dir)
-        for pattern in required_files_patterns:
-            this_required_files = runSet.expand_filename_pattern(
-                pattern, runSet.benchmark.base_dir, rel_sourcefile
+
+        self.should_be_skipped = False
```
PhilippWendler, line 1086: It is not really ideal design if we create an object and afterwards have to check whether the object can/should be used. It is usually better to not create the respective object when it should not be used.

Here, the way you did it is certainly easier because otherwise we would need to do a lot of restructuring, so we can keep it. But it would be good to add some documentation about this unusual pattern, e.g., mention in the docstring of the constructor that callers should check `should_be_skipped` afterwards.

And we could for example add some `assert not self.should_be_skipped` statements to the methods in the `Run` class to guard against someone forgetting the check.

PhilippWendler (commented):
Nice tests! Thanks!

Overall the code looks great! I haven't tried it out yet but I am optimistic.

My main points are about some better names for the new attribute. Sorry that this comes late, but sometimes it is just not so obvious before seeing the actual code and thinking more deeply about it. But naming is important because it is hard to change afterwards.

There is now a merge conflict in the test file, because I changed something on the main branch. But this should be easy to resolve, likely it is just the logging import (we now enforce a certain import order). Please try merging main into your branch to resolve it. And afterwards please make sure to run `ruff check` because on the main branch I have now enabled many more ruff checks than before.

We should also use the opportunity to add some documentation for `<requiredfiles>`, potentially in `doc/benchexec.md` and some `doc/benchmark*.xml` examples. Could you please have a look where it fits?

benchexec/model.py
martinpycha: Changed to `skip-run`.

benchexec/model.py
martinpycha: Because `skip-run` drops individual runs and is only meaningful where required files are declared per task.

benchexec/model.py
martinpycha: Changed:)

benchexec/model.py
martinpycha: Done.

benchexec/model.py
```diff
@@ -216,6 +216,23 @@ def relpath(path):
     return args
 
 
+_REQUIREDFILES_MODES = ("fail", "warn", "ignore", "skip-run")
+# "skip-run" drops individual runs and is only meaningful where required files
+# are declared per task. A benchmark-level pattern either matches for all runs
+# or for none, so skipping there would mean skipping the whole benchmark.
```
PhilippWendler: This is a good idea, I wouldn't have thought about this. But I think what is explained here is not correct because patterns can contain variables that are expanded by `substitute_vars`, and these can refer to the task name (like `${taskdef_name}`). So a global pattern with a task-specific variable can match depending on the task.

Can you please add a test case with `<requiredfiles ifmissing="skip-run">` on the global level and two tasks, one which is skipped and one that is not?

PhilippWendler:
I tried it out with a real benchmark definition, and with `skip-run` I still got many log messages of the form `No files found matching '...'.` We can see these as well [here](https://buildbot.sosy-lab.org/cpachecker/#/builders/68/builds/485/steps/5/logs/stdio). The messages come from `expand_filename_pattern`. We should clean this up as well, such that with `skip-run` there is no message, and with `warn` there is only one message, and not two. I see two possibilities: either we add a parameter to `expand_filename_pattern` that controls whether a non-matching pattern should be logged, or we remove the logging from `expand_filename_pattern` and add it to all callers where it makes sense.

Could you please also add a test case where `<requiredfiles>` contains a variable expansion like `${taskdef_name}` and make it such that one task is skipped and the other matches? This is the actual real-world use case that we have and it would be good to cover it in a test.

Apart from this, I think everything looks good and works as it should.

doc/benchmark.dtd
```diff
@@ -29,6 +29,8 @@ SPDX-License-Identifier: Apache-2.0
 
 <!ATTLIST rundefinition name CDATA #IMPLIED>
 
+<!ATTLIST requiredfiles mode (fail|warn|ignore|skip) #IMPLIED>
```
PhilippWendler, line 32: This line still needs to be updated.

PhilippWendler:
Note to myself: Now `<requiredfiles>` has actually become a relevant BenchExec feature and is not just for BenchCloud users. So we should document it together with the `ifmissing` attribute.

martinpycha:
I tried to add documentation for the `<requiredfiles>` into the `doc/bencbexec.md` file underneath the documentation for the tag `<resultfiles>`, it seemed to me to be the most logical place for it. Also, I tried to document in further detail the usage of the optional `ifmissing` attribute and added a small example, to hopefully make its usage clear to everybody. 

I also added documentation into the `doc/benchmark.xml` file, since I figured this file represents the _format reference_. Furthermore, commented example has been added to the `doc/benchmark-example-cmbc.xml` file, since this example actually works with some files, making it quite a reasonable candidate for such comments.

Regarding the logging, the `log_if_empty=True` has been added to the `RunSet.expand_filename_pattern`. Also, the added test `test_requiredfiles_skip_run_with_variable_expansion`, which test the variable expansion.

Should be ready for merging now.

PhilippWendler (approved):
Thanks, I just had some minor cleanup.

Issue #1278, opened 2026-07-09
Add tests for table-generator on GitHub Actions


PhilippWendler:
Right now the primary CI for all Python tests is GitLab, because it has the environment that supports running containers etc. and because we can use our own CI runner there and thus have no resource constraints. Because `table-generator` is supposed to also work on Windows, we added executions for its tests on AppVeyor, which provides Windows runners. But AppVeyor is slow and often the CI jobs have to wait.

GitHub Actions also provides Windows runners, and we use GitHub Actions already for other tests. So we could run the `table-generator` tests (those currently executed by `.appveyor.yml`) also on GitHub Actions, at least on Windows or potentially on Windows and Linux.

If this works well we can remove AppVeyor. This would also have the advantage that we have to configure only two instead of three CI systems.

PhilippWendler:
Done by #1296.

PR #1296, opened 2026-07-30
Adding table-generator-tests to GitHub actions 

martinpycha:
#1278 

Adding GitHub actions workflow for running the table generator tests, runs "python -m pytest benchexec/tablegenerator/" on windows and ubuntu (Python 3.10 - 3.13).


.github/workflows/table-generator-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  test-table_generator:
+    name: ${{ matrix.os }} (Python ${{ matrix.python-version }})
+    runs-on: ${{ matrix.os }}
+  
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest]
+        python-version: ["3.10", "3.13"]
```
PhilippWendler: I think we want 3.10 and 3.14 (oldest one supported by BenchExec and latest version).

.github/workflows/table-generator-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
PhilippWendler: I think we don't need to upgrade pip.

.github/workflows/table-generator-tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Table generator tests
```
PhilippWendler: Please always write table-generator with a hyphen.

PhilippWendler:
Could we use our own container images for the Ubuntu runs? The image `registry.gitlab.com/sosy-lab/software/benchexec/test:python-3.10` is the same that we use for tests on GitLab, and it has the necessary dependencies, so we could skip installing `libsystemd-dev`.

PhilippWendler:
I just realized something: Does GitHub also provide free MacOS runners maybe? Could you please check and if yes, add these tests on Mac as well?

And you can also remove the `.appveyor.yml` file as part of this PR and update the respective reference in `doc/DEVELOPMENT.md`.

martinpycha:
Hi, the _zizmor_ check is failing due to the unpinned container image. Should the alert be suppressed or pinned?
Also, since the _appveyor.yml_ file was deleted, the AppVeyor build fails.

@PhilippWendler

PhilippWendler:
> Hi, the _zizmor_ check is failing due to the unpinned container image. Should the alert be suppressed or pinned?

I guess we can suppress it. It is our image, we update it weekly, and we use the same image for our GitLab runs.

> Also, since the _appveyor.yml_ file was deleted, the AppVeyor build fails.

No problem. We can merge nevertheless and once this is done I disable AppVeyor completely.

.github/workflows/table-generator-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+#
+# SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Table-generator tests
+
+on:
+  push:
+    branches: [main]
```
PhilippWendler, line 14: I thought about it whether this makes sense or whether we want all branches, but I agree. But please let's document the reason:
```suggestion
    # Skip on branches because these tests are also run on GitLab,
    # but run on main for Windows + Mac.
    branches: [main]
```

PhilippWendler (approved):

Issue #1289, opened 2026-07-25
Fix typos with `codespell`


Rotzbua:
I noticed a few typos and so checked the codebase using [`codespell`](https://github.com/codespell-project/codespell). There were quite a few, so I created several PRs for easier review process.

PRs

- [x] #1286
- [x] #1287
- [x] #1288

PhilippWendler:
Thanks for this suggestion and the PRs! Seems like a nice tool.

We definitively want to have this in CI as well, so I'll keep the issue open for this. Potentially we can use the official [GitHub action](https://github.com/codespell-project/actions-codespell).

PhilippWendler:
Done by #1299.

PR #1299, opened 2026-08-06
Adding codespell to CI

martinpycha:
#1289 

Added Codespell configuration file added to check for typos, it is now included in CI. 

Furthermore, another configuration file, .codespellrc, was added to enable ignoring some files/words, which should not be checked by codespell tool. 
These include (but are not limited to):

**files**:
- _./.git, BenchExec.egg-info_ - metadata regarding the repository
- _react-table/build, *.min.js_ - web table build
- _.js, .json, .snap_ - instances of javascript compiled code
- _AProVE_ - termination prover
- _GRAT_ - proof-checking tool chain
- _./benchexec/tablegenerator/test_integration/results/test.2015-03-03_1815.logfiles_ - logfiles from test run in 2015 (typo: breaked), seems unreasonable to correct typos in logfiles

**words**:
- _aprove, grat, notin, Donig, te, infor, wit, othe_ - names, identifiers, variables, ...


PhilippWendler:
Could you please add it on GitLab as well? I think for consistency it is good to have everything there (except Zizmor) that we have in other CI systems as well?

PhilippWendler:
#1303 fixes some more codespell warnings, so we can update this PR with the latest main branch in order to get the fixes.

.codespellrc
```diff
@@ -0,0 +1,16 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
+# https://github.com/sosy-lab/benchexec
+#
+# SPDX-FileCopyrightText: 2007-2026 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[codespell]
```
PhilippWendler: Is it possible to add the codespell config to `pyproject.toml`? If yes, I would prefer this because then we clutter the main directory with less files.

.codespellrc
```diff
@@ -0,0 +1,16 @@
...
+# skipping compiled javascript code
+skip = 	./.git,./BenchExec.egg-info,
```
PhilippWendler: Do we really need to manually list `.git`? Many common tools like ruff skip it automatically, maybe codespell as well?

.codespellrc
```diff
@@ -0,0 +1,16 @@
...
+		./benchexec/tablegenerator/react-table/build,
+		./benchexec/tablegenerator/react-table/node_modules,
+		*.min.js,package-lock.json,*.snap,
+		./benchexec/tablegenerator/test_integration/results/test.2015-03-03_1815.logfiles
+
+ignore-words-list = aprove,grat,notin,Donig,te,infor,wit,othe
```
PhilippWendler: `infor` seems to be a typo and should be fixed.

Does codespell support suppressing a single warning wit a comment? Then we likely want to do this for `Donig` and the stuff in `property_files_multiline_string.yml`.

And if it supports suppressing a warning for a whole file with a single comment in the file, then we would also want to do this for `aprove`, `grat`, and potentially `notIn`.

`benchexec/tablegenerator/test_integration/expected/` can be skipped completely, and then we can probably also remove `wit` from the ignore-words list. Ideally that list is empty afterwards.

PhilippWendler (commented):
Thanks! Looks good, just some questions about the config.

PhilippWendler:
Oh, where did the CI config go?

.codespellrc
martinpycha: `infor` has been fixed, `Donig` and stuff in `property_files_multiline_string.yml` suppressed. Suppressing a warning for a whole file with a single comment does not seem to be possible, so I just added the files concerned to the list of files which should not be subject to the control.
Also, some of the typos cannot be ignored using inline comment, because such comment clashes with Ruff.

martinpycha:
The CI config file should be back now.

.codespellrc
martinpycha: `.codespellrc` has been added to `pyproject.toml`.

.codespellrc
martinpycha: Removed `.git` from the list.

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
 ]
 python_files = ["test_*.py", "test_integration/__init__.py", "test.py"]
 norecursedirs = ["contrib/p4/docker_files", "build", "benchexec/tablegenerator/react-table"]
+
+[tool.codespell]
+skip = 	"""\
```
PhilippWendler: Could you please try turning this into a list of strings (with []) instead of a multiline string? Like for pytype's `exclude` setting? Each pattern should be its own string.

And please add `*.pdf`, there are often some in `contrib/plots/` that trigger codespell.

And the leading `./` is not necessary.

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
...
+        ./BenchExec.egg-info,\
+		./benchexec/tablegenerator/react-table/build,\
```
PhilippWendler: Please avoid tabs (in all our files) and make the indentation consistent with the rest of the file.

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
...
+		./benchexec/tablegenerator/react-table/node_modules,\
+        ./benchexec/tablegenerator/test_integration/expected,\
+		*.min.js,package-lock.json,*.snap,\
+        ./pyproject.toml,\
+        ./benchexec/tools/aprove.py,\
+        ./benchexec/tools/gratchk.py,\
+        ./benchexec/tools/gratgen.py,\
```
PhilippWendler: Please add a comment "because of AProVE" / "because of GRAT".

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
...
+        ./benchexec/model.py,\
+        ./contrib/vcloud-benchmark.py,\
```
PhilippWendler: These two files should not be ignored.

To fix the `te` problem, just rename `te` to `type_error` (also in `table-generator/__init__.py`).

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
...
+        ./benchexec/tablegenerator/react-table/src/tests/utils.test.tsx,\
+        ./benchexec/tablegenerator/react-table/src/utils/utils.tsx,\
```
PhilippWendler: These two files should also not be ignored.

It is better to ignore `notIn` (which likely would also occur in other JavaScript tests in the future) than the whole test files.

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
...
 norecursedirs = ["contrib/p4/docker_files", "build", "benchexec/tablegenerator/react-table"]
+
+[tool.codespell]
+skip = 	"""\
+        ./BenchExec.egg-info,\
+		./benchexec/tablegenerator/react-table/build,\
+		./benchexec/tablegenerator/react-table/node_modules,\
+        ./benchexec/tablegenerator/test_integration/expected,\
+		*.min.js,package-lock.json,*.snap,\
+        ./pyproject.toml,\
```
PhilippWendler: This seems unnecessary?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
```
PhilippWendler: Why 3.13 here and 3.10 in GitLab? I guess it does not matter, but we could be consistent. We could simply use 3.14.

pyproject.toml
```diff
@@ -266,3 +266,22 @@ filterwarnings = [
 ]
 python_files = ["test_*.py", "test_integration/__init__.py", "test.py"]
 norecursedirs = ["contrib/p4/docker_files", "build", "benchexec/tablegenerator/react-table"]
+
+[tool.codespell]
```
PhilippWendler, line 270: Please add:
```toml
builtin = "clear,rare,informal,usage,names"
summary = true
```

This will trigger a few more suggestions, please resolve them as follows:
- Ignore `.../dependencies.json`.
- Use `blocklist`.
- In URLs where `master` occurs, replace with `main` if the target repo uses `main`. If not, add a comment for ignoring.
- For the "master host" of the VerifierCloud we can just say "host".

pyproject.toml
martinpycha: Unfortunately, the list-of-strings format does not work. Multiline string seems to be the only working option, does not work without the leading `./`.  Added `*.pdf`.

pyproject.toml
martinpycha: Should be more consistent now.

pyproject.toml
martinpycha: Added.

pyproject.toml
martinpycha: Renamed.

pyproject.toml
martinpycha: Now, we ignore `notIn`.

pyproject.toml
martinpycha: Without some of those ignores, errors arise, which we may want to ignore, such as keywords (to be ignored) in `.toml` file.

.github/workflows/codespell.yml
martinpycha: Fixed to 3.14.

PhilippWendler:
Ah, and I forgot: Please pin codespell to a specific version in CI. They add dictionary entries in every release, and this could mean CI suddenly starts to fail on existing branches if we do not pin the version.

martinpycha:
Now pinned to the Codespell version 2.4.3.

pyproject.toml
martinpycha: Should be included in the commit.

.github/workflows/codespell.yml
PhilippWendler: Please also in `.gitlab-ci.yml`.

PhilippWendler (approved):
Looks good, thanks! Is it ready from your point of view as well?

PR #1301, opened 2026-08-07
Implement periodic writes for task-set result files

### LINKED ISSUE #1215: Periodic writes to task-set-specific result files

PhilippWendler:
This is a replacement of #1247 because #1247 also has code from a different feature mixed.

Fixes #1215

### DIFF
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -423,6 +423,9 @@ class OutputHandler:
             not self.results_per_rundefinition and len(runSet.blocks) > 1
         ):
             runSet.block_xml_files = {}
+            taskset_files_only = (
+                self.results_per_taskset and not self.results_per_rundefinition
+            )
             block_names = collections.Counter(block.name for block in runSet.blocks)
             duplicate_block_names = {
                 block_name for block_name, count in block_names.items() if count > 1
@@ -445,7 +448,8 @@ class OutputHandler:
                     "last_modified_time": time.monotonic(),
                 }
                 self.all_created_files.add(blockFileName)
-                self.xml_file_names.append(blockFileName)
+                if taskset_files_only:
+                    self.xml_file_names.append(blockFileName)
 
     def output_for_skipping_run_set(self, runSet, reason=None):
         """
--- a/benchexec/outputhandler.py
+++ b/benchexec/outputhandler.py
@@ -415,8 +415,37 @@ class OutputHandler:
             self.all_created_files.add(runSet.xml_file_name)
             self.xml_file_names.append(runSet.xml_file_name)
         else:
-            # make sure to never write intermediate files
+            # make sure to never write intermediate files for rundefinition
             runSet.xml_file_last_modified_time = math.inf
+
+        # Initialize tracking for task-set (block) files if needed for periodic writes
+        if self.results_per_taskset or (
+            not self.results_per_rundefinition and len(runSet.blocks) > 1
+        ):
+            runSet.block_xml_files = {}
+            block_names = collections.Counter(block.name for block in runSet.blocks)
+            duplicate_block_names = {
+                block_name for block_name, count in block_names.items() if count > 1
+            }
+
+            for block in runSet.blocks:
+                if block.name in duplicate_block_names:
+                    continue
+
+                blockFileName = self.get_filename(runSet.name, block.name + ".xml")
+                # Create initial empty XML for this block
+                block_xml = self.runs_to_xml(runSet, [], block.name)
+                block_xml.set("starttime", runSet.xml.get("starttime"))
+
+                # Write initial empty file
+                self._write_rough_result_xml_to_file(block_xml, blockFileName)
+
+                runSet.block_xml_files[block.name] = {
+                    "filename": blockFileName,
+                    "last_modified_time": time.monotonic(),
+                }
+                self.all_created_files.add(blockFileName)
+                self.xml_file_names.append(blockFileName)
 
     def output_for_skipping_run_set(self, runSet, reason=None):
         """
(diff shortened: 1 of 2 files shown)


PhilippWendler:
@martinpycha Please make sure to add @saurabh12nxf to the authors list in the README as part of this PR (and yourself as well if not yet done).

martinpycha:
Should be ready for merge now.

PhilippWendler:
Any ideas how to add tests for this?

martinpycha:
Firstly, tried to test the feature manually, creating a `.xml` file with the following content.

<?xml version="1.0"?>
<benchmark tool="dummy" timelimit="200s">
  <rundefinition name="rd1">
    <tasks name="setA"><include>test/tasks/true_task.yml</include></tasks>
    <tasks name="setB"><include>test/tasks/false_task.yml</include></tasks>
  </rundefinition>
</benchmark>

This proved the behaviour to be correct. 

Furthermore, automatic test was added in the `benchexec/test_outputhandler.py`. The benchmark definition describes one run definition with two task sets (setA and setB), where each is holding a single task. The INTERMEDIATE_WRITE_INTERVAL is set to 0, so that one does not have to wait for 60 seconds. 

`@patch("benchexec.outputhandler._INTERMEDIATE_WRITE_INTERVAL", 0)` sets the write interval to zero for the duration of the test, so the "has enough time passed?" check always succeeds. Without it, nothing would be written for 60 seconds.
First assertion checks that both files parse and contain zero <run> elements (baseline). 'output_after_run_set' is never called, which means that the benchmark is still in progress. Second pair of assertions checks that file_a contains one <run>, file_b is still zero. Because the only thing that happened was only one run completing, that element must have been written incrementally - this is where we test the feature.

benchexec/outputhandler.py
```diff
@@ -65,6 +65,8 @@
 TIME_PRECISION = 2
 _BYTE_FACTOR = 1000  # byte in kilobyte
 
+_INTERMEDIATE_WRITE_INTERVAL = 0  # seconds between writes of intermediate results
```
PhilippWendler, line 68: This should be set to 60, right?

benchexec/test_outputhandler.py
```diff
@@ -0,0 +1,89 @@
...
+    walltimelimit = None
+    memorylimit = None
+    corelimit = None
+    num_of_threads = None
+    selected_run_definitions = None
+    selected_sourcefile_sets = None
+    description_file = None
+    start_time = None
+    compress_results = False
+    results_per_rundefinition = False
```
PhilippWendler, line 41: If we add such a nice test, I would say we can also use it to test both the `results_per_taskset` and the `results_per_rundefinition`. Likely just using one test for both is fine. So let's enable this, add assertions about the other file as well, and rename the test?

benchexec/test_outputhandler.py
```diff
@@ -0,0 +1,89 @@
...
+    results_per_taskset = True
+
+    def __init__(self, output_path):
+        self.output_path = output_path
+
+
+class TestIncrementalTasksetFiles(unittest.TestCase):
```
PhilippWendler, line 48: We don't need a class here, this was only necessary in the past.

When you remove it, replace all the `self.assert*` with standard Python `assert`s.

benchexec/test_outputhandler.py
```diff
@@ -0,0 +1,89 @@
...
+            self.assertEqual(self._count_runs(file_b), 0)
+
+            self._finish_run(block_a.runs[0])
+            handler.output_after_run(block_a.runs[0])
+
+            # the result of the finished run is on disk before the benchmark ends
+            self.assertEqual(self._count_runs(file_a), 1)
+            self.assertEqual(self._count_runs(file_b), 0)
+
+            handler.close()
```
PhilippWendler, line 80: At the end of the test, we could also finish the second run and test the file again. So it would test all expected steps of files being written (at least for `compress_results=False`).

We basically do this as part of our standard integration test as well, but this test is even better (faster, more robust). So maybe we eventually have more such tests and less integration tests, for example for features like `compress_results`.

benchexec/outputhandler.py
```diff
@@ -570,14 +605,41 @@ def output_after_run(self, run):
...
 
+            # Also write task-set (block) files periodically if enabled
+            if hasattr(run.runSet, "block_xml_files"):
+                for block in run.runSet.blocks:
+                    if block.name not in run.runSet.block_xml_files:
+                        continue
+
+                    # Check if this run belongs to this block
+                    if run not in block.runs:
+                        continue
```
PhilippWendler, line 627: I think this check is problematic. Because we do not write on every block (only every 60s), the last runs in a block might not get written (if the next 60s interval only elapses once we have started a new block).

I do like the fact, however, that no every block is written every time, only those blocks that have new results. We should keep this.

It would probably also be a good idea to always write on the last run of a block, even if the 60s have not yet passed.

Can you please think about how to implement this nicely?

benchexec/outputhandler.py
```diff
@@ -415,9 +417,42 @@ def output_before_run_set(self, runSet, start_time=None):
...
+                # Write initial empty file
+                self._write_rough_result_xml_to_file(block_xml, blockFileName)
+
+                runSet.block_xml_files[block.name] = {
+                    "filename": blockFileName,
+                    "last_modified_time": time.monotonic(),
+                }
+                self.all_created_files.add(blockFileName)
+                if taskset_files_only:
+                    self.xml_file_names.append(blockFileName)
```
PhilippWendler, line 454: The code here partially duplicates the existing code for writing block files from `output_after_run_set`. Maybe this is the best way, but I am not sure. Can you please think about whether there can be some code refactored/shared/removed?

PhilippWendler (commented):
The test is really nice! Perfect how you made it work without any sleeping/waiting/time-dependent flakiness.

There are some things I noticed which I would like to ask you to work on.

PR #1316, opened 2026-09-15
Sikraken coverage

echancrure:
The new sikraken.py adds two things. 
cmdline now appends --coverage branch or --coverage reach, plus --reach NAME only when the property file names a function other than reach_error. A new helper, _coverage_options, reads the property file to decide which to emit. The flags are only added when sikraken.sh -v reports version 2.1.0 or later, so older archives still get today's command line.


PhilippWendler:
We typically advise against adding such complex logic for building a command line to tool-info modules. Reasons are for example that it is not immediately obvious to BenchExec users that this is happening and that the actually executed command line is not what is defined in the benchmark definition. Transparency is generally good in benchmarking, as it helps detect mistakes. Another reason is that such logic is hard to change, both for users of the tool who may want to benchmark it with different options, and for future developers who should consider backwards compatibility. My recommendation would thus be to instead either add the desired command-line arguments to the benchmark definition, or extend the tool such that it can decide based on a property file what it should do.

Another thing is that parsing the property file and starting the tool to get its version number once per `cmdline` call is too costly for large benchmarks in practice (especially when combining BenchExec with a job distribution system, where actual run execution is parallelized on a cluster but preparations are local and sequential). So this would have to be cached.

echancrure:
Thanks, agreed. New version is straightforward and just passes the property_file on CLI

PR #1317, opened 2026-09-17
Add tool-info module for JavaSMT

gcarpio21:
Adds a tool-info module for JavaSMT (https://github.com/sosy-lab/java-smt), which runs its command-line interface with a chosen solver on one SMT-LIB 2 file. Answers are mapped like in smtlib2.py; uncaught Java exceptions are reported by class name. 

Depends on sosy-lab/java-smt#710, which adds the command-line interface and  the `javasmt` launcher script the module executes.


benchexec/tools/javasmt.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for JavaSMT, a unified interface for SMT solvers in Java.
+    https://github.com/sosy-lab/java-smt
+
+    JavaSMT decides the satisfiability of a single SMT-LIB 2 formula, so tasks
+    with more than one input file are not supported. If a memory limit is
+    specified for BenchExec, it is passed to the JVM with the parameter -Xmx.
```
PhilippWendler, line 22: I do not think this makes sense. `-Xmx` does not specify the maximum memory consumption of the JVM, nor that of the process with the JVM. Both can get higher, in particular the latter one. For useful benchmarking, users need to choose a maximum heap size that fits into the memory limit together with all the other memory consumption.

benchexec/tools/javasmt.py
```diff
@@ -0,0 +1,122 @@
...
+import os
+
+import benchexec.tools.template
+from benchexec import result
+
+
+class Tool(benchexec.tools.template.BaseTool2):
+    """
+    Tool info for JavaSMT, a unified interface for SMT solvers in Java.
+    https://github.com/sosy-lab/java-smt
```
PhilippWendler, line 18:
```suggestion
```
Redundant.

benchexec/tools/javasmt.py
```diff
@@ -0,0 +1,122 @@
...
+    - javasmt, the launcher script that assembles the classpath
+    - the classes to run, either below bin/ or as a java-smt-<version>.jar
+    - lib/java/core/ with the core JARs
+    - lib/java/runtime-<solver>/ with the JARs of the individual solvers
+    - lib/native/<architecture>/ with the JNI libraries of the native solvers
+    """
+
+    REQUIRED_PATHS = [
+        "javasmt",
+        "bin",
```
PhilippWendler, line 34: Why this directory if there is also the JAR?

`program_files` is used basically only for combining BenchExec with BenchCloud, and it is likely better to not send lots of individual files.

benchexec/tools/javasmt.py
```diff
@@ -0,0 +1,122 @@
...
+        # e.g. java-smt-5.0.1-1260-g7e5485a79.jar.
+        jars = [
+            jar
+            for jar in glob.glob(
+                os.path.join(os.path.dirname(executable), "java-smt-*.jar")
+            )
+            if not jar.endswith(("-sources.jar", "-javadoc.jar"))
+        ]
+        if len(jars) != 1:
+            return ""
```
PhilippWendler, line 61: I think the whole part with having potentially more than one JAR file needs to be thought through. What happens for execution if there is more than one? Which one will be taken? Is it useful for users what happens?

In particular, if there is more than one JAR lying around, getting the correct version number (of the one that will be actually executed) is more important than if there is just one JAR, but the code here just gives up.

benchexec/tools/javasmt.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def cmdline(self, executable, options, task, rlimits):
+        heap = [f"-Xmx{rlimits.memory}"] if rlimits.memory else []
+        return [executable, *heap, *options, task.single_input_file]
+
+    def determine_result(self, run):
+        # JavaSMT exits with code 1 after printing "unknown",
+        # so the answer is checked before the exit code.
+        for line in run.output:
+            line = line.strip()
```
PhilippWendler, line 75: The lines in `run.output` do not contain a trailing line break, so all the `strip()` calls here and below are useless unless you expect actual spaces in the output lines.

benchexec/tools/javasmt.py
gcarpio21: I have removed it.

benchexec/tools/javasmt.py
gcarpio21: bin/ is now only sent if there is no JAR, so only the JAR is sent if there is one.

benchexec/tools/javasmt.py
gcarpio21: If there are several JARs, BenchExec now stops before starting any runs and lists them.

benchexec/tools/javasmt.py
PhilippWendler: Is there an actual use case for running JavaSMT without the JAR? What is the goal?

It adds complexity and so there should be a reason for it. Adding features later on is also easier than removing features.

PhilippWendler:
AFAIU this depends on https://github.com/sosy-lab/java-smt/pull/710. So it is better to not merge this yet until the JavaSMT feature is merged and released, in order to not release something in BenchExec that immediately needs to be changed if something in JavaSMT changes before the release.

Please ping me once the feature in JavaSMT is merged and released.

PR #1318, opened 2026-09-21
Take category into account when comparing results.

### LINKED ISSUE #969: Status comparison for diff tables does not take category into account
If we compute a diff table by comparing the status column (the default mode for diff tables), only the status itself is taken into account, not the associated category. Usually, this is no problem because for a given tuple `(task, status)` the category is always the same. However, in some scenarios additional categories (`correct-unconfirmed`) are introduced into the result files (from outside of BenchExec) and then this no longer applies. But the diff table will omit the columns where the category is different but the status is the same.

Given that we provide some support for the category `correct-unconfirmed` and e.g. compute statistics for it, we should also take it into account for the diff table. Comparing `(status, category)` by default instead of just `status` is probably a good default.

martinpycha:
Fixes #969 

Changed the logic in the `all_equal_result` function - now compares the categories as well.
Added a Unit test to check the altered function. The `tablegenerator/test_diff.py` checks three cases, one of them probing the altered functionality.

The new behaviour was documented accordingly in `doc/table-generator.md`.


PhilippWendler (approved):
Very nice, I have absolutely nothing to comment on, only that code + test + docs are as good as I can imagine!

